PR #1832, opened 2022-06-03
Adding Redux to plots

sroy3:
Tests were great to catch bugs, but this is definitely the kind of PR that needs to be tried as well as checked.


webview/src/plots/components/App.test.tsx
```diff
@@ -134,32 +141,22 @@ describe('App', () => {
...
   })
 
   it('should render the loading state when given no data', async () => {
-    render(<App />)
-    const loadingState = await screen.findByText('Loading Plots...')
-
-    expect(loadingState).toBeInTheDocument()
-  })
-
-  it('should render the loading state when not initially provided with collapsed sections', async () => {
```
sroy3, line 151: Removed because default data is now set in the webview and it's not a specific case anymore.

webview/src/plots/components/ZoomablePlot.tsx
```diff
@@ -1,42 +1,35 @@
...
-    plot: VegaLiteProps,
-    id: string,
-    refresh?: boolean
-  ) => void
-}
-
-interface ZoomablePlotOwnProps extends ZoomablePlotProps {
-  plotProps: VegaLiteProps
+interface ZoomablePlotProps {
+  plotProps: string
```
sroy3, line 9: I might revisit this later, but this is the fix I found to make sure all plots get rendered. Otherwise, Vega complains that it cannot set value to read only property id (I'm paraphrasing) or group...

We lose type validation that's the downside, but other than getting all plots rendered, another plus side is that passing strings instead of objects is that if a parent re-renders, the string will stay the same while an object might be passed with a new reference.

webview/src/plots/components/checkpointPlots/checkpointPlotsSlice.ts
```diff
@@ -0,0 +1,73 @@
...
+      state.isCollapsed = action.payload
+    },
+    update: (state, action: PayloadAction<CheckpointPlotsData>) => {
+      if (action.payload) {
+        state.plots = action.payload.plots
+        state.colors = action.payload.colors
+        state.selectedMetrics = action.payload.selectedMetrics
+        state.size = action.payload.size
+        state.plotsIds = action.payload?.plots?.map(plot => plot.title) || []
+        state.plotsById = {}
```
sroy3, line 59: I haven't done so for template plots because there isn't the same filtering being done in `TemplatePlotsWrapper` as there is in `CheckpointPlotsWrapper`. For template plots, `plots` is used in `TemplatePlots` and isn't passed to any other components.

webview/src/plots/components/templatePlots/TemplatePlots.tsx
```diff
@@ -1,64 +1,61 @@
...
-      type: MessageFromWebviewType.REORDER_PLOTS_TEMPLATES
-    })
-  }
-
-  const setSectionEntries = (index: number, entries: TemplatePlotEntry[]) => {
-    sections[index] = {
-      ...sections[index],
-      entries
+  useEffect(() => {
+    if (shouldSendMessage.current) {
```
sroy3, line 38: Causes a state update loop otherwise

rogermparent (commented):
Great work! I spun up plots manually and tried to check around thoroughly, and I didn't find anything that seems outright broken.
Unfortunately, from what I can see it seems like the UI is a bit less responsive on the redux version than on our current one- I don't know if that's just due to my system or not, but the difference when toggling plots looks significant.

main:

https://user-images.githubusercontent.com/9111807/174154327-7baf311d-6f88-4396-a730-d60669dd12c5.mp4

redux: 

https://user-images.githubusercontent.com/9111807/174154359-68c381a9-3f59-4579-9fe1-8ddb55d00e36.mp4 

I won't quite call it a blocker, but I think it calls for some more investigation and optimization whether or not we merge.

mattseddon:
I'm seeing a degradation in performance as well:

Screen Shot 2022-06-17 at 8 47 05 pm

https://user-images.githubusercontent.com/37993418/174284521-6bd51bd0-aa02-46b2-b444-3446fe7cf0e0.mov

Could be that my laptop can't handle the webpack dev server being run at the same time as experiments and having the webview open but it is chugging.

webview/src/plots/constants.ts
```diff
@@ -0,0 +1,6 @@
+export enum ReducerName {
```
mattseddon, line 1: [N] upper case enum keys.

mattseddon (commented):
The code looks fine but we need to look into the performance issues. Can redux handle comparing the larger template plot objects? Here is the performance running an experiment when I get rid of the confusion matrix:

https://user-images.githubusercontent.com/37993418/174289805-69d8acf7-9def-4139-89c4-44be872d9f5b.mov

Seems like a big difference. Maybe we need to rework that part of the state?

sroy3:
I'll check on the performance. That quite unexpected.

sroy3:
It's weird, it seems like the performance problem happens after the **Redux** change (in the component itself).

https://user-images.githubusercontent.com/3683420/174351176-ce93bec0-eff5-4903-ba55-3b4e41807cf4.mov

See how "order changed" comes after a delay while "use effect triggered" is almost instantaneous? 

This is the part of the code I used for that test:

```
export const Ribbon: React.FC = () => {
  const revisions = useSelector(
    (state: RootState) => state.webview.selectedRevisions
  )
  const [order, setOrder] = useState<string[]>(revisions.map(revision => revision.id ||''))
  const reorderId = 'id'

  useEffect(() => {
    setOrder(pastOrder => performOrderedUpdate(pastOrder, revisions, reorderId))
    console.log('TEST: use effect triggered')
  }, [revisions])

  useEffect(() => {
    console.log('TEST: order changed')
  }, [order])
...
```

Still investigating what could be the problem. My computer does not show the same long delays as yours, so I didn't notice it at first, but after testing main and my branch one after another, it is noticeable.

webview/src/plots/components/ZoomablePlot.tsx
sroy3: This is the main cause of performance drop. Like I said before, this fixed the problem, but not definitely not at the root. I thought we could leave it as is for now, but that is definitely not an option now with results like this. `JSON.parse` can get quite expensive, who knew? I think I might have found the real solution for the original problem so that we can get rid of stringify'ed props.

sroy3:
Found the [cause](https://github.com/iterative/vscode-dvc/pull/1832#discussion_r901924304) of performance lags and I think I also found how to fix it.

sroy3:
> @sroy3 I'm late to the party, but why do we need to store the problematic objects in the Redux state?

We have many levels between the `App` component where the plots are received and the actual plot. Using Redux allows to store the plots until they are needed. The goal here is to improve the performance by cutting down the number of re-renders. Also, it cleans up a lot of our logic and makes code more readable.

webview/src/plots/components/Plots.tsx
```diff
@@ -1,34 +1,38 @@
...
 import { DragDropProvider } from '../../shared/components/dragDrop/DragDropContext'
-import { sendMessage } from '../../shared/vscode'
 import { GetStarted } from '../../shared/components/getStarted/GetStarted'
-
-interface PlotsProps {
-  state: PlotsWebviewState
-}
+import { RootState } from '../store'
 
 // eslint-disable-next-line sonarjs/cognitive-complexity
```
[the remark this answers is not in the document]
sroy3: Unfortunately, yes. The complexity got to 7, which is much better than it was.

sroy3:
https://user-images.githubusercontent.com/3683420/175325788-1799e04e-c7f4-41b7-9fe3-fa34ce1b60fe.mov

https://user-images.githubusercontent.com/3683420/175328092-223074a8-b8b1-4469-9935-38d5da5d2b4a.mov


Performance looks much better now, to me it seems faster than main:

wolmir (approved):

rogermparent (approved):
Performance is significantly improved for me as well, I wouldn't say it's better than main but it's not much worse if at all.

PR #1872, opened 2022-06-10
Add global sort and filter indicators to the Experiments Table webview

rogermparent:
This PR adds indicators to the Experiments Table webview that show if the user has any sorts or filters applied. We initially discussed using text mostly for ease of implementation, but that proved pretty hard to fit in so I opted for an icon+badge design similar to VSCode's Activity Bar but a bit smaller.

These indicators are also clickable, doing so will focus the TreeView the represents the same data, either sorts or filters. This will help users see details about the applied filters that these indicators show basic at-a-glance info for, allowing for ease of use while minimizing duplication.

The colors don't exactly match the activity bar because some light themes (like the default) use a light-on-dark scheme for the activity bar that doesn't translate into the table without an extra background for the indicators alone or a drastic change to the table headers- both are still on the table, but I opted for the path of least resistance and went with our normal foreground and watermark colors that are built to contrast with standard backgrounds.

https://user-images.githubusercontent.com/9111807/173494262-3cdb8a0f-12f5-4642-a75e-9afafe9c9d84.mp4

These indicators also have tooltips that more explicitly display the amount of sorts/filters added, and in the case of filters the amount of experiments/sorts that have been filtered.

https://user-images.githubusercontent.com/9111807/173746678-f6aa2128-4395-449d-b8f0-190f64f9766c.mp4


webview/src/experiments/components/table/Indicators.tsx
```diff
@@ -0,0 +1,62 @@
...
+  filters?: string[]
+}) => {
+  return (
+    <div className={styles.tableIndicators}>
+      <Indicator
+        count={sorts?.length}
+        aria-label="sorts"
+        onClick={focusSortsTree}
+      >
+        <Icon width={16} height={16} icon={SvgSortPrecedence} />
```
sroy3, line 51: There is an `AllIcon` const with all the icons in there (needs to add each one manually). I added this because it was an easy way to see all the available icons, but importing them directly like you did is way more convenient. The only thing I would say about this is that it'd be better to import it as `SortPrecedence` from the `icons/index.ts` file. Or we could change the SVGR script not to generate an index file, I don't really mind. I'll do a follow up removing `AllIcons` and importing the icons directly.

sroy3 (approved):

mattseddon:
@rogermparent tests seem genuinely broken.

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -700,6 +700,62 @@ suite('Experiments Test Suite', () => {
...
+      expect(mockSendTelemetryEvent).to.be.calledOnce
+      expect(
+        mockSendTelemetryEvent,
+        'should send a telemetry call that the sorts tree has been focused'
+      ).to.be.calledWithExactly(
+        EventName.VIEWS_EXPERIMENTS_TABLE_FOCUS_SORTS_TREE,
+        undefined,
+        undefined
+      )
+    })
```
mattseddon, line 730: [B] Need to wait for the event to be flushed. Dirty version of the required test is 
```
    it('should be able to handle a message to focus the sorts tree', async () => {
      const { experiments } = buildExperiments(disposable, expShowFixture)

      const mockSendTelemetryEvent = stub(Telemetry, 'sendTelemetryEvent')

      const messageReceived = new Promise(resolve =>
        disposable.track(mockMessageReceived.event(() => resolve(undefined)))
      )

      const executeCommandStub = stub(commands, 'executeCommand')
      mockMessageReceived.fire({
        type: MessageFromWebviewType.FOCUS_SORTS_TREE
      })

      await messageReceived

      expect(executeCommandStub).to.be.calledWith(
        'dvc.views.experimentsSortByTree.focus'
      )
      expect(mockSendTelemetryEvent).to.be.calledOnce
      expect(
        mockSendTelemetryEvent,
        'should send a telemetry call that the sorts tree has been focused'
      ).to.be.calledWithExactly(
        EventName.VIEWS_EXPERIMENTS_TABLE_FOCUS_SORTS_TREE,
        undefined,
        undefined
      )
    })

```

rogermparent:
Tooltips have just been added and pushed with tests! The current format is `${filtered} experiments filtered`, though I think the ideal format and the one we've discussed so far is `${filtered}/${total}`. That will require getting the total amount of experiments/checkpoints into the Indicators component which currently isn't there, I can do that in a follow-up if we want to merge this in its current state ASAP.

mattseddon (approved):

mattseddon:
Happy to merge if the approved diffs in the Storybook are expected.

webview/src/stories/Table.stories.tsx
```diff
@@ -46,7 +47,8 @@ export default {
     design: {
       type: 'figma',
       url: 'https://www.figma.com/file/AuQXbrFj60xA2QXOjo9Z65/Experiments-Panel-%E2%80%A2-496'
-    }
+    },
+    layout: 'fullscreen'
```
rogermparent, line 51: This allows the table stories to work with the `position:fixed`, as Storybook adds a padding to its `iframe` by default which is considered part of the viewport which messes with the table otherwise.

webview/src/experiments/components/table/Indicators.tsx
```diff
@@ -0,0 +1,108 @@
...
+                <div>
+                  {formatFilteredCountMessage(
+                    'experiment',
+                    filteredCounts.experiments
+                  )}
+                </div>
+                <div>
+                  {formatFilteredCountMessage(
+                    'checkpoint',
+                    filteredCounts.checkpoints
```
mattseddon, line 96: [B] @rogermparent are we always showing this? Even when the project doesn't have checkpoints?

rogermparent: Ah, I think so. I can do a quick fix to not show the line when 0 checkpoints are filtered.

Issue #1636, opened 2022-05-01
DVC errors that may occur during or after running experiment


maxagin:
Find solution for displaying errors when the experiment failed, broken parameters, metrics and more. 

### TODO

**EXPs table**
- [x] Error at the row level 
- [x] On a particular column within a row
- [x] A hint with title and error content

[Figma specs](https://www.figma.com/file/CycOpuUFNMum7TgUgEPiJu/1636-%E2%80%A2-DVC-errors-that-may-occur-during-or-after-running-experiment?node-id=0%3A1)

**Plots**
- [x] Plots: handle errors gracefully #1649

[Figma specs](https://www.figma.com/file/CycOpuUFNMum7TgUgEPiJu/Errors?node-id=1206%3A40473)

---

**To consider**

- The hint includes Error and How To Fix content example
- Hover style and active symbol style

**Helpful links**
- https://github.com/iterative/dvc/pull/5984
- https://github.com/iterative/dvc/blob/8cdcf9b8bb31a5badf7459efc3ec308d6d3ed62e/dvc/commands/experiments/show.py#L24-L25

mattseddon:
@maxagin I would like to start working on this again, I looked at the spec and everything has been moved back into WIP. LMK when the next iteration is ready.

maxagin:
Thank you @mattseddon for the heads up. I will try to finish it at the earliest convenience - there is still some work needed.

dberenbaum:
@pmrowla Will there be any circumstance in which the whole experiment fails and still shows in `dvc exp show` results after the queuing changes are merged?

That might no longer be relevant in this view (although it might be shown if VS Code adds a view of the queue).

maxagin:
>Will there be any circumstance in which the whole experiment fails and still shows in dvc exp show results after the queuing changes are merged? https://github.com/iterative/vscode-dvc/issues/1636#issuecomment-1176441285

@mattseddon  Does it mean we are not expecting to have any error at the row-level anymore?

pmrowla:
By default failed experiments will not show in the table after the queueing changes, but we have discussed adding flags to `exp show` to control what gets displayed (i.e `dvc exp show --failed`) since it may still be useful for users to be able to see what params & deps were used in the experiment that failed

maxagin:
[...]

PR #1874, opened 2022-06-10
Handle experiments which contain errors

mattseddon:
Relates to #1636

This is a prototype for dealing with experiment-level errors. In the screenshot we are dealing with the following data being presented for the workspace:

```
{
  "workspace": {
    "baseline": {
      "error": {
        "type": "YAMLFileCorruptedError",
        "msg": "unable to read: 'params.yaml', YAML file structure is corrupted"
      }
    }
  }
```

We need to be mindful that we have to block "selection" of these records because we will not be able to display any plots for these particular records.

### Screenshot:

image

Screen Shot 2022-06-13 at 7 24 33 pm


mattseddon:
Will reuse some of the ideas/code from this branch but basically have to start again because we do not want to suppress selection for plots if the experiment has an error.

PR #1906, opened 2022-06-15
Add Multi-Select context menu to Experiments Table

wolmir:


webview/src/experiments/components/table/Cell.tsx
```diff
@@ -44,6 +45,15 @@ export const FirstCell: React.FC<
           isPlaceholder && styles.groupPlaceholder
         )
       })}
+      tabIndex={0}
+      aria-checked={bulletColor !== undefined}
+      role="switch"
+      onClick={toggleExperiment}
+      onKeyDown={e => {
```
sroy3, line 52: Since you now have `clickAndEnterProps`, you could use it here.

sroy3 (changes requested):
I am not able to remove some rows. I get the error

```
[object Object]]Command failed with exit code 255: /Users/stephroy/Dev/vscode-dvc/demo/.env/bin/python -m dvc exp remove 9af3ebde1f370f46d2ca97db8b506cf2346b8648
ERROR: '9af3ebde1f370f46d2ca97db8b506cf2346b8648' are not valid experiment names
```

https://user-images.githubusercontent.com/3683420/174798647-416e4fa0-54ec-4e27-a15a-80e6b70da608.mov

wolmir:
Thanks for the review guys! This will change a bit, we'll use checkboxes instead

shcheklein:
> This will change a bit, we'll use checkboxes instead

We still need to highlight the selected rows, right? Checkboxes change the trigger as far as I understand (and clearly we'll have to show them), but they don't change the major part of this PR unless I'm missing something?

shcheklein:
@wolmir yes, that's totally fine and even better to split this. It might happen that there will be no mechanism to multi-select anything for now. That's also fine, or we can implement some workaround just for testing purposes - e.g. select by clicking on the empty left side of the first column.

webview/src/experiments/components/table/Row.tsx
```diff
@@ -40,6 +44,91 @@ const experimentMenuOption = (
...
+    () => getMultiSelectMenuOptions(selectedRowsList),
+    () =>
+      getSingleSelectMenuOptions(
+        id,
+        isWorkspace,
+        projectHasCheckpoints,
+        depth,
+        queued
+      )
+  )
```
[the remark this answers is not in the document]
wolmir: Yes, and it needs a better name.

But when I use a ternary or if block it adds +2 to Sonar's and CodeClimate's complexity metrics, and this doesn't add anything.
I know that maybe I'm avoiding a root cause here, but I also think a function like this is useful because we can use it as an expression and compose it with other functions without worrying about the metrics.

wolmir:
@rogermparent @sroy3 Can you take another look here while I fix the test coverage, please?

webview/src/experiments/components/table/Cell.tsx
```diff
@@ -44,6 +46,16 @@ export const FirstCell: React.FC<
...
+      tabIndex={0}
+      aria-checked={bulletColor !== undefined}
+      role="switch"
+      {...clickAndEnterProps(toggleExperiment)}
+      // onClick={toggleExperiment}
+      // onKeyDown={e => {
+      //   if (e.key === 'Enter' || e.key === ' ') {
+      //     toggleExperiment(e)
+      //   }
+      // }}
```
sroy3, line 58: Don't forget to remove the commented out code.

sroy3 (approved):
It took me a little while to understand that to select a row, you can click anywhere BUT the experiment column.For the select/deselect of the experiment itself, I'd make the target be the text and bullet only and allow selecting the row in the empty space.

webview/src/experiments/components/table/Cell.tsx
wolmir: Weird, this should have been caught by the linter

wolmir:
> It took me a little while to understand that to select a row, you can click anywhere BUT the experiment column.For the select/deselect of the experiment itself, I'd make the target be the text and bullet only and allow selecting the row in the empty space.

Thanks! I'll implement it now, it's a quick fix

wolmir:
@rogermparent @sroy3 Can you take another look, please?

rogermparent (approved):
Looking pretty good now! One small nitpick: clicking the copy button in a cell will toggle selected status, but all we need to do to suppress that is add `e.stopPropagation` to the `CopyButton`'s click handler. I still don't like `cond` much, but I can live with it.

sroy3 (approved):
Don't forget to fix the broken test, otherwise, it looks good.

maxagin:
> One small nitpick: clicking the copy button in a cell will toggle selected status

Not anymore as we are removing the copy button as well we are implementing a different way of selecting rows: Selecting Experiments redefining #1924

PR #1907, opened 2022-06-15
Only show filtered counts if they are above 0

rogermparent:
This PR is a quick fix that handles the situation where a user doesn't have checkpoints enabled in their project but would previously see "No checkpoints filtered". Now no line is shown, which is essentially the same logic the filters tree uses in its description.

https://user-images.githubusercontent.com/9111807/173927009-73b572a1-d1fc-4cfa-bce8-f67c2aa7ca94.mp4

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -784,7 +784,7 @@ describe('App', () => {
 
     const tooltip = screen.getByRole('tooltip')
 
-    expect(tooltip).toHaveTextContent('No filters applied')
+    expect(tooltip.textContent).toBe('No filters applied')
 
     const { columns } = tableDataFixture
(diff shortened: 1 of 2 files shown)


webview/src/experiments/components/App.test.tsx
```diff
@@ -784,7 +784,7 @@ describe('App', () => {
 
     const tooltip = screen.getByRole('tooltip')
 
-    expect(tooltip).toHaveTextContent('No filters applied')
+    expect(tooltip.textContent).toBe('No filters applied')
```
rogermparent, line 787: This change effectively makes the test check that _only_ the given text is there, whereas `toHaveTextContent` will pass even if there is other text present.

mattseddon:
[Q] can we end up with an empty tooltip?

PR #1915, opened 2022-06-16
Improve reset the workspace SCM command

mattseddon:
# 1/2 `main` <- this <- #1916

The second attempt at #1900 which was reverted in #1910 (reasons for the revert are inline).

I will write up the incident.


extension/src/git.ts
```diff
@@ -51,6 +51,17 @@ export const getAllUntracked = async (
   return new Set([...files, ...dirs])
 }
 
+export const getHasChanges = async (
+  repositoryRoot: string
+): Promise<boolean> => {
+  const output = await executeProcess({
+    args: ['diff', '--name-status', '-z'],
```
mattseddon, line 58: [F] The first bug was here. `git status` touches `.git/index` which fired the watcher which ran `git status` which...

extension/src/repository/index.ts
```diff
@@ -45,7 +48,10 @@ export class Repository extends DeferredDisposable {
   }
 
   public getState() {
-    return this.model.getState()
+    return {
+      ...this.model.getDecorationState(),
+      ...this.model.getSourceControlManagementState()
```
mattseddon, line 53: [F] This function is only for testing. When I rewrite this I will get rid of the need for it.

extension/src/repository/index.ts
```diff
@@ -89,7 +95,9 @@ export class Repository extends DeferredDisposable {
   }
 
   private setState() {
-    this.sourceControlManagement.setState(this.getState())
-    this.decorationProvider.setState(this.getState())
+    this.sourceControlManagement.setState(
+      this.model.getSourceControlManagementState()
+    )
+    this.decorationProvider.setState(this.model.getDecorationState())
```
mattseddon, line 101: [F] Bug two. 

Was a great example of trying to do something clever that turned out to be incredibly dumb.

Previous code:

```
export type DecorationState = Record<Status, Set<string>>

export interface DecorationModel {
  getState: () => DecorationState
}

enum Status {
  ADDED = 'added',
  DELETED = 'deleted',
  MODIFIED = 'modified',
  NOT_IN_CACHE = 'notInCache',
  RENAMED = 'renamed',
  GIT_MODIFIED = 'gitModified',
  TRACKED = 'tracked'
}

export class DecorationProvider
  extends Disposable
  implements FileDecorationProvider
{
...

  public setState(state: DecorationState) {
    const urisToUpdate = this.getUnion(this.state, state)
    this.state = state
    this.decorationsChanged.fire(urisToUpdate)
  }

  private getUnion(existingState: DecorationState, newState: DecorationState) {
    return flattenUnique([
      ...Object.values(existingState).map(status => [...(status || [])]),
      ...Object.values(newState).map(status => [...(status || [])])
    ]).map(path => Uri.file(path))
  }
```

```
export class RepositoryModel
  extends Disposable
  implements DecorationModel, SourceControlManagementModel
{
...
  public getState() {
    const acc = []
    for (const relTrackedOut of this.relTrackedOuts) {
      acc.push(join(this.dvcRoot, relTrackedOut))
    }

    return {
      ...omit(this.state, ['trackedLeafs', 'trackedNonLeafs']),
      hasRemote: new Set([...acc, ...this.state.trackedLeafs]),
      tracked: this.getTracked()
    }
  }
```

The way that this was setup meant that the compiler couldn't/didn't warn me that I was putting a boolean into `Uri.file` which causes a silently rejected promise and stopped any decorations from being provided.

I've changed the code to be very explicit in terms of what the `DecorationProvider` and `SourceContorlManagement` expect for state.

extension/src/git.ts
```diff
@@ -51,6 +51,17 @@ export const getAllUntracked = async (
   return new Set([...files, ...dirs])
 }
 
+export const getHasChanges = async (
+  repositoryRoot: string
+): Promise<boolean> => {
+  const output = await executeProcess({
+    args: ['diff', '--name-only', '-z'],
```
mattseddon, line 58: [F] The first bug was here. git status touches .git/index which fired the watcher which ran git status which...

extension/src/repository/index.ts
```diff
@@ -45,7 +48,10 @@ export class Repository extends DeferredDisposable {
   }
 
   public getState() {
```
mattseddon, line 50: I'm going to follow up and remove this. It is only used for testing purposes. This change is big enough now though.

sroy3 (approved):

rogermparent (approved):

Issue #1919, opened 2022-06-16
Windows: check all file watchers work as expected.


mattseddon:
Follow up from #1904.

Need a Windows cloud instance to test.

This should be part of cross-platform end-to-end testing.

mattseddon:
https://github.com/microsoft/vscode-discussions/discussions/9

mattseddon:
Saw this whilst running an experiment in the demo project:

```
ERROR: Error running 'checkpoint' task, 'stage: 'train'' will be aborted - [WinError 5] Access is denied: 'C:\\Code\\vscode-dvc\\demo\\misclassified.jpg'
SUCCESS: The process with PID 1624 (child process of PID 7020) has been terminated.
SUCCESS: The process with PID 7020 (child process of PID 4984) has been terminated.
SUCCESS: The process with PID 4984 (child process of PID 6480) has been terminated.
```

mattseddon:
Things tested:

- [x] Running experiments - all the things update
- [x] Reset the workspace
- [x] Checkout of individual items
- [x] Making commits
- [x] Changing branches
- [x] Open params file from experiments webview

PR #1921, opened 2022-06-17
Fix DVC cwd for Windows

mattseddon:
Attempt to fix https://github.com/iterative/dvc/issues/7909 on our side.

The issue is that when we call the CLI it uses Python's `os.getcwd()` to get the directory that the command is being run in. This returns a case-sensitive path. It seems that by default Node (and VS Code) will always return a lower-case letter for a drive on Windows. As Windows is a case-sensitive file system this causes bizarre behaviour (see issue above). 

The fix is to use `fs.realpath.native` on the cwd before we send it to the CLI. I have tested it manually on both Mac and Windows and it seems to behave as expected. I would call this a patch and not a long-term solution.

### DIFF
--- a/extension/src/cli/cwd.ts
+++ b/extension/src/cli/cwd.ts
@@ -1,6 +1,6 @@
 import { realpathSync } from 'fs'
 
-export const getCwd = (path: string): string => {
+export const getCaseSensitiveCwd = (path: string): string => {
   try {
     const cwd = realpathSync.native(path)
     if (cwd) {
--- a/extension/src/cli/cwd.ts
+++ b/extension/src/cli/cwd.ts
@@ -1,0 +1,11 @@
+import { realpathSync } from 'fs'
+
+export const getCwd = (path: string): string => {
+  try {
+    const cwd = realpathSync.native(path)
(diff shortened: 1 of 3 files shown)


extension/src/cli/cwd.ts
```diff
@@ -0,0 +1,11 @@
+import { realpathSync } from 'fs'
+
+export const getCaseSensitiveCwd = (path: string): string => {
+  try {
+    const cwd = realpathSync.native(path)
```
mattseddon, line 5: [Docs](https://nodejs.org/api/fs.html#fsrealpathnativepath-options-callback).

mattseddon:
Relates to #1919

extension/src/cli/cwd.ts
```diff
@@ -0,0 +1,11 @@
...
+    if (cwd) {
+      return cwd
+    }
+  } catch {}
+  return path
```
mattseddon, line 10: [F] In this function I am trying to pre-empt everything that could go wrong when attempting to call `fs.realpathSync.native`.

rogermparent (approved):

PR #1935, opened 2022-06-21
Use vertical line as drop target indicator for plots

### LINKED ISSUE #1604: Improve drag and drop feedback in plots comparison table
This follows on from the work done in 
- #1543
- #1551 
- #1572 
- #1594 

The current feedback for dragging and dropping an entire column looks like this: 

https://user-images.githubusercontent.com/3683420/162769693-dff23987-2fc8-4af9-8e87-4cb2afecd2a7.mov

We received this feedback as part of the UX review in https://github.com/iterative/vscode-dvc/issues/1565:

image

We need to add further feedback to the column to let the user know which column they are currently working with.

Feedback [figmajam](https://www.figma.com/file/RkSYR0tMeYy9CDeU5mte8l/dvc-vscode-beta-review?node-id=0%3A1).

SkyLeite:
Fixes #1604

This was a tricky one to do initially. My first idea was to simply change the DropTarget to a vertical line and change the dragged column to not be hidden, but that didn't work because the vertical line was still part of the table flow, meaning it would consume the entire width of a header.

My current approach passes the header element itself to DropTarget so it can render the header as well as the drop indicators.

TODO:

- [ ] Fix not being able to drag a column to the first or last spot
- [ ] Fix tests

### DIFF
--- a/webview/src/plots/components/comparisonTable/DropTargetIndicator.tsx
+++ b/webview/src/plots/components/comparisonTable/DropTargetIndicator.tsx
@@ -1,0 +1,17 @@
+import React from 'react'
+import { DragEnterDirection } from '../../../shared/components/dragDrop/util'
+import styles from '../styles.module.scss'
+
+export const DropTargetIndicator: React.FC<Props> = ({ direction }) => (
+  <span
+    className={styles.dropTargetIndicator}
+    data-testid="comparison-drop-target-indicator"
+    style={{
+      [direction.toLowerCase()]: -4
+    }}
+  />
+)
+
+interface Props {
+  direction: DragEnterDirection
+}
--- a/webview/src/plots/components/styles.module.scss
+++ b/webview/src/plots/components/styles.module.scss
@@ -268,12 +268,24 @@ $gap: 20px;
 
 .dropSectionMaximized,
 .dropTarget {
+  position: relative;
   height: auto;
-  opacity: 0.5;
-  border: 3px dashed $accent-color;
+  width: 100%;
   display: flex;
   justify-content: center;
   align-items: center;
+
+  & > th {
+    display: block;
+    width: 100%;
+  }
+}
+
+.dropTargetIndicator {
+  position: absolute;
+  top: 2px;
+  min-height: 32px;
+  border: 2px dotted $accent-color;
 }
 
 .dropIcon {
--- a/webview/src/shared/components/dragDrop/DragDropContainer.tsx
+++ b/webview/src/shared/components/dragDrop/DragDropContainer.tsx
@@ -42,7 +42,8 @@ export const makeTarget = (
   dropTarget: DropTargetInfo,
   handleDragOver: DragEventHandler<HTMLElement>,
   handleOnDrop: DragEventHandler<HTMLElement>,
-  id: string
+  id: string,
+  props?: Partial<{ children: JSX.Element; direction: DragEnterDirection }>
 ) => (
   <dropTarget.wrapperTag
     data-testid="drop-target"
@@ -51,7 +52,7 @@ export const makeTarget = (
     onDrop={handleOnDrop}
     id={`${id}__drop`}
   >
-    {dropTarget.element}
+    {React.cloneElement(dropTarget.element, props)}
   </dropTarget.wrapperTag>
 )
 interface DragDropContainerProps {
@@ -126,7 +127,6 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
     })
     draggedOverIdTimeout.current = window.setTimeout(() => {
       setDraggedId(id)
-      setDraggedOverId(order[toIdx])
     }, 0)
   }
 
@@ -204,7 +204,7 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
       onDragEnter={handleDragEnter}
       onDrop={handleOnDrop}
       draggable={!disabledDropIds.includes(id)}
-      style={(id === draggedId && { display: 'none' }) || draggable.props.style}
+      style={draggable.props.style}
     />
   )
 
@@ -213,11 +213,12 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
     const item = id && buildItem(id, draggable)
 
     if (id === draggedOverId) {
-      const target = makeTarget(dropTarget, handleDragOver, handleOnDrop, id)
+      const target = makeTarget(dropTarget, handleDragOver, handleOnDrop, id, {
+        children: draggable,
+        direction
+      })
 
-      return direction === DragEnterDirection.RIGHT
-        ? [item, target]
-        : [target, item]
+      return target
     }
 
     return item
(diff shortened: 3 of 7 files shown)


sroy3:
> * Test `should move a template plot from one type in another section of the same type and show two drop targets` fails
>   I need some guidance for this one. I couldn't find the scenario it describes in storybook, and it also seems like it would no longer be relevant with the new Drop Target mechanism for plots, but I'm not entirely sure.

You modified the behaviour of drag and drop for other plots (Trends and Data Series), it should stay the same. If the test still fails after, this is shown here by the last drag in the video:


https://user-images.githubusercontent.com/3683420/174911520-aff1821c-5a9e-49b5-856e-dd3257e6d301.mov

sroy3:
There is a little bug when dropping on an invalid target. The column will stay dark:

https://user-images.githubusercontent.com/3683420/174911930-968bcd8b-ed0d-4ce0-9f89-adfc5b1137d3.mov

Other than that, your implementation looks great. Great job!

Like I said previously, be careful not to change the behaviour of drag and drop elsewhere (checkpoints and template plots and column headers in the table) and the PR will be pretty solid after that.

SkyLeite:
@sroy3 Thank you for the feedback and help! I managed to fix both issues, and even added back the border that was missing for the drop target of Trends and Data Series. As such I'm marking this as ready for review :)

rogermparent:
> * nvalid DOM tree
>       Due to the way the indicator is placed in the DOM, it generates a DOM tree that fails react-dom's check as a `div` is not supposed to be a child of `tr`. I can't think of a way to get around this, as the indicator needs to be a child of the row to be appropriately placed.

Does using `div`s with `display: table` instead of `tr`s get around this? If not, maybe we can re-create the styles with flexbox much like how `react-table` does so with the experiments table.

webview/src/shared/components/dragDrop/DragDropContainer.tsx
```diff
@@ -189,22 +195,30 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
...
+      onDragEnd={handleDragEnd}
       onDragOver={handleDragOver}
       onDragEnter={handleDragEnter}
       onDrop={handleOnDrop}
       draggable={!disabledDropIds.includes(id)}
-      style={(id === draggedId && { display: 'none' }) || draggable.props.style}
+      style={
+        (id === draggedId && hideDragged && { display: 'none' }) ||
+        draggable.props.style
+      }
```
rogermparent, line 221: I see it was like this before this PR, but is the value of this style prop not just a weirdly written ternary? i.e. `id === draggedId && hideDragged ? {display: 'none'} : draggable.props.style`

webview/src/plots/components/App.test.tsx
```diff
@@ -856,13 +856,10 @@ describe('App', () => {
...
-    expect(plotsWithDropTarget.map(plot => plot.id)).toStrictEqual([
-      'plot-drop-target',
-      plots[0].id,
-      plots[1].id,
-      plots[2].id
-    ])
+    const target = screen.getByTestId('drop-target')
+
+    // eslint-disable-next-line testing-library/no-node-access
+    expect(target.nextElementSibling?.id).toStrictEqual(plots[0].id)
```
rogermparent, line 862: Could these just compare the elements' references? i.e. `expect(target.nextElementSibling).toBe(plots[0])`

Not super important, just makes things a tiny bit simpler. This understandably apes the existing implementation, so it's not like it comes from nowhere.

webview/src/plots/components/comparisonTable/DropTargetIndicator.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { DragEnterDirection } from '../../../shared/components/dragDrop/util'
+import styles from '../styles.module.scss'
+
+export const DropTargetIndicator: React.FC<Props> = ({ direction }) => (
+  <span
+    className={styles.dropTargetIndicator}
+    data-testid="comparison-drop-target-indicator"
+    style={{
+      [direction.toLowerCase()]: -4
```
rogermparent, line 10: Using the transformed value of an enum directly as a string makes me a little uneasy, though it's not like avoiding doing so makes the code that much more elegant.

webview/src/plots/components/styles.module.scss
```diff
@@ -268,12 +268,28 @@ $gap: 20px;
...
-  border: 3px dashed $accent-color;
+  width: 100%;
   display: flex;
   justify-content: center;
   align-items: center;
+
+  & > th {
+    display: block;
+    width: 100%;
+  }
```
rogermparent, line 281: Does this still apply to anything? It seems like the drop target was turned into a `div` in another part of the code, and changing a `th`'s `display` seems like it would cause a lot more chaos if it were applying.

rogermparent (commented):
Great work so far! I've found a couple tiny nitpicks from a first read-over, but it seems like all the big issues were caught by Stephanie and you've already handled them! I'll be taking another look early tomorrow.

webview/src/shared/components/dragDrop/DragDropContainer.tsx
sroy3: Not sure in this particular case, but I think our linter favours this syntax over ternaries using `?` and `:` sometimes.

sroy3 (commented):
The drag and drop elsewhere is still not quite how it was before:

https://user-images.githubusercontent.com/3683420/175039449-0b8a7248-16fa-45cd-81c4-46f69ee9d034.mov

There is a double border when the drop target is at the start:

Screen Shot 2022-06-22 at 9 07 51 AM

Did you try changing the style of the dragged element like the mock? Was it too complicated?

webview/src/plots/components/styles.module.scss
SkyLeite: Good catch. It is applying because the current implementation re-renders the dragged cell as the drop target, which in this case is a `th`. The `display: block` is indeed unnecessary, but the `width: 100%` is there so the cell expands to fill the drop target.

webview/src/shared/components/dragDrop/DragDropContainer.tsx
SkyLeite: Agreed. I changed it to a ternary and ESLint seemed to be fine with it in this case.

webview/src/plots/components/App.test.tsx
SkyLeite: Good point. I agree with comparing by reference instead of id 👍

webview/src/plots/components/comparisonTable/DropTargetIndicator.tsx
SkyLeite: Something I've seen done in Elm on cases like this, since in Elm you can't derive a string from an enum, is a function that explicitly returns a string from an enum value. I thought about doing the same thing here, but it gave me "early optimization" vibes. Does team have any ideas or precedents in the code base for tackling this?

webview/src/plots/components/App.test.tsx
```diff
@@ -856,13 +856,10 @@ describe('App', () => {
...
     dragEnter(plots[1], plots[0], DragEnterDirection.LEFT)
 
-    const plotsWithDropTarget = screen.getAllByTestId(/^plot_/)
-    expect(plotsWithDropTarget.map(plot => plot.id)).toStrictEqual([
-      'plot-drop-target',
-      plots[0].id,
-      plots[1].id,
-      plots[2].id
-    ])
+    const target = screen.getByTestId('drop-target')
```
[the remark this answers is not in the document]
SkyLeite: Agreed. This should've been reverted in ba9c2844e5d03f6154fa5a761b73546ffe83758c along with the other changes. It's done now.

PR #1942, opened 2022-06-23
Remove problematic tests

wolmir:


sroy3 (commented):
Are they flaky or just not needed anymore?

PR #1943, opened 2022-06-24
Fix CI cross-platform tests

wolmir:


wolmir:
Yep, seems to be fixed now

webview/src/experiments/components/App.test.tsx
```diff
@@ -691,7 +695,11 @@ describe('App', () => {
       const target = screen.getByTestId('workspace-row')
       fireEvent.contextMenu(target, { bubbles: true })
 
-      const menu = await screen.findByTestId('messages-menu')
+      const menu = await screen.findByTestId(
```
sroy3, line 698: `findBy` [queries](https://testing-library.com/docs/queries/about#types-of-queries):
> The promise is rejected if no element is found or if more than one element is found after a default timeout of 1000ms.

It means that the options aren't shown for one whole second after the event is fired. That seems like a lot. I feel like there might be an underlying problem here.

wolmir: I imagine it's because the CI machine is not very powerful and the tests are run in parallel.
I only have a Mac here, and the manual tests aren't slow at all.

I'll try with a smaller timeout and see if it works still

sroy3: Still 1 second is very long to wait for an element to appear.
The tooltip has a delay of 100ms. You probably just need to add `jest.advanceTimersByTime(100)` just before the query.

wolmir: Thank you! I'll try that

sroy3 (approved):

PR #1944, opened 2022-06-24
Use Redux for drag and drop state

sroy3:
Simple replacement of the `DragDropProvider` for a Redux state. Turns out the experiments table does not use `DragDropContainer` at all so it's fine to remove the context for Redux in there.


rogermparent:
Comparison table stories are broken both on Chromatic and locally, maybe they need to be wrapped with Redux?
image

sroy3:
> Comparison table stories are broken both on Chromatic and locally, maybe they need to be wrapped with Redux? image

The previous merge should have fixed this, but I'll check.

rogermparent (approved):
DnD works well, and the storybook issue is fixed for me too!

PR #1949, opened 2022-06-27
Patch plots for branches containing path separators

mattseddon:


extension/src/plots/model/index.ts
```diff
@@ -168,9 +173,9 @@ export class PlotsModel extends ModelWithPersistence {
       ...Object.keys(this.revisionData)
     ])
 
-    return this.getSelectedRevisions().filter(
-      revision => !cachedRevisions.has(revision)
-    )
+    return this.getSelectedRevisions()
+      .filter(label => !cachedRevisions.has(label))
+      .map(label => this.getCLIId(label))
```
mattseddon, line 178: [F] Whenever we get the missing revisions to call the CLI with we replace the branch label with the short sha.

extension/src/plots/model/index.ts
```diff
@@ -96,10 +96,15 @@ export class PlotsModel extends ModelWithPersistence {
   }
 
   public async transformAndSetPlots(data: PlotsOutput, revs: string[]) {
-    this.fetchedRevs = new Set([...this.fetchedRevs, ...revs])
+    const cliIdToLabel = this.getCLIIdToLabel()
```
mattseddon, line 99: [F] When we get the data back from the CLI we substitute the label for the short sha that we provided.

extension/src/plots/model/index.ts
```diff
@@ -330,6 +339,20 @@ export class PlotsModel extends ModelWithPersistence {
...
+
+    for (const rev of this.getSelectedRevisions()) {
+      mapping[this.getCLIId(rev)] = rev
+    }
+
+    return mapping
+  }
+
+  private getCLIId(label: string) {
+    return this.branchRevisions[label] || label
```
mattseddon, line 353: [F] At the moment this is the only substitution that we do but we can extend this if we have to.

extension/src/plots/data/index.ts
```diff
@@ -72,4 +71,15 @@ export class PlotsData extends BaseData<{ data: PlotsOutput; revs: string[] }> {
   public setModel(model: PlotsModel) {
     this.model = model
   }
+
+  private getArgs(revs: string[]) {
+    if (
+      this.model &&
+      (sameContents(revs, ['workspace']) || sameContents(revs, []))
+    ) {
+      return this.model.getDefaultRevs()
```
mattseddon, line 80: [F] We can no longer call the CLI with no arguments because the default behaviour is to return the `workspace` & branch name as the revisions. The rest of the code will then fail.

shcheklein:
@mattseddon https://github.com/iterative/dvc/issues/7934 got fixed - can we revert some of the changes here?

Issue #1676, opened 2022-05-08
Star experiments in table


maxagin:
Description: We want to be able to label/mark a single experiment in the table to be able to find it easter.
Original inspiration - gmail stars.
This task is a part of the bigger story #1720

### User story

Iterating on get started repo recently I got lost quickly with all the combinations. Since the dataset name is the same, size in that case was not signalling to me much, etc. I had "good enough" iterations, but it was impossible to put some label, pin, group, save filter, etc.

### TODO

- [x] Be able to add and remove the mark/s (star) from exp/s
- [x]  Checkpoints can have stars.
- [x]  Select a few checkpoints to remove stars.
- [x] Horizontal scroll. To keep the “stars” section (and exp title?) fixed at the left side
- [x]  If a row with checkpoints is folded, we need a sort of way to show that there are checkpoints with stars hidden
- [ ] quick pick - show quick find a starred experiment and select it in the table
- [ ] quick pick - start / unstar. Possible scenario: show quick find exp + star this exp
- [x] Check how does it impact other panels, quick pick, etc? 

[Figma specs](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/1720-%E2%80%A2-Story%3A-Help-navigate-in-a-lot-of-experiments?node-id=808%3A21347)

wolmir:
@maxagin Regarding this comment, what do you mean by "modify the menu options"?

![Screen Shot 2022-06-20 at 14 34 55](https://user-images.githubusercontent.com/1231848/174653871-0808c19b-7cc6-4358-a301-66beb124f1d2.png)

shcheklein:
@maxagin it means that we'll need checkboxes for nested experiments?

shcheklein:
> for example, include actions with the stars
>  for the checkpoints as there are no actions that can be taken

looks like a contradiction? Or I'm missing something ...

maxagin:
You are right. This is my bad, this is the only action (in the context menu), for now, that can be associated with the stars and yes if we want to remove multiple stars by selecting rows, then the checkboxes will be in every row of the table. Great !

maxagin:
>Check how does it impact other panels, quick pick, etc?

@shcheklein  I do not see how this may impact the quick pick or the sidebar. Am I missing something?
@wolmir any ideas in regards to this?

shcheklein:
Some ideas:

- show stars / or highlight in some other way starred experiments in the table
[...]

PR #1950, opened 2022-06-27
Star experiments in table

wolmir:
This is the first PR for #1676 

This PR will introduce the Star Experiments feature to the experiments table webview.

This option will be available in the form of star icons in the first cells, as well as context-menu options.

Demo:

https://user-images.githubusercontent.com/1231848/176021943-5f744280-3eb3-4f3c-833a-3bee0948d84e.mov


rogermparent (changes requested):
There's an issue with the context menu where invoking it on a multi-select that contains both starred and unstarred items will create an empty context menu:

https://user-images.githubusercontent.com/9111807/176047323-f3b19880-a1b4-45c1-88a4-96e3bc59d12d.mp4

Beyond that, I think we'll want to show star/unstar options in this situation, especially if we add more multi-select options like selecting ranges or groups. 
I think the most useful logic would be simply showing group star if any are unstarred and group unstar if any are starred, allowing for both to be present. Having a single option would make it very inconvenient for users to use the option that has less priority, especially where the selection clears when any option is taken.

extension/src/experiments/model/index.ts
```diff
@@ -108,6 +114,19 @@ export class ExperimentsModel extends ModelWithPersistence {
...
   }
 
+  public toggleStars(ids: string[]) {
+    for (const id of ids) {
+      const current = this.starredExperiments[id]
+
+      if (current) {
+        delete this.starredExperiments[id]
+      } else {
+        this.starredExperiments[id] = STARRED
```
mattseddon, line 124: [Q] Do you think there will be some in-between state in the future? Why not start with a boolean value?

extension/src/experiments/model/index.ts
```diff
@@ -505,10 +528,30 @@ export class ExperimentsModel extends ModelWithPersistence {
...
+  private addStarred(experiment: Experiment) {
+    const { id } = experiment
+
+    if (!this.isStarred(id)) {
+      return experiment
+    }
+
+    return {
+      ...experiment,
+      starred: true
```
mattseddon, line 551: [C] Changing to a boolean could simplify this method.

extension/src/experiments/model/index.ts
```diff
@@ -524,6 +567,10 @@ export class ExperimentsModel extends ModelWithPersistence {
     }
   }
 
+  private addSelectedAndStarred(experiment: Experiment) {
+    return this.addStarred(this.addSelected(experiment))
```
mattseddon, line 571: [Q] Is `addSelected` used anymore? Can we combine these methods?

mattseddon: rename to `addDetails`?

webview/src/experiments/components/table/Row.tsx
```diff
@@ -98,6 +127,18 @@ const getSingleSelectMenuOptions = (
...
       MessageFromWebviewType.VARY_EXPERIMENT_PARAMS_AND_QUEUE,
       !isNotCheckpoint
+    ),
+    experimentMenuOption(
+      [id],
+      starred ? 'Unstar Experiment' : 'Star Experiment',
+      MessageFromWebviewType.TOGGLE_EXPERIMENT_STAR,
+      isWorkspace
+    ),
+    withId(
```
mattseddon, line 137: [Q] Should there be dividers between the VARY_AND_ commands and toggle star and also remove?

webview/src/experiments/components/App.test.tsx
```diff
@@ -750,7 +750,8 @@ describe('App', () => {
       expect(itemLabels).toStrictEqual([
         'Modify and Resume',
         'Modify, Reset and Run',
-        'Modify and Queue'
+        'Modify and Queue',
+        'Star Experiment'
       ])
     })
```
mattseddon, line 756: [Q] New test for new behavior?

extension/src/experiments/model/index.ts
```diff
@@ -81,6 +83,10 @@ export class ExperimentsModel extends ModelWithPersistence {
       PersistenceKey.EXPERIMENTS_STATUS,
       {}
     )
+    this.starredExperiments = this.revive<StarredExperiments>(
```
mattseddon, line 86: [Q] What will happen if we move between branches/commits and experiments are still starred? I can't see anything that would ever flush stale experiments out of this list. An ever-expanding object is something that we should try to avoid.

shcheklein: I don't see a better option tbh, any ideas? I hope this list won't be growing too fast?

It should be probably split by commits for faster access though.

wolmir:
@rogermparent @maxagin 

What about something like this for heterogeneous rows?

https://user-images.githubusercontent.com/1231848/176250520-3d656e54-69d0-481a-b8cd-14c816100807.mov

wolmir:
@rogermparent @mattseddon Can you re-review this please?

rogermparent (approved):
The issue I found was handled nicely, and from a user standpoint feature works enough that I'd consider it good enough to approve. I'm not sure about the other concerns, those may have been handled in the meeting but otherwise we may want to wait for Matt's approval before merging.

extension/src/experiments/model/index.ts
mattseddon: [N] Could be 

```
return { ...this.addSelected(experiment), starred: !!this.isStarred(id) }
```

mattseddon (approved):

mattseddon:
@wolmir after you merge this tomorrow please make a release to the marketplace 👍🏻.

PR #1953, opened 2022-06-28
Show errors when interacting with the CLI from the experiments webview or tree

### LINKED ISSUE #1934: While I was trying to create a new branch in order to answer Max the command failed because the branch name was already taken.
While I was trying to create a new branch in order to answer Max the command failed because the branch name was already taken.
But the extension didn't report it.
Maybe it's a bug?

_Originally posted by @wolmir in https://github.com/iterative/vscode-dvc/issues/1924#issuecomment-1160668339_

mattseddon:
Closes #1934.

This took a lot more effort than I anticipated. Mostly because the code has evolved/diverged from a consolidated approach over time. I have consolidated commands and registered them all in a central location. This gives us a few benefits:

1. Any CLI failures will result in a toast message being shown with a link to the output channel.
2. Any CLI output from user-initiated actions will be shown in toast messages, e,g `Queued experiment '6cbc21f' for future execution.`.
3. Data updates will be paused when experiments are started/queued using the modify and... path.
4. We know that commands have telemetry events attached.

One drawback is that the code is not always linear and it may not be obvious exactly where a command has been registered. Another is that the telemetry events will not currently tell us the location a command has been invoked from. This is something that we can live with given the lack of maturity around what we do with the telemetry data.

### Demo (original issue)

https://user-images.githubusercontent.com/37993418/176356560-fd3b09ce-e2ff-4e1e-b677-bf5ddf1a315e.mov

**Note: I will be manually testing this tomorrow. Please take a look at the approach and LMK what you think.**


extension/src/commands/external.ts
```diff
@@ -11,6 +11,14 @@ export enum RegisteredCliCommands {
   EXPERIMENT_RESET_AND_RUN = 'dvc.resetAndRunCheckpointExperiment',
   QUEUE_EXPERIMENT = 'dvc.queueExperiment',
 
+  EXPERIMENT_TREE_APPLY = 'dvc.views.experimentsTree.applyExperiment',
+  EXPERIMENT_TREE_BRANCH = 'dvc.views.experimentsTree.branchExperiment',
+  EXPERIMENT_TREE_QUEUE = 'dvc.views.experimentsTree.queueExperiment',
+  EXPERIMENT_TREE_REMOVE = 'dvc.views.experimentsTree.removeExperiment',
+  EXPERIMENT_TREE_RESUME = 'dvc.views.experimentsTree.resumeCheckpointExperiment',
+  EXPERIMENT_TREE_RUN = 'dvc.views.experimentsTree.runExperiment',
+  EXPERIMENT_TREE_RESET_AND_RUN = 'dvc.views.experimentsTree.resetAndRunCheckpointExperiment',
```
mattseddon, line 20: These are problematic. We need to call the CLI from inside experiments after we get the params. There needs to be two different behaviours. One for the runner and one for the executor (QUEUE)

extension/src/experiments/index.ts
```diff
@@ -384,14 +389,24 @@ export class Experiments extends BaseRepository<TableData> {
...
+      return
+    }
+
+    const promise = this.internalCommands.executeCommand<string>(
+      commandId,
+      this.dvcRoot,
+      ...paramsToModify
+    )
+    if (commandId === AvailableCommands.EXPERIMENT_QUEUE) {
+      Toast.showOutput(promise)
```
mattseddon, line 407: Needs split

extension/src/commands/external.ts
mattseddon: We want to offer to show the error when trying to Queue but not when trying to run in the runner.

extension/package.json
```diff
@@ -409,39 +409,12 @@
...
-        "command": "dvc.views.experimentsTree.resumeCheckpointExperiment",
-        "category": "DVC",
-        "icon": "$(debug-continue-small)"
-      },
-      {
-        "title": "%command.views.experimentsTree.resetAndRunCheckpointExperiment%",
-        "command": "dvc.views.experimentsTree.resetAndRunCheckpointExperiment",
-        "category": "DVC",
-        "icon": "$(play)"
-      },
```
mattseddon, line 444: [F] For these 4 commands we need to pause updates. Which means that have to go through the `WorkspaceExperiments` path

mattseddon: One downside of this is that the context menu titles blow out:

image

But an upside is that the correct toasts are always shown if there is output or an error that the user needs to see.

extension/src/experiments/index.ts
```diff
@@ -609,28 +633,6 @@ export class Experiments extends BaseRepository<TableData> {
     return this.notifyChanged()
   }
 
-  private async createBranchFromExperiment(experimentId: string) {
```
mattseddon, line 612: [F] None of these had telemetry events

rogermparent (approved):
I like the more `executeCommand`-centric approach, maybe we could add an argument to tell analytics where the command is being fired from?

mattseddon:
> I like the more `executeCommand`-centric approach, maybe we could add an argument to tell analytics where the command is being fired from?

Good future consideration

mattseddon:
Did a lot of manual testing to confirm the context menus/actions still work as expected.

PR #1965, opened 2022-07-01
Upgrade React to v18

### LINKED ISSUE #1716: Upgrade React to v18
This will let us upgrade all of our `@testing-library` dependencies to the latest version.

🐉: there are some breaking changes with the `children` property no longer being implicitly declared.

mattseddon:
Closes #1716.

"It's just a one-line change"


package.json
```diff
@@ -63,9 +63,10 @@
     "prettier-config-standard": "^5.0.0",
     "ts-node": "^10.8.1",
     "turbo": "^1.3.1",
-    "typescript": "^4.6.3"
+    "typescript": "^4.7.4"
   },
   "resolutions": {
-    "trim": "0.0.3"
+    "trim": "0.0.3",
+    "@types/react": "^18.0.14"
```
mattseddon, line 70: [F] This does scare me but without it we cannot upgrade (webview toolkit and react-virtualization have incompatible types).

webview/src/experiments/components/table/Table.test.tsx
```diff
@@ -318,25 +318,36 @@ describe('Table', () => {
...
-        clientX: 0
+        bubbles: true
+      })
+      fireEvent.mouseUp(experimentColumnResizeHandle)
+
+      expect(mockedPostMessage).toBeCalledWith({
+        payload: { id: 'id', width: 333 },
+        type: MessageFromWebviewType.RESIZE_COLUMN
       })
-      fireEvent.mouseMove(document, {
```
mattseddon, line 336: [F] Seems like this no longer works because the DOM has no size.

mattseddon: Test is changed to accommodate not being a able to move the mouse (still tests the functionality just less directly)

webview/src/experiments/util/buildDynamicColumns.tsx
```diff
@@ -107,16 +107,15 @@ const buildDynamicColumns = (
...
         Cell,
         Header,
         accessor: pathArray && buildAccessor(pathArray),
         columns: childColumns.length > 0 ? childColumns : undefined,
         group: type,
         id: path,
         name: label
-      }
-      return column
+      } as ColumnGroup<Experiment> | TableColumn<Experiment>
```
mattseddon, line 118: [F] Another type error, I have manually tested that the table works as expected but I'd appreciate everyone trying it out.

webview/src/plots/components/App.test.tsx
```diff
@@ -769,8 +770,13 @@ describe('App', () => {
 
     expect(topDropIcon).not.toBeInTheDocument()
 
-    multiViewPlot.dispatchEvent(createBubbledEvent('dragstart'))
-    topSection.dispatchEvent(createBubbledEvent('dragenter'))
+    act(() => {
+      multiViewPlot.dispatchEvent(createBubbledEvent('dragstart'))
+    })
```
mattseddon, line 775: all of these dispatched events now need to be wrapped in act

package.json
mattseddon: Our use of the toolkit still seems to work. I've check the virtualized plots story in storybook and it seems to render just fine. Drag and drop is a bit janky.

mattseddon, line 70: [F] This does scare me but without it we cannot upgrade (webview toolkit and react-virtualization have incompatible types). 

Our use of the toolkit still seems to work. I've checked the virtualized plots story in storybook and it seems to render just fine. Drag and drop is a bit janky.

.github/workflows/continuous-integration.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           coverageCommand: xvfb-run -a yarn run cover
           coverageLocations: ${{github.workspace}}/coverage/lcov.info:lcov
 
-      - run: yarn build-storybook
+      - run: yarn build-test-storybook
```
mattseddon, line 48: [F] The visual components with any `act` behaviour failed to render on Chromatic because `act(…) is not supported in production version of React`. I have added another addon and set the NODE_ENV to test before building the storybook.

webview/src/experiments/components/table/Table.test.tsx
mattseddon, line 336: [F] This event no longer does anything. I think it is because the elements no longer have any size.

mattseddon: Test is changed to accommodate not being a able to move the mouse (still tests the functionality just less directly)

webview/src/experiments/util/buildDynamicColumns.tsx
mattseddon, line 118: [F] Another type error, I have manually tested that the table works as expected but I'd appreciate everyone trying it out.

webview/src/stories/Plots.stories.tsx
```diff
@@ -208,20 +211,22 @@ MultiviewZoomedInPlot.parameters = {
...
-  chromatic: { delay: 500 }
+  chromatic: { delay: 1000 }
 }
 CheckpointZoomedInPlot.play = async ({ canvasElement }) => {
   const canvas = within(canvasElement)
   const plot = await canvas.findByText('summary.json:val_accuracy')
 
-  plot.scrollIntoView()
+  await waitFor(() => {
+    plot.scrollIntoView()
```
mattseddon, line 228: [F] Waiting for this is what has changed the visual test.

webview/src/stories/Plots.stories.tsx
```diff
@@ -208,20 +211,7 @@ MultiviewZoomedInPlot.parameters = {
 MultiviewZoomedInPlot.play = async ({ canvasElement }) => {
   const canvas = within(canvasElement)
   const plot = await canvas.findByTestId('plots-section_template-multi_1')
-  const plotButton = within(plot).getByRole('button')
+  const plotButton = await waitFor(() => within(plot).getByRole('button'))
 
   fireEvent.click(plotButton)
 }
-
-export const CheckpointZoomedInPlot = Template.bind({})
```
mattseddon, line 216: [F] This test started flaking out.

webview/src/experiments/util/buildDynamicColumns.tsx
mattseddon: same for plots

.github/workflows/continuous-integration.yml
rogermparent: Why would the storybook build be trying to compile something with `act`? Are we importing test modules in our stories?

rogermparent: Just looked into it and seems it's only the use of `fireEvent` in stories that's the problem, they render correctly without this script if you replace `fireEvent.click(el)` with `el.click()` and we don't have to deploy a big ol' test build for every chromatic. 
I'll leave whatever solution you prefer up to you, but using `testing-library` utils for the fake dom in the real dom of storybook does feel weird to me.

mattseddon: I should have left a better helper comment here.

I will try that. We need to be aware that not having the events wrapped in `act` could make the test flaky.

mattseddon: That might not actually be true because we use the delay option to control when we take the screenshot.

wolmir (approved):
Experiments table work as expected, as far as I tested it out

PR #1967, opened 2022-07-02
Small table cell tooltip changes

### LINKED ISSUE #1954: Small issues with the table tooltips
- [x] Remove prefixes (Metrics, Param) - they are not needed yet in our case
- [x] Can we try to keep tooltips if I hover on it?
- [ ] Bug with booleans (check strings, arrays?): 

Screen Shot 2022-06-27 at 7 31 54 PM

rogermparent:
Fixes #1954

https://user-images.githubusercontent.com/9111807/176987842-41bc3542-e4ab-469e-b0dc-8229c154c78c.mp4

- [x] Remove prefixes (Metrics, Param) - they are not needed yet in our case.
- [x] Table cell tooltips have `interactive` set on them, so users can do actions like copying text. Currently has a bug where clicking the tooltip in any way registers as a row click for selection.
- [x] Tooltip values are cast to `String`, confirmed booleans, arrays, and 0 work.

note: seems like `null`, even when put into yaml, breaks table parsing.


webview/src/experiments/util/buildDynamicColumns.tsx
```diff
@@ -58,10 +38,11 @@ const Cell: React.FC<Cell<Experiment, string | number>> = cell => {
   return (
     <Tooltip
       className={tooltipStyles.padded}
-      content={<CellTooltip cell={cell} />}
+      content={String(value)}
```
rogermparent, line 41: We'll need a component when we display anything more complex, but with this specific implementation for label-less tooltips it's never been less justified to have a separate component.

rogermparent: Turns out "anything more complex" also includes adding `stopPropagation` so selection clicks don't register as row clicks.

webview/src/shared/components/tooltip/styles.module.scss
```diff
@@ -17,7 +17,7 @@ $font: var(--vscode-font-family);
   white-space: nowrap;
   & > :global(.tippy-arrow) {
     color: $background;
-    z-index: 2;
+    z-index: 1;
```
rogermparent, line 20: Fixes a small issue with the arrow appearing over text highlighting in content

shcheklein:
@rogermparent is it still a draft?

wolmir (approved):

rogermparent:
I wanted to write some real tests for the other tooltip data types, but I could do that as an immediate follow-up as well to get this fix in since it's at better tested and working than what we currently have.

wolmir (approved):
I'm ok with adding the tests as a follow-up

mattseddon:
Please do follow up.

PR #1970, opened 2022-07-04
Increase timeout of integration tests again

mattseddon:
Have seen more intermittent failures creeping in recently from timeouts.

### DIFF
--- a/extension/src/test/suite/experiments/index.test.ts
+++ b/extension/src/test/suite/experiments/index.test.ts
@@ -364,7 +364,7 @@ suite('Experiments Test Suite', () => {
         { path: mockColumnId },
         undefined
       )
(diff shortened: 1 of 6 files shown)


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -824,7 +824,7 @@ suite('Experiments Test Suite', () => {
         'experiments have been starred'
       ).to.be.true
     })
-  })
+  }).timeout(WEBVIEW_TEST_TIMEOUT)
```
mattseddon, line 827: [F] All of these tests open a webview which means that they need the extended timeout.

wolmir (approved):

PR #1969, opened 2022-07-04
More dependency updates

mattseddon:
Clearing down some more dependency updates from #1938

PR #1971, opened 2022-07-04
Use create root to instantiate React apps

mattseddon:
Follow up from https://github.com/iterative/vscode-dvc/pull/1969#issuecomment-1173894768

https://stackoverflow.com/questions/71668256/deprecation-notice-reactdom-render-is-no-longer-supported-in-react-18

- [x] Build VSIX and manually test

### DIFF
--- a/extension/src/webview/index.ts
+++ b/extension/src/webview/index.ts
@@ -140,20 +140,22 @@ export class BaseWebview<T extends WebviewData> {
 
     // TODO make CSP more strict!
     return `
-			  <html>
-				  <head>
-				  <meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


extension/src/webview/index.ts
```diff
@@ -140,20 +140,22 @@ export class BaseWebview<T extends WebviewData> {
...
+          <head>
+          <meta charset="UTF-8">
+          <meta http-equiv="Content-Security-Policy" content="default-src * 'unsafe-inline' 'unsafe-eval'; script-src ${this.webviewPanel.webview.cspSource} * 'unsafe-inline' 'unsafe-eval'; connect-src * 'unsafe-inline'; img-src https: ${this.webviewPanel.webview.cspSource}; frame-src *; style-src * 'unsafe-inline'; worker-src * data: blob: data: 'unsafe-inline' 'unsafe-eval'; font-src * 'unsafe-inline' 'unsafe-eval' 'self' data: blob:;">
+          <style>
+            html { height: 100%; width: 100%; padding: 0; margin: 0; }
+            body { height: 100%; width: 100%; padding: 0; margin: 0; }
+          </style>
+          </head>
+          <body>
+            <div id="root">
```
mattseddon, line 153: [F] This is the actual change. The rest is converting tabs to spaces.

wolmir (approved):

webview/src/shared/style.scss
```diff
@@ -5,14 +5,14 @@
 
 html,
 body,
-.react-root {
+#root {
```
mattseddon, line 8: Changing to an ID selector so we don't have to set the class name on the element

mattseddon:
I will merge this into dev now so it gets a good run before the next release.

PR #1986, opened 2022-07-05
Fix checkbox selection issues

wolmir:
This PR will fix some issues with the current checkbox selection of rows in the experiments table.

It will also be possible to clear the selection with the Esc key.

https://user-images.githubusercontent.com/1231848/177400574-f2e60c2e-9ea4-4188-bb84-79024f489119.mov

### TODO

- [x] Remove the context-menu suppression when an experiment is running (apply rule per-option)
- [x] Only select the parent row if all of its children are selected

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -842,6 +842,35 @@ describe('App', () => {
       const itemLabels = menuitems.map(item => item.textContent)
       expect(itemLabels).toContain('Star Experiments')
     })
+
+    it('should clear the row selection when the Escape key is pressed', () => {
+      render(<App />)
+
+      fireEvent(
+        window,
+        new MessageEvent('message', {
+          data: {
+            data: tableDataFixture,
+            type: MessageToWebviewType.SET_DATA
+          }
+        })
+      )
+
+      const firstRowCheckbox = within(getRow('4fb124a')).getByRole('checkbox')
+      fireEvent.click(firstRowCheckbox)
+
+      const tailRow = within(getRow('42b8736')).getByRole('checkbox')
+      fireEvent.click(tailRow, { shiftKey: true })
+
+      const selectedRows = () =>
+        screen.queryAllByRole('row', { selected: true })
+      expect(selectedRows().length).toBe(4)
+
+      fireEvent.keyUp(tailRow, { bubbles: true, key: 'Escape' })
+
+      jest.advanceTimersByTime(100)
+      expect(selectedRows().length).toBe(0)
+    })
   })
 
   describe('Star Experiments', () => {
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -843,6 +843,42 @@ describe('App', () => {
       expect(itemLabels).toContain('Star Experiments')
     })
 
+    it('should not include collapsed experiments in the bulk selection', () => {
+      render(<App />)
+
+      fireEvent(
+        window,
+        new MessageEvent('message', {
+          data: {
+            data: {
+              ...tableDataFixture,
+              hasRunningExperiment: false
+            },
+            type: MessageToWebviewType.SET_DATA
+          }
+        })
+      )
+
+      const testBranchContractButton = within(getRow('42b8736')).getByTitle(
+        'Contract Row'
+      )
+      fireEvent.click(testBranchContractButton)
+
+      jest.advanceTimersByTime(100)
+      const firstRowCheckbox = within(getRow('4fb124a')).getByRole('checkbox')
+      fireEvent.click(firstRowCheckbox)
+
+      const tailRow = within(getRow('22e40e1')).getByRole('checkbox')
+      fireEvent.click(tailRow, { shiftKey: true })
+
+      fireEvent.click(testBranchContractButton)
+
+      jest.advanceTimersByTime(100)
+      expect(getRow('42b8736')).toHaveAttribute('aria-selected', 'true')
+      expect(getRow('2173124')).not.toHaveAttribute('aria-selected', 'true')
+      expect(getRow('9523bde')).not.toHaveAttribute('aria-selected', 'true')
+    })
+
     it('should present the Clear selected rows option when multiple rows are selected', () => {
       render(<App />)
 
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -746,8 +746,8 @@ describe('App', () => {
       const menuitems = screen.getAllByRole('menuitem')
       const itemLabels = menuitems.map(item => item.textContent)
       expect(itemLabels).toStrictEqual([
-        'Modify and Resume',
         'Modify, Reset and Run',
+        'Modify and Resume',
         'Modify and Queue',
         'Star Experiment'
       ])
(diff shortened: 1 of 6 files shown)


maxagin:
Do we have somewhere documented list of the supported hotkeys? How the users will know about this option? 
We may want to think about educating the user in regards to this small, but important details and functionality.

wolmir:
> Do we have somewhere documented list of the supported hotkeys? How the users will know about this option?
We may want to think about educating the user in regards to this small, but important details and functionality.

I'll add the option to the context-menu as well

What about showing the keyboard shortcut (Esc) in that option @maxagin ?

wolmir:
@maxagin I updated the video above with the menu option

mattseddon:
> Do we have somewhere documented list of the supported hotkeys? How the users will know about this option? We may want to think about educating the user in regards to this small, but important details and functionality.

@maxagin https://code.visualstudio.com/docs/getstarted/keybindings

webview/src/experiments/components/App.test.tsx
```diff
@@ -842,6 +842,72 @@ describe('App', () => {
...
+
+      const firstRowCheckbox = within(getRow('4fb124a')).getByRole('checkbox')
+      fireEvent.click(firstRowCheckbox)
+
+      const tailRow = within(getRow('42b8736')).getByRole('checkbox')
+      fireEvent.click(tailRow, { shiftKey: true })
+
+      const selectedRows = () =>
+        screen.queryAllByRole('row', { selected: true })
+      expect(selectedRows().length).toBe(4)
```
mattseddon, line 904: [N] I think you can use `toHaveLength` here.

webview/src/experiments/components/table/Row.tsx
```diff
@@ -67,7 +70,8 @@ const getMultiSelectMenuOptions = (selectedRowsList: RowProp[]) => {
     .filter(value => value.row.depth === 1)
     .map(value => value.row.values.id)
 
-  const hideRemoveOption = removableRowIds.length !== selectedRowsList.length
+  const hideRemoveOption =
+    removableRowIds.length !== selectedRowsList.length || hasRunningExperiment
```
mattseddon, line 74: [C] IMO if one of the selected rows is removable we should offer the remove option and remove only the ones we can.

webview/src/experiments/components/table/Row.tsx
```diff
@@ -92,6 +96,46 @@ const getMultiSelectMenuOptions = (selectedRowsList: RowProp[]) => {
...
+
+  return [
+    withId(
+      projectHasCheckpoints ? 'Modify and Resume' : 'Modify and Run',
+      MessageFromWebviewType.VARY_EXPERIMENT_PARAMS_AND_RUN,
+      !isNotCheckpoint,
+      !hideVaryAndRun
+    ),
+    withId(
+      'Modify, Reset and Run',
```
mattseddon, line 131: [Q] Can we reorder these to match the tree context menu? 

image

mattseddon: The reason being "Reset and Run" is the most likely candidate for the next action.

webview/src/experiments/components/table/Row.tsx
```diff
@@ -100,53 +144,51 @@ const getSingleSelectMenuOptions = (
...
+      hidden || hasRunningExperiment,
+      divider
+    )
 
   return [
     withId(
       'Apply to Workspace',
       MessageFromWebviewType.APPLY_EXPERIMENT_TO_WORKSPACE,
-      canApplyOrCreateBranch
+      hideApplyAndCreateBranch
```
mattseddon, line 172: [Q] The name would suggest the logic has been reversed here (but it looks like the same condition). Was the name wrong before and the logic correct or has something else happened?

webview/src/experiments/components/table/Row.tsx
```diff
@@ -92,6 +96,46 @@ const getMultiSelectMenuOptions = (selectedRowsList: RowProp[]) => {
...
+  hideVaryAndRun: boolean,
+  depth: number
+) => {
+  const isNotCheckpoint = depth <= 1 || isWorkspace
+
+  return [
+    withId(
+      projectHasCheckpoints ? 'Modify and Resume' : 'Modify and Run',
+      MessageFromWebviewType.VARY_EXPERIMENT_PARAMS_AND_RUN,
+      !isNotCheckpoint,
```
mattseddon, line 127: [C] Might be easier to read if you change the condition to `isCheckpoint` instead of the double negative.

webview/src/experiments/components/table/Row.tsx
```diff
@@ -92,6 +96,46 @@ const getMultiSelectMenuOptions = (selectedRowsList: RowProp[]) => {
...
+    {
+      divider: true,
+      id: 'clear-selection',
+      keyboardShortcut: 'Esc',
+      label: 'Clear row selection'
+    }
+  ]
+}
+
+const getWorkspaceOptions = (
```
mattseddon, line 109: [C] Doesn't appear that this only applies to the workspace.

webview/src/experiments/components/table/Table.tsx
```diff
@@ -139,29 +147,37 @@ export const Table: React.FC<TableProps & WithChanges> = ({
   useClickOutside(tableRef, clickOutsideHandler)
 
   const batchRowSelection = React.useCallback(
```
mattseddon, line 149: [Q] Can we further test this?

mattseddon (commented):
Have gone through and left some comments, IMO we will want to suppress the selection of checkpoints on collapsed experiments before merging:

https://user-images.githubusercontent.com/37993418/177449656-4e3ee355-7529-4706-81d3-5b4d9f0a9fc2.mov

webview/src/experiments/components/table/Row.tsx
mattseddon: try and simplify this boolean logic wherever you can, it is getting pretty difficult to understand at first glance.

mattseddon:
If you followed the instructions in the link you would be able to see all of the actions in VS Code currently mapped to `Escape`

image

Hint: generally they relate to closing/unselecting/hiding

webview/src/experiments/components/table/Row.tsx
wolmir: Yes, that seems like a simple change

webview/src/experiments/components/table/Row.tsx
wolmir: Yes, because the property is called `hidden`, so these are assertions that will hide the option when true. It was wrong before

webview/src/experiments/components/table/Row.tsx
wolmir: I don't like it either, but I have a feeling this refactoring will take some time. I can tackle it in the next PR, if that's ok.

webview/src/experiments/components/table/Table.tsx
wolmir: Sure, I think I can add some more tests today

rogermparent:
https://user-images.githubusercontent.com/9111807/177836594-3d7c1244-a4ab-4849-ad79-351f735367ac.mp4

I think it'd be best to make the tooltips dismiss on esc (maybe even for all interactive ones). I don't think it's a blocker but warrants a follow-up or an issue.

rogermparent (approved):
I think [this comment](https://github.com/iterative/vscode-dvc/pull/1986/files#r914334330) is both important and simple enough to be considered before merging, but I don't see any other showstoppers right now. Nice work!

shcheklein:
@wolmir good stuff! 😎 please create a housekeeping / followup ticket for the comments / improvements we need to address still.

PR #1988, opened 2022-07-05
Fix plot colors

### LINKED ISSUE #1957: bad colours in VS code (compared to rendering in browser)
with the default VS code dark mode, some plots (e.g. the confusion matrix) are hard to read (see the first picture). They are not much better with other VS code themes, unfortunately. However, they are just fine when opened in a browser (see the second picture)
image

image

The browser is Firefox, but it should not matter.

rogermparent:
Fixes #1957 

This PR changes our Vega embed config to be less aggressive. The general goal is allowing for the plot's original color to show through on internal parts, like the cells of a confusion matrix, while changing the color of labels that are laid over the transparent background to adapt to light/dark mode via CSS variables.

The new config was adapted from [vega's built in dark theme](https://github.com/vega/vega-themes/blob/next/src/theme-dark.ts), but with colors replaced with CSS variables and some of our old settings added back in.

image
image

Videos:

https://user-images.githubusercontent.com/9111807/177435467-8f7a1e1b-26cc-4772-bf17-0204c53e8532.mp4

https://user-images.githubusercontent.com/9111807/177435470-3bcd2953-957e-4534-b7c6-e03c36d0366e.mp4

### DIFF
--- a/webview/src/plots/components/constants.ts
+++ b/webview/src/plots/components/constants.ts
@@ -15,10 +15,10 @@ const title = {
 export const config: Config = {
   axis: {
     domain: false,
-    domainColor: foregroundColor,
     gridColor: foregroundColor,
     gridOpacity: 0.25,
     tickColor: foregroundColor,
+    titleColor: foregroundColor,
     titlePadding: 15
   },
   background: backgroundColor,
--- a/webview/src/plots/components/constants.ts
+++ b/webview/src/plots/components/constants.ts
@@ -25,6 +25,9 @@ export const config: Config = {
 
   padding: 20,
   style: {
+    cell: {
+      stroke: foregroundColor
+    },
     'group-title': title,
     'guide-label': {
       fill: foregroundColor,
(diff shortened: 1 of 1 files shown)


webview/src/plots/components/constants.ts
```diff
@@ -1,12 +1,12 @@
 import { Config, FontWeight } from 'vega'
-import { getThemeValue, ThemeProperty } from '../../util/styles'
 
-const foregroundColor = getThemeValue(ThemeProperty.FOREGROUND_COLOR)
+const foregroundColor = 'var(--vscode-editor-foreground)'
```
rogermparent, line 3: `getThemeValue` causes an issue here where the color it sets does not get updated when the color theme changes, requiring a webview re-open. Switching to a plain CSS variable fixes that.

mattseddon (approved):
Thanks for doing this Roger, looks like a good solution. 

Can you please push this one all the way out to the marketplace as a user is waiting for it. 

Thanks!

Issue #1814, opened 2022-06-02
Error creating an experiment when `params.yaml` contains a list of values


alex000kim:
My `params.yaml` contains a parameter that's a list of two values:
```yaml
...
data_load:
  ...
  code_names: [0, 1]
...
```
When I try to create an experiment (either run immediately or put in the queue), there are two problems:
1. the experiment doesn't get created
2. there's no error message

When it look at the output of the extension, I see this error:
`ERROR: 2 parameters are missing from 'params.yaml': data_load.code_names.1, data_load.code_names.0`

The extension tries to run `dvc exp run` with the following options:
`dvc exp run --queue ... -S params.yaml:data_load.code_names.0=0 -S params.yaml:data_load.code_names.1=1 ...`

PR #1991, opened 2022-07-06
Fix tooltip casting

rogermparent:
Had a regression while adding `stopPropagation` to the tooltips where I forgot to re-add the `String` conversion.

I suppose this really emphasizes the need for regression tests. Also refactored things to be slightly more efficient.

https://user-images.githubusercontent.com/9111807/177436373-9a3ae893-d9fb-454b-a138-9848531c7ee2.mp4

### DIFF
--- a/webview/src/experiments/util/buildDynamicColumns.tsx
+++ b/webview/src/experiments/util/buildDynamicColumns.tsx
@@ -13,7 +13,6 @@ import Tooltip, {
   CELL_TOOLTIP_DELAY
 } from '../../shared/components/tooltip/Tooltip'
 import styles from '../components/table/styles.module.scss'
-import tooltipStyles from '../../shared/components/tooltip/styles.module.scss'
 import { CopyButton } from '../../shared/components/copyButton/CopyButton'
 import { OverflowHoverTooltip } from '../components/overflowHoverTooltip/OverflowHoverTooltip'
 const UndefinedCell = (
@@ -23,11 +22,11 @@ const UndefinedCell = (
 )
 
 const CellTooltip: React.FC<{
-  cell: Cell<Experiment, string | number>
-}> = ({ cell }) => {
-  const { value } = cell
+  stringValue: string
+}> = ({ stringValue }) => {
   return (
     <div
+      className={styles.cellTooltip}
       role="textbox"
       tabIndex={0}
       onClick={e => {
@@ -37,7 +36,7 @@ const CellTooltip: React.FC<{
         e.stopPropagation()
       }}
     >
-      {value}
+      {stringValue}
     </div>
   )
 }
@@ -57,8 +56,7 @@ const Cell: React.FC<Cell<Experiment, string | number>> = cell => {
 
   return (
     <Tooltip
-      className={tooltipStyles.padded}
-      content={<CellTooltip cell={cell} />}
+      content={<CellTooltip stringValue={stringValue} />}
       placement="bottom"
       arrow={true}
       delay={[CELL_TOOLTIP_DELAY, 0]}
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -663,3 +663,7 @@ $badge-size: 0.85rem;
   vertical-align: middle;
   font-size: 0.5rem;
 }
+
+.cellTooltip {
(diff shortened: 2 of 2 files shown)


mattseddon:
Consider: https://github.com/iterative/vscode-dvc/issues/1814 (this may already be catered for but just popped into my head when watching the video)

extension/src/cli/reader.ts
```diff
@@ -49,14 +49,8 @@ export type StatusesOrAlwaysChanged = StageOrFileStatuses | 'always changed'
...
-export type Value =
-  | string
-  | number
-  | boolean
-  | null
-  | number[]
-  | string[]
-  | boolean[]
+type SingleValue = string | number | boolean | null
+export type Value = SingleValue | SingleValue[]
```
rogermparent, line 53: This allows the `Value` type to hold mixed but not nested lists.

extension/src/experiments/columns/model.test.ts
```diff
@@ -10,21 +10,32 @@ import {
...
+import {
+  dataTypesOutput,
+  columns as dataTypesColumns
+} from '../../test/fixtures/expShow/dataTypes'
 
 describe('ColumnsModel', () => {
   const exampleDvcRoot = 'test'
 
-  it('should return columns that equal the columns fixture when given the output fixture', async () => {
+  it('should return expected columns when given the default output fixture', async () => {
```
rogermparent, line 21: Made all the output tests for row and column outputs of the fixtures to use similar titles.

extension/src/test/fixtures/expShow/dataTypes.ts
```diff
@@ -0,0 +1,193 @@
+import { ExperimentsOutput } from '../../../cli/reader'
+import {
+  Column,
+  ColumnType,
+  Row,
+  TableData
+} from '../../../experiments/webview/contract'
+
+export const dataTypesOutput: ExperimentsOutput = {
```
rogermparent, line 9: We'll probably add more types here, I just included what has been asked for and we can work from there- I haven't tried nested lists or objects, and `null` breaks the table when given but it doesn't occur in JSON.

webview/src/experiments/components/App.test.tsx
```diff
@@ -641,31 +639,57 @@ describe('App', () => {
...
+        expectedTooltipResult: 'true'
+      })
+      expectTooltipValue({
+        cellLabel: 'false',
+        expectedTooltipResult: 'false'
+      })
+      expectTooltipValue({
+        cellLabel: '[true, false, string, 2]',
+        expectedTooltipResult: '[true, false, string, 2]'
+      })
```
rogermparent, line 692: Multiple test titles would have more clean messages, but I don't have the heart to re-render the same fixture 3+ more times to do it.

webview/src/experiments/util/buildDynamicColumns.tsx
```diff
@@ -57,8 +53,7 @@ const Cell: React.FC<Cell<Experiment, string | number>> = cell => {
 
   return (
     <Tooltip
-      className={tooltipStyles.padded}
-      content={<CellTooltip cell={cell} />}
+      content={<CellTooltip stringValue={stringValue} />}
```
rogermparent, line 56: Padding style moved into the base element

webview/src/stories/Table.stories.tsx
```diff
@@ -79,6 +80,9 @@ WithNoRunningExperiments.args = {
   }
 }
 
+export const WithAllDataTypes = Template.bind({})
+WithAllDataTypes.args = { tableData: dataTypesTableData }
```
mattseddon, line 84: [Q] Can you add the boolean tooltip into this test with the `play` functionality?

mattseddon (approved):

extension/src/experiments/columns/model.test.ts
rogermparent: Yeah, that reads a bit more naturally

mattseddon:
Looks like the breakage in Storybook is from this PR: https://github.com/iterative/vscode-dvc/pull/2005#issuecomment-1180925942

mattseddon:
Did you reshuffle the PRs and not update the description titles? #2005-> this is the wrong way round, right?

PR #2002, opened 2022-07-08
Add e2e tests into CI

### LINKED ISSUE #1471: Use `wdio-vscode-service` for e2e tests
Problems to solve

- [x] Add `wdio-vscode-service` with simple tests
- [x] Build webview page object selector
- [x] Test that webview is open and working
- [x] Structure job (needs to build venv)
- [x] Shared VS Code instance with `.vscode-test` - won't do (cost/benefit)
- [x] Investigate screenshot regression tests - need somewhere to store/compare baseline
- [x] If using screenshots - work out how to update the baseline
- [x] Schedule for running in CI

Decided against messing around with screenshots for the time being.

mattseddon:
# 1/2 `main` <= this <= #2013 

Closes #1471.

Please review the approach. Not much has changed from the last PR (#1993). We just have to `dvc pull` the required data and ensure we checkout the entire git history to get things working in the CI.

Comments inline.

There is one remaining test that I would like to write. I'll produce a follow up PR.


.github/workflows/end-to-end-test.yml
```diff
@@ -0,0 +1,36 @@
+name: End-To-End Test
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
mattseddon, line 6: [I] Schedule this and run on request (it is expensive).

extension/src/test/e2e/util.ts
```diff
@@ -35,28 +56,42 @@ export const waitForViewContainerToLoad = async () => {
...
-    }
+    )
+  })
+
+  await waitForDvcToFinish()
 
-    for (const progress of currentProgressBars) {
-      if ((await progress.getAttribute('aria-hidden')) !== 'true') {
+  const workbench = await browser.getWorkbench()
+  await workbench.executeCommand('DVC: Pull')
```
mattseddon, line 72: [F] In the CI we need to pull all of the DVC tracked data. This was a headache to get working.

.github/workflows/end-to-end-test.yml
```diff
@@ -0,0 +1,45 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
mattseddon, line 20: image

mattseddon: This took a few hours of my life.

rogermparent (commented):
Ideally we would add this to the Continuous Integration CI, it's just a question of how liberal we want to be with our GitHub Actions minutes. We could add a few conditions to save minutes, maybe something like skipping the e2e run on draft PRs?

I'm also curious how much time we could save on e2e tests by stripping down the dvc project that drives it to something that only layers the bare minimum above dvc, removing even `torch` in favor of something like static fixture data; what we lose in authenticity would easily be made up for by how huge it would be to be able to run e2e tests as liberally as our other suites.

A hyper-minimal testbed project wouldn't necessarily have to replace our current demo, but there could also arguments that it should.

As far as cross-platform tests go, that's a whole other question. Considering windows and mac runners consume 2 and 10 times as much minutes as linux runners, I think our best bet is to save cross-platform e2e runs for situations when they're specifically requested through either `workflow-dispatch` or some more convenient form of trigger like a special label.

mattseddon:
I am going to add to the Continuous Integration workflow and strip the `macOS-latest` from the `matrix.os` property in `.github/workflows/cross-platform-test.yml`.

For me, having a realistic project is important as it is an end to end test. It is also important to use something realistic when we are manually testing the project.

.github/workflows/cross-platform-test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        os: [macOS-latest, windows-latest]
+        os: [windows-latest]
```
mattseddon, line 18: [F] By removing the less valuable `macOS-latest` tests we make get back more minutes than we expend running the E2E tests in the main workflow.

extension/src/test/e2e/pageObjects/locators.ts
```diff
@@ -7,7 +7,7 @@ export const experiments = {
   ...webview,
   expandRowButton: 'button[title="Expand Row"]',
   row: '[role=row]',
-  table: '[role=table]'
+  table: '[role=tree]'
```
mattseddon, line 10: [F] This changed when we move up to v18 of React

extension/src/test/e2e/wdio.conf.ts
```diff
@@ -1,13 +1,30 @@
...
+import { Logger } from '../../common/logger'
+
+const screenshotDir = join(__dirname, 'screenshots')
 
 export const config: Options.Testrunner = {
   after: async function () {
     await browser.switchToFrame(null)
     await browser.switchToFrame(null)
   },
+  afterTest: async (test, __, { passed }) => {
```
mattseddon, line 14: [F] This is close to what is provided in the example tests for the service. If I get time I will extend the way that we do timeouts to also take screenshots.

rogermparent (approved):

wolmir (approved):

PR #2005, opened 2022-07-08
Use @storybook/testing-library for Storybook play functions

rogermparent:
# this -> #1991 & #2010

This official storybook library should allow us to use the full testing-library query suite in `play` functions, particularly `userEvent` which errors out when imported from `@testing-library/react`.

### DIFF
--- a/webview/src/stories/Plots.stories.tsx
+++ b/webview/src/stories/Plots.stories.tsx
@@ -193,9 +193,6 @@ VirtualizedPlots.args = {
 VirtualizedPlots.parameters = chromaticParameters
 
 export const ZoomedInPlot = Template.bind({})
-ZoomedInPlot.parameters = {
-  chromatic: { delay: 300 }
-}
 ZoomedInPlot.play = async ({ canvasElement }) => {
   const plots = await within(canvasElement).findAllByTestId(/^plot_/)
   const plot = await within(plots[0]).findByRole('button')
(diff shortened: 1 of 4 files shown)


wolmir (approved):
Thanks!

mattseddon:
Can we use user-event directly?

rogermparent:
> Can we use the testing library user-event package directly?

I tried it out and it seems we cannot, doing so makes `play` functions not execute in Chromatic.

mattseddon (approved):

mattseddon:
image

@rogermparent Storybook actually looks broken.

webview/src/stories/Plots.stories.tsx
```diff
@@ -193,25 +193,20 @@ VirtualizedPlots.args = {
...
-  chromatic: { delay: 300 }
-}
 MultiviewZoomedInPlot.play = async ({ canvasElement }) => {
-  const canvas = within(canvasElement)
-  const plot = await canvas.findByTestId('plots-section_template-multi_1')
-  const plotButton = await waitFor(() => within(plot).getByRole('button'))
+  const plot = await within(canvasElement).findByTestId(
+    'plots-section_template-multi_1'
+  )
+  await within(plot).findByRole('graphics-document')
```
rogermparent, line 208: This waits until the initial plot is rendered before we click on it, and fixes the weird Storybook issue we were having where the multiview plot disappeared. I think that bug was more due to vega instead of storybook, and the new Storybook config is just bringing a symptom to surface.

PR #2010, opened 2022-07-11
Conditional sticky table head shadows

### LINKED ISSUE #1826: Add shadow to sticky header and footer
Now that we have both a sticky table head and experiments column, we'll want to increase the contrast between them and the page with shadows.

This issue has two parts, depending on what one thinks constitutes a completed shadow

- [x] Adding CSS shadows added to each sticky bar
- [ ] Making those shadows only visible when the related component is "sticking"

rogermparent:
# #2005 <- this 

fixes #1826 

This PR makes the shadow on the sticky table header only appear when the table is scrolled a bit (set as 15px, but can be changed to anything if we want via `rootMargin`).

https://user-images.githubusercontent.com/9111807/178364986-73b21dbb-e132-4435-940d-b58947a21464.mp4

Opted to rely on Chromatic for testing, as the result is purely visual. To do this, I added a new "Scrolled" story to the table stories that uses a custom viewport and scrolls to the bottom-right in its `play` function.


webview/setup-tests.js
```diff
@@ -15,4 +15,13 @@ const mutationObserverMock = jest.fn().mockImplementation(() => {
...
 global.MutationObserver = mutationObserverMock
 
+const intersectionObserverMock = jest.fn().mockImplementation(() => {
+  return {
+    disconnect: jest.fn(),
+    observe: jest.fn(),
+    unobserve: jest.fn()
+  }
+})
+global.IntersectionObserver = intersectionObserverMock
```
rogermparent, line 25: Could be made better if we wanted to actually puppet the observer, but I think we have better ways to test most scenarios it would apply to.

mattseddon (approved):

webview/src/experiments/components/table/TableHead.tsx
```diff
@@ -85,9 +89,14 @@ export const TableHead: React.FC<TableHeadProps> = ({
       type: MessageFromWebviewType.REORDER_COLUMNS
     })
   }
+  const [ref, scrolled] = useInView({
```
mattseddon, line 92: [C] Would be good to have an indicator here to know this is for the sticky header shadow. `[ref, headerNeedsShadow]`? Just something for the next person that comes along.

rogermparent: My first thought was `scrolled` because that's describing what the variable contains regardless of how we use it, but I can see how that would be helpful where it's only used in one place now.

PR #2011, opened 2022-07-11
Fix race condition in manual plot refresh (M1 macs)

mattseddon:
(surprisingly) Only two tests from the entire test suite didn't work when I switched to M1. 

The tests in question are both from the integration suite:

1. `should handle a message to manually refresh a revision from the webview`
2. `should handle a message to manually refresh all visible plots from the webview`

The issue was caused by a race condition which this PR fixes.


wolmir (approved):
Thanks!

rogermparent (approved):

PR #2013, opened 2022-07-12
Add end to end test for SCM and file decorations

mattseddon:
# 2/2 `main` <- #2002 <- this

This PR adds an end to end test for the SCM view and our custom file decorations.

### DIFF
--- a/extension/src/test/e2e/extension.test.ts
+++ b/extension/src/test/e2e/extension.test.ts
@@ -125,8 +125,6 @@ suite('DVC Extension For Visual Studio Code', () => {
   // eslint-disable-next-line sonarjs/cognitive-complexity
   describe('Source Control View', () => {
     it('should show the expected changes after running an experiment', async () => {
-      const treeItems = await findScmTreeItems()
-
       const expectedScmItemLabels = [
         'demo DVC',
         'training_metrics',
@@ -140,6 +138,7 @@ suite('DVC Extension For Visual Studio Code', () => {
       await browser.waitUntil(
         async () => {
           dvcTreeItemLabels = []
+          const treeItems = await findScmTreeItems()
           for (const treeItem of treeItems) {
             const treeItemLabel = await getLabel(treeItem)
             if (!expectedScmSet.has(treeItemLabel)) {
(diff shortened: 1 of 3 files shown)


extension/src/test/e2e/extension.test.ts
```diff
@@ -117,4 +121,48 @@ suite('DVC Extension For Visual Studio Code', () => {
...
+        {
+          interval: 5000,
+          timeout: 60000
+        }
+      )
+
+      expect(expectedScmItemLabels.sort()).toStrictEqual(
+        dvcTreeItemLabels.sort()
+      )
+    }).timeout(60000)
```
wolmir, line 166: Isn't this timeout too long?

wolmir (approved):

extension/src/test/e2e/extension.test.ts
mattseddon: The SCM state relies on `dvc status`. Whenever an experiment is running `dvc status` will fail. There is an exponential back off in place for retries which means that after the experiment finishes there can be a significant lag before the SCM view is updated with the correct details. Locally this is less of a problem but in the CI running an experiment takes longer. Leading to a longer lag in the subsequent update. I'd rather that this test is reliable with a longer timeout than flaky with a shorter one. 

Does that make sense? WDYT?

wolmir: It does for me 👍
I know this is a can of worms, but is there a way to only `dvc status`  after all the runs? Like using a subscription or something like that?

shcheklein: Q: can it be done as a global setting - one minute timeout per test?

shcheklein: probably it applies to all tests that run DVC internally?

mattseddon: > I know this is a can of worms, but is there a way to only `dvc status` after all the runs? Like using a subscription or something like that?

That would mean we would have to introduce stubs into this test suite, making them less e2e. I would rather test the extension as is. We should be able to improve the situation once we get `data:status`.

> Q: can it be done as a global setting - one minute timeout per test?

This was actually already set to 1m

PR #2016, opened 2022-07-12
Add storybook-addon-themes

rogermparent:
https://user-images.githubusercontent.com/9111807/178782062-073ece6d-b6a8-4092-babc-0bf30c8fc6e1.mp4

This PR adds color themes to Storybook, allowing the developer to preview the Light and Dive Bar themes in Storybook on top of the normal dark theme.

These themes are just scraped from VSCode's dev tools by grabbing every rule that sets CSS variables and tossing it into a conditional classname that the plugin adds UI for toggling.


mattseddon (approved):
Would be good to add some more of the default themes + One Dark Pro before merging.

rogermparent:
Added some more default themes, here's the current list

- Dark
- Light
- One Dark Pro
- Red
- Dive Bar
- High Contrast Light
- High Contrast Dark

PR #2021, opened 2022-07-14
Ignore packages in Renovate config

mattseddon:
There are three packages that we want to manually manage:

| package | reason | 
| -- | -- |
| dvc | make sure new features are backwards compatible, should be updated when we update the min version |
| @types/vscode | updating means we limit users to the most recent version of VS Code |
| execa | next version is ESM only |

[Docs](https://docs.renovatebot.com/configuration-options/#ignoredeps).


julieg18 (approved):
Looks good to me!

rogermparent (approved):

PR #2026, opened 2022-07-14
Remove webview message handling from experiments

mattseddon:
# 1/2 `main` <- this <- #2027 

This PR pulls out webview message handling from the now overcrowded `Experiments` "glue" class.

Reason for targeting this file:

Tl;dr from a book called [Software Design X-Rays](https://pragprog.com/titles/atevol/software-design-x-rays/) (which is a good book that I recommend reading): Prioritise code hotspots when working on technical debt. 

From Code Climate:

image

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -1,13 +1,4 @@
-import { join } from 'path'
-import {
-  commands,
-  Event,
-  EventEmitter,
-  Memento,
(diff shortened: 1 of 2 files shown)


rogermparent (approved):
The title made me think this PR would be _way_ more extreme than it actually is

wolmir (approved):

PR #2029, opened 2022-07-14
Highlight changed hashes for Deps columns

wolmir:
This PR will highlight the dependency cells with changes in relation to HEAD.

Demo:

https://user-images.githubusercontent.com/1231848/179546412-de3c7884-e3b3-4428-a1a8-2ae5da88e5fb.mov

### DIFF
--- a/extension/src/experiments/columns/extract.ts
+++ b/extension/src/experiments/columns/extract.ts
@@ -38,15 +38,8 @@ const extractDeps = (
   for (const [path, { hash }] of Object.entries(columns)) {
     const value = shortenForLabel(hash)
     acc[path] = {
-      changes: false,
+      changes: !!value && !!branch && branch?.deps?.[path].value !== value,
       value
-    }
-
-    if (value && branch && branch?.deps?.[path].value !== value) {
-      acc[path] = {
-        changes: true,
-        value
-      }
     }
   }
 
--- a/extension/src/experiments/columns/extract.ts
+++ b/extension/src/experiments/columns/extract.ts
@@ -37,9 +37,12 @@ const extractDeps = (
 
   for (const [path, { hash }] of Object.entries(columns)) {
     const value = shortenForLabel(hash)
-    acc[path] = value
+    acc[path] = {
+      changes: false,
+      value
+    }
 
-    if (value && branch && branch?.deps?.[path] !== value) {
+    if (value && branch && branch?.deps?.[path].value !== value) {
       acc[path] = {
         changes: true,
         value
--- a/extension/src/experiments/columns/extract.ts
+++ b/extension/src/experiments/columns/extract.ts
@@ -1,6 +1,10 @@
 import { Deps, ExperimentFields, ValueTreeRoot } from '../../cli/reader'
 import { shortenForLabel } from '../../util/string'
-import { DepColumns, MetricOrParamColumns } from '../webview/contract'
+import {
+  DepColumns,
+  Experiment,
+  MetricOrParamColumns
+} from '../webview/contract'
 
 const extractMetricsOrParams = (
   columns?: ValueTreeRoot
@@ -21,7 +25,10 @@ const extractMetricsOrParams = (
   return acc
 }
 
-const extractDeps = (columns?: Deps): DepColumns | undefined => {
+const extractDeps = (
+  columns?: Deps,
+  branch?: Experiment
+): DepColumns | undefined => {
   if (!columns) {
     return
   }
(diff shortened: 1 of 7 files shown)


extension/src/experiments/columns/extract.ts
```diff
@@ -21,28 +25,40 @@ const extractMetricsOrParams = (
...
   }
 
   const acc: DepColumns = {}
 
   for (const [path, { hash }] of Object.entries(columns)) {
-    acc[path] = shortenForLabel(hash)
+    const value = shortenForLabel(hash)
+    acc[path] = value
+
+    if (value && branch && branch?.deps?.[path] !== value) {
```
mattseddon, line 42: [Q] Is this definitely accessing the right property in the branch?

mattseddon: I can see from 
```
export interface DepColumns {
  [path: string]: string | ValueWithChanges
}
```
that you've got mixed types. It would be a good idea to standardise the type now as we are moving towards Studio style tooltips for deps:

image

webview/src/experiments/components/table/Cell.tsx
```diff
@@ -107,15 +108,24 @@ export const CellWrapper: React.FC<
...
+  const cellhasChanges = (cell.value as ValueWithChanges)?.changes
+
+  return (
+    <div
+      {...cell.getCellProps({
+        className: cx(
+          styles.td,
+          cell.isPlaceholder && styles.groupPlaceholder,
+          {
+            [styles.workspaceChange]:
```
mattseddon, line 121: [I] Name needs to be updated. Personally I think it would be ok to have two different (currently duplicate) styles. One for workspace changes and another for dep changes against HEAD.

webview/src/experiments/util/buildDynamicColumns.tsx
```diff
@@ -38,17 +42,21 @@ const CellTooltip: React.FC<{
...
+const Cell: React.FC<
+  Cell<Experiment, string | number | ValueWithChanges>
+> = cell => {
   const { value } = cell
   if (value === undefined) {
     return UndefinedCell
   }
 
-  const stringValue = String(value)
+  const rawValue = (value as ValueWithChanges).value ?? value
```
mattseddon, line 53: [I] Would be good to have a couple of helper functions for 

```
const rawValue = (value as ValueWithChanges).value ?? value
const cellhasChanges = (cell.value as ValueWithChanges)?.changes
```

so that we can keep them together and have the concept all in one place (easier to extend/reason about later)

mattseddon (commented):
Functionality looks good:

image

extension/src/experiments/columns/extract.ts
```diff
@@ -21,28 +25,43 @@ const extractMetricsOrParams = (
...
   if (!columns) {
     return
   }
 
   const acc: DepColumns = {}
 
   for (const [path, { hash }] of Object.entries(columns)) {
-    acc[path] = shortenForLabel(hash)
+    const value = shortenForLabel(hash)
+    acc[path] = {
```
sroy3, line 40: You could move this below the if clause and add a `continue` in the if clause. That way you wouldn't assign to `acc[path]` twice.

extension/src/experiments/webview/contract.ts
```diff
@@ -6,8 +6,13 @@ export interface MetricOrParamColumns {
   [filename: string]: ValueTree
 }
 
+export interface ValueWithChanges {
+  value: string | number
```
sroy3, line 10: Could this ever be anything else than a string or number? I think we have a similar interface somewhere and we also have `string[]`, `number[]`,  `boolean`, `boolean[]` in there.

webview/src/experiments/components/table/Cell.tsx
```diff
@@ -189,15 +190,22 @@ export const CellWrapper: React.FC<
...
-    {cell.render('Cell')}
-  </div>
-)
+> = ({ cell, cellId, changes }) => {
+  return (
+    <div
+      {...cell.getCellProps({
+        className: cx(
+          styles.td,
+          cell.isPlaceholder && styles.groupPlaceholder,
```
sroy3, line 199: It doesn't really matter, but since we have a block with conditionals just below, this could be moved there as `[styles.groupPlaceholder]: cell.isPlaceholder`

sroy3 (approved):

mattseddon:
@wolmir please remember to add one of the three labels to your PRs or they won't make it into the release changelog.

The labels are :product: or :housekeeping: or :bug:

Issue #2036, opened 2022-07-15
When applying a star to the exp, some strange styles appear on the sidebar


maxagin:
https://user-images.githubusercontent.com/98249521/179128304-0a47be1f-288b-458d-a6c8-d53bbb5ad1fa.mov

shcheklein:
Interesting, seems like it tries to run a progress bar or something?

mattseddon:
Those are progress bars they will be triggered whenever the tree is being updated/loading. They are feature of VS Code.

PR #2042, opened 2022-07-15
Do not update trees on star update

mattseddon:
# 4/3 `main` <- #2030 <- #2033 <- #2038 <- this

Closes https://github.com/iterative/vscode-dvc/issues/2036

### DIFF
--- a/extension/src/experiments/webview/messages.ts
+++ b/extension/src/experiments/webview/messages.ts
@@ -156,7 +156,7 @@ export class WebviewMessages {
 
   private setExperimentStars(ids: string[]) {
     this.experiments.toggleStars(ids)
-    this.notifyChanged()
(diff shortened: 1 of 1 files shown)


extension/src/experiments/webview/messages.ts
```diff
@@ -156,7 +156,7 @@ export class WebviewMessages {
 
   private setExperimentStars(ids: string[]) {
     this.experiments.toggleStars(ids)
-    this.notifyChanged()
+    this.sendWebviewMessage()
```
mattseddon, line 159: [F] Stars are not shown in any of the trees so we only need to update the webview.

sroy3 (approved):

wolmir (approved):

julieg18 (approved):

PR #2004, opened 2022-07-08
Show middle state badges for running, starred and checked sub-rows

wolmir:
This PR will introduce indicators for the state of sub-rows in the experiments table.

There will an indicator for selected, starred and running sub-rows.

See https://github.com/iterative/vscode-dvc/issues/1714 for context on the issue.

https://user-images.githubusercontent.com/1231848/179523374-1acbe6c8-d3e4-43c0-8c5e-6e82aecb8334.mov

PR #2056, opened 2022-07-18
Move some App.test.tsx helpers to utils

wolmir:
Based on https://github.com/iterative/vscode-dvc/pull/2004/files#r921740235

### DIFF
--- a/webview/src/test/experimentsTable.tsx
+++ b/webview/src/test/experimentsTable.tsx
@@ -73,12 +73,15 @@ export const clickRowCheckbox = (label: string, multiSelection?: boolean) => {
   })
 }
 
+export const toggleExpansion = (label: string, btnTitle: string) => {
+  const button = within(getRow(label)).getByTitle(`${btnTitle} Row`)
+  fireEvent.click(button)
+}
+
 export const contractRow = (label: string) => {
-  const contractButton = within(getRow(label)).getByTitle('Contract Row')
-  fireEvent.click(contractButton)
+  toggleExpansion(label, 'Contract')
 }
 
 export const expandRow = (label: string) => {
-  const expandButton = within(getRow(label)).getByTitle('Expand Row')
-  fireEvent.click(expandButton)
+  toggleExpansion(label, 'Expand')
 }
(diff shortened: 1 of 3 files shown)


webview/src/test/experimentsTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const renderTableWithWorkspaceRowOnly = () => {
+  renderTable({ ...tableDataFixture, rows: [tableDataFixture.rows[0]] })
+}
+
+export const renderTableWithSortingdata = () => {
+  renderTable(sortingTableDataFixture)
+}
+
+export const renderTableWithoutRuningExperiments = () => {
```
julieg18, line 44: Did you mean `renderTableWithoutRunningExperiments`?

webview/src/test/experimentsTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const renderTableWithNoColumns = () => {
+  renderTable({ ...tableDataFixture, columns: [] })
+}
+
+export const renderTableWithWorkspaceRowOnly = () => {
+  renderTable({ ...tableDataFixture, rows: [tableDataFixture.rows[0]] })
+}
+
+export const renderTableWithSortingdata = () => {
```
julieg18, line 40: Did you mean `renderTableWithSortingData`?

mattseddon (approved):

webview/src/test/experimentsTable.tsx
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable unicorn/filename-case */
```
sroy3, line 1: In theory, it should already be in the ignore pattern of the rule (all files in test folders are ignored for PascalCase rule, but I think we had that because we used to have something like `util.tsx`. I'm torn between fixing the rule so that your file gets ignored and you can remove that line and just using PascalCase here (and even removing the ignore pattern on the test files).

wolmir: I forgot about this issue!
I prefer to ignore the rule for cases like this (no just tests).
My reason is that a PascalCase implies that the file is either a component (Like IconMenu.tsx) or a family of related components (like Row(s).tsx).
But for buildDynamicColumns.tsx and this file the jsx code is just ancillary stuff.
They are not the main purpose of the file.

sroy3: Then it should be in the ignore pattern of the rule. It is already the case, but it must be broken.

PR #2062, opened 2022-07-19
Add conditional shadow to sticky experiments column

### LINKED ISSUE #2017: Add a shadow to the sticky experiments column
image

The strong border between the sticky experiments column and the content under it can be a bit jarring, and we may want to consider giving it a conditional shadow like #2010.

We could also consider another effect, like monday.com's semi-transparent sticky column that slightly shows contents under it and terminates in a normal border line.
image

julieg18:
* Adds a shadow to the experiments column if you start scrolling horizontally. 
* Takes off the gray border on the experiments column

https://user-images.githubusercontent.com/43496356/180019902-dd3f10fa-41bf-4ae4-9dd6-4f0c505964b8.mov

Fixes #2017


demo/dvc.lock
```diff
@@ -31,4 +31,4 @@ stages:
       nfiles: 2
     - path: training_metrics.json
       md5: 2a592caac70651d359b2331ff674beec
-      size: 69
```
julieg18, line 34: Accidently committed some `demo` changes! Not sure what's causing this file change though 🤔

webview/src/experiments/components/table/Table.tsx
```diff
@@ -205,6 +219,7 @@ export const Table: React.FC<TableProps & WithChanges> = ({
             hasRunningExperiment={hasRunningExperiment}
             projectHasCheckpoints={hasCheckpoints}
             batchRowSelection={batchRowSelection}
+            root={tableRef.current}
```
julieg18, line 222: Since the experiment column's cells needed access to `tableRef.current`, I ended up needing to carry the `root` prop across multiple table components. Though maybe there is an easier way that doesn't involve having to touch so many components 🤔

shcheklein:
@julieg18 could you please add a video (on mac you can do video screenshots) and publish it to the description? plus put more context in the description if needed. Thanks! Good stuff.

webview/src/experiments/components/table/Cell.tsx
```diff
@@ -129,27 +130,38 @@ export const RowActions: React.FC<RowActionsProps> = ({
...
-
   const {
     subRowStates: { plotSelections }
   } = rowActionsProps
 
+  const [ref, needsShadow] = useInView({
+    root,
+    rootMargin: '0px 0px 0px -15px',
+    threshold: 1
+  })
```
rogermparent, line 154: Are we able to do this without adding an observer to every row's first cell?

julieg18: Good point! This solution requires us to add multiple observers which isn't optimal 🤔  I'll look into a way where we'd only have to rely on one observer!

julieg18: Updated the solution! Now only using one observer and ended up touching less files!

mattseddon:
[Q] Is the vertical line (from the video) in the design? Can it be removed?

shcheklein:
Feedback - on the video shadow doesn't look like a shadow on my end?

And yes, we can now remove the gray vertrical line?

julieg18:
> Feedback - on the video shadow doesn't look like a shadow on my end?

Is it not showing up on a different theme, or are you not seeing it at all? 

> [Q] Is the vertical line (from the video) in the design? Can it be removed?

Not sure what line you are referring to 🤔  The gray one?

julieg18:
On further inspection, the shadow actually doesn't have any `blur` to it which is why it doesn't look very shadow-like. Going to look into adding the `blur`!

julieg18:
Added a blur to the box shadow and removed the gray border on the experiments column!

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -282,6 +290,12 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
     }
   }
 
+  .table.withExpColumnShadow {
+    .tr:not(.rowSelected) > *:first-child:after {
+      box-shadow: 3px 0px 3px var(--vscode-widget-shadow);
```
julieg18, line 295: Should we add some `transition` to the `box-shadow` or is it fine as is? 

Also, currently, the header and experiments column are using two different box-shadow designs. Should I adjust the header box-shadow so they are using the same kind?

maxagin:
Yeah. Great stuff @julieg18 ! The video in the description has been updated I guess? 

> Should we add some transition to the box-shadow or is it fine as is?

`transition` very cool!

> Should I adjust the header box-shadow so they are using the same kind like in Figma

No, please. The current shadows in the code are consistent. It's the same as notification shadow (I hope :) However, we may change it in the future.

maxagin:
The `transition` for the header shadow is good I think. Let's use it for both then. Thanks @julieg18

webview/src/experiments/components/table/TableHeader.tsx
```diff
@@ -152,10 +226,12 @@ const TableHeaderCell: React.FC<{
   menuDisabled,
   onDragOver,
   onDragStart,
-  onDrop
+  onDrop,
+  root,
+  isFirst,
```
mattseddon, line 231: [C] I think we need a better name for this "FirstCell". We would want to convey with the name that it contains controls/metadata about the rest of the row. I am not sure what the correct name would be.
Perhaps do a bit of research and follow up with a simple renaming PR. Rename everything around the "First" concept e.g `<FirstCell/>` etc.

webview/src/experiments/components/table/MergeHeaderGroups.tsx
```diff
@@ -28,16 +31,21 @@ export const MergedHeaderGroups: React.FC<{
...
   return (
     <div
       {...headerGroup.getHeaderGroupProps({
         className: cx(styles.tr, styles.headRow)
       })}
     >
-      {headerGroup.headers.map((column: HeaderGroup<Experiment>) => (
+      {headerGroup.headers.map((column: HeaderGroup<Experiment>, ind) => (
         <TableHeader
+          isFirst={isFirst && ind === 0}
```
mattseddon, line 47: [Q] Why do we need both of these criteria? 
[Q] Can we pull out the `index === 0` into a function so we don't duplicate it in two different places?

mattseddon (approved):

mattseddon:
Storybook has got a wacky diff. I checked the branch out locally and it looked fine. Please take a look before merging.

webview/src/experiments/components/table/MergeHeaderGroups.tsx
julieg18: > [Q] Why do we need both of these criteria?

We need to select just one experiment column cell (aka one of the rendered `TableHeader`s) so we can add the observer to it. The PR does this by finding the first `MergedHeaderGroups`, then the first `TableHeader`. Personally, I felt like this solution felt overcomplicated but I couldn't think of an alternative 🤔 

> [Q] Can we pull out the index === 0 into a function so we don't duplicate it in two different places?

Sure!

julieg18:
> Storybook has got a wacky diff. I checked the branch out locally and it looked fine. Please take a look before merging.

Looks like it is due to the storybook comparing the diff to an earlier version of the pr instead of `main`.

shcheklein:
@julieg18 q - is the video up to date? I see that the horizontal shadow doesn't look the same now ... is it intentional?

julieg18:
> @julieg18 q - is the video up to date? I see that the horizontal shadow doesn't look the same now ... is it intentional?

The video is up to date and I didn't edit the horizontal shadow as far as I know... Can you specify what is exactly different on the horizontal shadow? They are looking the same from my end 🤔

shcheklein:
From the video:

Screen Shot 2022-07-21 at 2 57 10 PM

Top / horizontal shadow is not fading out? and looks a bit like a delimiter?

julieg18:
If I'm understanding correctly, looking at the horizontal shadow from before this pr and how it looks currently, they seem to look the same. Though it does seem that neither is fading out very well 🤔 

Current:
image

Before #2062 merge:
image

mattseddon:
I think the part of the shadow in question might be the top of the vertical shadow:

image

shcheklein:
@julieg18 sorry for the confusion. I mixed up the part of the tables's row with the shadow (also it starts with some delay after you scroll?) :) I see that it's the same more or less now.

julieg18:
> also it starts with some delay after you scroll?

Could be due to the `transition` animation or the code waiting till you've scrolled `15px`...

PR #2064, opened 2022-07-19
Comparison table drag and drop feedback

sroy3:
First 2 tasks of #1604 

https://user-images.githubusercontent.com/3683420/179813423-6d7865c9-31e9-4ef7-854a-892689231b93.mov


webview/src/test/nodes.ts
```diff
@@ -0,0 +1,3 @@
+export const idToNodeNode = (id: string) =>
+  // eslint-disable-next-line unicorn/prefer-query-selector
```
sroy3, line 2: This is needed to pass Windows tests. When not using `getElementById` we need a `CSS.escape` because of all the funky ids we have in both the experiments table and plots. `CSS.escape` does not seem to create a valid id query on Windows unfortunately.

webview/src/test/nodes.ts
```diff
@@ -0,0 +1,3 @@
+export const idToNodeNode = (id: string) =>
```
julieg18, line 1: Is this supposed to be `idToNode`?

julieg18 (approved):
Looks good to me!

webview/src/test/nodes.ts
sroy3: Thanks. I renamed it and must have not selected the whole thing.

webview/src/experiments/components/App.test.tsx
```diff
@@ -182,7 +182,12 @@ describe('App', () => {
     const headerB = screen.getByText('B')
     const headerD = screen.getByText('D')
 
-    dragAndDrop(headerB, headerD, DragEnterDirection.AUTO)
+    dragAndDrop(
+      headerB,
+      // eslint-disable-next-line testing-library/no-node-access
+      headerD.parentElement?.parentElement || headerD,
```
mattseddon, line 188: [Q] Can we follow up and wrap these calls in a function so we don't have `eslint-disable` all through the tests?

```
// eslint-disable-next-line testing-library/no-node-access
      headerD.parentElement?.parentElement || headerD
```

mattseddon: limit them to a single place like we try to do when using `any`

mattseddon (approved):

PR #2065, opened 2022-07-20
Add selected for plotting indicator to experiments webview

mattseddon:
Based on the conversation we had in the planning meeting regarding how we link filled circles with the fact that those experiments have been selected for plotting. 

From my understanding, this was one of the @sroy3 ideas.

We can display an indicator with the graph icon that shows how many experiments are selected for plotting. 

We can also provide an action to open the plots webview:

https://user-images.githubusercontent.com/37993418/179870922-1750d7a0-a19f-4f6f-b356-8776afc29d75.mov

~Code is only a prototype and needs to be restructured.~

What does everyone think?

@sroy3 @wolmir @julieg18 @shcheklein @maxagin

### DIFF
--- a/webview/src/experiments/components/table/Indicators.tsx
+++ b/webview/src/experiments/components/table/Indicators.tsx
@@ -123,6 +123,25 @@ const formatFilteredCountMessage = (filteredCounts: FilteredCounts): string =>
     .filter(Boolean)
     .join(', ')} Filtered`
 
+const addToSelected = (
+  selectedForPlotsCount: number,
+  row: Row<Experiment>
+): number => selectedForPlotsCount + (row.original?.selected ? 1 : 0)
+
+const getSelectedForPlotsCount = (rows: Row<Experiment>[]): number => {
+  let selectedForPlotsCount = 0
+
+  for (const row of rows) {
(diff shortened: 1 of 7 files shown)


maxagin:
Very good idea, I think. Can we split the screen when the icon clicks, so the user will have the context?

julieg18 (approved):
Looks great!

webview/src/experiments/components/table/Indicators.tsx
```diff
@@ -115,19 +123,54 @@ const formatFilteredCountMessage = (filteredCounts: FilteredCounts): string =>
...
 export const Indicators = ({
+  rows,
   sorts,
   filters,
   filteredCounts
 }: {
   sorts?: SortDefinition[]
   filters?: string[]
   filteredCounts: FilteredCounts
+  rows: Row<Experiment>[]
```
mattseddon, line 153: [C] I don't think I should be giving this access to every single row. I will update to calculate the `selectedForPlotsCount` higher up.

mattseddon:
> Very good idea, I think. Can we split the screen when the icon clicks, so the user will have the context?

IMO it's less opinionated to just open the view. The user can make the decision on how to position everything inside of their IDE for themselves. Let's see if users want this/ask them before implementing that change.

PR #2067, opened 2022-07-20
Introduce the latest DVC version extension is tested with

### LINKED ISSUE #2063: CLI version not matching need of the extension
When having version 2.6.4 of `dvc` installed I get the following error when opening VScode:

image

Updating `dvc` to version 2.13.0 I get:

image

So, now, I don't know which version of `dvc` the VScode extension is expecting. I have version 0.3.11 of the extension installed.

shcheklein:
Fixes #2063

- Fixes tests (we were not testing properly since we were using `join(',')` instead of `join('.')` which gives `NaN`s ... and NaNs are terrible in JS
- Introduces the variable that defines the latest version of DVC that we know is compatible with. Now if the detected version is between MIN_VERSION and BUILD_VERSION we are not firing any warnings

*Things to improve / discuss:*

- Name of the variable. Can be TESTED_WITH_VERSION or something. Any thoughts?
- Lifecycle of this variable. Probably we should move it to the package.json, listen to DVC updates and automatically release a new version of the extension (can be done in the future, I guess). For now we can keep an eye and see if that works.
- I'm not familiar with TS - can semver be wrapped to avoid things like NaNs, etc? (minor, I think)


extension/src/cli/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const MIN_CLI_VERSION = '2.11.0'
+export const BUILD_CLI_VERSION = '2.12.0'
```
mattseddon, line 2: [Q] Should this be `2.13.0`?
[C] Could name it `LATEST_TESTED_CLI_VERSION`.

extension/src/cli/version.ts
```diff
@@ -25,12 +29,12 @@ const getTextAndSend = (version: string, update: 'CLI' | 'extension'): void => {
...
-  minMajor: number,
+  buildMajor: number,
   currentMinor: number,
-  minMinor: number
+  buildMinor: number
 ) => {
-  if (currentMajor === minMajor && currentMinor > minMinor) {
-    Toast.warnWithOptions(`The located version of the CLI is at least a minor version ahead of the expected version. 
+  if (currentMajor === buildMajor && currentMinor > buildMinor) {
+    Toast.warnWithOptions(`The located version of the CLI is at least a minor version ahead of the latest version this version of extension was tested with.
```
mattseddon, line 37: [C] Too many `versions` in this sentence. 

How about 
```
The located CLI is at least a minor version ahead of the latest version the extension was tested with.
```

mattseddon (approved):
👍🏻

extension/src/cli/constants.ts
shcheklein: done!

extension/src/cli/version.ts
shcheklein: done, thanks!

julieg18 (approved):
Looks good to me!

Issue #1649, opened 2022-05-04
Handle DVC errors gracefully in plots


shcheklein:
This depends on https://github.com/iterative/dvc/issues/7692

[Figma specs](https://www.figma.com/file/CycOpuUFNMum7TgUgEPiJu/Errors?node-id=1206%3A40780)

### Motivation 

1. Will be missing data for a specific experiment within one of the plots. This will be most apparent in the comparison table. We need a placeholder and some way to refresh the missing experiment's data.
2. Currently, if an experiment completely fails to load plots data there will be nothing to let the user know the experiment is missing. Add an indicator into the webview's top experiment selection section in order to let the user know that an experiment has something wrong.

### TODO

Will be handled in relation to the: DVC errors that may occur during or after running experiment #1636
- [ ] Placeholder and refresh https://github.com/iterative/vscode-dvc/issues/1649#issuecomment-1128348855 
- [ ] Missing experiment indicator https://github.com/iterative/vscode-dvc/issues/1649#issuecomment-1128387551

arcticbear:
@maxagin 
There must be another red color for showing an error on the label which will be better to read. This one is really tough to read, it's painful for the eyes. A suggestion is to use a more light red tone.

Tooltip aligning by the center is confusing. Currently, all the tooltips are aligned by their left side to the parent element, which is consistent with menus and other elements' behavior.

maxagin:
> There must be another red color for showing an error on the label which will be better to read. @arcticbear 

We are trying to follow the native roles as much as possible - VSCode guidelines 

<img width="791" alt="Screen Shot 2022-07-07 at 1 59 21 AM" src="https://user-images.githubusercontent.com/98249521/177702612-3c9dcf23-e913-4d85-a316-fcf344e7cb36.png">

However, for this particular case, we probably won't use red at all (WIP).

---

>Tooltip aligning by the center is confusing.

The context menus are always aligned by the right side, unless it's very close to the right age of the screen, in this case, the context menu will appear on the left side of the mouse. However, in some rear cases, the context menu appears always on the right side.

Regarding the tooltips with content that can be copied or having other actions such as links will be placed at the center of the element. This is consistent with common interaction patterns including GH and Studio

shcheklein:
When `dvc plots show --json` fails like this:

```
(.venv) ?254 Projects/ensemble-dvc-template % dvc plots show --json
ERROR: unexpected error - Field 'epoch' does not exist in provided data.

[...]

PR #2072, opened 2022-07-20
Highlight experiments with errors

mattseddon:
This PR contains the first iteration of #1636.

Users will now be able to see experiment errors in the UI.

### Demo

https://user-images.githubusercontent.com/37993418/180121230-8bc49856-577b-48bc-a69e-6da4b57e33ad.mov

Things to note:

1. We are not suppressing the selection of an experiment for plotting even if the experiment is completely broken.
2. We are currently not highlighting experiment errors in the plots webview. This will be covered under #1649.
3. I updated the test fixture to contain error records which blew out the size of this PR (+~200 is for the test fixture by itself).
4. Styles match between the webview and the tree.
5. Errors cannot currently be filtered from the experiments table due to the hack we put in to stop queued items from being filtered by missing metrics.


extension/src/experiments/columns/extract.ts
```diff
@@ -46,15 +66,34 @@ const extractDeps = (
...
+  experiment: ExperimentFields,
+  branch?: Experiment
+): Columns => {
+  const metricsData = extractMetricsOrParams(experiment.metrics)
+  const paramsData = extractMetricsOrParams(experiment.params)
+
+  const error = [
+    ...(metricsData?.errors || []),
+    ...(paramsData?.errors || [])
+  ].join('\n')
```
mattseddon, line 86: [I] Test this bit and the tooltip showing up in the UI as markdown.

extension/src/test/suite/experiments/model/filterBy/tree.test.ts
```diff
@@ -95,7 +95,7 @@ suite('Experiments Filter By Tree Test Suite', () => {
               )
             })
             .map(experiment =>
-              experiment.queued
+              experiment.queued || experiment.error
```
mattseddon, line 98: [F] This points out that the way we don't filter queued will spill over onto error records. You cannot filter by an undefined value which means that "full error" records will be "unfilterable".

extension/src/experiments/webview/contract.ts
```diff
@@ -16,19 +16,20 @@ export interface DepColumns {
 }
 
 export interface Experiment extends BaseExperimentFields {
+  deps?: DepColumns
```
mattseddon, line 19: [F] Sorted and added `error?: string`

extension/src/test/fixtures/expShow/noErrors.ts
```diff
@@ -0,0 +1,24 @@
...
+      }
+    }
+    expShowFixtureWithoutErrors[sha] = { baseline, ...experiments }
+  }
+  return expShowFixtureWithoutErrors
+}
+
+const data = excludeErrors()
+
+export default data
```
mattseddon, line 24: [F] Without this extra fixture there would be a lot of changes to plots tests which are currently unrelated to this change.

webview/src/experiments/components/table/Errors.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
```
mattseddon, line 1: [Q] Better placement for this? Should it be a shared component? We will probably use something very similar for plots.

extension/src/test/fixtures/expShow/noErrors.ts
mattseddon: And this is why the likes of `extension/src/test/suite/plots/index.test.ts` are now pointed at this.

mattseddon: When I come back to do plots errors I should be able to remove this. As we'll want to highlight the behaviour that we can have experiment errors but still try to plot those experiments.

extension/src/experiments/index.ts
```diff
@@ -16,7 +16,7 @@ import { ExperimentsData } from './data'
 import { askToDisableAutoApplyFilters } from './toast'
 import { Experiment, ColumnType, TableData } from './webview/contract'
 import { WebviewMessages } from './webview/messages'
-import { DecorationProvider } from './model/filterBy/decorationProvider'
+import { DecorationProvider } from './model/decorationProvider'
```
mattseddon, line 19: [F] Move to a more central location as the decorations are now for both errors and filtered experiments

extension/src/experiments/model/decorationProvider.ts
```diff
@@ -38,19 +43,27 @@ export class DecorationProvider
   }
 
   public provideFileDecoration(uri: Uri): FileDecoration | undefined {
+    if (this.errors.has(uri.fsPath)) {
+      return DecorationProvider.DecorationError
```
mattseddon, line 47: [F] Take priority but under the current implementation, we can't have both.

extension/src/test/fixtures/expShow/output.ts
```diff
@@ -1545,6 +1550,119 @@ const data: ExperimentsOutput = {
         timestamp: '2020-12-29T15:26:36'
       }
     },
+    [errorShas[0]]: {
+      error: {
+        type: 'YAMLFileCorruptedError',
+        msg: "unable to read: 'params.yaml', YAML file structure is corrupted"
```
mattseddon, line 1556: [F] The fixture now contains one "full error" and one "partial error".

maxagin:
> Styles match between the webview and the tree.

@mattseddon this is only for now? Right? Will we have the styles for the table as we planned?

mattseddon:
> > Styles match between the webview and the tree.
> 
> @mattseddon this is only for now? Right? Will we have the styles for the table as we planned?

I have implemented everything as per my latest comments in the spec. The main point here is to highlight to users that there is an error and give them information about the error. We can adjust later when we get plot errors or user feedback.

webview/src/experiments/components/table/Errors.tsx
sroy3: We can move it in the shared folder when it's used in both webviews.

webview/src/shared/components/icons/Error.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
```
sroy3, line 1: We decided to remove the`AllIcons` enum that gave us the ability to view all available icons since it was duplicating the index file in the icons folder. We should either export this inside the index file or simply remove the index file.

sroy3 (approved):

julieg18 (approved):

webview/src/shared/components/icons/Error.tsx
mattseddon: I think we have a couple of other direct imports. I'll track them down today and move all imports to come from the index file. Let me know if you'd like to go the other way.

PR #2082, opened 2022-07-22
Add more resolutions for security advisories in dev dependencies

mattseddon:
Should close 4 security advisories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,9 @@
     "typescript": "4.7.4"
   },
   "resolutions": {
+    "undici": "5.8.0",
     "trim": "1.0.1",
+    "trim-newlines": "3.0.1",
     "@types/react": "18.0.15"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -67,7 +67,9 @@
     "typescript": "4.7.4"
   },
   "resolutions": {
+    "undici": "5.8.0",
     "trim": "1.0.1",
+    "trim-newlines": "3.0.1",
```
mattseddon, line 72: [F] This comes in through storybook. Looks like we don't hit the path anywhere.

package.json
```diff
@@ -67,7 +67,9 @@
     "typescript": "4.7.4"
   },
   "resolutions": {
+    "undici": "5.8.0",
```
mattseddon, line 70: [F] Introduced by `wdio`. As long as the e2e tests pass we are ok.

julieg18 (approved):

sroy3 (approved):

PR #2083, opened 2022-07-22
Move divider to top of group in experiments table context menu

mattseddon:
When demoing the experiments table yesterday I noticed that one of the context menu dividers was in the wrong place. This PR fixes the issue and adds a context menu story.

### Screenshot

image

### DIFF
--- a/webview/src/experiments/components/table/Row.tsx
+++ b/webview/src/experiments/components/table/Row.tsx
@@ -124,13 +124,13 @@ const getRunResumeOptions = (
     withId(
       'Modify, Reset and Run',
       MessageFromWebviewType.VARY_EXPERIMENT_PARAMS_RESET_AND_RUN,
-      !isNotCheckpoint || !projectHasCheckpoints
+      !isNotCheckpoint || !projectHasCheckpoints,
+      !hideVaryAndRun
     ),
     withId(
       projectHasCheckpoints ? 'Modify and Resume' : 'Modify and Run',
       MessageFromWebviewType.VARY_EXPERIMENT_PARAMS_AND_RUN,
-      !isNotCheckpoint,
-      !hideVaryAndRun
+      !isNotCheckpoint
     ),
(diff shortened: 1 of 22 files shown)


webview/src/stories/Table.stories.tsx
```diff
@@ -101,22 +118,19 @@ WithMiddleStates.play = async ({ canvasElement }) => {
...
+  tableData: noRunningExperiments
+}
+
+export const WithContextMenu = Template.bind({})
+WithContextMenu.args = {
+  tableData: noRunningExperiments
+}
+WithContextMenu.play = async ({ canvasElement }) => {
+  const experiment = await within(canvasElement).findByText('[exp-e7a67]')
+  userEvent.click(experiment, {
```
mattseddon, line 130: The context menu appears to the left of the element clicked on. I spent about 30 minutes looking into a fix. I don't think there is much value in spending any more time looking.

julieg18 (approved):

sroy3 (approved):

webview/src/experiments/components/table/Row.tsx
```diff
@@ -124,13 +124,13 @@ const getRunResumeOptions = (
...
-      !isNotCheckpoint || !projectHasCheckpoints
+      !isNotCheckpoint || !projectHasCheckpoints,
+      !hideVaryAndRun
     ),
     withId(
       projectHasCheckpoints ? 'Modify and Resume' : 'Modify and Run',
       MessageFromWebviewType.VARY_EXPERIMENT_PARAMS_AND_RUN,
-      !isNotCheckpoint,
-      !hideVaryAndRun
+      !isNotCheckpoint
```
mattseddon, line 133: This needs a divider when there are no checkpoints.

mattseddon: This is hard to read.

PR #2085, opened 2022-07-22
Refactor experiments column conditional shadow logic

julieg18:
* updates naming, making logic more clear
* reduces complexity with `headerGroup` data

Followup to [#2062 (comment)](https://github.com/iterative/vscode-dvc/pull/2062#discussion_r926129993)


webview/src/experiments/components/table/TableHead.tsx
```diff
@@ -102,6 +101,8 @@ export const TableHead = ({
 
   const selectedForPlotsCount = getSelectedForPlotsCount(rows)
 
+  const firstExpColumnCellId = headerGroups[0].headers[0].id
```
julieg18, line 104: Instead of relying on array indexes to find the first experiments column cell, I realized I could find the cell with the `id` of the cell inside `headerGroup`.

webview/src/experiments/components/table/TableHeader.tsx
```diff
@@ -130,7 +130,7 @@ const getIconMenuItems = (
   }
 ]
 
-const FirstTableHeaderCellWrapper: React.FC<{
+const WithExpColumnNeedsShadowUpdates: React.FC<{
```
julieg18, line 133: I tried to make the naming more clear but this sounds weird to me as well... Open to other suggestions 😅

mattseddon (approved):

PR #2086, opened 2022-07-22
Add getParentelem helper for tests

sroy3:
Following https://github.com/iterative/vscode-dvc/pull/2064#discussion_r927200095

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -50,6 +50,7 @@ import {
   selectedRows,
   setTableData
 } from '../../test/experimentsTable'
+import { getParentElement } from '../../test/nodes'
 
 jest.mock('../../shared/api')
 jest.mock('../../util/styles')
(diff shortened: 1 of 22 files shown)


julieg18 (approved):

webview/src/experiments/components/App.test.tsx
```diff
@@ -182,12 +183,7 @@ describe('App', () => {
     const headerB = screen.getByText('B')
     const headerD = screen.getByText('D')
 
-    dragAndDrop(
-      headerB,
-      // eslint-disable-next-line testing-library/no-node-access
-      headerD.parentElement?.parentElement || headerD,
-      DragEnterDirection.AUTO
-    )
+    dragAndDrop(headerB, getParentElement(headerD, 2), DragEnterDirection.AUTO)
```
rogermparent, line 186: I think a [testing-library custom query](https://testing-library.com/docs/dom-testing-library/api-custom-queries/) might be a better level of abstraction for this, both in ease of use and the ability to easily adapt to changes in the experiments table's DOM in the future.

mattseddon (approved):

webview/src/experiments/components/App.test.tsx
sroy3: Thanks for the suggestion! I'll change this.

sroy3:
Following @rogermparent's suggestion, I removed `getParentElem` and used a custom query instead. It does not need to disable any eslint rules, it is easier to reuse for that use case (does what it says it does) and gives meaningful error messages when failing.

PR #2087, opened 2022-07-22
Fix cell tooltip position on table column resize

### LINKED ISSUE #2037: If a column has resized, the hint is showing not in the right place
image

julieg18:
https://user-images.githubusercontent.com/43496356/180488182-17354f71-758c-4aa3-9238-a8b84a748725.mov

Fixes #2037


wolmir (approved):

mattseddon (approved):
We just need to be wary of anything that gets left aligned. Not sure we have a test in place that will let us know we have a new bug. Might not be an issue. Just something to consider.

julieg18:
> We just need to be wary of anything that gets left aligned. Not sure we have a test in place that will let us know we have a new bug. Might not be an issue. Just something to consider.

This PR only updates the tooltips within the resizable table cells so we shouldn't be misaligning anything else :)

PR #2088, opened 2022-07-22
Style the ghost image of the comparison table dragged item

sroy3:
Last task for #1604 

https://user-images.githubusercontent.com/3683420/180842339-99f707eb-3e17-4c0c-8a1c-17f6c7eda298.mov


webview/src/util/styles.ts
```diff
@@ -21,3 +21,6 @@ export const alphaToHex = (color: string, alpha: number): string => {
     .toString(16)
     .slice(-2)}`
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
sroy3, line 25: Spent way to much time trying to find a type that work 100% in the context I'm using this and simply couldn't find anything. If anyone knows of something, I'd be happy to change it.

wolmir: I think you added an `Any` helper a while ago @sroy3 in util/objects.ts

sroy3: The problem here is not that it could be `any` as `string`, `number`... It's that there is a type which I'm currently unable to figure out. The closest I got was something like `keyof Omit<CSSStyleDeclaration, 'length' | 'parentRule'>`, but it does not work when assigning strings to it. `elem.style[prop as keyof Omit<CSSStyleDeclaration, 'length' | 'parentRule'>] = ''` will give the following error: `Type 'string' is not assignable to type '() => IterableIterator<string>'`

wolmir: I'm personally ok with using any here, unless you feel the type will be useful or needed soon.

Typescript feels sadistic sometimes

mattseddon (approved):

mattseddon:
🚢

PR #2097, opened 2022-07-25
Redux for experiments table drag and drop

sroy3:
As this reducer already existed, it was easy to remove the `DragDropContext` and just use the reducer. Things seem to work just as they did for drag and drop. @wolmir can you test and tell me if I missed anything. All tests were passing, so it should be fine, but we never know...

Starting work for #2103


julieg18 (approved):
Looks good to me!

wolmir:
Sorry @sroy3 I'll test it now

wolmir (approved):
Works great, thanks!

webview/src/stories/Table.stories.tsx
```diff
@@ -196,9 +203,11 @@ WithNoSortsOrFilters.args = {
...
 export const Scrolled: ComponentStory<typeof Experiments> = ({ tableData }) => {
   return (
-    <div style={{ height: 400, width: 600 }}>
-      <Experiments tableData={tableData} />
-    </div>
+    <Provider store={configureStore({ reducer: experimentsReducers })}>
+      <div style={{ height: 400, width: 600 }}>
+        <Experiments tableData={tableData} />
+      </div>
+    </Provider>
```
rogermparent, line 210: Maybe `Provider` could be added with a [decorator](https://storybook.js.org/docs/react/writing-stories/decorators)?

sroy3: That is a good idea, I'll add this in my follow up PR

PR #2101, opened 2022-07-26
Fix `yarn build` by reverting "chore(deps): update dependency vsce to v2.9.3"

julieg18:
Reverts iterative/vscode-dvc#2084 since updating the `vsce` package breaks `yarn build`

Error thrown when you try to run `yarn build`:
```
dvc:package: $ vsce package --yarn -o ./dvc.vsix
dvc:package:  ERROR  Command failed: yarn list --prod --json
dvc:package: {"type":"error","data":"No lockfile in this directory. Run `yarn install` to generate one."}
```

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -1556,7 +1556,7 @@
     "sinon-chai": "3.7.0",
     "ts-jest": "28.0.7",
     "ts-loader": "9.3.1",
-    "vsce": "2.9.3",
+    "vsce": "2.9.2",
     "vscode-uri": "3.0.3",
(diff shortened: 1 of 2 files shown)


rogermparent (approved):
I found [an interesting GitHub thread](https://github.com/microsoft/vscode-vsce/issues/300#issuecomment-978256849) about others running into and trying to work around issues with vsce and workspaces. Weird that we're only hitting it with this patch update.

sroy3 (approved):

PR #2108, opened 2022-07-27
Upgrade vsce

mattseddon:
I have made some changes that enable us to move to the most recent version of VSCE (see https://github.com/iterative/vscode-dvc/pull/2101 for why we need to make changes).

I have tested the build locally. Would be good if someone else can also verify.

Please LMK if you would like to see anything done differently.

Thanks.


extension/scripts/ensureYarnLock.js
```diff
@@ -0,0 +1,6 @@
+const { resolve } = require('path')
+const { ensureFileSync } = require('fs-extra')
+
+const yarnLock = resolve(__dirname, '..', 'yarn.lock')
+
+ensureFileSync(yarnLock)
```
mattseddon, line 6: [F] Added this instead of using `touch` for mythical 🪟s contributors.

extension/.vscodeignore
```diff
@@ -1 +1,17 @@
...
+src/**
+.eslintrc.js
+.gitignore
+.prettierignore
+.lintstagedrc.js
+.vscodeignore
+jest.config.js
+tsconfig.json
+webpack.config.ts
+yarn.lock
```
mattseddon, line 17: After upgrading...

With old setup:

image

With this:

image

julieg18 (approved):

sroy3 (approved):

PR #2112, opened 2022-07-27
Table data in redux

### LINKED ISSUE #2103: Add Redux to experiments table
Now that the plots state reside in Redux, we should do the same with the experiments table so that we can work on both webviews in the same way.

sroy3:
Closes #2103


webview/src/experiments/components/table/tableDataSlice.ts
```diff
@@ -0,0 +1,45 @@
...
+  initialState: tableDataInitialState,
+  name: 'tableData',
+  reducers: {
+    update: (state, action: PayloadAction<TableData>) => {
+      if (action.payload) {
+        return {
+          ...state,
+          ...action.payload,
+          hasData: true
+        }
```
sroy3, line 36: This is pretty much the same as it was before. An improvement here would be to go through previous columns, rows and objects to change the reference of only what changed. I will do that as a follow up.

julieg18 (approved):
Great work! Things look more readable with less props being passed around :)

mattseddon (approved):
[Q] How does performance compare? Looks good to go otherwise.

sroy3:
> [Q] How does performance compare? Looks good to go otherwise.

I'm not seeing any difference. It's pretty much just a drop in replacement since the whole thing changes when the data changes. Like I said I'll follow up to try to limit the changes to only what changes. That could help with really big project, but that is not something we will notice otherwise.

PR #2126, opened 2022-08-01
Simplify drag and drop integrations

### LINKED ISSUE #1743: Consolidate the shared/dragDrop module to fix all the duplicated code

sroy3:
Closes #1743 
Would be quite hard to make just one integration after all. I didn't realize that the one for the experiments table does not work as a container since draggable items are deeply nested inside markup. What I was able to do though is to clean up that integration to use only what is necessary and not depend on the other integration (mainly the `makeTarget` function).

### DIFF
--- a/webview/src/shared/components/dragDrop/Draggable.tsx
+++ b/webview/src/shared/components/dragDrop/Draggable.tsx
@@ -26,13 +26,12 @@ export const Draggable: React.FC<DraggableProps> = ({
   onDrop,
   onDragEnter,
   onDragStart
+  // eslint-disable-next-line sonarjs/cognitive-complexity
 }) => {
-  const groupStates = useSelector(
-    (state: ExperimentsState) => state.dragAndDrop.groups
+  const groupState = useSelector(
+    (state: ExperimentsState) => state.dragAndDrop.groups[group] || {}
   )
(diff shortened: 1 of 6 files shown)


webview/src/shared/components/dragDrop/Draggable.tsx
```diff
@@ -26,16 +24,14 @@ export const Draggable: React.FC<DraggableProps> = ({
   disabled,
   dropTarget,
   onDrop,
-  onDragOver,
+  onDragEnter,
   onDragStart
   // eslint-disable-next-line sonarjs/cognitive-complexity
```
mattseddon, line 29: [Q] Is this still needed?

mattseddon: from the looks probably but thought I would check

mattseddon (approved):
Let's close https://github.com/iterative/vscode-dvc/issues/1743 if this is as far as we are going.

webview/src/shared/components/dragDrop/Draggable.tsx
sroy3: It passes sonar tests but not codeclimate. I checked.

Issue #606, opened 2021-06-29
Handle gracefully DVC-tracked/SCM actions when DVC is already busy


shcheklein:
## Details

<img width="437" alt="Screen Shot 2021-06-29 at 2 49 56 PM" src="https://user-images.githubusercontent.com/3659196/123876551-626d1a00-d8f0-11eb-91b4-03681b4354af.png">

## Reproduce

Take a repo with large DVC-tracked directories, pull one, while it's running trigger the second pull.

## Expected

I think we should handle this gracefully. Block actions? Show that something else is running? Other ideas?

## Extension version

`0.1.0-7ed06bac1b0a1cbce9454d566a53a3f53c1056b5`

## DVC Version

```
DVC version: 2.4.2+61835f
---------------------------------
Platform: Python 3.8.9 on macOS-10.15.6-x86_64-i386-64bit
Supports: azure, gdrive, gs, hdfs, webhdfs, http, https, s3, ssh, oss
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s1
Caches: local
Remotes: https
Workspace directory: apfs on /dev/disk1s1
Repo: dvc, git
```

## Logs

<details>
[version: 0.1.0-7ed06bac1b0a1cbce9454d566a53a3f53c1056b5, 2021-06-29T21:29:53.477Z] > /Users/ivan/Projects/get-started-experiments/.env/bin/python dvc -h 
[version: 0.1.0-7ed06bac1b0a1cbce9454d566a53a3f53c1056b5, 2021-06-29T21:29:56.053Z] > /Users/ivan/Projects/get-started-experiments/.env/bin/python dvc diff --show-json 
[version: 0.1.0-7ed06bac1b0a1cbce9454d566a53a3f53c1056b5, 2021-06-29T21:29:56.577Z] > /Users/ivan/Projects/get-started-experiments/.env/bin/python dvc list .  --dvc-only --show-json 
[version: 0.1.0-7ed06bac1b0a1cbce9454d566a53a3f53c1056b5, 2021-06-29T21:29:58.356Z] > /Users/ivan/Projects/get-started-experiments/.env/bin/python dvc exp show --show-json failed. ERROR: failed to show experiments - Unable to acquire lock. Most likely another DVC process is running or was terminated abruptly. Check the page <https://dvc.org/doc/user-guide/troubleshooting#lock-issue> for other possible reasons and to learn how to resolve this.
[...]

PR #2128, opened 2022-08-02
Block users from attempting to run concurrent SCM commands

mattseddon:
Fixes https://github.com/iterative/vscode-dvc/issues/606.

The main approach here is to block all of the user-facing SCM commands from being available whenever one is running. This works well with the exception of the `"scm/title"` & `"view/title"` commands. Having these removed from the UI whilst commands are running looked unstable (almost like another bug). I've copied what the git extension does and turned these into no-ops for the time being.

### Demo

https://user-images.githubusercontent.com/37993418/182267033-b1a92c7b-f4fe-4c9c-853f-02bdde86a046.mov

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -561,7 +561,7 @@
         },
         {
           "command": "dvc.checkout",
-          "when": "dvc.commands.available && dvc.project.available && !dvc.commands.running"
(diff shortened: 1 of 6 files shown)


julieg18 (approved):
Looks good to me!

PR #2132, opened 2022-08-02
Update CODEOWNERS

mattseddon:
Rather than leaving @rogermparent  in the CODEOWNERS file and having him ignore all PRs in this repo, I think it would be better to remove him and specifically request reviews when we need help.

WDYT?


rogermparent:
I was thinking the same thing recently, I could be more helpful by only getting notifications when requested. If I want to see every PR, I can just look at the repo's PR page.

rogermparent (approved):

wolmir (approved):

Issue #2070, opened 2022-07-20
Experiments table: improve styles


maxagin:
EXPs table improvements. Original ticket https://github.com/iterative/design/issues/28

[Figma](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1631%3A119506)

### Rest state

1. Add a bottom line to every row and remove even rows BG style
2. Keep the header BG colour the same as the table body BG colour
3. Use the same line style for header item separators
4. Keep the caret the same colour as the other icons on the left side

### Hover state

1. Show hover styles (column, row) when the cell is on hover 
2. "lr" on hover. Also, show the hover style for the column
3. "global" on hover. Also show the hover style for the  "seed", "lr", "weight_decay" and three columns
4. The first column cell on hover. Checkbox and Star rest styles (same like without hover)

### Resize columns

1. When resizing the cell sideline grows from 1px to 2xp and changes colour
2. When resizing the hover styles are applied the same way as if the block is on hover.
3. The resize is always applied to the right side of the column
4. The resize is activated only in the header of the table when hovering on the sidelines

### Select and star EXP/s

1. When a selected row is on hover, add a shadow to show the hovered row
2. New (lighter) color for the selected row background. One of the reasons: we do not want to have it similar to the color that is used for the buttons
3. Selected rows have different bottom borders and left/right for selected cell color

### Star and checkbox behaviour

1. Star 
2. Checkbox

maxagin:
Note: In a composition with these changes, we may want to rethink the sticky panel shadow and implement one from the design. I think the best if we could see both examples. Screen recording (video) probably will do the job. cc @julieg18

julieg18:
> Note: In a composition with these changes, we may want to rethink the sticky panel shadow and implement one from the design. I think the best if we could see both examples. Screen recording (video) probably will do the job. cc @julieg18

When you say "stick panel shadow", are you referring to the header shadow? Currently, the experiments column shadow should match the design on Figma while the header shadow is a little different from the Figma design :)

maxagin:
> When you say "stick panel shadow", are you referring to the header shadow?

Yes @julieg18 . Thanks

> Currently, the experiments column shadow should match the design on Figma while the header shadow is a little different from the Figma design :)

Cool. I just wanted to say that it's going to be easier to decide about the shadow after other changes are implemented. However, if you want, you can implement it right now like in Figma :) 

[...]

PR #2133, opened 2022-08-02
Update table styles

### LINKED ISSUE #2032: Include the Workspace row into the fixed section along with the table header
Include both workspace and main (branch) into the fixed section on top https://github.com/iterative/vscode-dvc/issues/2032#issuecomment-1185097951

Will be done after  https://github.com/iterative/vscode-dvc/issues/2070 completed!

---

https://user-images.githubusercontent.com/98249521/175386685-67d7e96f-6326-46c3-8e1a-5c0a2bdda10d.mov

julieg18:
* add cell hover style
* add row hover styles
* add a bottom border to every row
* remove even row bg color
* keep header bg color the same as body bg
* use the same line style for header item separators
* keep row arrow the same color as left icons
* update workspace row, making it sticky and using default row styles

https://user-images.githubusercontent.com/43496356/184659276-ae458eeb-aec2-48f5-a164-4480c66e5352.mov

Part of #2070, Fixes #2032


shcheklein:
@julieg18 please, put a small demo / video in the description.

mattseddon:
Looks like there are some changes on the plots side: 

image

Were these intended changes?

julieg18:
> Were these intended changes?

Accidental! Should be fixed now!

webview/src/plots/components/ribbon/styles.module.scss
```diff
@@ -10,7 +10,7 @@
 .block {
   border-left-style: solid;
   border-left-width: 3px;
-  background-color: $header-bg-color;
+  background-color: var(--vscode-dropdown-background);
```
sroy3, line 13: We try to keep all colors inside the `variables.scss` file so that it's easy to see what's available and what we are using (so that we don't start having too many different colors). The fact that the new `header-bg-color` impacts plots as well as experiments might not be a big problem if the color fits.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -370,44 +357,15 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
...
-
-      &:last-child {
-        padding-right: $workspace-row-edge-margin;
-        .innerCell {
-          padding-right: $cell-padding;
-          border-right: $row-border;
-        }
-      }
+    &.withShadow {
+      box-shadow: 0 5px 8px -2px var(--vscode-widget-shadow);
```
sroy3, line 368: Another variable to move to `variables.scss`

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -298,6 +283,8 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
   }
 
   .bodyRow {
+    border-bottom: $row-border;
```
julieg18, line 286: I used our original border color for the row lines (which isn't black compared to the Figma). Do we want to use more a black border color to better match the Figma?

Figma:

image

Current: 

image

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -445,6 +401,11 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
     font-size: 0.8rem;
     line-height: 2rem;
     align-items: center;
+
+    &:hover {
```
julieg18, line 405: I finished most of the tasks in #2070 except for the column hover styles. Should I convert this pr to a draft until the column hover styles are added as well? 

image

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -461,6 +422,11 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
     display: flex;
     flex-flow: row nowrap;
     text-align: left;
+
+    &:hover {
```
julieg18, line 426: Do we want our rows to be highlighted on hover similar to the column hover styles? 

image

julieg18 (commented):
Made some comments based on the discussion in our demo meeting!

webview/src/experiments/components/table/styles.module.scss
maxagin: Obviously, Figma is the guide. We can jump on a call tomorrow to discuss the small details. However, the base theme needs to look like in Figma (including 0.5px borders). Thanks, @julieg18!

webview/src/experiments/components/table/styles.module.scss
maxagin: In this case, let's try to stick to the plan as much as possible. Thanks

webview/src/plots/components/ribbon/styles.module.scss
julieg18: > We try to keep all colors inside the variables.scss file so that it's easy to see what's available and what we are using (so that we don't start having too many different colors). 

Oh, makes sense! Thanks for letting me know!

> The fact that the new header-bg-color impacts plots as well as experiments might not be a big problem if the color fits.

Color did not fit (made them have no background at all) 😄

webview/src/experiments/components/table/Table.tsx
```diff
@@ -74,29 +75,57 @@ export const ExperimentGroup: React.FC<
...
+      ref={ref}
       {...instance.getTableBodyProps({
         className: cx(
           styles.rowGroup,
           styles.tbody,
-          row.values.id === 'workspace'
-            ? styles.workspaceRowGroup
-            : styles.normalRowGroup
+          styles.workspaceRowGroup,
+          needsShadow && styles.withShadow
```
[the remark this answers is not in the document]
julieg18: Thansk! I prefer the `&&` syntax for this instance :)

webview/src/experiments/components/table/TableHead.tsx
```diff
@@ -86,21 +86,22 @@ export const TableHead = ({
...
-    root,
-    rootMargin: '-15px 0px 0px 0px',
-    threshold: 1
-  })
 
   const selectedForPlotsCount = getSelectedForPlotsCount(rows)
 
   const firstExpColumnCellId = headerGroups[0].headers[0].id
 
+  const wrapper = useRef<HTMLDivElement>(null)
```
sroy3, line 94: Hooks are usually used at the start of a component. It may not happen here, but it's a good practice to avoid things like "rendered more hooks than during the previous render". This also includes the `useEffect` below. (I think that comes from the branch I pushed that we did while pairing, we didn't go back and clean stuff, oops sorry)

webview/src/experiments/components/table/styles.module.scss
sroy3: What does this look like on a high contrast theme? If we do not offer high contrast for at least these themes, it's not very good on our part.

sroy3 (approved):

webview/src/experiments/components/table/styles.module.scss
julieg18: >  However, the base theme needs to look like in Figma (including 0.5px borders, BG color).

Updated the border width to `0.5`! As for border colors, I looked into an alternative color that was more like black in the dark theme but couldn't find a color variable darker than the current table background color. I did look into using just `#000` but that didn't look good in some of the darker themes :(

> What does this look like on a high contrast theme? If we do not offer high contrast for at least these themes, it's not very good on our part.

On storybook, contrast themes looks good to me... 

image

image

webview/src/experiments/components/table/TableHead.tsx
julieg18: Whoops! Thanks for spotting that!

julieg18:
Had a call with Maxim and went over some desired changes to the design. Going to convert this into a draft for now!

julieg18:
With column styles taken off and styles updated (based off the [latest Figma](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1528%3A127749)), this pr is ready for another review!

sroy3:
This is default Light High Contrast theme. I don't think we should spend too much time on one specific theme, but I believe high contrast themes should stay high contrast for accessibility
Screen Shot 2022-08-15 at 9 17 22 AM

sroy3 (approved):

julieg18:
> I don't think we should spend too much time on one specific theme, but I believe high contrast themes should stay high contrast for accessibility

True! Updated the background color to a variable that is viewable on all themes! Though It does end up being the same color as the table on the light themes: 

image
image
image

webview/src/shared/variables.scss
```diff
@@ -3,7 +3,7 @@ $bg-color: var(--vscode-editor-background);
 $bullet-size: 9px;
 $watermark-color: var(--vscode-descriptionForeground);
 
-$border-color: var(--vscode-dropdown-border);
+$border-color: var(--checkbox-border);
```
mattseddon, line 6: [Q] Any ideas why some of these aren't prefixed with vscode?

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -150,18 +147,11 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
...
+
     .workspaceWithChanges.runningExperiment & {
       border-right-color: $changed-color;
       border-top-color: $changed-color;
     }
-
-    .runningExperiment.oddRow > .experimentCell & {
-      border-left: 1.5px solid $row-bg-alt-color;
-      border-bottom: 1.5px solid $row-bg-alt-color;
-    }
```
mattseddon, line 164: [I] Couple of cases to follow up on:

1. Middle of an unselected experiment's circle whilst the row is highlighted

Screen Shot 2022-08-16 at 9 33 44 am

2. Middle of a running experiment's spinner whilst highlighted

Screen Shot 2022-08-16 at 9 32 42 am

mattseddon (approved):
Happy to merge this now, would be good to get the follow-up items addressed before we ship this.

webview/src/shared/variables.scss
julieg18: Just realized this actually 😄, I'll update this variable with the vscode prefixed version in the follow-up.

Issue #1933, opened 2022-06-20
Plots sections: enable linking in the plots tooltips, other improvements


jorgeorpinel:
Plots section descriptions.

- [x] put links to DVC docs in those tooltips, probably means they should toggle by a click / ESC, not disappear
- [x] keep all three sections visible (for better education(, if section is empty, show the same message instead of it's content

_Originally posted by @shcheklein in https://github.com/iterative/vscode-dvc/pull/1922#pullrequestreview-1011585467

jorgeorpinel:
> I feel like this repo needs a `ux` issue label.

maxagin:
We had it before. I just want to understand what is the reason for changing as it makes sense if the description is dynamic. Please clarify it for me. Thanks

<img width="842" alt="Screen Shot 2022-06-21 at 4 22 44 PM" src="https://user-images.githubusercontent.com/98249521/174891003-6495d0cd-2fff-495c-9d6b-58e917706d01.png">

shcheklein:
> I think @shcheklein meant that instead of tooltips we should have the text visible under each Plot type header

No, we don't want this text to be permanently visible. Only as (potential) way for placeholders when there are not plots in the section.

tooltip should stay as a tooltip but react properly considering that there will be links

shcheklein:
I've updated the description, there was a bug -  `should not toggle` -> ` should toggle`

maxagin:
I think I have a good solution for this. I will share the concept soon

maxagin:
@jorgeorpinel , here is a solution I believe in :) Sharing the example below. Tomorrow I will design the example of a hint including a button (or another link example).

Concept: show hint on hover with the mouse over the transparent area so the user can move the cursor into the hint area without it disappearing. When the mouse is out of the hint area - we hide it.

This will work also for other similar sections, such as coming errors: DVC errors that may occur during or after running experiment #1636

Example from our favourite GitHub

https://user-images.githubusercontent.com/98249521/175198240-0b9379ca-967d-46e6-83d0-b645e4959874.mov

maxagin:
Minor: the hints are not context menus, this is why I suggest we will have them centred in relation to the hovered element. Thanks

maxagin:
[...]

PR #2140, opened 2022-08-03
Links in section description tooltips

sroy3:
Part of #1933 

This is an example with a link (this link is not in the actual code, links will need to be provided later)

https://user-images.githubusercontent.com/3683420/182685733-d89d1378-b4ae-4278-a243-3416fb39efb0.mov


demo/dvc.lock
```diff
@@ -17,18 +17,18 @@ stages:
         weight_decay: 0
     outs:
     - path: misclassified.jpg
-      md5: 5edd7dd13144d95ef72e270570d9d338
```
julieg18, line 20: Were the changes to `demo` unintentional?

julieg18 (commented):
The CI tests are failing, but is that unrelated to the PR?

demo/dvc.lock
sroy3: Thanks, I went too fast on that one, I didn't even played with the demo project so I didn't think of checking for a simple change 🤦🏼‍♀️

mattseddon (approved):

PR #2145, opened 2022-08-04
All plots sections visible

sroy3:
Part of #1933 
Screen Shot 2022-08-04 at 3 06 09 PM
Screen Shot 2022-08-04 at 3 05 30 PM
Screen Shot 2022-08-04 at 3 04 32 PM


webview/package.json
```diff
@@ -47,7 +47,6 @@
     "@storybook/manager-webpack5": "6.5.9",
     "@storybook/preset-scss": "1.0.3",
     "@storybook/react": "6.5.9",
-    "@storybook/testing-library": "0.0.13",
```
sroy3, line 50: This is there twice in the list

mattseddon, line 50: I think you'll need to pin the other version (maybe just remove that one as it is out of order)

webview/src/plots/components/App.test.tsx
```diff
@@ -192,29 +212,27 @@ describe('App', () => {
...
-    })
-
-    sendSetDataMessage({
       template: templatePlotsFixture
     })
 
     expect(screen.queryByText('Loading Plots...')).not.toBeInTheDocument()
     expect(screen.getByText('Trends')).toBeInTheDocument()
-    expect(screen.getByText('Data Series')).toBeInTheDocument()
+    expect(screen.getByText('No Metrics Selected')).toBeInTheDocument()
```
mattseddon, line 235: [Q] Can we hide the context menu dropdown(s) when there are no actual plots for the project that we are showing? I think that at the moment the "filter metrics" dropdown would show as empty?

webview/src/plots/components/App.test.tsx
```diff
@@ -246,6 +264,7 @@ describe('App', () => {
   })
 
   it('should toggle the checkpoint plots section in state when its header is clicked', async () => {
+    // Fails when other test above
```
mattseddon, line 267: [Q] Missed this one?

webview/src/plots/components/App.test.tsx
```diff
@@ -1108,7 +1147,11 @@ describe('App', () => {
 
         beforeEach(async () => {
           // eslint-disable-next-line testing-library/no-render-in-setup
```
mattseddon, line 1149: [C] If we're ok with this we should disable the rule globally.

mattseddon (approved):
For checkpoint plots, I think we need to differentiate between no metrics being selected and there being no plots. Extending that we should probably have different text based on whether or not the project has those types of plots.

mattseddon:
@sroy3 could be unrelated to this particular change but virtualized plots seems broken in the storybook

julieg18 (approved):

sroy3:
> @sroy3 could be unrelated to this particular change but virtualized plots seems broken in the storybook

The only difference is the Images section being added between the two other sections.

sroy3:
> For checkpoint plots, I think we need to differentiate between no metrics being selected and there being no plots. Extending that we should probably have different text based on whether or not the project has those types of plots.

Changed it for when there is no data for checkpoint plots

webview/src/plots/components/App.test.tsx
sroy3: Done

webview/src/plots/components/App.test.tsx
sroy3: 🤦🏼‍♀️

webview/src/plots/components/App.test.tsx
sroy3: Done

PR #2146, opened 2022-08-04
Dvc yaml support - Autocomplete + Snippets

wolmir:
What's done:

Params suggestions and Vars

What's left:

Suggest folders and files
Maybe more snippets
Fix bugs


wolmir:
We also need to parse TOML and .py files for parameters

.vscode/dvcYamlSchema.json
```diff
@@ -0,0 +1,472 @@
+{
+  "title": "dvc.yaml",
```
wolmir, line 2: I will delete these things, we're not using it at the moment

extension/src/dvcYaml/index.ts
```diff
@@ -0,0 +1,130 @@
+import { join } from 'path'
+import {
```
wolmir, line 2: Lots of changes here, but the plan is to move this code to the dvcYaml module, like it was done for all the other modules we have.

extension/src/dvcYaml/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  TextDocument,
+  Uri,
+  workspace,
+  WorkspaceEdit
+} from 'vscode'
+import { DvcYamlSupport, DvcYamlSupportWorkspace } from './support'
+import { loadText } from '../fileSystem'
+import { findFiles } from '../fileSystem/workspace'
+
+export class DvcYamlDocumentLinkProvider implements DocumentLinkProvider {
```
wolmir, line 30: This provider will be deleted for this PR.

extension/src/dvcYaml/index.ts
```diff
@@ -0,0 +1,130 @@
...
+      completion.insertText = new SnippetString(`${item.completion}`)
+      completion.filterText = item.completion
+      completions.push(completion)
+    }
+
+    return new CompletionList(completions)
+  }
+}
+
+export class DvcYamlHoverProvider implements HoverProvider {
```
wolmir, line 101: Same here

extension/src/experiments/context.ts
```diff
@@ -36,7 +36,7 @@ const setContextOnDidChangeActiveEditor = (
       return
     }
 
-    const isParamsFile = getParamsFiles().has(path)
+    const isParamsFile = getParamsFiles().has(path) || path.includes('dvc.yaml')
```
wolmir, line 39: I don't think I used this change in the end. I'll remove it too

extension/src/extension.ts
```diff
@@ -432,7 +446,35 @@ export class Extension extends Disposable implements IExtension {
...
+      dvcYamlSelector,
+      new DvcYamlDocumentLinkProvider()
+    ),
+    languages.registerCompletionItemProvider(
+      dvcYamlSelector,
+      new DvcYamlCompletionProvider(),
+      '.',
+      '/'
+    ),
+    languages.registerHoverProvider(
```
wolmir, line 465: These extra providers will be removed for this PR

extension/src/fileSystem/dvcYaml/dvcYaml.ts
```diff
@@ -0,0 +1,240 @@
+import { load } from 'js-yaml'
+import { Any } from '../../util/object'
+
+export interface DvcYAML {
```
wolmir, line 4: This was auto-generated based on the schema, but I had to fix some of the definitions.

extension/src/fileSystem/dvcYaml/dvcYamlSchema.json
```diff
@@ -0,0 +1,472 @@
+{
```
wolmir, line 1: Will delete this too

extension/src/fileSystem/dvcYaml/index.ts
```diff
@@ -0,0 +1,55 @@
+import { join } from 'path'
```
wolmir, line 1: Didn't use this either, will remove it too

extension/src/fileSystem/index.ts
```diff
@@ -102,6 +102,14 @@ export const loadJson = <T>(path: string): T | undefined => {
   }
 }
 
+export const loadText = (path: string): string | undefined => {
```
wolmir, line 105: This will be removed as well

extension/src/util/object.ts
```diff
@@ -26,3 +26,7 @@ export const createTypedAccumulator = <T extends string>(
   }
   return acc
 }
+
+type BaseType = string | number | boolean | Object | undefined | null
```
wolmir, line 30: I needed these guys to fix the auto-generated types, so I copied them from the webview module

extension/src/vscode/workspace.ts
```diff
@@ -1,4 +1,7 @@
-import { workspace, WorkspaceFolder } from 'vscode'
+import { TextDocument, workspace, WorkspaceFolder } from 'vscode'
 
 export const getWorkspaceFolders = (): readonly WorkspaceFolder[] =>
   workspace.workspaceFolders || []
+
+export const getWorkspaceFiles = (): readonly TextDocument[] =>
```
wolmir, line 6: Didn't use; will remove

extension/syntaxes/dvcYaml.tmGrammar.json
```diff
@@ -0,0 +1,50 @@
+{
+  "scopeName": "source.yaml",
```
wolmir, line 2: This was an experiment to add highlighting to specific dvc.yaml keywords, but it needs improvement/thinking/discussion so I'll remove it for this PR too

webview/src/shared/components/dragDrop/DragDropContainer.tsx
```diff
@@ -12,10 +12,10 @@ import { changeRef, setDraggedOverGroup } from './dragDropSlice'
 import styles from './styles.module.scss'
 import { DropTarget } from './DropTarget'
 import { getIDIndex, getIDWithoutIndex } from '../../../util/ids'
-import { Any } from '../../../util/objects'
 import { PlotsState } from '../../../plots/store'
 import { getStyleProperty } from '../../../util/styles'
 import { idToNode } from '../../../util/helpers'
+import { Any } from 'dvc/src/util/object'
```
wolmir, line 18: This was collateral damage, will revert it

extension/snippets.json
```diff
@@ -0,0 +1,40 @@
+{
+  "Basic": {
+    "prefix": ["basic"],
```
mattseddon, line 3: For these, I think we should have a super short version as well. The [Jest Snippets](https://marketplace.visualstudio.com/items?itemName=andys8.jest-snippets) extension that I use has 2/3 letter acronyms for each snippet.

image

The longer prefix should also mention the fact that it is a snippet for a stage.

Here are some more examples:

- [ES7+ React/Redux/React-Native snippets](https://github.com/dsznajder/vscode-react-javascript-snippets/blob/HEAD/docs/Snippets.md) (5.6m)
- [VS Code JavaScript (ES6) snippets](https://github.com/xabikos/vscode-javascript)

extension/src/dvcYaml/index.ts
```diff
@@ -0,0 +1,130 @@
+import { join } from 'path'
```
mattseddon, line 1: It would be good to rename all `dvcYaml` directories to `pipelines` as that is the high-level concept that these YAML files belong to. We will be able to add DAG-related code, etc to these directories later.

extension/src/extension.ts
```diff
@@ -432,7 +446,35 @@ export class Extension extends Disposable implements IExtension {
...
 export function activate(context: ExtensionContext): void {
   extension = new Extension(context)
-  context.subscriptions.push(extension)
+  const dvcYamlSelector = {
+    language: 'yaml'
+  }
+
+  context.subscriptions.push(
+    extension,
+    languages.registerDocumentLinkProvider(
```
mattseddon, line 455: [Q] Does this return a disposable? Can we use the normal `dispose.track` pattern?

extension/src/util/object.ts
mattseddon: I think we can put them into a central contract file and import them into the webview instead of duplicating them.

extension/src/fileSystem/dvcYaml/index.ts
mattseddon: We would have only wanted one instance of `extension/src/fileSystem/data/index.ts` per repository anyway.

extension/src/dvcYaml/index.ts
mattseddon: I would try to avoid adding files to `extension/src/fileSystem` as well if possible.

extension/src/extension.ts
mattseddon: [Q] Did you look at `registerDefinitionProvider`? Might be able to use that to step into a file?

mattseddon: or you were planning to use `registerDocumentSymbolProvider`?

wolmir: Both, actually, I think these definition/symbol providers will be easier to do

extension/src/dvcYaml/index.ts
wolmir: Thanks! I was looking for the right name but I thought it would be better to just be literal and ask about it later.

extension/src/extension.ts
wolmir: Yes, this was the fastest way to get it working, but I'm currently integrating this with the Extension class like the other modules

mattseddon:
@wolmir can you please either close this or pull out the snippets so we can review and merge.

PR #2091, opened 2022-07-25
Use new data status command

mattseddon:
# 1/5 `main` <- this <- #2151 <- #2266 <- #2267 <- #2299

This PR integrates our DVC Tracked tree, SCM View and File Decorations with the new `dvc data status` command. That means we no longer have to call `dvc status`, `dvc list` & `dvc diff` 🎉. We also don't need the help of the `exp show` data to work out which paths we can feed directly into `dvc pull` without causing errors.

**I have found a few issues with the way that the new command behaves and I have raised issues in the DVC repository**:

1. https://github.com/iterative/dvc/issues/8052
2. https://github.com/iterative/dvc/issues/8053
3. https://github.com/iterative/dvc/issues/8062

~None of the above issues are blockers as I have been able to mostly patch over the problems.~ The blocking issue that I cannot fix is not all paths are returned when there is no cache (part of https://github.com/iterative/dvc/issues/8062).

### Other Blockers:

We need https://github.com/iterative/dvc/issues/8051 to be fixed and released by DVC before we can safely increment our minimum version of the CLI. That needs to happen before we merge and release this PR. 

**Note:** There are some other fixes in DVC's `main` branch that we also need to be made available before we can safely increment our minimum version.

### Demos:

**Not in cache**

https://user-images.githubusercontent.com/37993418/181851488-8e8d4961-8c71-442e-ac0f-8d37c6ff0216.mov

**Add Data**

https://user-images.githubusercontent.com/37993418/181852487-435ca9a6-7855-41aa-9693-847790ce2473.mov

**Commit modified data**

https://user-images.githubusercontent.com/37993418/181853496-d513f445-52a4-4cf2-985a-3f8c9e66cee8.mov

PR #2151, opened 2022-08-06
Refine data status consumption

mattseddon:
# 2/5 `main` <- #2091 <- this <- #2266 <- #2267 <- #2299

This PR improves our use of the new data status command. 

We now:

- Show resources in multiple SCM groups when they are returned by data status.
- Apply a single decoration to each resource based on a prioritized list. We apply not in cache first, then uncommitted, then committed statuses the implementation matches the git extension's approach. 
- Do not show checkout inline for resources that are uncommitted and also not in cache (action would not work).

### Demo

https://user-images.githubusercontent.com/37993418/183329636-d9b3e429-0a92-4c61-8cab-95e66c2d63ea.mov

**Note:** Still need https://github.com/iterative/dvc/issues/8062 to be fixed before I can provide a non-confusing demo of not in cache behaviour.

### DIFF
--- a/extension/src/repository/model/collect.ts
+++ b/extension/src/repository/model/collect.ts
@@ -2,7 +2,7 @@ import { join, relative, resolve } from 'path'
 import { Uri } from 'vscode'
 import { Resource } from '../commands'
 import { addToMapSet } from '../../util/map'
-import { DataStatusOutput } from '../../cli/dvc/reader'
+import { Changes, DataStatusOutput } from '../../cli/dvc/reader'
 import { DecorationDataStatus } from '../decorationProvider'
 import { relativeWithUri } from '../../fileSystem'
 import {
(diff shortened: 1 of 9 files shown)


extension/src/repository/model/collect.ts
```diff
@@ -105,83 +105,122 @@ const addToTracked = (
...
+}
 
-  return dataStatus
+export const collectDataStatus = (
+  dvcRoot: string,
+  dataStatusOutput: DataStatusOutput & { untracked?: string[] }
+): DataStatusAccumulator => {
+  const acc = createDataStatusAccumulator()
+
+  for (const groupedStatuses of mapGroupedStatuses(dataStatusOutput)) {
```
mattseddon, line 209: [F] We now pre-process committed & uncommitted differently than we do unchanged, untracked & not in cache. We do this because paths should be unique in each of those groups but they can be duplicated across groups. After creating an object of the format `{ [path: string]: Status }` for each group we pass it into the old process and fill in missing ancestors. We then transform the data into `{ [status: Status]: Set<string> }` for use in the rest of the subsystem.

sroy3 (approved):

julieg18 (approved):

Issue #1720, opened 2022-05-16
Story:  Help navigate in a lot of experiments


maxagin:
These comments are the only meaningful thing left in this ticket

![image](https://user-images.githubusercontent.com/37993418/223578217-a1de5a87-2945-42f1-a57b-84128e583e64.png)

----

outdated:

This epic is created so we can discuss the concepts together with the DVC team.

### Ideas

- [x] Star experiments in table #1676
- [ ] Pin experiments to show them at the top (won't do)
- [ ] Be able to drag and drop rows pinned expS (won't do)
- [ ] Naming/renaming experiments after execution (won't do)
- [ ] Able to add context to an experiment (labels https://github.com/iterative/vscode-dvc/issues/1720#issuecomment-1174473366, custom labels, pin, group, comments, free text, notes about experiments etc) (won't do)
- [x] Collapse exps and checkpoints #2071 & Show a limited amount of checkpoints that can optionally be expanded further #1269

### Figma specs

- [Star experiments in table](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1004%3A18606)
- [Label experiments in the table](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1075%3A20307)
- [Collapse exps and checkpoints](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1230%3A36058)
- [Group interesting runs in sections with a title](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1232%3A40813)
- [Able to add context to an experiment (comments)](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1227%3A40148)

maxagin:
Labels (using GH colour scheme), just so we can keep this ticket alive. Would probably make sense to sort by label, the same way we are planning to show only starred expS.

<img width="1586" alt="Screen Shot 2022-07-04 at 8 19 01 PM" src="https://user-images.githubusercontent.com/98249521/177227927-7b2a19e8-866a-4f9b-8d1b-6bed6b19b85d.png">

maxagin:
Hey @shcheklein . As a part of the task "Help navigate in a lot of experiments" I think that being able to fold not interesting EXPs can be useful. Please let me know WDYT. 
[...]

PR #2164, opened 2022-08-09
Add option to filter experiments to starred

mattseddon:
# 1/4 `main` <- this <- #2169 <- #2170 <-#2171

This PR gives users the ability to filter experiments by their starred status.

### Demo

https://user-images.githubusercontent.com/37993418/183581276-2464f171-174f-47c8-bc80-b1d582f8142d.mov

### DIFF
--- a/extension/src/experiments/columns/like.ts
+++ b/extension/src/experiments/columns/like.ts
@@ -1,0 +1,21 @@
+import { ColumnLike } from './quickPick'
+import { Column } from '../webview/contract'
+
+const starredColumnLike = {
+  label: '$(star-full)',
+  path: 'starred',
+  types: ['boolean']
+}
+
+export const addStarredToColumns = (
+  columns: Column[] | undefined
+): ColumnLike[] | undefined => {
+  if (!columns?.length) {
+    return
+  }
+
+  return [
+    ...columns.map(({ label, path, types }) => ({ label, path, types })),
+    starredColumnLike
+  ]
+}
--- a/extension/src/experiments/columns/like.ts
+++ b/extension/src/experiments/columns/like.ts
@@ -1,7 +1,8 @@
-import { ColumnLike } from './quickPick'
 import { Column } from '../webview/contract'
 
-const starredColumnLike = {
+export type ColumnLike = { label: string; path: string; types?: string[] }
+
+const starredColumnLike: ColumnLike = {
   label: '$(star-full)',
   path: 'starred',
   types: ['boolean']
--- a/extension/src/experiments/model/filterBy/quickPick.ts
+++ b/extension/src/experiments/model/filterBy/quickPick.ts
@@ -4,7 +4,7 @@ import { getInput } from '../../../vscode/inputBox'
 import { quickPickManyValues, quickPickValue } from '../../../vscode/quickPick'
 import { Title } from '../../../vscode/title'
 import { Toast } from '../../../vscode/toast'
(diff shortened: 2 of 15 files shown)


extension/src/experiments/model/filterBy/quickPick.ts
```diff
@@ -84,9 +83,11 @@ const addFilterValue = async (path: string, operator: Operator) => {
 }
 
 export const pickFilterToAdd = async (
-  columns: Column[] | undefined
+  columns:
+    | { label: string; path: string; types: string[] | undefined }[]
```
mattseddon, line 87: [F] needs updated

extension/src/experiments/model/index.ts
```diff
@@ -477,12 +479,14 @@ export class ExperimentsModel extends ModelWithPersistence {
   }
 
   private splitExperimentsByQueued(getQueued = false) {
-    return this.flattenExperiments().filter(({ queued }) => {
-      if (getQueued) {
-        return queued
-      }
-      return !queued
-    })
+    return this.flattenExperiments()
```
mattseddon, line 482: [F] this all needs tidied up

extension/src/experiments/model/index.ts
```diff
@@ -128,7 +128,7 @@ export class ExperimentsModel extends ModelWithPersistence {
```
mattseddon, line 128: [F] The main change in this file was to encapsulate access to experiments and checkpoints so that details (selected and starred) are always added.

extension/src/experiments/columns/like.ts
```diff
@@ -0,0 +1,22 @@
+import { Column } from '../webview/contract'
+
+export type ColumnLike = { label: string; path: string; types?: string[] }
```
mattseddon, line 3: [F] This definition is used in the quick pick. It means that we don't have to create a full mocked `Column` for "starred".

extension/src/experiments/model/index.ts
```diff
@@ -128,7 +128,7 @@ export class ExperimentsModel extends ModelWithPersistence {
 
   public toggleStatus(id: string) {
     if (
-      this.flattenExperiments().find(({ id: queuedId }) => queuedId === id)
+      this.getFlattenedExperiments().find(({ id: queuedId }) => queuedId === id)
```
mattseddon, line 131: [F] The main change in this file was to encapsulate access to experiments and checkpoints so that details (selected and starred) are always added.

mattseddon:
Related to #1720

extension/src/experiments/columns/like.ts
```diff
@@ -0,0 +1,22 @@
...
+const starredColumnLike: ColumnLike = {
+  label: '$(star-full)',
+  path: 'starred',
+  types: ['boolean']
+}
+
+export const addStarredToColumns = (
+  columns: Column[] | undefined
+): ColumnLike[] | undefined => {
+  if (!columns?.length) {
```
sroy3, line 14: Didn't we have a lint rule where it said we have to compare length to 0?

julieg18 (approved):

sroy3 (approved):

extension/src/experiments/columns/like.ts
mattseddon: I think the optional chain will be messing with it. You would need to write 
`columns?.length === undefined || columns.length === 0`

PR #2165, opened 2022-08-09
Split up `Table.tsx`

julieg18:
* splits up `Table.tsx`, giving each component its own file


webview/src/experiments/components/table/Table.tsx
```diff
@@ -3,123 +3,12 @@ import { useSelector } from 'react-redux'
 import cx from 'classnames'
```
julieg18, line 3: I put each file in the `table` directory, but we could also put each file in one folder called `table`.

sroy3 (approved):

Issue #2177, opened 2022-08-10
When DVC is only initialized experiments and plots break


shcheklein:
Get project, uncommit or just copy all files to a new dir and initialize DVC there. Run experiments via CLI. Try to see them in the extension. 

For me experiments panel was empty (and basis message with the links to docs). Plots got stuck on "Loading plots" and nothing in the terminal.

Extension version: 0.3.16

DVC version:

```
DVC version: 2.17.0 (pip)
---------------------------------
Platform: Python 3.9.12 on macOS-12.4-arm64-arm-64bit
Supports:
        azure (adlfs = 2022.7.0, knack = 0.9.0, azure-identity = 1.10.0),
        gdrive (pydrive2 = 1.14.0),
        gs (gcsfs = 2022.5.0),
        hdfs (fsspec = 2022.5.0, pyarrow = 9.0.0),
        webhdfs (fsspec = 2022.5.0),
        http (aiohttp = 3.8.1, aiohttp-retry = 2.7.0),
        https (aiohttp = 3.8.1, aiohttp-retry = 2.7.0),
        s3 (s3fs = 2022.5.0, boto3 = 1.21.21),
        ssh (sshfs = 2022.6.0),
        oss (ossfs = 2021.8.0),
        webdav (webdav4 = 0.9.7),
        webdavs (webdav4 = 0.9.7)
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk3s1s1
Caches: local
Remotes: None
Workspace directory: apfs on /dev/disk3s1s1
Repo: dvc, git
```

mattseddon:
Yep, I hit this as well. Will need to put a safe guard in. It will be something to do with the experiments data not looking complete.

shcheklein:
I also hit it again, now with an existing commit. Looks like if something is wrong with one of the commands (and I don't see actually what is happening in logs) based on the previous commit or an existing one- it fails and get stuck in some random state. Doing a few times `git commit` helps to cleanup the history.

mattseddon:
Looks like a bug introduced in https://github.com/iterative/vscode-dvc/pull/2029. The code expects are `deps` to be constant across revisions and could not handle missing data. Working on a fix now.

mattseddon:
I think this is a combination of a few things. The first two things that current jump out are 

1. Before the initial git commit we get `ERROR: unexpected error - Empty git repo`.
2. The bug that I fixed yesterday with #2187.

[...]

PR #2194, opened 2022-08-16
Handle associated git repo having no commits

mattseddon:
Part of #2177.

When a DVC repo is first created there is the possibility that both `exp show` and `plots diff` will throw `ERROR: unexpected error - Empty git repo`. This PR blocks the `cliReader` from endlessly retrying if that error is received. For this particular error default data is now returned and the extension continues as normal. The user will then end up in this state until they make a commit:

image

### Demo

https://user-images.githubusercontent.com/37993418/184805533-4f2e5460-fb7f-4f35-adf1-7069b62cc6d6.mov

**Note**: `list`, `diff` and `status` all return empty objects under these circumstances.

### DIFF
--- a/extension/src/experiments/model/tree.ts
+++ b/extension/src/experiments/model/tree.ts
@@ -268,7 +268,14 @@ export class ExperimentsTree
 
   private getDescription() {
     const dvcRoots = this.experiments.getDvcRoots()
-    if (!definedAndNonEmpty(dvcRoots)) {
+
+    const total = sum(
+      dvcRoots.map(dvcRoot =>
+        this.experiments.getRepository(dvcRoot).getExperimentCount()
+      )
+    )
+
+    if (!total) {
       return
     }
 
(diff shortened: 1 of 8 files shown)


extension/src/experiments/model/tree.ts
```diff
@@ -268,7 +268,14 @@ export class ExperimentsTree
...
     const dvcRoots = this.experiments.getDvcRoots()
-    if (!definedAndNonEmpty(dvcRoots)) {
+
+    const total = sum(
+      dvcRoots.map(dvcRoot =>
+        this.experiments.getRepository(dvcRoot).getExperimentCount()
+      )
+    )
+
+    if (!total) {
```
mattseddon, line 278: [F] If we have no experiments then all of the repos in the workspace are empty and we should not bother to show a description (previously `0 of 0 (max 7)` would have been shown).

mattseddon:
In an ideal world, we would get this error returned as part of the normal data structures for `exp show` and `plots diff`. Could be addressed in https://github.com/iterative/dvc/issues/7692 for `plots diff`.

julieg18 (approved):

mattseddon:
I'll branch off this and update with the 255 error code.

PR #2197, opened 2022-08-16
Improve table styles

julieg18:
# 1/2 `main` <- this <- #2198 
* add vscode prefix to `--checkbox-border` variable
* keep cell spinners and unselected circles the same bg color as cells
* keep the sticky column from losing opacity on row hover 

**Example of cell spinners and unselected circles**

https://user-images.githubusercontent.com/43496356/184926283-3adc63b2-3fa6-4e2f-9771-01fbd682d27e.mov

**Example of now fully opaque experiment column**

https://user-images.githubusercontent.com/43496356/184946778-bc6421e6-ba2d-415e-aa1f-99db7cb738c1.mov

Followup to #2133

### DIFF
--- a/webview/src/shared/variables.scss
+++ b/webview/src/shared/variables.scss
@@ -3,7 +3,7 @@ $bg-color: var(--vscode-editor-background);
 $bullet-size: 9px;
 $watermark-color: var(--vscode-descriptionForeground);
 
-$border-color: var(--checkbox-border);
+$border-color: var(--vscode-checkbox-border);
 $metrics-color: var(--vscode-dvc-metrics);
 $params-color: var(--vscode-dvc-params);
 $deps-color: var(--vscode-dvc-deps);
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -126,7 +126,7 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
       height: 4px;
       vertical-align: middle;
       border: 1px solid currentColor;
-      background-color: $bg-color;
+      background-color: inherit;
     }
 
     .workspaceWithChanges.unselectedExperiment & {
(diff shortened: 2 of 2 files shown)


sroy3 (approved):
Should we change the border to be the same as the row background as well?

julieg18:
> Should we change the border to be the same as the row background as well?

Apologies, I'm not sure what border you're referring to?

sroy3:
> > Should we change the border to be the same as the row background as well?
> 
> Apologies, I'm not sure what border you're referring to?
Screen Shot 2022-08-16 at 11 05 14 AM
 The black border on the spinner where the cursor is.

julieg18:
> The black border on the spinner where the cursor is.

Oh, I see what you're saying! I'll need to rethink this solution a bit more. Now that I'm looking at things a second time, I just noticed that our checkboxes are not visible on the default theme on hover as well: 

image

We may need to rethink our hover background colors again 😅

julieg18:
Properly fixed the background for spinners/bullets and made sure that the exp column always retains full opacity no matter the background color(some row background colors are partially transparent) causing this issue in some themes: 

image

sroy3 (approved):

shcheklein:
I think things like this should be fixed, folks. Unless I'm missing something and it's an actual edge case (e.g. there are places in VS Code with this theme where it breaks the interface).

Screen Shot 2022-08-16 at 10 37 24 AM

I'm not sure why is it happening btw. @julieg18 could you give a bit more details. What variable do we use for background that causes this? In what case (hover, active, etc?)

shcheklein:
Good, thanks @julieg18 !

mattseddon:
Still a slight issue with the row being selected:

image

webview/src/shared/variables.scss
```diff
@@ -3,7 +3,7 @@ $bg-color: var(--vscode-editor-background);
 $bullet-size: 9px;
 $watermark-color: var(--vscode-descriptionForeground);
 
-$border-color: var(--checkbox-border);
+$border-color: var(--vscode-checkbox-border);
```
mattseddon, line 6: There is at least one more of these in this file. `--button-primary-background`

julieg18: Weird, that variable actually doesn't have a `vscode` prefixed version 🤔

julieg18:
> Still a slight issue with the row being selected:

Fixed!

PR #2198, opened 2022-08-16
Update cell hover styles

julieg18:
# 2/2 `main` <- #2197 <- this
* add cell horizontal borders instead of updating cell background color, fixing invisible (or near invisible) checkboxes in some themes.

**Example of `main` cell hover style:** 

image

**Example of PR cell hover style:**

https://user-images.githubusercontent.com/43496356/184978628-df891c38-c2a1-46be-8f0a-890302709822.mov

Part of #2133 followup

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -133,10 +133,6 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
       background-color: $row-hover-background-color;
     }
 
-    .unselectedExperiment:not(.rowSelected) .experimentCell:hover & {
-      background-color: $cell-hover-background-color;
-    }
-
     .workspaceWithChanges.unselectedExperiment & {
       border: 1px solid $changed-color;
     }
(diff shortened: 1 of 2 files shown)


shcheklein:
I don't like tbh, how text is moving a few pixels as you hover. Is it a bug or a feature?

shcheklein:
Color also doesn't change - is it expected?

julieg18:
> I don't like tbh, how text is moving a few pixels as you hover. Is it a bug or a feature?

Whoops, wrong video. Text shouldn't be moving now. 

> Color also doesn't change - is it expected?

Yes, we originally wanted to add a background color to the cells, but I'm having trouble finding a background color that works well across themes 🤔

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -166,12 +162,6 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
       border-bottom-color: $row-hover-background-color;
     }
 
-    .runningExperiment:not(.rowSelected) .experimentCell:hover & {
```
mattseddon, line 169: High contrast light theme

image

Dive bar: 

image

mattseddon (approved):

sroy3 (approved):

webview/src/experiments/components/table/styles.module.scss
julieg18: Figured out a way to fix the background, though the spinners border are still a slight different color. You can't really notice unless you squint, so should be good to merge :) 

image

PR #2202, opened 2022-08-17
Handle newly introduced deps (before dvc commit)

mattseddon:
# 1/2 `main` <- this <- #2203 

Related to https://github.com/iterative/vscode-dvc/issues/2177.

When a `dep` is first introduced it with be returned under the workspace record in the `exp show` data like this: 

```
  workspace: {
    baseline: {
      data: {
        ...
        deps: {
          'train.py': {
            hash: null,
            size: null,
            nfiles: null
          },
     ...
```

This previously made the extension fall over. I have changed the types/code to accommodate this situation. See comments inline for details.

### Screenshot

Screen Shot 2022-08-17 at 1 36 33 pm

### DIFF
--- a/extension/src/experiments/columns/extract.ts
+++ b/extension/src/experiments/columns/extract.ts
@@ -56,10 +56,12 @@ const extractDeps = (
   const acc: DepColumns = {}
 
   for (const [path, { hash }] of Object.entries(columns)) {
-    const value = shortenForLabel(hash)
-    acc[path] = {
-      changes: !!value && !!branch && branch?.deps?.[path]?.value !== value,
-      value
+    const value = shortenForLabel<string | null>(hash)
+    if (value) {
+      acc[path] = {
+        changes: !!branch && branch?.deps?.[path]?.value !== value,
+        value
+      }
     }
   }
 
--- a/extension/src/util/string.ts
+++ b/extension/src/util/string.ts
@@ -1,8 +1,3 @@
 export const shortenForLabel = <T extends string | null>(
   strOrNull: T
-): T | string => {
-  if (typeof strOrNull === 'string') {
-    return strOrNull?.slice(0, 7)
-  }
-  return strOrNull
-}
+): T | string => (strOrNull ? strOrNull.slice(0, 7) : strOrNull)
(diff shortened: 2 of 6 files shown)


extension/src/util/string.ts
```diff
@@ -1 +1,8 @@
-export const shortenForLabel = (str: string): string => str.slice(0, 7)
```
mattseddon, line 1: [F] This is where the error was happening, you cannot slice `null`, simply adding an optional chain seemed a little dicey so I have updated the incoming types and flushed out some issues in `extension/src/experiments/columns/extract.ts`. The resulting function is a complete monstrosity but backwards compatible so I can live with it. Please LMK if you have better ideas... especially for the file name.

extension/src/experiments/columns/extract.ts
```diff
@@ -56,10 +56,12 @@ const extractDeps = (
   const acc: DepColumns = {}
 
   for (const [path, { hash }] of Object.entries(columns)) {
-    const value = shortenForLabel(hash)
-    acc[path] = {
-      changes: !!value && !!branch && branch?.deps?.[path]?.value !== value,
-      value
+    const value = shortenForLabel<string | null>(hash)
+    if (value) {
```
mattseddon, line 60: [F] By not collecting `null` here we end up with the desired output in webview (show `...`).

extension/src/util/string.ts
```diff
@@ -1 +1,8 @@
...
+export const shortenForLabel = <T extends string | null>(
+  strOrNull: T
+): T | string => {
+  if (typeof strOrNull === 'string') {
```
julieg18, line 4: Is there a reason not to use a ternary operator?
```
return typeof strOrNull === 'string' ? strOrNull?.slice(0, 7) : strOrNull
```

julieg18 (approved):

extension/src/util/string.ts
```diff
@@ -1 +1,8 @@
...
+    return strOrNull?.slice(0, 7)
```
sroy3, line 5: Why not simply check for `strOrNull` to be truthy instead of checking for the type to be a string? A conditional wouldn't be necessary here.

sroy3 (approved):

extension/src/util/string.ts
mattseddon: image

Will use a ternary.

PR #2204, opened 2022-08-17
Fix partially hidden header context menus

julieg18:
stops context menu from being blocked by the workspace row

**Current:**

image

**PR:**

image

Part of #2133 followup

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -680,7 +680,7 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
   box-shadow: none;
   position: sticky;
   top: 0;
-  z-index: 4;
+  z-index: 5;
 }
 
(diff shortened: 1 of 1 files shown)


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -680,7 +680,7 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
   box-shadow: none;
   position: sticky;
   top: 0;
-  z-index: 4;
+  z-index: 5;
```
julieg18, line 683: Noticed yet another small bug that I missed in #2133 😅. Looks like there was an issue with the `z-index`, causing the workspace row to have a higher stacking context than the head tooltips. Fixed by increasing the head `z-index`, making it higher than the workspace row.

sroy3 (approved):

mattseddon (approved):

PR #2206, opened 2022-08-17
Make drop zones larger when dragging in the same section

### LINKED ISSUE #2115: Problems with drag and drop targets in plots
There are currently a few little problems with the drop targets in the plots section.

- [ ] A drop target will appear only if we drag a plot over another plot and it's only possible to drop the plot in the drop section or the adjacent plot. The drop section should extend to the whole vertical line of plot invisibly while the style should stay the same
- [x] Dropping an item to "cancel" the re-positioning should clear all states related to drag and drop.

https://user-images.githubusercontent.com/3683420/181608651-ce1d58fb-8e38-4deb-813e-8636c0c7156e.mov

sroy3:
Closes #2115 

https://user-images.githubusercontent.com/3683420/185230228-1a038f87-875b-4a87-98cc-3d37ec2c4174.mov

https://user-images.githubusercontent.com/3683420/185230592-248ac530-18ad-4b1c-b66f-fae388da2cb6.mov

### DIFF
--- a/webview/src/plots/components/App.test.tsx
+++ b/webview/src/plots/components/App.test.tsx
@@ -808,6 +808,68 @@ describe('App', () => {
     expect(screen.getByTestId('plot_drop-target')).toBeInTheDocument()
   })
 
+  it('should show a drop target at the end of the section when moving a plot inside of one section but not over any other plot', () => {
+    renderAppWithOptionalData({
+      template: complexTemplatePlotsFixture
+    })
+
+    const aSingleViewPlot = screen.getByTestId(join('plot_other', 'plot.tsv'))
+
+    dragEnter(aSingleViewPlot, 'template-single_0', DragEnterDirection.LEFT)
+
+    expect(screen.getByTestId('plot_drop-target')).toBeInTheDocument()
+  })
+
+  it('should drop a plot at the end of the section when moving a plot inside of one section but not over any other plot', () => {
+    renderAppWithOptionalData({
+      template: complexTemplatePlotsFixture
+    })
+
+    const aSingleViewPlot = screen.getByTestId(join('plot_other', 'plot.tsv'))
+    const topSection = screen.getByTestId('plots-section_template-single_0')
+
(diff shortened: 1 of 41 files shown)


webview/src/plots/components/App.test.tsx
```diff
@@ -801,13 +801,75 @@ describe('App', () => {
...
-      'plots-section_template-single_0',
-      DragEnterDirection.RIGHT
+      'template-single_0',
+      DragEnterDirection.LEFT
     )
 
     expect(screen.getByTestId('plot_drop-target')).toBeInTheDocument()
   })
 
+  it('should show a drop target at the end of the section when moving a plot inside of one section but not over any other plot', () => {
```
mattseddon, line 811: [Q] Should mention template plots? (like the last two tests mention checkpoints)

webview/src/shared/components/dragDrop/DragDropContainer.tsx
```diff
@@ -72,33 +72,44 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
   // eslint-disable-next-line sonarjs/cognitive-complexity
 }) => {
   const [draggedOverId, setDraggedOverId] = useState('')
+  const [hoveringSomething, setHoveringSomething] = useState(false)
```
mattseddon, line 75: [Q] what is the difference between `hoveringSomething` and `isHovering`, why do we need both?

mattseddon (approved):

webview/src/shared/components/dragDrop/DragDropContainer.tsx
sroy3: They are working together. And it's because of:
```
const handleDragLeave = () => {
    isHovering.current = false
    hoveringTimeout.current = window.setTimeout(() => {
      if (!isHovering.current) {
        setHoveringSomething(false)
      }
    }, 500)
  }
  ```
`onDragLeave` is a little too reactive and changing the state every time is a lot. Not only that but there is also a race condition with `onDragEnter` of other elements.

PR #2212, opened 2022-08-18
Fix sticky workspace row obscuring branch

### LINKED ISSUE #2211: Sticky workspace row can obscure branch record
If there are columns in the table that have more than two levels and those columns are hidden then the sticky workspace record will obscure the branch information.

### Demo

https://user-images.githubusercontent.com/37993418/185348881-25073f79-e938-4297-bd05-832eb254ca81.mov

mattseddon:
Closes #2211 

### Demo

https://user-images.githubusercontent.com/37993418/185356120-fd22d1ed-5bf2-40ea-b00f-937e328c5292.mov

### DIFF
--- a/webview/src/experiments/components/table/TableHead.tsx
+++ b/webview/src/experiments/components/table/TableHead.tsx
@@ -49,7 +49,7 @@ export const TableHead = ({
     if (wrapperHeight) {
       setTableHeadHeight(wrapperHeight)
     }
-  }, [setTableHeadHeight])
+  }, [setTableHeadHeight, headerGroups])
 
   const onDragStart: DragFunction = ({ currentTarget }) => {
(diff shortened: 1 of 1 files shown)


webview/src/experiments/components/table/TableHead.tsx
```diff
@@ -49,7 +49,7 @@ export const TableHead = ({
     if (wrapperHeight) {
       setTableHeadHeight(wrapperHeight)
     }
-  }, [setTableHeadHeight])
+  }, [setTableHeadHeight, headerGroups])
```
mattseddon, line 52: The alternative is to use `allHeaders` here and wrap its creation in `useMemo` but I think this is fine.

sroy3 (approved):

PR #2216, opened 2022-08-18
Add a copy button for comparison table rows

sroy3:
https://user-images.githubusercontent.com/3683420/185461084-45bc63fa-1a5d-4f51-8108-a836dbed42cb.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.3.20] - 2022-08-18
+
+### 🚀 New Features and Enhancements
+
+- Show min tested version in toast warning message [#2209](https://github.com/iterative/vscode-dvc/pull/2209) by [@mattseddon](https://github.com/mattseddon)
+
+### 🐛 Bug Fixes
+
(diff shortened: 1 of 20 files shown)


shcheklein:
Thanks @sroy3 for taking care of this ... can we also make names "selectable" as a regular text?

sroy3:
> Thanks @sroy3 for taking care of this ... can we also make names "selectable" as a regular text?
Just added. Might be a little annoying because it's a button that toggles the row, but it isn't a destructive action.
https://user-images.githubusercontent.com/3683420/185461757-e41bd5b6-e3ac-46f5-be18-0095852d1b10.mov

shcheklein:
Thanks! QQ - is there a way to recognize that it was in the "selection mode" and do not activate the action?

sroy3:
> Thanks! QQ - is there a way to recognize that it was in the "selection mode" and do not activate the action?

Wasn't easy, but I got it!

https://user-images.githubusercontent.com/3683420/185469462-b38839e8-7bf5-4515-8dd0-443080b02cb3.mov

julieg18 (approved):

mattseddon (approved):

PR #2219, opened 2022-08-19
Share an experiment from context menus

mattseddon:
# 1/2 `main` <- this <- #2221 

This PR gives the user the option to share an experiment as a branch from the context menus available in the experiments tree and table.

### Demo

https://user-images.githubusercontent.com/37993418/185542576-243bc77d-8a28-4592-a9a6-c6ebb77a5b16.mov

Note: There are a few things to follow up with in terms of outstanding housekeeping, I also want to add all calls to git into our output channel and add a new (matching) command to the palette. Next step after that is to add a "Commt and Share" command.

### DIFF
--- a/demo/dvc.yaml
+++ b/demo/dvc.yaml
@@ -2,21 +2,21 @@ stages:
   train:
     cmd: python train.py
     deps:
-      - data/MNIST/raw
-      - train.py
+    - data/MNIST
+    - train.py
     params:
-      - params.yaml:
+    - params.yaml:
     outs:
-      - model.pt:
-          checkpoint: true
+    - model.pt:
+        checkpoint: true
     metrics:
       - training_metrics.json:
           persist: true
           cache: false
     plots:
-      - training_metrics
-      - misclassified.jpg
-      - predictions.json:
-          template: confusion
-          x: actual
-          y: predicted
+    - training_metrics
+    - misclassified.jpg
+    - predictions.json:
+        template: confusion
+        x: actual
+        y: predicted
--- a/demo/dvc.yaml
+++ b/demo/dvc.yaml
@@ -2,21 +2,21 @@ stages:
   train:
     cmd: python train.py
     deps:
-    - data/MNIST
-    - train.py
+      - data/MNIST/raw
+      - train.py
     params:
-    - params.yaml:
+      - params.yaml:
     outs:
-    - model.pt:
-        checkpoint: true
+      - model.pt:
+          checkpoint: true
     metrics:
       - training_metrics.json:
           persist: true
           cache: false
     plots:
-    - training_metrics
-    - misclassified.jpg
-    - predictions.json:
-        template: confusion
-        x: actual
-        y: predicted
+      - training_metrics
+      - misclassified.jpg
+      - predictions.json:
+          template: confusion
+          x: actual
+          y: predicted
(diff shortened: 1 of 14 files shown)


demo/dvc.yaml
```diff
@@ -2,21 +2,21 @@ stages:
   train:
     cmd: python train.py
     deps:
-    - data/MNIST
-    - train.py
+      - data/MNIST/raw
```
mattseddon, line 5: [F] Change proposed here https://github.com/iterative/dvc/issues/8113#issuecomment-1219099594

extension/src/experiments/commands/register.ts
```diff
@@ -153,19 +156,56 @@ const registerExperimentInputCommands = (
...
+            name,
+            input
+          )
+          await experiments.runCommand(
+            AvailableCommands.EXPERIMENT_APPLY,
+            dvcRoot,
+            name
+          )
+          await experiments.runCommand(AvailableCommands.PUSH, dvcRoot)
+          const std = gitPushBranch(dvcRoot, input)
```
mattseddon, line 206: [C] This makes me think that we should be showing all of the git commands that we run inside of our DVC output channel

demo/dvc.yaml
mattseddon, line 5: [F] Change proposed here https://github.com/iterative/dvc/issues/8113#issuecomment-1219099594

extension/src/experiments/commands/register.ts
```diff
@@ -153,19 +156,55 @@ const registerExperimentInputCommands = (
...
+            name,
+            input
+          )
+          await experiments.runCommand(
+            AvailableCommands.EXPERIMENT_APPLY,
+            dvcRoot,
+            name
+          )
+          await experiments.runCommand(AvailableCommands.PUSH, dvcRoot)
+          return Toast.showOutput(gitPushBranch(dvcRoot, input))
```
mattseddon, line 206: [C] This makes me think that we should be showing all of the git commands that we run inside of our DVC output channel.

I will follow up and give Git its own reader/executor.

webview/src/experiments/components/App.test.tsx
```diff
@@ -679,6 +679,27 @@ describe('App', () => {
...
+      const menuitems = screen.getAllByRole('menuitem')
+      const itemLabels = menuitems.map(item => item.textContent)
+      expect(itemLabels).toStrictEqual([
+        'Apply to Workspace',
+        'Create new Branch',
+        'Share',
+        'Modify, Reset and Run',
+        'Modify and Resume',
+        'Modify and Queue',
+        'Star Experiment',
```
mattseddon, line 698: I don't think 'Experiment' needs to be here. I'll follow up

extension/package.json
```diff
@@ -460,6 +460,12 @@
         "category": "DVC",
         "icon": "$(play)"
       },
+      {
+        "title": "%command.views.experiments.shareExperiment%",
+        "command": "dvc.views.experiments.shareExperiment",
+        "category": "DVC",
+        "icon": "$(repo-push)"
```
mattseddon, line 467: image

demo/dvc.yaml
mattseddon, line 5: will make this change separately

sroy3 (approved):

julieg18 (approved):

PR #2225, opened 2022-08-22
Add Git cli calls into output channel

mattseddon:
# 1/2 `main` <- this <- #2226

This PR moves all calls made to Git into the DVC output channel. This will make the user aware of these calls and should let them debug should something go wrong.

### Demo

https://user-images.githubusercontent.com/37993418/185831174-2e480fb5-f36f-40f3-9116-d0559cc5148d.mov

PR #2226, opened 2022-08-22
Rename DVC CLI files and classes

mattseddon:
# 2/2 `main` <- #2225 <- this

This PR cleans up after #2225 by moving and renaming the classes that interact with the DVC CLI.


extension/src/cli/dvc/constants.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export enum SubCommand {
+  SHOW = 'show'
+}
+
+export enum Flag {
+  FORCE = '-f',
+  OUTPUT_PATH = '-o',
+  RECURSIVE = '-R',
+  SHOW_JSON = '--show-json',
```
skshetry, line 34:
```suggestion
  SHOW_JSON = '--json',
```

We have softly deprecated the `--show-json` flag for the shorter `--json` flag, since 2.8.0 (see https://github.com/iterative/dvc/pull/6711, and https://github.com/iterative/dvc/releases/2.8.0).

mattseddon: Has `--json` been available for all of the required commands since `2.11.0`?

mattseddon: I'll have to bump the min required cli version when we start to consume `data status`. I can make the change then.

skshetry: yes, since 2.8.0, all of the commands that have `--show-json` should have `--json` alias. `plots --json` was introduced in 2.9.3 with both flags, and `data status --json` since 2.15.0.

sroy3 (approved):

PR #2227, opened 2022-08-22
Replace vary with modify in variable names

mattseddon:
Housekeeping follow-up from #2219. Noticed the inconsistency then but didn't want to bloat the PR anymore.

### DIFF
--- a/extension/src/experiments/webview/messages.ts
+++ b/extension/src/experiments/webview/messages.ts
@@ -84,17 +84,17 @@ export class WebviewMessages {
           RegisteredCliCommands.EXPERIMENT_VIEW_BRANCH,
           { dvcRoot: this.dvcRoot, id: message.payload }
         )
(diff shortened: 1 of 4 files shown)


extension/src/experiments/webview/messages.ts
```diff
@@ -84,17 +84,17 @@ export class WebviewMessages {
           RegisteredCliCommands.EXPERIMENT_VIEW_BRANCH,
           { dvcRoot: this.dvcRoot, id: message.payload }
         )
-      case MessageFromWebviewType.VARY_EXPERIMENT_PARAMS_AND_QUEUE:
```
mattseddon, line 87: [F] VARY is left over from a previous naming convention

julieg18 (approved):

sroy3 (approved):

PR #2239, opened 2022-08-23
Make timestamp column hideable and draggable

### LINKED ISSUE #2031: Timestamp column improvements
Related to https://github.com/iterative/vscode-dvc/issues/1444

---

### EXPs table improvements

- [x] Align cell content to the left (Reference https://github.com/iterative/vscode-dvc/issues/1700)
- [ ] The column needs to be moveable and hideable https://github.com/iterative/vscode-dvc/issues/2031#issuecomment-1185032012

### Needs to be fixed

- [x] Timestamp label alignment
- [x] Cell information alignment

Screen-Shot-2022-08-09-at-12 06 32-PM

julieg18:
* makes timestamp column hideable
* makes timestamp column moveable

## Demo

https://user-images.githubusercontent.com/43496356/186498796-274403a2-53e2-44d4-9b59-a500e6fa598b.mov

Fixes #2031


extension/src/experiments/columns/quickPick.ts
```diff
@@ -14,11 +15,13 @@ export const pickFromColumnLikes = (
     return Toast.showError('There are no columns to select from.')
   }
   return quickPickValue<ColumnLike>(
-    columnLikes.map(columnLike => ({
-      description: columnLike.path,
-      label: columnLike.label,
-      value: columnLike
-    })),
+    columnLikes
+      .filter(columnLike => columnLike.path !== timestampColumn.path)
```
mattseddon, line 19: [C] I do not think we want to filter this here. We will actually want to make it possible to sort/filter by timestamp in the future.

extension/src/experiments/columns/collect/timestamp.ts
```diff
@@ -0,0 +1,14 @@
+import { ColumnAccumulator } from './util'
+import { ColumnType } from '../../webview/contract'
+
+export const timestampColumn = {
+  hasChildren: false,
+  label: 'Timestamp',
+  parentPath: 'timestamp',
+  path: 'timestamp:timestamp',
```
mattseddon, line 8: When there is lots of `type === ColumnType.TIMESTAMP ? ` (or similar) around in the code that is a smell.

I think if you had `path: 'timestamp'` and `parentPath: ''` the rest of the code might be simpler. You could also consider making both the path and label `'Timestamp'` and whilst collecting the variable [here](https://github.com/iterative/vscode-dvc/pull/2239/files#diff-d9605ebc61e90bded81b48018d5d07110af18b48d720bec74e060bd0b8c111cbR134) you could use 

```
if (timestamp) {
    experiment.Timestamp = timestamp
}
```

FYI the reason we have the colon separated path is because of the (usually) nested data structure that looks something like this

```
{
	params: {
		params.yaml: {
			lr: value,
			acc: value
		}
	}
...
}
```

mattseddon: Raised https://github.com/iterative/vscode-dvc/pull/2247 to show you want I am talking about.

Tests should also be fixed on that branch 🙂

extension/src/experiments/columns/quickPick.ts
julieg18: > [C] I do not think we want to filter this here. We will actually want to make it possible to sort/filter by timestamp in the future.

I did test out sorting with timestamp but it wasn't working so I disabled both for now :)

extension/src/experiments/columns/collect/timestamp.ts
julieg18: Makes sense and simplifies things for sure!

extension/src/experiments/columns/collect/timestamp.ts
```diff
@@ -0,0 +1,14 @@
+import { ColumnAccumulator } from './util'
+import { ColumnType } from '../../webview/contract'
+
+export const timestampColumn = {
```
julieg18, line 4: Is there a better place to keep this variable? It is used in a couple different files 🤔 

image

webview/src/experiments/components/table/TableHeader.tsx
```diff
@@ -60,17 +60,18 @@ export const TableHeader: React.FC<TableHeaderProps> = ({
...
   const contextMenuOptions: MessagesMenuOptionProps[] = React.useMemo(() => {
     const menuOptions: MessagesMenuOptionProps[] = [
       {
-        hidden: !!column.headers,
+        hidden: !isTimestamp && !!column.headers,
         id: 'hide-column',
         label: 'Hide Column',
         message: {
-          payload: column.id,
+          payload: isTimestamp && column.name ? column.name : column.id,
```
mattseddon, line 74: [Q] Can we change the timestamp's id so that we can get rid of this ternary?

extension/src/experiments/columns/collect/timestamp.ts
mattseddon: TBH I would make a `extension/src/experiments/columns/constants.ts` file. I would also not collect this column for every iteration in `extension/src/experiments/columns/collect.ts`. I would create the initial accumulator object with this column in it.

Importing into the test fixtures breaks the `turborepo` workflow (a change in this file will not necessarily trigger a rerun of the webview tests but they do depend on this).

mattseddon: You should be able to update the dependencies for turborepo,

extension/src/experiments/columns/collect/timestamp.ts
```diff
@@ -0,0 +1,14 @@
...
+  hasChildren: false,
+  label: 'Timestamp',
+  parentPath: ColumnType.TIMESTAMP,
```
mattseddon, line 7: [Q] Did you get a chance to look at making this optional? Having this field for the timestamp column does not really make sense.

webview/src/test/sort.ts
```diff
@@ -1,7 +1,7 @@
 import { screen } from '@testing-library/react'
 import { ColumnType, TableData } from 'dvc/src/experiments/webview/contract'
 
-export const defaultColumns = ['Experiment', 'Timestamp']
+export const defaultColumns = ['Experiment']
```
mattseddon, line 4: [N] This should be `defaultColumn` and not an ary now.

extension/src/experiments/columns/collect/index.ts
```diff
@@ -19,6 +20,7 @@ const collectFromExperiment = (
 ) => {
   const { data } = experiment
   if (data) {
+    collectTimestamp(acc)
```
mattseddon, line 23: [I] If you look at the rest of this file we are now collecting a static column (constant) for every experiment and every branch. We don't need to do that. We can simply create the accumulator with the constant included.

extension/src/experiments/columns/extract.ts
```diff
@@ -88,6 +89,7 @@ export const extractColumns = (
   ].join('\n')
 
   const columns: Columns = {
+    Timestamp: experiment?.timestamp,
```
sroy3, line 92: Is there a reason for using the capitalized `Timestamp`?

webview/src/experiments/components/table/TableHeaderCell.tsx
```diff
@@ -104,8 +104,7 @@ export const TableHeaderCell: React.FC<{
   setExpColumnNeedsShadow
 }) => {
   const [menuSuppressed, setMenuSuppressed] = React.useState<boolean>(false)
-  const isDraggable =
-    !column.placeholderOf && !['id', 'timestamp'].includes(column.id)
+  const isDraggable = !column.placeholderOf && !['id'].includes(column.id)
```
sroy3, line 107: Since there is only one item in `['id'].includes(column.id)` we can change it to be `column.id !== id`

extension/src/experiments/columns/collect/timestamp.ts
julieg18: When I attempted it, the timestamp disappeared entirely. Would need to debug further to figure out the issue 🤔

extension/src/experiments/columns/extract.ts
julieg18: Using a capitalized Timestamp helps us to write less code since that's what the label is referred to as :)

mattseddon: I think now would be a good time to rename this to `Created` to match the CLI table:

image

Can do it in follow up if necessary.

extension/src/experiments/columns/collect/timestamp.ts
julieg18: > You should be able to update the dependencies for turborepo.

Makes sense! I've created a constants file but I'm not quite sure how to update the dependencies for turborepo 🤔

mattseddon: There is a `globalDependencies` key in `turbo.json`. Docs: https://turborepo.org/docs/reference/configuration#globaldependencies

julieg18: Thanks! Updated!

julieg18:
Resolved most of the comments and I think the rest can be in a follow-up! Anything else that needs to be done before we merge this?

mattseddon (approved):

extension/src/experiments/columns/collect/timestamp.ts
mattseddon: You'd need to update this method:

```
  public filterChildren(path?: string) {
    return this.data.filter(element =>
      path
        ? element.parentPath === path
        : Object.values<string>(ColumnType).includes(element.parentPath)
    )
  }

```

in `extension/src/experiments/columns/model.ts`.

The entry point to find that out would be `getChildren` in or around `extension/src/experiments/columns/tree.ts`.

PR #2242, opened 2022-08-23
DVC Language Server

wolmir:
# DVC Language Server

## Current Specification and Documentation:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/


languageServer/src/features/StageCompletions.ts
```diff
@@ -0,0 +1,26 @@
...
+  TextDocumentPositionParams
+} from 'vscode-languageserver/node'
+import { BaseFeature } from './BaseFeature'
+
+export class StageCompletions extends BaseFeature {
+  protected setup(): void {
+    this.connection?.onCompletion(
+      (params: TextDocumentPositionParams): CompletionItem[] => {
+        const stageNames = this.query({
+          dvcYamls: '$.workspace[?(@.uri.match(/dvc.yaml$/))]',
```
wolmir, line 13: "Get the text documents from workspace that have a uri like dvc.yaml"

languageServer/src/features/StageCompletions.ts
```diff
@@ -0,0 +1,26 @@
...
+          stages: '$.dvcYamls[*].parsed.stages.*~'
```
wolmir, line 14: "Then for every dvcYaml it could find, get the parsed contents, look up the stages property and get their names (the ~ is to get the property names only)"

languageServer/src/features/StageCompletions.ts
```diff
@@ -0,0 +1,26 @@
...
+        }).getValue('stages') as string[]
+
+        return stageNames.map(stage => {
+          return {
+            kind: CompletionItemKind.Method,
+            label: stage
```
wolmir, line 20: And then wrap them in a minimal completion item for this first case

languageServer/src/server.ts
```diff
@@ -0,0 +1,118 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
```
mattseddon, line 4: [Q] Which extensions/docs did you use to come up with this?

languageServer/src/features/StageCompletions.ts
mattseddon: Did you build the `vsix` to demo this? I'm unable to get any completions when running the `dev-server` (Disabled the red-hat extension).

extension/src/extension.ts
```diff
@@ -437,15 +451,79 @@ export class Extension extends Disposable implements IExtension {
...
+  // Create the language client and start the client.
+  client = new LanguageClient(
+    'languageServerExample',
+    'Language Server Example',
+    serverOptions,
+    clientOptions
+  )
+
+  // Start the client. This will also launch the server
+  client.start()
```
mattseddon: this returns a disposable so you can `this.dispose.track` it. I would assume that disposing of it stops the server. Which means we can just use the normal pattern.

Source: https://github.com/redhat-developer/vscode-yaml/blob/df90855d7e280ebc2b01599747d9223ab37e46f1/src/extension.ts#L131

mattseddon: Dispose does stop:
```
    dispose(timeout = 2000) {
        try {
            this._disposed = 'disposing';
            return this.stop(timeout);
        }
        finally {
            this._disposed = 'disposed';
        }
    }
```

languageServer/src/features/StageCompletions.ts
wolmir: No, I'll include it in today's tasks, but it shouldn't affect the server. I'll try it in the demo project today, I was testing with the getting-started project

languageServer/src/server.ts
wolmir: https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-sample

I missed these, sorry. In my defence It's hard to notice comments in most highlight schemes.
I installed this https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments

It seems to be working now

extension/src/extension.ts
wolmir: Thanks!

languageServer/src/features/StageCompletions.ts
mattseddon: I was using example-get-started

mattseddon:
Every time you ask for completions the server adds the same completions to the existing list:

image

languageServer/src/features/BaseFeature.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  protected queryWith(
+    transformer: (
+      utils: Utilities,
+      data: Record<string, unknown>
+    ) => Record<string, unknown>
+  ): this {
+    this.refreshWorkspace()
+    this.refreshWorkspace()
```
mattseddon, line 68: Call is duplicated

turbo.json
```diff
@@ -7,7 +7,7 @@
       "outputs": ["dist/**"]
     },
     "dvc#build": {
-      "dependsOn": ["dvc-vscode-webview#build"],
+      "dependsOn": ["dvc-vscode-webview#build", "dvc-vscode-lsp#build"],
```
mattseddon, line 10: Note to self: We will need to remember to move the built server next to/into the extension/dist folder before packaging.

languageServer/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
sroy3, line 1: Maybe we should create a `tsconfig.base.json` at root and extend from it so that we share as many options between projects as possible.

PR #2247, opened 2022-08-24
Make some timestamp processing suggestions

mattseddon:
- Make timestamp column hideable
- make some suggestions

### DIFF
--- a/extension/src/experiments/columns/collect/timestamp.ts
+++ b/extension/src/experiments/columns/collect/timestamp.ts
@@ -4,8 +4,8 @@ import { ColumnType } from '../../webview/contract'
 export const timestampColumn = {
   hasChildren: false,
   label: 'Timestamp',
-  parentPath: 'timestamp',
-  path: 'timestamp:timestamp',
+  parentPath: ColumnType.TIMESTAMP,
+  path: 'Timestamp',
   type: ColumnType.TIMESTAMP
 }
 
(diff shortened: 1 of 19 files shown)


webview/src/experiments/components/table/TableHeader.tsx
```diff
@@ -228,8 +228,7 @@ const TableHeaderCell: React.FC<{
   setExpColumnNeedsShadow
 }) => {
   const [menuSuppressed, setMenuSuppressed] = React.useState<boolean>(false)
-  const isDraggable =
-    !column.placeholderOf && !['id', 'timestamp'].includes(column.id)
+  const isDraggable = !column.placeholderOf && !['id'].includes(column.id)
```
mattseddon, line 231: [F] And now can magically drag and drop the column....

extension/src/experiments/columns/collect/timestamp.ts
```diff
@@ -4,8 +4,8 @@ import { ColumnType } from '../../webview/contract'
 export const timestampColumn = {
   hasChildren: false,
   label: 'Timestamp',
-  parentPath: 'timestamp',
-  path: 'timestamp:timestamp',
+  parentPath: ColumnType.TIMESTAMP,
```
mattseddon, line 7: I would see if it is possible to change this to an optional property and drop it from here.

julieg18 (approved):
Thanks for the suggestions! Makes the solution way cleaner!

PR #2257, opened 2022-08-24
Display the smoothness slider correctly

### LINKED ISSUE #2248: Issue with plots: template: smooth
When building a plot with the smooth template and showing the results in VS-code the "smoothness indicator" shows on top of the plot(s) and cannot be modified. If I open the plots in the browser it works fine!

Screenshot 2022-08-24 at 12 00 50
Screenshot 2022-08-24 at 12 01 48

sroy3:
Closes #2248 

https://user-images.githubusercontent.com/3683420/186522954-a214ba75-9904-4183-87f0-2d7852f73872.mov

Try it yourself: https://5ffc93f46e9344002155f496-hvagdnhbph.chromatic.com/?path=/story/plots--smooth-template

### DIFF
--- a/webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
+++ b/webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
@@ -67,6 +67,9 @@ export const TemplatePlotsGrid: React.FC<TemplatePlotsGridProps> = ({
     }
 
     for (const panel of Object.values(panels)) {
+      panel.removeEventListener('mouseenter', addDisabled)
+      panel.removeEventListener('mouseleave', removeDisabled)
+      panel.removeEventListener('click', disableClick)
       panel.addEventListener('mouseenter', addDisabled)
       panel.addEventListener('mouseleave', removeDisabled)
       panel.addEventListener('click', disableClick)
--- a/webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
+++ b/webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
@@ -1,7 +1,7 @@
 import cx from 'classnames'
 import { TemplatePlotEntry } from 'dvc/src/plots/webview/contract'
 import { reorderObjectList } from 'dvc/src/util/array'
-import React, { useEffect, useState } from 'react'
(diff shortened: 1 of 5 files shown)


shcheklein:
Good fix, thanks @sroy3 . Can we control its color / position? Now it looks foreign a bit (clearly let's merge and ship this first).

sroy3:
> Good fix, thanks @sroy3 . Can we control its color / position? Now it looks foreign a bit (clearly let's merge and ship this first).

Color, it's possible. I simply added a background, but did not try changing the default of the slider. Position is a little harder. Adding the slider into the plot rectangle makes the plot really small. Like you mentioned, I was thinking we should iterate on this after merging the fix.

sroy3:
Screen Shot 2022-08-25 at 2 15 39 PM

With theme accent color

sroy3:
Do not merge yet. Seems like there is a empty panel that pops up sometimes on other plots

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
```diff
@@ -43,11 +43,47 @@ export const TemplatePlotsGrid: React.FC<TemplatePlotsGridProps> = ({
...
+      e.stopPropagation()
+    }
+
+    for (const panel of Object.values(panels)) {
+      panel.removeEventListener('mouseenter', addDisabled)
+      panel.removeEventListener('mouseleave', removeDisabled)
+      panel.removeEventListener('click', disableClick)
+      panel.addEventListener('mouseenter', addDisabled)
+      panel.addEventListener('mouseleave', removeDisabled)
+      panel.addEventListener('click', disableClick)
```
mattseddon, line 75: [Q] What is the reason for recreating these?

mattseddon (approved):

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
sroy3: Since it's being created by vega the `vega-bindings` and that the `DragDropContainer` rebuilds the plots, setting these events on re-render is the only way to hook on them. Doing so once would create them too early. Maybe I can set up an observer or something to add the listeners once the elements are ready. I'll look into it and do a follow up if possible.

PR #2269, opened 2022-08-26
Rename "Timestamp" column label to "Created"

julieg18:
Renames the timestamp column's name to "Created", matching DVC CLI tables.

### Main: 
image

### PR:

image

Followup to [#2239 comment](https://github.com/iterative/vscode-dvc/pull/2239#discussion_r955448883)


sroy3 (approved):

maxagin:
@julieg18, @shcheklein Shouldn't we request the DVC team change it since the Timestamp term is more efficient?

mattseddon (approved):

mattseddon:
> @julieg18, @shcheklein Shouldn't we request the DVC team change it since the Timestamp term is more efficient?

"Timestamp" does not tell you what the timestamp represents. Created does.

julieg18:
Going to go ahead and merge this! We can always revert if we decide that we want to keep the "Timestamp" name :)

shcheklein:
@maxagin if we want to change this, let's raise an issue in the DVC repo first please. My take both are fine for now, there are pros and cons, so let's keep it at move forward.

mattseddon:
> A timestamp is a digital record of the time of occurrence of a particular event.

Which event? When the experiment was created, when it was first modified, when it was last modified, when it was completed, last time it was selected?

maxagin:
>A timestamp is a digital record of the time of occurrence of a particular event.

>>Which event? When the experiment was created, when it was first modified, when it was last modified, when it was completed, last time it was selected?

The entrance of the experiment.

---

Question. Why we do not display the following information in the table?

- Created
- First modified
- Last modified
- Completed
- Last time it was selected

PR #2271, opened 2022-08-26
Comparison table rows drag and drop

sroy3:
https://user-images.githubusercontent.com/3683420/187252248-34c9c159-1960-46cc-ba7a-2b1a9ca00858.mov


webview/src/shared/components/dragDrop/DragDropContainer.tsx
```diff
@@ -100,7 +115,11 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
 
   useEffect(() => {
     cleanup()
-  }, [order])
+  }, [order, cleanup])
+
+  useLayoutEffect(() => {
+    onLayoutChange?.()
```
sroy3, line 121: This is going to be really helpful for [this](https://github.com/iterative/vscode-dvc/pull/2257/files#r955440333) as well

julieg18 (approved):

mattseddon (commented):
There are a couple of weird things:

https://user-images.githubusercontent.com/37993418/187329436-120eb289-df58-466a-9a1e-ef01dcf22692.mov

1. depending on where I pick up the "row" a different component will be shown as dragged
2. after spending time re-arranging my rows any closing of the section/webview will reset the order

Dragging across different screens can also be painful:

https://user-images.githubusercontent.com/37993418/187329770-78553300-5625-4c32-8ba0-d9602cdcbc2b.mov

This one is a blocker:

https://user-images.githubusercontent.com/37993418/187330322-b2bacbac-5ee7-4b0b-91ac-4751c484b237.mov

Any attempt to remove a comparison plot whilst the webview is open results in a crash. Adding works, removing does not. This also seemed to work this morning before you merged `main`.

sroy3:
> 1. depending on where I pick up the "row" a different component will be shown as dragged

I'm looking and have no idea on how to fix this, other than adding a complete different style on drag that would always look the same. I'll try researching more and testing stuff.

> 2. after spending time re-arranging my rows any closing of the section/webview will reset the order

Was going to comment that I forgot to ask "Should we persist this?". That answers my question.

> Dragging across different screens can also be painful:

Let me see if there is something I can do. Like you said, there is probably nothing that can be done, but it's worth checking. One solution would be to obscure the images and make them smaller on drag so that there is less distance to travel, but that might be a weird effect, we'd have to try.

> This one is a blocker:
> Any attempt to remove a comparison plot whilst the webview is open results in a crash. Adding works, removing does not. This also seemed to work this morning before you merged `main`.
> 
> Happy to iterate on 1 & 2. There is not much that can be done about 3 but we need to fix the last one before merging.

Once #2290 is merged then or is there still something else? It is also unrelated to that PR (but something we do need to fix ASAP).

mattseddon (approved):
Probably best to make [this change](https://github.com/iterative/vscode-dvc/pull/2295#issuecomment-1232328007) in this PR.

PR #2280, opened 2022-08-28
Account for exp show returning an empty object

mattseddon:
After https://github.com/iterative/dvc/pull/8177 was merged lack of a Git commit will no longer create an unexpected error in `exp show` or `plots diff`. Both calls will now return an empty object under those circumstances. This PR reworks the way that the default value for `exp show` is returned.

Note: This should go out with the `data status` update as it will require a version of the CLI higher than `2.22.1` to work consistently. Related bug was https://github.com/iterative/dvc/issues/8163


julieg18 (approved):

mattseddon:
Will merge and ship with #2299

extension/src/cli/dvc/reader.ts
```diff
@@ -126,6 +127,10 @@ export class DvcReader extends DvcCli {
       ...flags,
       Flag.JSON
     )
+    if (isEmpty(output)) {
+      return { workspace: { baseline: output as {} } }
```
mattseddon, line 131: [C] This looks ugly/unnecessary now but makes more sense to do it this way with #2301 coming.

Issue #2241, opened 2022-08-23
Review: table styles


maxagin:
I have reviewed the EXPs table and would like to share with you some comments.
I will be referring to this [Figma file](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1631%3A119506 )

## TODO 

- [x] Add tooltips to checkboxes and stars
- [x] Hide the right border of the hovered first cell after the horizontal scroll is activated and the shadow appears 
- [x] Increase the workspace row height
- [x] Take off table header right closing line
- [x] Update on-hover header separator line background color
- [x] Take off the first cell's right border when the cell is sticky
- [x] Selected first cell does not have a shadow when scrolled horizontally 
- [x] Keep checkpoint lines on top of the row border
- [x] Update the view background color
- [x] Inconsistent borders on head cells vs body cells
- [x] Remove sticky section bottom border after the vertical scroll is activated
- [x] Keep the resize icon and the vertical line (resize indicator) visible during column resize
- [x] Header cell resizer bar doesn't always reach the top of a cell
- [x] Remove the EXP name from being a toggle
- [x] When star an experiment the weird progress lines show on the sidebar. (Bug?)

### Hint/Tooltip 

- [ ] Add hint "Add/Remove to plots" (maybe different wording) when the EXP's circle has hovered
- [ ] Hints like Select, Add Star and similar. The text is too big. Try to make it the same as when you hover on the Experiments tab
- [ ] Hints need to be closer to the cursor
- [x] Hints do not need to have a triangle at the top

 
### Row

- [x] Update row selection styles
- [x] A number of hidden element indicators (circle with the number) styles for rested and selected row states
- [x] There is a little spacing between the elements in the design, for example, checkbox and star, it is important to show the user where the active area ends and where the new starts
- [x] Chevron on hover - cursor pointer
- [x] The table's last row does not need a button border
- [x] Checkbox and star styles when the row selected

### Context menu

- [x] All context menus need to appear on the right side of the target point.
The context menu is usually placed on the right/bottom side of the mouse click unless it is close to the right age of the screen, then it will appear on the left side. However, in VSCode it seems like always on the right side, but we can do it correctly I think.
- [x] Remove the triangle (on top) from the context menus

maxagin:
In relation to the Experiments table: improve styles #2070

maxagin:
[...]

PR #2289, opened 2022-08-29
Improve table styles

julieg18:
* add tooltips to row actions (checkbox and stars)
* take off first cell right border when the cell is sticky 
* increase workspace row height
* take off right border on the head rows last children
* update column resizer hover background color

## Demo

https://user-images.githubusercontent.com/43496356/187478759-e24444d3-0ca0-4404-8736-17707d784ad7.mov

Part of #2241

### DIFF
--- a/webview/src/experiments/components/table/CellRowActions.tsx
+++ b/webview/src/experiments/components/table/CellRowActions.tsx
@@ -6,6 +6,9 @@ import styles from './styles.module.scss'
 import { clickAndEnterProps } from '../../../util/props'
 import { StarFull, StarEmpty } from '../../../shared/components/icons'
 import { pluralize } from '../../../util/strings'
+import Tooltip, {
+  CELL_TOOLTIP_DELAY
+} from '../../../shared/components/tooltip/Tooltip'
 
 export type CellRowActionsProps = {
   isRowSelected: boolean
@@ -27,6 +30,7 @@ export type CellRowActionProps = {
   children: React.ReactElement
   hidden?: boolean
   testId?: string
+  tooltipContent: string
 }
 
 export const CellRowAction: React.FC<CellRowActionProps> = ({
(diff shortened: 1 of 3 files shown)


mattseddon:
@julieg18 what happens at ~10 seconds of the video?

julieg18:
> From watching the whole video I can assume that selecting the row removes the stickiness of the first column. This looks broken to me.

Yes, selecting the row turns off the stickiness. It's always been like that though (picture below was taken on the latest extension release):

image

Agreed though, it does look buggy... I can look into keeping selected rows sticky if we'd like!

maxagin:
> Agreed though, it does look buggy... I can look into keeping selected rows sticky if we'd like!

Please share the updated version, when you have it. Thank you @julieg18

maxagin:
Good work @julieg18 
Very minor: consistent vertical lines

Screen Shot 2022-08-29 at 9 26 03 PM

maxagin:
In relation to the

> take off first cell right border when the cell is sticky

Screen Shot 2022-08-29 at 9 36 12 PM

maxagin:
Just FYI. The selected first cell does not have a shadow:

Screen Shot 2022-08-29 at 9 39 16 PM

maxagin:
Q: is the blue border when on hover 2px?

Screen Shot 2022-08-29 at 9 42 10 PM

julieg18:
> Q: is the blue border when on hover 2px? 

Yes, it is 2px!

sroy3 (approved):
Since what was previously mentioned by Matt is out of scope. This LGTM! I would still follow up on it though.

julieg18:
Ok, merged #2294 which fixed the none sticky selected rows issue! 

Also, per @maxagin comments, added opacity to the resizer bar and took off the borders when the header cells are sticky since those were both really simple fixes. Video is updated and I think we're good to merge once tests are passing!

maxagin:
Looks like "The selected first cell does not have a shadow" also was fixed :) https://github.com/iterative/vscode-dvc/pull/2289#issuecomment-1231043608

PR #2290, opened 2022-08-30
Show empty state when final comparison plot is de-selected

mattseddon:
Identified whilst testing #2271.

There was a bug in the comparison table slice where a message to update the comparison table state to `null` would not result in a state update.

### Demo

https://user-images.githubusercontent.com/37993418/187327940-6aaa772d-cc7f-4085-bb3d-49ae19041860.mov

### DIFF
--- a/webview/src/plots/components/checkpointPlots/checkpointPlotsSlice.ts
+++ b/webview/src/plots/components/checkpointPlots/checkpointPlotsSlice.ts
@@ -40,20 +40,20 @@ export const checkpointPlotsSlice = createSlice({
       state.isCollapsed = action.payload
     },
     update: (state, action: PayloadAction<CheckpointPlotsData>) => {
-      if (action.payload) {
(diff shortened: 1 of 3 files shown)


webview/src/plots/components/checkpointPlots/checkpointPlotsSlice.ts
```diff
@@ -40,20 +40,20 @@ export const checkpointPlotsSlice = createSlice({
...
-        return {
-          ...state,
-          ...statePayload,
-          hasData: !!action.payload,
-          plotsIds: plots?.map(plot => plot.title) || [],
-          plotsSnapshots: snapShots
-        }
+      if (!action.payload) {
+        return checkpointPlotsInitialState
+      }
```
mattseddon, line 45: [C] I thought it made sense to standardise the approach across all three.

PR #2293, opened 2022-08-30
Give option to sort or filter by Created timestamp

mattseddon:
This PR gives users the ability to sort/filter by the created column. Follow up from https://github.com/iterative/vscode-dvc/pull/2239#discussion_r953246119.

### Demo

~https://user-images.githubusercontent.com/37993418/187358368-d7bdd74d-123d-42ec-8f0b-60a49f4d0754.mov~

https://user-images.githubusercontent.com/37993418/187565674-6d284ce6-4371-43eb-92a3-d9c893d2ef73.mov


extension/src/experiments/columns/quickPick.ts
```diff
@@ -15,13 +14,11 @@ export const pickFromColumnLikes = (
...
   }
   return quickPickValue<ColumnLike>(
-    columnLikes
-      .filter(columnLike => columnLike.path !== timestampColumn.path)
-      .map(columnLike => ({
-        description: columnLike.path,
-        label: columnLike.label,
-        value: columnLike
-      })),
+    columnLikes.map(columnLike => ({
```
mattseddon, line 17: [F] Sorting takes care of itself because the timestamp is an ISO string.

extension/src/util/date.ts
```diff
@@ -0,0 +1,33 @@
+export const getIsoDate = () => new Date().toISOString().slice(0, 10)
```
mattseddon, line 1: [I] Test

extension/src/experiments/model/filterBy/quickPick.ts
```diff
@@ -67,11 +69,43 @@ export const operators = [
...
+    value: Operator.AFTER_DATE
+  },
+  {
+    description: 'Before Date',
+    label: Operator.BEFORE_DATE,
+    types: [ColumnType.TIMESTAMP],
+    value: Operator.BEFORE_DATE
+  },
+  {
+    description: 'On Day',
```
julieg18, line 86: Maybe it's an issue on my end but the on day filter doesn't work for me:


https://user-images.githubusercontent.com/43496356/187434628-24eb97d0-a885-4eeb-b0fd-79c98b4d1de1.mov

extension/src/experiments/model/filterBy/quickPick.ts
```diff
@@ -67,11 +69,43 @@ export const operators = [
...
+      (text?: string): null | string =>
+        isFreeTextDate(text)
+          ? null
+          : 'please enter a valid date of the form yyyy-mm-dd',
+      getIsoDate()
+    )
+  }
+  return getInput(Title.ENTER_FILTER_VALUE)
+}
+
```
julieg18, line 106: Also, probably minor but due to the the "Created" text being left aligned on the column, the filter covers up the text no matter how large the column is: 

image

julieg18 (commented):
Since timestamps are sortable now, do we want to update their context menu to include sorting options: 

image

extension/src/experiments/model/filterBy/quickPick.ts
sroy3: It won't remove the current branch row.

julieg18: > It won't remove the current branch row.

Yes, but shouldn't experiment `exp-f659a` stay because it was created on August 30th? 

image

extension/src/experiments/model/filterBy/quickPick.ts
maxagin: Right side for the left-aligned content?

mattseddon:
Thanks, @julieg18 I'll fix everything.

julieg18 (approved):

PR #2294, opened 2022-08-30
Make the first cells of selected rows sticky

julieg18:
## Demo

https://user-images.githubusercontent.com/43496356/187464292-c21e1657-e213-4388-8e57-b4ad0df49123.mov

Resolves [comments in #2289 ](https://github.com/iterative/vscode-dvc/pull/2289#issuecomment-1230963960)

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -309,31 +309,37 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
       }
     }
 
-    &:not(.rowSelected) {
-      & > *:first-child {
-        position: sticky;
-        left: 0;
-        z-index: 3;
+    > *:first-child {
+      position: sticky;
+      left: 0;
+      z-index: 3;
 
-        &:after {
-          content: '';
-          height: 100%;
-          position: absolute;
-          top: 0;
-          width: 6px;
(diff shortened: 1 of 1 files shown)


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -309,31 +309,37 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
...
       }
     }
 
     &.rowSelected {
       background-color: $row-bg-selected-color;
+
+      > *:first-child {
+        background-color: $row-bg-color;
+
+        &:before {
```
julieg18, line 334: I'm assuming the reason we chose to make selected rows not sticky originally was that the selected rows background color was partially transparent, leading to something like this if you don't use `:before` elements:

image

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -369,7 +375,7 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
       background-color: $header-bg-color;
     }
 
-    &:not(.rowSelected) > *:first-child:after {
```
julieg18, line 372: Notice that this line shouldn't need a `:not(.rowSelected)` since it's inside a `headRow`, which can't get selected :)

PR #2299, opened 2022-08-30
Bump min version of DVC to 2.23.0 (--with-dirs removed from data status)

mattseddon:
# 5/5 `main` <- #2091 <- #2151 <- #2266 <- #2267 <- this

This PR removes the use of a `--with-dirs` from our call to `data status`. The flag was removed in `2.23.0` in favour of having that as the default behaviour.

### DIFF
--- a/demo/requirements.txt
+++ b/demo/requirements.txt
@@ -1,3 +1,3 @@
-dvc[s3]==2.21.1
+dvc[s3]==2.23.0
 torch==1.12.0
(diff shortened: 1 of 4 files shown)


sroy3 (approved):

sroy3:
Is #2266 needed then if you're going to merge this into it?

julieg18 (approved):

PR #2300, opened 2022-08-31
Remove retries from reader (all commands now lockless)

### LINKED ISSUE #2255: Data trees - recover faster after problem is fixed
If there is a problem in the repo and tree is staying long in this state (loading):

Screen Shot 2022-08-24 at 1 31 21 PM

even resolving the problem doesn't help then or at least one would have to wait long time for tree to refresh itself.

Quick workaround show a refresh button in the tree panel when there is a problem (+ related show a proper message - that something is wrong, related to #2254 

Better solutions to explore.

- reset exponential backoff counter on new file changes
- trigger some command from time to time - kinda `dvc health` or something

mattseddon:
# 6/5 `main` <- #2091 <- #2151 <- #2266 <- #2267 <- #2299 <- this 

Closes #2255 by removing retries from our CLI reader. We can do this because the three commands are all now lockless (enough). 

### Demo

https://user-images.githubusercontent.com/37993418/187575463-1d6b5c65-1f16-4cf4-aff8-dcc91b6568fd.mov

Note: I'll be following up by starting to display errors in the DVC tracked tree.


extension/src/cli/dvc/reader.ts
```diff
@@ -156,11 +156,16 @@ export class DvcReader extends DvcCli {
...
-      )) || defaultValue
+    let output
+    try {
+      output = await this.executeDvcProcess(cwd, ...args)
+    } catch (error: unknown) {
+      const errorMessage = (error as Error).message
+      Logger.error(`${args} failed with ${errorMessage} retrying...`)
+    }
+
+    output = output || defaultValue
```
mattseddon, line 167: [F] I'll have to change all of the types to include an optional error which I can then parse/collect and display in the UI.

sroy3 (approved):
6/5 will this ever end or will this overflow to infinity? 😉

julieg18 (approved):

mattseddon:
> 6/5 will this ever end or will this overflow to infinity? 😉

∞

PR #2304, opened 2022-08-31
Only drag rows not single imagges

sroy3:
Turned out to be way simpler than I thought.

https://user-images.githubusercontent.com/3683420/187698877-142aa90a-d740-4fe1-ab5d-0985774d13bc.mov


mattseddon:
👍🏻

PR #2305, opened 2022-08-31
Improve table column resizing styles

julieg18:
* show `columnResizer` bar active styles on hover and when the column is being resized
* keep the resizer bar the same height of the cell(s)

## Demo

https://user-images.githubusercontent.com/43496356/187734504-a6f9e544-02fe-48e2-a7fa-fc6a1630f9e8.mov

Part of #2241

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -1085,7 +1085,7 @@ describe('App', () => {
   it('should disable text selection while resizing', async () => {
     renderTable()
 
-    expect(document.body).not.toHaveClass(styles.noSelect)
+    expect(document.body).not.toHaveClass(styles.isColumnResizing)
 
     const [experimentColumnResizeHandle] = await screen.findAllByRole(
       'separator'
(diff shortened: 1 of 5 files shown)


sroy3 (approved):
Out of scope, I thought the resizer would take the whole border when it's the last item of the group since it's the whole group that is resizing:

https://user-images.githubusercontent.com/3683420/187740873-71586548-402e-4723-a80c-f65b4f77a65b.mov

I believe that it was the original behaviour and that it was changed along the way. I'll raise an issue for that (does not seem to be in the current issues).

maxagin:
> show columnResizer bar active styles on hover and when the column is being resized

Good stuff @julieg18 . Minor: can we keep the resize cursor till the user will release the mouse button


https://user-images.githubusercontent.com/98249521/189009528-48a1f95c-536a-43a3-88d1-e30bd806b366.mov

julieg18:
> Good stuff @julieg18 . Minor: can we keep the resize cursor till the user won't release the mouse button?

Should be easily doable for most elements 🤔

PR #2316, opened 2022-09-01
Fix bugs in table styles

julieg18:
* keep checkpoint dotted lines above border
* keep head cell and body cell borders lining up correctly
* turn off workspace row border when sticky
* update table container background

## Demo

https://user-images.githubusercontent.com/43496356/188031600-6beea2ff-1252-4026-b3a6-97b6adc79cd8.mov

Part of #2241


webview/src/shared/variables.scss
```diff
@@ -10,6 +10,7 @@ $deps-color: var(--vscode-dvc-deps);
 $changed-color: var(--vscode-dvc-workspaceChanged);
 $error-color: var(--vscode-errorForeground);
 
+$experiments-view-bg-color: var(--vscode-menu-background);
 $row-bg-color: $bg-color;
 $header-fg-color: $fg-color;
 $row-bg-alt-color: var(--vscode-sideBar-background);
```
mattseddon, line 16: [Q] Unrelated to this PR but can we change the name of this? It is only use for on-hover in the table.

mattseddon (approved):
- [x] keep checkpoint dotted lines above border
- [x] keep head cell and body cell borders lining up correctly
- [x] turn off workspace row border when sticky*
- [x] update table container background

* in the storybook can now see through the sticky workspace row to contents underneath (see above workspace row):

image

mattseddon:
Note: I cannot recreate the issue shown in the storybook in VS Code.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -243,8 +243,8 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
       .bullet {
         &::after {
           content: '';
-          height: 100%;
-          top: 0;
+          height: 105%;
```
sroy3, line 246: Is it possible to use `calc` here? (same for line256) Like `calc(100% + 5px)`. I guess the 5% here is so the line can overflow the cell?  I feel like `calc` makes this more obvious and more precise.

sroy3 (approved):

julieg18:
> in the storybook can now see through the sticky workspace row to contents underneath (see above workspace row):

Weird! I can't see any issues with workspace row in storybook locally or in chromatic: 

image

@sroy3, are you seeing what Matt spotted?

webview/src/experiments/components/table/styles.module.scss
julieg18: Good idea!

sroy3:
> > in the storybook can now see through the sticky workspace row to contents underneath (see above workspace row):
> 
> Weird! I can't see any issues with workspace row in storybook locally or in chromatic:
> 
> image
> 
> @sroy3, are you seeing what Matt spotted?

I can recreate only when zooming. Normal zoom looks fine.

PR #2321, opened 2022-09-02
Stop row experiment name from being a toggle

julieg18:
## Demo

https://user-images.githubusercontent.com/43496356/188204553-c8815901-533d-444a-8f75-f6a44dc9321f.mov

Part of #2241

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -348,13 +348,26 @@ describe('App', () => {
   })
 
   describe('Toggle experiment status', () => {
+    const getRowByLabel = (label: string) => {
+      const rows = screen.getAllByRole('row')
+
+      return rows.find(element => element.innerHTML.includes(label))
+    }
+
     it('should send a message to the extension to toggle an experiment when the row is clicked', () => {
       renderTable()
 
       const testClick = (label: string, id = label) => {
         mockPostMessage.mockReset()
 
-        fireEvent.click(screen.getByText(label))
+        const row = getRowByLabel(label)
+        if (row) {
+          const bulletToggleElement = within(row).getByLabelText(
+            'Toggle experiment for plots'
+          )
+
+          fireEvent.click(bulletToggleElement)
+        }
 
         expect(mockPostMessage).toBeCalledTimes(1)
         expect(mockPostMessage).toBeCalledWith({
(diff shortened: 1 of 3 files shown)


webview/src/experiments/components/App.test.tsx
```diff
@@ -348,13 +348,26 @@ describe('App', () => {
...
+      return rows.find(element => element.innerHTML.includes(label))
+    }
+
     it('should send a message to the extension to toggle an experiment when the row is clicked', () => {
       renderTable()
 
       const testClick = (label: string, id = label) => {
         mockPostMessage.mockReset()
 
-        fireEvent.click(screen.getByText(label))
```
julieg18, line 357: Since the label is no longer a toggle, we can't just get an element by the label text to check if things are working. I came up with a solution but it's probably over complicated and there is an easier way 😅

sroy3 (approved):
I still believe it's better to keep the toggle on the text and allow selection. The click target of the bullet might be enough, but it does not look like a toggle in my opinion and I don't think people will purposefully click on it. With the experiment name being in the toggle target, there is a lot more chance for discoverability.

maxagin:
Good stuff @julieg18 ! Have you increased the hover area? Can not really see it in the video :)

maxagin:
>I still believe it's better to keep the toggle on the text and allow selection. The click target of the bullet might be enough, but it does not look like a toggle in my opinion and I don't think people will purposefully click on it. With the experiment name being in the toggle target, there is a lot more chance for discoverability.

@sroy3, we have no other option than to point the user to how exactly to select the EXPs for plots. This is something that is planned to be done.

> there is a lot more chance for discoverability.

I am not sure that accidental discovery is good enough :) We need to. make sure it's intuitive. When we will decide how we handle the EXPs in the side panel this problem also going to be unlocked.

mattseddon:
Agree with Stephanie

maxagin:
@mattseddon, my arguments are below.

**What**: we need to tell the user to add EXP to the Plots view. 
**How**: we have the control that does it. Anything besides it is not clear (unless the user was educated before). 
**Conclusion**: Have a special icon/control that does the job - not a cell or group of elements.

In addition:
1. We have no other option than to point the user to how exactly to select the EXPs for plots. `This is something that is planned to be done`.
2. Use a better symbol (eye). for example. We need to. make sure it's intuitive. When we will decide how we handle the EXPs in the side panel this problem also going to be unlocked.

maxagin:
In addition to the discussion, here is the basic row anatomy as I see it.

Screen Shot 2022-09-02 at 6 03 46 PM

mattseddon:
I still agree with Stephanie. All of the behaviour should change together. We should not be making it worse in the interim because `This is something that is planned to be done`. Design/approach has not been agreed.

maxagin:
> All of the behaviour should change together.

By sharing the basic row anatomy I just wanted to express my vision. This particular ticked is about removing the toggle from the Name, which covers everything. What is "All of the behaviour"?

> We should not be making it worse

Why worse? This is more logical I think https://github.com/iterative/vscode-dvc/pull/2321#issuecomment-1235898740 No? If this is not clear, would you like me to rephrase it?

maxagin:
> Design/approach has not been agreed.

@mattseddon, this is a part of #2241 
Please review the scope and let me know if there is something that needs to be re-discussed. Thank you!

shcheklein:
I also take Matt's and Stephanie side on it. Even though I don't like the current implementation - selection is broken - it's a terrible and a very serious UX bug + it's far from being obvious what it toggles + when you toggle is selects/deselects text. It's really bad. 

We should approach a bit more holistically I think (instead of patching it this way and spending time on making special zones to toggle, etc). 

We can probably solve this by showing a tooltip on hover with actions + explanation. At least for now.

maxagin:
> We can probably solve this by showing a tooltip on hover with actions + explanation. At least for now.

This is already included in the #2241
"Add tooltip: Add/Remove to plots (may be different wording) when the EXP's circle has hovered"

PR #2301, opened 2022-08-31
Show data status errors in DVC Tracked tree

mattseddon:
# 2/2 `main` <- #2318 <- this 

Addresses some of https://github.com/iterative/vscode-dvc/issues/2254 by showing error messages in the DVC tracked tree. I have matched the style of the error shown to what we show in the experiments tree. I have also added a command to the solitary tree item which will open the output channel.

### Demo

https://user-images.githubusercontent.com/37993418/187844795-4590b378-d5af-4a25-8797-94ca50b6371b.mov

This was the first iteration which was discarded:

https://user-images.githubusercontent.com/37993418/187592685-665e1a2e-5614-4c88-abe0-1c8537ce3075.mov

PR #2330, opened 2022-09-05
Consolidate use of decoration providers with tree views

mattseddon:
# 1/3 `main` <- this <- #2331 <- #2332

This PR consolidates how we provide non-SCM text decoration and custom tooltips in the DVC tracked and Experiments trees. This is important because we will no doubt iterate on these styles in the future, we want to make it as easy as possible to keep them consistent.


extension/src/experiments/model/decorationProvider.ts
```diff
@@ -1,53 +1,38 @@
 import {
```
mattseddon, line 1: [F] I moved this decoration provided back next to the tree it is associated with. I did the same thing for the other provider. I will be moving `extension/src/fileSystem/tree.ts` => `extension/src/repository/model/tree.ts` in the next PR to match the pattern we use in the experiments folder.

extension/src/experiments/model/decorationProvider.ts
```diff
@@ -0,0 +1,44 @@
+import { EventEmitter, FileDecoration, ThemeColor, Uri } from 'vscode'
```
mattseddon, line 1: [F] I moved this decoration provided back next to the tree it is associated with. I did the same thing for the other provider. I will be moving `extension/src/fileSystem/tree.ts` => `extension/src/repository/model/tree.ts` in the next PR to match the pattern we use in the experiments folder.

extension/src/tree/errorDecorationProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+  FileDecorationProvider,
+  ProviderResult,
+  ThemeColor,
+  Uri,
+  window
+} from 'vscode'
+import { DecoratableTreeItemScheme } from '.'
+import { Disposable } from '../class/dispose'
+
+export abstract class ErrorDecorationProvider
```
mattseddon, line 15: This is the base class that I mentioned [here](https://github.com/iterative/vscode-dvc/pull/2301/files#r961303414).

wolmir (approved):

PR #2333, opened 2022-09-05
Switch from jest from ts-jest to swc

mattseddon:


extension/src/common/logger.test.ts
```diff
@@ -1,5 +1,13 @@
 import { Logger } from './logger'
 
+jest.mock('console', () => {
+  const actualModule = jest.requireActual('console')
+  return {
+    __esModule: true,
+    ...actualModule
+  }
+})
```
mattseddon, line 9: [F] Unfortunately we cannot pull out a helper from this boilerplate as the compiler behaves differently if `jest.mock` in not written like this in the code.

extension/src/processExecution.ts
```diff
@@ -59,16 +58,3 @@ export const executeProcess = async (
   const { stdout } = await createProcess(options)
   return stdout
 }
-
-const sendOutput = (process: Process) =>
-  process.all?.on('data', chunk =>
-    Logger.log(chunk.toString().replace(/(\r?\n)/g, ''))
-  )
-
-export const createProcessWithOutput = (options: ProcessOptions) => {
```
mattseddon, line 68: [F] moving this removed the need for a spy which was breaking some tests. It is currently only used in a single place so was ok to move.

extension/src/common/logger.test.ts
mattseddon, line 9: [F] Unfortunately we cannot pull out a helper from this boilerplate as the compiler behaves differently if `jest.mock` in not written like this in the code.

extension/src/processExecution.ts
mattseddon, line 68: [F] moving this removed the need for a spy which was breaking some tests. It is currently only used in a single place so was ok to move.

package.json
```diff
@@ -56,7 +56,7 @@
     "eslint-plugin-testing-library": "5.6.0",
     "eslint-plugin-unicorn": "43.0.2",
     "husky": "8.0.1",
-    "jest": "29.0.1",
+    "jest": "^29.0.2",
```
mattseddon, line 59: don't even know how this happened

webview/jest.config.js
```diff
@@ -17,8 +17,10 @@ module.exports = {
   moduleNameMapper: {
     '\\.(scss|css|less)$': 'identity-obj-proxy'
   },
-  preset: 'ts-jest',
   setupFiles: ['jest-canvas-mock', '<rootDir>/setup-tests.js'],
-  testEnvironment: 'node',
-  testTimeout: 20000
+  testEnvironment: 'jsdom',
```
mattseddon, line 21: [F] Has to be `jsdom` or everything blows up. Which is fine, we were only doing this to try and get some performance improvements.

webview/src/plots/components/comparisonTable/ComparisonTable.test.tsx
```diff
@@ -496,7 +503,12 @@ describe('ComparisonTable', () => {
 
       const [, firstRow, secondRow] = screen.getAllByRole('rowgroup') // First rowgroup is the thead
 
-      dragAndDrop(secondRow, firstRow, DragEnterDirection.BOTTOM)
+      dragAndDrop(
+        secondRow,
+        firstRow,
+        DragEnterDirection.BOTTOM,
+        EventCurrentTargetDistances
```
mattseddon, line 510: [C] This is gross but I could not find any other way to successfully spy on a single call of `getEventCurrentTargetDistances`. I had to

1. give the function its own module
3. move the import to inside the test file (instead of the util)
4. mock inside the test file for the usual workaround
5. pass the module down into the test helper. 

I had to do this in 4 places.

webview/src/shared/components/iconMenu/IconMenu.test.tsx
```diff
@@ -1,6 +1,3 @@
-/**
- * @jest-environment jsdom
- */
```
mattseddon, line 3: [F] No longer required as the whole suite now runs in `jsdom`.

extension/src/common/logger.test.ts
```diff
@@ -1,5 +1,13 @@
 import { Logger } from './logger'
 
+jest.mock('console', () => {
```
mattseddon, line 3: https://github.com/iterative/vscode-dvc/pull/2333#discussion_r962563536

extension/src/python/index.ts
```diff
@@ -2,7 +2,21 @@ import { join } from 'path'
...
+    Logger.log(chunk.toString().replace(/(\r?\n)/g, ''))
+  )
+
+export const createProcessWithOutput = (options: ProcessOptions) => {
+  const process = createProcess(options)
+
+  sendOutput(process)
+
+  return process
+}
```
wolmir, line 19: Are we allowed to use this to do things like [getting the ast from the project's Python files](https://docs.python.org/3/library/ast.html#command-line-usage) ?

wolmir (approved):

extension/src/python/index.ts
mattseddon: You could definitely try. Take a look at https://github.com/microsoft/vscode-python/tree/main/src/client/languageServer to see what those guys do (outside of the closed source language server)

webview/src/shared/components/dragDrop/currentTarget.ts
```diff
@@ -0,0 +1,11 @@
+import { DragEvent } from 'react'
```
mattseddon, line 1: @sroy3 any suggestions for the name of this file?

extension/src/python/index.ts
wolmir: Why is that closed, btw?

mattseddon: 💰

sroy3 (approved):

Issue #2340, opened 2022-09-06
Long text in labels cause overflow


daavoo:
<img width="803" alt="Captura de Pantalla 2022-09-06 a las 16 34 12" src="https://user-images.githubusercontent.com/12677733/188663751-6fd1640f-1d41-4504-9499-574907e358f4.png">

Depending on the selected zoom, which alters the grid layout, this can get better or worse.

shcheklein:
@daavoo why does it have `\.` btw? is it expected?

daavoo:
> @daavoo why does it have \. btw? is it expected?

It's expected but I have just noticed that it's unnecessary. I added a patch for escaping special characters that were breaking vega-lite rendering. However, it appears that only the `field` needs to be escaped and the `labels`/`title` can be unescaped.

shcheklein:
Yep, I was asking specifically about `\` (escaping) - should it be visible in the titles or not, was it part of the name?

mattseddon:
@sroy3 spotted another issue with some text overflowing on zoom:

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/189015110-95f470d9-02de-4d05-ba1e-23000a6dea7a.png">

daavoo:
> should it be visible in the titles or not, was it part of the name?

`dvc-render` needs to take care of not including the escaping in the title so it doesn't show on the UI

mattseddon:
> > should it be visible in the titles or not, was it part of the name?
> 
> `dvc-render` needs to take care of not including the escaping in the title so it doesn't show on the UI

@daavoo do you have a ticket for that?

daavoo:
> > > should it be visible in the titles or not, was it part of the name?
> > 
> > 
> > `dvc-render` needs to take care of not including the escaping in the title so it doesn't show on the UI
> 
> @daavoo do you have a ticket for that?

https://github.com/iterative/dvc-render/pull/82

Will be most likely available on next dvc release

PR #2344, opened 2022-09-06
Truncate long plot titles

sroy3:
Fixes part of #2340 
Screen Shot 2022-09-06 at 1 53 03 PM
Screen Shot 2022-09-06 at 1 50 00 PM
Screen Shot 2022-09-06 at 1 48 52 PM

I went with the min width of each plot size (there is not much flexibility here, we cannot use css or dynamically change the size with percentages).


wolmir (approved):

mattseddon:
[Q] What changed the title of the smooth plot in the storybook? I can't see anything obvious from a glance.

sroy3:
> [Q] What changed the title of the smooth plot in the storybook? I can't see anything obvious from a glance.

The only thing I can think of is that we truncate with the param `'left'`, which might cause everything to shift to the left side of the label (by what looks like barely a pixel) even though we keep the whole label as is.

webview/src/plots/components/ZoomablePlot.tsx
```diff
@@ -12,12 +13,20 @@ interface ZoomablePlotProps {
   spec: VisualizationSpec
   data?: PlainObject
   id: string
+  size: PlotSize
+}
+
+const TitleLimit = {
+  [PlotSize.LARGE]: 70,
```
sroy3, line 20: I have a follow up for the axis labels that will fine tune these values a little bit more.

mattseddon:
> > [Q] What changed the title of the smooth plot in the storybook? I can't see anything obvious from a glance.
> 
> The only thing I can think of is that we truncate with the param `'left'`, which might cause everything to shift to the left side of the label (by what looks like barely a pixel) even though we keep the whole label as is.

I know but we had 
image

and now we have

image

sroy3:
> > > [Q] What changed the title of the smooth plot in the storybook? I can't see anything obvious from a glance.
> > 
> > 
> > The only thing I can think of is that we truncate with the param `'left'`, which might cause everything to shift to the left side of the label (by what looks like barely a pixel) even though we keep the whole label as is.
> 
> I know but we had image
> 
> and now we have
> 
> image

That's weird, when I checked the diff, things moved a few pixels only, now I'm seeing what you mean. Will look into it, I probably mutated the fixture somehow

sroy3:
Since I'm rebuilding the title and `spec.title` as a type of `Title | Text | TitleParams<ExprRef | SignalRef>'` (`string[]`, `string`). I thought (naively maybe) it might be simpler to use the id if a title is present as they are the same. For the smooth template plot, I copied another plot (id included) and override it with the content of a smooth template plot with a different title. That's why it changed in Storybook. I'll change to just use the title since it's safer and future proof if we ever do something with ids or titles.

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
```diff
@@ -44,49 +46,47 @@ export const TemplatePlotsGrid: React.FC<TemplatePlotsGridProps> = ({
...
   const [disabledDrag, setDisabledDrag] = useState('')
-  const nbVegaPanels = useRef(0)
+  const size = useSelector((state: PlotsState) => state.template.size)
 
-  useEffect(() => {
-    setOrder(entries.map(({ id }) => id))
-  }, [entries])
-
-  // eslint-disable-next-line sonarjs/cognitive-complexity
-  const layoutEffect = () => {
```
sroy3, line 54: With truncation, we somehow entered a race condition where this would be called before vega is ready. I found the `onNewView` method on the `<VegaLite />` component that gets called after the view is ready, which makes things much simpler and cleaner for us. That `onLayoutEffect` is also used elsewhere so I'll do a follow up after to use `onNewView` instead.

mattseddon (approved):

PR #2346, opened 2022-09-07
Remove extra sending of data to webview on change of star selection

mattseddon:
See inline comment

### DIFF
--- a/extension/src/experiments/webview/messages.ts
+++ b/extension/src/experiments/webview/messages.ts
@@ -170,7 +170,6 @@ export class WebviewMessages {
 
   private setExperimentStars(ids: string[]) {
     this.experiments.toggleStars(ids)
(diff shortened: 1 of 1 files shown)


extension/src/experiments/webview/messages.ts
```diff
@@ -170,7 +170,6 @@ export class WebviewMessages {
 
   private setExperimentStars(ids: string[]) {
     this.experiments.toggleStars(ids)
-    this.sendWebviewMessage()
```
mattseddon, line 173: `this.notifyChanged()` below already sends the data. We have to send the data to the webview and update the trees because we added the filter to starred action.

julieg18 (approved):

sroy3 (approved):

PR #2348, opened 2022-09-07
Switch ts-loader to swc-loader

mattseddon:
Having a look at making this switch. There are a few things that are still handled by `tsc` that maybe we can also switch.


mattseddon:
Complication does not seem much faster, no point in making the switch right now.

PR #2351, opened 2022-09-07
Update row styles

julieg18:
* update row selection styles
* update indicator count styles
* change cursor to a pointer on cursor hover
* add some space between checkbox and star
* take off last table row border

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -205,6 +205,7 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
   position: relative;
   border: none;
   background: none;
+  cursor: pointer;
 }
 
 .expandedRowArrow {
@@ -221,6 +222,14 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
   bottom: 0;
 }
 
+.rowGroup:last-child > .tr:last-child {
+  border-color: $row-bg-color;
+
+  &.rowSelected {
+    border-color: $row-bg-selected-color;
+  }
+}
+
 .experiments {
   box-sizing: border-box;
   height: 100%;
@@ -228,6 +237,14 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
   flex-flow: column nowrap;
 
   .experimentGroup {
+    &:last-child .tr:last-child {
+      border-color: $row-bg-color;
+
+      &.rowSelected {
+        border-color: $row-bg-selected-color;
+      }
+    }
+
     .nestedRow {
       .experimentCell {
         padding-left: $nested-row-padding;
@@ -296,10 +313,33 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
     position: relative;
     background-color: $row-bg-color;
 
+    &:hover.rowSelected {
+      &:after {
+        box-shadow: inset 0 0 20px $shadow;
+        content: '';
+        top: 0;
+        right: 0;
+        width: 100%;
+        height: 100%;
+        z-index: 4;
+        position: absolute;
+        pointer-events: none;
+      }
+
+      .td:hover,
+      .td:hover + .td {
+        border-left-color: $row-border-selected-color;
+      }
+    }
+
     &:hover:not(.rowSelected) {
       .td:not(.experimentCell),
       .experimentCell:before {
         background-color: $row-hover-background-color;
+      }
+
+      .indicatorCount {
+        border-color: $row-hover-background-color;
       }
 
       .td:hover,
@@ -325,6 +365,8 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
     }
 
     &.rowSelected {
+      border-color: $row-border-selected-color;
+
       .td {
         background-color: $row-bg-selected-color;
       }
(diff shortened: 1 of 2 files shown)


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -336,6 +378,21 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
...
+        background-color: $indicator-selected-badge-background;
+        color: $indicator-selected-badge-foreground;
+        border: 1px solid $indicator-selected-badge-border;
+      }
+
+      .starSwitch:not([aria-checked='true']) svg {
+        fill: $star-switch-selection-fill-color;
+      }
+
+      .expandedRowArrow,
```
julieg18, line 392: Went ahead and updated the cursor color when selected as well since it's initially the same color as the stars:

image

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -336,6 +378,21 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
...
       }
+
+      .indicatorCount {
+        background-color: $indicator-selected-badge-background;
+        color: $indicator-selected-badge-foreground;
+        border: 1px solid $indicator-selected-badge-border;
+      }
+
+      .starSwitch:not([aria-checked='true']) svg {
+        fill: $star-switch-selection-fill-color;
```
julieg18, line 389: With this update, the star and cursor are now more viewable on the light theme 🎉 

Current:

image

PR:

image

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -325,6 +365,8 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
     }
 
     &.rowSelected {
+      border-color: $row-border-selected-color;
```
julieg18, line 368: Originally, we wanted to take the final border of multiple selected rows off, but I couldn't find a way to do that without writing some complex JS solution: 

Screen Shot 2022-09-07 at 6 41 52 PM
Screen Shot 2022-09-07 at 6 42 03 PM

webview/src/experiments/components/table/styles.module.scss
maxagin: This is very good !

webview/src/experiments/components/table/styles.module.scss
maxagin: Q: CSS :last-child selector? Or something similar?

maxagin:
This is just for the record, I just saw your comment regarding it: We want to have a bottom border for the selected rows, except the last one or only one, in those cases, there is no border bottom needed.

I completely forgot to include the Chevron. Good stuff @julieg18 !

> change cursor to a pointer on cursor hover

Chevron hover probably :)

> add some space between checkbox and star

Just for your information, it is 2px currently everywhere in design.

### :pray:

maxagin:
Probably partly includes "Checkbox and star styles when the row selected"?

webview/src/experiments/components/table/styles.module.scss
mattseddon: Feels like we are overcomplicating this. Chevrons and badges do not need to change color:

image

julieg18: > Feels like we are overcomplicating this. Chevrons and badges do not need to change color:

By badges, I'm assuming you mean the indicator counts? Agreed, that these would probably be ok to leave as is! cc @maxagin. 

As for chevrons, updating the color does improve contrast on a couple different tested themes, including [VSCode's default light theme](https://github.com/iterative/vscode-dvc/pull/2351#discussion_r965383603) 🤔

webview/src/experiments/components/table/styles.module.scss
julieg18: > Q: CSS :last-child selector? Or something similar?

A CSS selector doesn't work correctly due to how the table is built out of nested rows vs all rows being siblings.

sroy3 (approved):

sroy3:
Do we also want to change these badges?


https://user-images.githubusercontent.com/3683420/189134381-b1a49cc7-eaac-44fd-8688-1c27d6bd7df0.mov

maxagin:
> Do we also want to change these badges?

Should be the same everywhere if possible IMO

wolmir (approved):

sroy3:
> > Do we also want to change these badges?
> 
> Should be the same everywhere if possible IMO

So leave them grey? It looks disabled.

wolmir:
Yes, I think we should choose a different color for those badges.
They are supposed to draw attention, anyway

mattseddon:
> Should be the same everywhere if possible IMO
> > > Do we also want to change these badges?
> > 
> > 
> > Should be the same everywhere if possible IMO
> 
> So leave them grey? It looks disabled.

> Yes, I think we should choose a different color for those badges. They are supposed to draw attention, anyway

My $0.02 -> it would be a lot simpler to stick with the UI elements as provided by VS Code. The whole row selection styling is overcomplicated and multi-select in the table has very limited functionality. 

Discuss in the retro.

julieg18:
So it looks like we want to keep the indicators as is and update the chevrons/stars to match the VSCode tree chevrons. Should we go ahead and merge this and do those things in a follow-up?

mattseddon:
Tree behaviour:

https://user-images.githubusercontent.com/37993418/189243819-36cb3c4a-e308-4d96-8ffa-cd4d43ca6ac2.mov


https://user-images.githubusercontent.com/37993418/189243849-af0c9f44-1d7c-474e-921d-6fb8551dfeaa.mov

Issue #2028, opened 2022-07-14
Checkboxes/Context-menu follow-ups and pending issues


wolmir:
- [x] Dismiss interactive tooltips on ESC (https://github.com/iterative/vscode-dvc/pull/1986#issuecomment-1177989980)
- [ ] On a bulk selection, if any of the experiments is removable, then show the remove option and remove only the ones we can (https://github.com/iterative/vscode-dvc/pull/1986/files#r914334330). **NOTE**: This needs discussion with team, maybe
- [x] Refactor the context-menu options logic (https://github.com/iterative/vscode-dvc/pull/1986#discussion_r914336887)

PR #2354, opened 2022-09-08
Dismiss tooltip/context menus on Escape

mattseddon:
One item from https://github.com/iterative/vscode-dvc/issues/2028:

> - [ ] Dismiss interactive tooltips on ESC (https://github.com/iterative/vscode-dvc/pull/1986#issuecomment-1177989980)

### Demo

https://user-images.githubusercontent.com/37993418/189049026-99192485-7a07-4ebe-a8dc-46a5a69660b9.mov

(trust me I am pressing escape)


webview/src/shared/components/tooltip/Tooltip.tsx
```diff
@@ -6,6 +6,27 @@ import 'tippy.js/dist/tippy.css'
 export const HEADER_TOOLTIP_DELAY = 100
 export const CELL_TOOLTIP_DELAY = 1000
 
+const hideOnEsc = {
```
mattseddon, line 9: Found this [here](https://github.com/atomiks/tippyjs/issues/750). It is also shown in the official docs: https://atomiks.github.io/tippyjs/v6/plugins/#hideonesc

mattseddon: I can't think of a reason not to have this on any of the tooltips/context menus but LMK if you think differently.

webview/src/shared/components/tooltip/Tooltip.tsx
```diff
@@ -6,6 +6,27 @@ import 'tippy.js/dist/tippy.css'
...
+  defaultValue: true,
+  fn({ hide }: { hide: () => void }) {
+    function onKeyDown(event: KeyboardEvent) {
```
julieg18, line 12: Things appear to be working correctly for the body cells but the `esc` key isn't working for the header cells on my end:

https://user-images.githubusercontent.com/43496356/189126147-69985ffe-bf05-4a68-8eb7-4f37e5071061.mov

wolmir (approved):

webview/src/shared/components/tooltip/Tooltip.tsx
mattseddon: Yep, doesn't work... isn't obvious why either.

mattseddon: I'll fix it today.

webview/src/experiments/components/table/TableHeaderCellContents.tsx
```diff
@@ -41,9 +41,6 @@ export const ColumnDragHandle: React.FC<{
     <span
       data-testid="rendered-header"
       className={cx(styles.cellContents)}
-      onKeyDown={e => {
```
mattseddon, line 44: [F] This was the cause of the breakage in the header context menu. From memory, we put this in to stop tab from causing all kinds of problems before we had a context menu. I could not find any reason to keep it around. 

@wolmir do you remember why you added this?

wolmir: @mattseddon No, I can't remember why I added it here

Sorry for the delay in the reply!

PR #2358, opened 2022-09-08
Truncate axis titles

sroy3:
Part of #2340 
Screen Shot 2022-09-08 at 10 30 11 AM

This technically works, but I'm not really happy about this because I previously got it to work by changing the titles on `spec.encoding.x` and `spec.encoding.y` and I feel like these could be found pretty much anywhere in the object structure.  I'm going to make a util that finds axes anywhere in the spec instead.


sroy3:
The reason why it was on `spec.encoding.x` before and not now is because it was in another project, not the demo one. All the more reason to close this and go with the other approach (already started working on it).

PR #2359, opened 2022-09-08
Create language server for dvc.yaml

wolmir:


languageServer/src/DvcLanguageServer.ts
```diff
@@ -0,0 +1,94 @@
...
+    language: 'json'
+  },
+  {
+    language: 'toml'
+  },
+  {
+    language: 'python'
+  }
+]
+export class DvcLanguageServer {
```
wolmir, line 29: I'm think about removing the `Dvc*` prefix from these names. It seems redundant and three more characters to type

mattseddon:
Needs functionality demos.

extension/src/lsp-client/language-client.ts
```diff
@@ -0,0 +1,69 @@
...
+      clientOptions
+    )
+
+    // Start the client. This will also launch the server
+    this.dispose.track(this.start())
+  }
+
+  start() {
+    this.client.start()
+    findFiles('**/*.py', '.??*').then(files => {
```
mattseddon, line 54: [C] This will need to be updated when the filesystem is updated with `.py` files (needs a watcher) can be done in follow up.

languageServer/src/test/fixtures/examples/make.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs-jetpack')
```
mattseddon, line 1: [Q] Did you use this to generate the test data? Is it usable anymore?

languageServer/src/DvcTextDocument.ts
```diff
@@ -0,0 +1,319 @@
...
+  symbolAt(position: Position): DocumentSymbol | undefined {
+    return this.symbolScopeAt(position).pop()
+  }
+
+  getTextDocuments() {
+    return this.documents
+  }
+
+  createFinder(txtDoc: TextDocument) {
+    return new DvcTextDocument(txtDoc, this.documents)
```
mattseddon, line 286: geTextDocuments is right there, any other usage you could replace? Or just drop the accessor?

languageServer/src/DvcTextDocument.ts
```diff
@@ -0,0 +1,319 @@
...
+    documents: TextDocuments<TextDocument>,
+    pythonFilePaths?: string[]
+  ) {
+    this.textDocument = textDocument
+    this.documents = documents
+    this.uri = this.textDocument.uri
+    this.pythonFilePaths = pythonFilePaths
+  }
+
+  offsetAt(position: Position) {
```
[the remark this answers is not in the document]
mattseddon: The main reason for doing this is to raise the public methods (that are used as entry points for `onCodeAction` etc to the top of the class. That means that anyone reviewing the code (or seeing it for the first time) should be able to get to grips with what it is doing at the top level quickly.

For example I just started trying to workout what is going on using `yamlScalarNodeToDocumentSymbols` as a starting point.... my brain melted.

languageServer/src/DvcYamlModel.ts
```diff
@@ -0,0 +1,177 @@
...
+      valueNode.range &&
+      inRange(offset, valueNode.range[0], valueNode.range[2])
+    )
+  }
+}
+
+export class DvcYaml implements IDvcYamlModel {
+  private stages: IDvcYamlStage[] = []
+
+  constructor(private document: IDvcTextDocument) {
```
mattseddon, line 145: For the sake of consistency can we not use this constructor shorthand, please.

languageServer/src/DvcYamlModel.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+    }
+  }
+
+  getStageAt(position: Position): IDvcYamlStage | undefined {
+    const offset = this.document.offsetAt(position)
+    return this.stages.find(stage => stage.contains(offset))
+  }
+
+  toString(): string {
```
mattseddon, line 163: Can you please go through these classes and remove any code that is not currently being used to serve something to the client. We need to use YAGNI.

mattseddon (commented):
Please go through and clean everything up. I will come back to this next week.

languageServer/jest.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  coverageDirectory: 'coverage/jest',
+  coveragePathIgnorePatterns: ['<rootDir>/src/test/'],
+  coverageReporters: ['json'],
+  preset: 'ts-jest',
```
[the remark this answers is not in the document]
sroy3: Maybe we could have a main config at root that we could extend.

sroy3 (commented):
We've seen demos of what this looks like, but I think an introduction into this code would be beneficial. There is a lot of new stuff in here.

languageServer/.lintstagedrc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('../.lintstagedrc')
```
mattseddon, line 1: Also, missing `prettier` altogether?

languageServer/src/DvcLanguageServer.ts
```diff
@@ -0,0 +1,94 @@
...
+      return this.getDvcTextDocument(params)?.getCodeActions(params) ?? null
+    })
+
+    connection.onRequest('sendPythonFiles', (files: string[]) => {
+      this.pythonFilePaths = files
+    })
+
+    this.documents.listen(connection)
+
+    connection.listen()
```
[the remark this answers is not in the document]
wolmir: Documents is a helper from the vscode-languageserver SDK.
It implements the handlers needed to synchronise the open text documents between the client and the server so we don't have to.

But we still need to manually handle the workspace file events like create/delete/rename and documents that are in the workspace, but the user didn't open them during the current session.

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -0,0 +1,294 @@
...
+  private findLocationsFor(aSymbol: DocumentSymbol) {
+    const parts = aSymbol.name.split(/\s/g)
+    const txt = this.getText()
+
+    return parts
+      .map(str => txt.indexOf(str))
+      .filter(index => index > 0)
+      .map(index => this.positionAt(index))
+      .map(pos => this.symbolAt(pos)?.range)
+      .filter(Boolean)
```
mattseddon, line 291: Could this be
```
    const acc: Location[] = []
    for (const str of parts) {
      const index = txt.indexOf(str)
      if (index <= 0) {
        continue
      }
      const pos = this.positionAt(index)
      const range = this.symbolAt(pos)?.range
      if (!range) {
        continue
      }
      acc.push(Location.create(this.uri, range as Range))
    }
    return acc
```

mattseddon: (tests still pass after that change)

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    return []
+  }
+
+  private symbolScopeAt(position: Position): DocumentSymbol[] {
+    const cursorOffset: number = this.offsetAt(position)
+
+    const symbolsFound: Array<DocumentSymbol | null> = []
+
+    visit(this.getYamlDocument(), (_key, node) => {
```
mattseddon, line 254: `_key` can just be `_`. That's what we normally use for an unused variable

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -0,0 +1,294 @@
...
+      if (isNode(node)) {
+        const range = node.range
+        if (!!range && cursorOffset >= range[0] && cursorOffset <= range[2]) {
```
mattseddon, line 257: Are you looking to see if the cursor is in the range here? Can we add a variable to explain what is going on?

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -0,0 +1,294 @@
...
+    )
+
+    return [edit]
+  }
+
+  private yamlNodeToDocumentSymbols(
+    node: Node | Pair,
+    range: [number, number, number]
+  ): DocumentSymbol[] {
+    if (isNode(node)) {
```
mattseddon, line 238: early return might make this easier to read

```
  private yamlNodeToDocumentSymbols(
    node: Node | Pair,
    range: [number, number, number]
  ): DocumentSymbol[] {
    if (isNode(node)) {
      if (!isScalar(node)) {
        return []
      }
      return this.yamlScalarNodeToDocumentSymbols(node, range)
    }
    if (isPair(node)) {
      return this.yamlNodeToDocumentSymbols(node.value as Node | Pair, range)
    }

    return []
  }
```

mattseddon: still pretty hard

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -0,0 +1,294 @@
...
+  ) {
+    const nodeValue = `${node.value}`
+
+    const symbolsSoFar: DocumentSymbol[] = [
+      DocumentSymbol.create(
+        nodeValue,
+        undefined,
+        SymbolKind.String,
+        Range.create(this.positionAt(nodeStart), this.positionAt(nodeEnd)),
+        Range.create(this.positionAt(nodeStart), this.positionAt(valueEnd))
```
mattseddon, line 216: [Q] Why are the `range` and `selectionRange` not equal here but they are equal when you create `DocumentSymbols` in `getTemplateExpressionSymbolsInsideScalar`?

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -0,0 +1,294 @@
...
+    scalarValue: string,
+    nodeOffset: number
+  ) {
+    const templateSymbols: DocumentSymbol[] = []
+
+    const templates = scalarValue.matchAll(variableTemplates)
+    for (const template of templates) {
+      const expression = template[1]
+      const expressionOffset: number = nodeOffset + (template.index ?? 0) + 2 // To account for the '${'
+      const symbols = expression.matchAll(/[\dA-Za-z]+/g)
```
mattseddon, line 181: Looks a lot like the regex [here](https://github.com/iterative/vscode-dvc/pull/2359/files#diff-f6824a6c404b5ff658c1c62a45c9ea9cb5a571cab1951e3adfa52d64174144ebR54). Same comment about trying to explain what you're matching and why

languageServer/src/test/completions.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        mockContents: '',
+        mockPath: 'params.yaml'
+      }
+    ])
+    const response = await requestCompletions(dvcYaml)
+
+    expect(response).toBeTruthy()
+    expect(response).toStrictEqual([])
+  })
+})
```
mattseddon, line 54: Should at least test the other completions using one of the other `dvc.yaml` files from the fixture.

languageServer/src/test/fixtures/examples/invalid/index.ts
```diff
@@ -0,0 +1,84 @@
+export const foreach_extra_dvc_yaml = `
```
mattseddon, line 1: all of these fixtures should be getting used.

languageServer/src/test/utils/setup-test-connections.ts
```diff
@@ -0,0 +1,33 @@
...
+import { LanguageServer } from '../../LanguageServer'
+
+class TestStream extends Duplex {
+  _write(chunk: string, _encoding: string, done: () => void) {
+    this.emit('data', chunk)
+    done()
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _read(_size: number) {}
```
mattseddon, line 12: I think this can be `_read() {}` and you can drop the disable line.

mattseddon (commented):
Good iteration.

languageServer/index.d.ts
```diff
@@ -0,0 +1,4 @@
+import { DocumentFilter } from 'vscode-languageserver'
```
sroy3, line 1: Why not just let `tsc` emit the declarations for us? This file is pretty simple, so I don't think it's really necessary, but would it help making it future proof?

languageServer/src/ITextDocumentWrapper.ts
```diff
@@ -0,0 +1,7 @@
+import { Position } from 'vscode-languageserver-textdocument'
+import { Document } from 'yaml'
+
+export interface ITextDocumentWrapper {
```
sroy3, line 4: We're not prefixing our interfaces with `I` anywhere else. I think it would be best to stick to the current format (no prefix).

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+  label: string
+  body: CompletionTemplateBody
+}
+export class TextDocumentWrapper implements ITextDocumentWrapper {
+  uri: string
+
+  private textDocument: TextDocument
+  private pythonFilePaths: string[] = []
+
+  constructor(textDocument: TextDocument, pythonFilePaths?: string[]) {
```
sroy3, line 42: Could be `constructor(textDocument: TextDocument, pythonFilePaths: string[] = []) {` instead of `this.pythonFilePaths = pythonFilePaths ?? []`

sroy3 (commented):
This is starting to look really good. I had a few comments while testing it:

- Really not a big deal, but I had to run `yarn dev-server` twice before I could get it to work. I guess it might be building after the extension or something like that. 
- How do you work the debug mode? I'm inspecting port 6009 with multiple `debugger` and `console.log` but nothing seems to be happening. I'm only seeing source maps missing warnings (would be great to create them btw).
- Is the big gap around the title normal? It looks like a param is missing in the creation of the popover:
Screen Shot 2022-09-15 at 10 23 55 AM

languageServer/src/ITextDocumentWrapper.ts
mattseddon: I do use the I prefix in the extension in at least one place where multiple things implement the interface (`ICli`)

wolmir: Yes, I looked it up beforehand to see if there were precedents 😁

But It's just a cosmetic quirk from me, I don't know if there is any real benefit from it.

I can take them out for consistency

PR #2362, opened 2022-09-08
Suggest existing files as parameter files

wolmir:


extension/src/lsp-client/language-client.ts
```diff
@@ -60,6 +61,19 @@ export class DVCLanguageClient extends Disposable {
...
     })
 
+    findFiles('**/*.{yaml,toml,json}', '.??*').then(files => {
+      this.client.sendRequest(
+        'sendParamsFiles',
+        files.map(file => {
+          return {
+            languageId: path.extname(file),
+            text: readFileSync(file),
+            uri: Uri.file(file).toString()
```
mattseddon, line 71: [C] This one looks correct but the relative path above should also be absolute so that we can suggest the relative path inside of the file that is currently being edited.

PR #2363, opened 2022-09-09
Ensure welcome screen is not shown when only one column is provided

mattseddon:
The initial problem of the table not being shown has been fixed. However, the indicators are mashed in with the experiments header name if the shown columns have a depth of 1:

### Demo

https://user-images.githubusercontent.com/37993418/189264864-ae724051-757b-4047-b795-9fd5c582028a.mov

I spent a good chunk of time looking at this. Can someone point me in the right direction to fix it?

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -136,6 +136,18 @@ describe('App', () => {
     expect(noColumnsState).toBeInTheDocument()
   })
 
+  it('should not show the no columns selected empty state when only the timestamp column is provided', () => {
+    renderTable({
+      ...tableDataFixture,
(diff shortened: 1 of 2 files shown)


sroy3 (approved):

julieg18 (approved):
Looks good! 

> However, the indicators are mashed in with the experiments header name if the shown columns have a depth of 1:
> I spent a good chunk of time looking at this. Can someone point me in the right direction to fix it?

One solution that comes to mind is checking for length of head rows in `TableHead`. If there is only one, add a class that adjusts the head cells to be twice as tall, giving enough room for the indicators in the head! Might also may be able to check for a single head row with just CSS and `:only-of-type` 🤔

PR #2367, opened 2022-09-09
Update table indicators and chevrons/stars styles

julieg18:
* update count indicators, matching style to VSCode
* update row chevrons and stars, updating colors to match VSCode tree chevrons

## Demo

https://user-images.githubusercontent.com/43496356/189404082-0c23c291-d0ee-4b1b-972d-535b7d2ab798.mov

Followup to #2351


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -338,10 +337,6 @@ $workspace-row-edge-margin: $edge-padding - $cell-padding;
         background-color: $row-hover-background-color;
       }
 
-      .indicatorCount {
-        border-color: $row-hover-background-color;
```
julieg18, line 342: @maxagin, I took off the border of the indicators since VSCode doesn't have one for their indicators, but if we want to keep that effect, I don't see an issue with adding it back :) 

Figma:
image

webview/src/shared/variables.scss
```diff
@@ -27,14 +27,14 @@ $row-hover-background-color: var(--vscode-list-hoverBackground);
...
-$indicator-selected-badge-border: $row-bg-selected-color;
-$indicator-selected-badge-background: $accent-color;
-$indicator-selected-badge-foreground: var(--button-primary-foreground);
-
-$star-switch-selection-fill-color: var(--vscode-list-focusHighlightForeground);
+$indicator-badge-background: var(--vscode-activityBarBadge-background);
+$indicator-badge-foreground: var(--vscode-activityBarBadge-foreground);
+
+$row-action-icon-color: var(--vscode-icon-foreground);
+$row-action-selected-icon-color: var(
```
julieg18, line 34: `--vscode-icon-foreground` appears to be the variable used for tree chevrons! 

Also, I realized that the default theme trees don't actually have their chevrons stay the same color when selected. Looks like they actually change their color based on the variable `--vscode-list-activeSelectionIconForeground`. The light theme and a couple other themes use this variable as well:

image
image

Matching how vscode themes work, I had our table update its chevrons if it's selected with `--vscode-list-activeSelectionIconForeground`, defaulting to the `--vscode-icon-foreground` since not all themes use the active variable.

sroy3 (approved):

webview/src/shared/variables.scss
mattseddon: Looks like the text also changes color.

mattseddon (approved):

webview/src/shared/variables.scss
julieg18: Yes! Probably also relies on some kind of `active` variable 🤔. Can look into it and open a pr at a later date!

mattseddon: 🙏🏻

PR #2376, opened 2022-09-12
Ensure table indicators do not overlap with experiment column title

mattseddon:
This PR follows on from https://github.com/iterative/vscode-dvc/pull/2363 by making sure that the table indicators never overlap with the experiments column text.

The approach taken is to add a placeholder to the experiments column. This means that the table header will never be < 2 rows high.

### Demo

https://user-images.githubusercontent.com/37993418/189558361-344576b1-8203-4a7f-93e3-3020e0b88b13.mov

### DIFF
--- a/webview/src/experiments/components/table/TableHead.tsx
+++ b/webview/src/experiments/components/table/TableHead.tsx
@@ -97,8 +97,6 @@ export const TableHead = ({
 
   const selectedForPlotsCount = getSelectedForPlotsCount(rows)
 
-  const firstExpColumnCellId = headerGroups[0].headers[0].id
-
   return (
     <div className={styles.thead} ref={wrapper}>
       <Indicators selectedForPlotsCount={selectedForPlotsCount} />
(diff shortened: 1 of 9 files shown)


webview/src/experiments/components/table/TableHead.tsx
```diff
@@ -97,8 +97,6 @@ export const TableHead = ({
 
   const selectedForPlotsCount = getSelectedForPlotsCount(rows)
 
-  const firstExpColumnCellId = headerGroups[0].headers[0].id
```
mattseddon, line 100: [F] This caused me much pain. The only way that the bug showed itself was in two stories. Seems more sensible to use a util function instead of passing this variable down through all of the components.

mattseddon: Bug was that the experiment column shadow was shown when it didn't need to be.

webview/src/experiments/util/columns.ts
```diff
@@ -114,3 +114,8 @@ export const leafColumnIds: (
 
   return [column.id]
 }
+
+export const EXPERIMENT_COLUMN_ID = 'id'
+
+export const isExperimentColumn = (id: string): boolean =>
```
julieg18, line 120: Good idea!

julieg18 (approved):
Great work!

sroy3 (approved):

wolmir (approved):

PR #2377, opened 2022-09-12
Move experiment table context menus to right-start

mattseddon:
Addresses the following point from https://github.com/iterative/vscode-dvc/issues/2241

>  - [ ] All context menus need to appear on the right side of the target point.
The context menu is usually placed on the right/bottom side of the mouse click unless it is close to the right age of the screen, then it will appear on the left side. However, in VSCode it seems like always on the right side, but we can do it correctly I think.

### Demo

https://user-images.githubusercontent.com/37993418/189559146-9b12a998-6009-459b-8b5d-1436256104e7.mov

### DIFF
--- a/webview/src/shared/components/contextMenu/ContextMenu.tsx
+++ b/webview/src/shared/components/contextMenu/ContextMenu.tsx
@@ -56,7 +56,7 @@ export const ContextMenu: React.FC<ContextMenuProps> = ({
   <Tooltip
     trigger={trigger}
     delay={[100, 200]}
-    placement={'bottom'}
+    placement={'right-start'}
     interactive
     isContextMenu={true}
     onTrigger={positionContextMenuAndDisableEvents}


webview/src/shared/components/contextMenu/ContextMenu.tsx
```diff
@@ -56,7 +56,7 @@ export const ContextMenu: React.FC<ContextMenuProps> = ({
   <Tooltip
     trigger={trigger}
     delay={[100, 200]}
-    placement={'bottom'}
+    placement={'right-start'}
```
mattseddon, line 59: All possible placements are [here](https://atomiks.github.io/tippyjs/#placements).

shcheklein: should we do right-end?

mattseddon: You would think so but:


https://user-images.githubusercontent.com/37993418/189559868-9a5be8ec-d7f0-419e-9e18-cb93f2f394e5.mov

julieg18 (approved):

sroy3 (approved):

wolmir (approved):

webview/src/shared/components/contextMenu/ContextMenu.tsx
maxagin: Normally it's bottom-end. No?

mattseddon: bottom-end:


https://user-images.githubusercontent.com/37993418/189771989-477e192d-e26c-4425-b143-8cea497ce1bd.mov

maxagin: Maybe I was not clear enough, sorry. This is an example from Figma 

Screen Shot 2022-09-12 at 7 26 32 PM

Issue #1757, opened 2022-05-24
Accomodate flexible plots


mattseddon:
DVC are about to introduce flexible plots via the `plots diff` command. 

### Example of a flexible plot:

![image](https://user-images.githubusercontent.com/37993418/169924818-f94c0f14-9be7-4b60-ae5e-27cc7f8ff33f.png)

In this particular example, we can see that there are multiple lines for the HEAD commit and multiple lines for the workspace. 

### Current VS Code UI:

![image](https://user-images.githubusercontent.com/37993418/169924663-2a6e0299-13a5-4a33-83d9-e02c26484609.png)

When comparing the two it should become clear that there is an issue. In VS Code we have a single line per experiment. If we had a flexible plot and `exp-84614` (from screenshot) needed to show train vs test (and a line for each) then what would we do? We currently do not show a legend as we want to "globally link" each color back to an individual experiment. We cannot simply show two red lines. We could potentially show one as dashed and the other solid but we should still need to add an indicator somewhere in the view to detail which is which. It also would not be long before we run out of line types. 

In order to accommodate the new plots data we need to do a few things (with different priorities):

 - [x] Make sure no breaking changes - p1
 - [x] Consume new data - p1
 - [x] Update the design to be able to accommodate the new visuals - p2

mattseddon:
Update: We've verified that flexible plots will not be shown in VS Code (and nothing will break) after the PR is merged.

We still need to discuss how we will cater to these plots in VS Code. The initial thought is to go with the different line types but we will still need a way to show this in VS Code. We can maybe bring back the legend specifically for these types of plots. 

If any changes need to be made to the default templates then we should consider #1957.

We should discuss this at the next cross-team planning meeting and talk about priorities 👍🏻.

wolmir:
Would it be possible to divide the color space in two complementary regions? One for each domain?

mattseddon:
> Would it be possible to divide the color space in two complementary regions? One for each domain?

There are potentially more than two domains for each revision. I think line types can handle up to 5 but at that point, the plot is going to be confusing no matter what we do.

dberenbaum:
Update on this: @pared is working on making it possible to determine from the JSON which plots are "flexible" so that it's possible to continue treating the traditional ones the same way and have separate rendering logic for the flexible ones. Then either he or the VS Code team can work on how to render those flexible plots in VS Code.

pared:
[...]

PR #2379, opened 2022-09-12
[WIP]: flexible plots support

pared:
Creating as draft, 

Current implementation details:
- currently we do will `rev` for complex plots on DVC side, and don't fill `rev` for simple plots - this was done as POC, and I think it cannot stay this way in final implementation. We cannot fill all `rev` on DVC side as vscode might fill `rev` with different value than DVC does (for example vscode fills with `main` string instead of exact commit SHA, which dvc does).

I am currently working on how to resolve this problem.
Related #1757


pared:
Seems like most work happen in #2403 . Closing

PR #2384, opened 2022-09-12
Update cell hint styles and context menu position

### LINKED ISSUE #2241: Review: table styles

julieg18:
* make cell hint text smaller
* add hint to first cell bullet
* take off cell indicator tooltips
* have context menu appear based off of cursor position

## Demo

https://user-images.githubusercontent.com/43496356/189775237-19885109-6001-43e7-bac5-f9cf16b292ef.mov

Fixes #2241


webview/src/experiments/components/table/CellRowActions.tsx
```diff
@@ -45,26 +40,16 @@ export const CellRowAction: React.FC<CellRowActionProps> = ({
...
         data-testid={testId}
       >
-        <Indicator
-          count={count}
-          tooltipContent={
-            count &&
-            `${count} ${pluralize('sub-row', count)} ${actionAppliedLabel}.`
-          }
-        >
+        <Indicator tooltipContent="" count={count}>
```
julieg18, line 48: Originally, each indicator has a tooltip but that caused some overlap:

image

I just took it off for now, but we could also just keep it as is or have a single tooltip that gives information on action and sub items.

webview/src/shared/components/contextMenu/ContextMenu.tsx
```diff
@@ -66,6 +66,8 @@ export const ContextMenu: React.FC<ContextMenuProps> = ({
     onShow={onShow}
     disabled={!content || disabled}
     appendTo={'parent'}
+    followCursor={'initial'}
+    offset={[0, 0]}
```
julieg18, line 70: By using `followCursor` and `offset`, we can make our context menu behave pretty similarly to vscode context menus 🤔 

image

Though can be taken off if we prefer the original :)

mattseddon: 👍🏻 

Can you take a look at the storybook to see if there is anything that can be done?

webview/src/experiments/components/table/Cell.tsx
```diff
@@ -63,21 +63,24 @@ export const FirstCell: React.FC<
       <div className={styles.innerCell}>
         <CellRowActions {...rowActionsProps} />
         <RowExpansionButton row={row} />
-        <span
-          className={styles.bullet}
-          style={{ color: bulletColor }}
-          {...clickAndEnterProps(toggleExperiment)}
+        <CellHintTooltip
+          tooltipContent={`${
+            bulletColor ? 'Remove from plots' : 'Add from plots'
```
mattseddon, line 68: [Q] Can we go with `Unplot` / `Plot`? Otherwise `Add from plots` should be `Add to plots`.

mattseddon: Related to the choice made for context menu options [here](https://github.com/iterative/vscode-dvc/pull/2388/files#diff-7f59061e48c6626575b99739398edf57257f7650eafb4af90f3a7792911fcf46R76).

webview/src/experiments/components/table/CellRowActions.tsx
```diff
@@ -93,7 +77,6 @@ export const CellRowActions: React.FC<CellRowActionsProps> = ({
       <CellRowAction
         showSubRowStates={showSubRowStates}
         subRowsAffected={stars}
-        actionAppliedLabel={'starred'}
         testId={'row-action-star'}
         tooltipContent={starred ? 'Remove Star' : 'Add Star'}
```
mattseddon, line 81: [C] This could be `Unstar` / `Star` as well

webview/src/experiments/components/table/CellHintTooltip.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Tooltip, {
+  CELL_TOOLTIP_DELAY
+} from '../../../shared/components/tooltip/Tooltip'
+
+export const CellHintTooltip: React.FC<{
+  tooltipContent: ReactNode
+  children: ReactElement
+}> = ({ tooltipContent, children }) => {
+  return (
+    <Tooltip
```
mattseddon, line 12: [Q] Do we want to move the indicator tooltips over to this?

mattseddon (approved):

webview/src/experiments/components/table/CellHintTooltip.tsx
julieg18: > [Q] Do we want to move the indicator tooltips over to this?

Not quite sure what you mean 🤔 Could you expand a bit more on the question?

webview/src/experiments/components/table/CellRowActions.tsx
sroy3: Maybe `tooltipContent` should be optional in `<Indicator />` and the markup should change whether the prop is passed or not.

sroy3 (approved):

webview/src/shared/components/contextMenu/ContextMenu.tsx
```diff
@@ -56,7 +56,7 @@ export const ContextMenu: React.FC<ContextMenuProps> = ({
   <Tooltip
     trigger={trigger}
     delay={[100, 200]}
-    placement={'right-start'}
+    placement={'bottom-start'}
```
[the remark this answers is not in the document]
julieg18: I believe so but `bottom-start` gives the most accurate result when you add `followCursor` as far as could tell when testing 🤔

webview/src/experiments/components/table/CellHintTooltip.tsx
mattseddon: We have two sizes of tooltip hints in the table. Do we want the tooltips for the indicators in the top left-hand corner to be this size?

https://user-images.githubusercontent.com/37993418/189995145-5b086ca8-73e7-41fc-a502-a73c4c5ee2f9.mov

julieg18: I don't have a preference 🤔 cc @maxagin

maxagin: Just to clarify.
We are talking about a tooltip text when the mouse moves over the element. Right?

Untitled-1

If yes. Is there a reason we use different sizes?

maxagin: The placement of the tooltip in relation to the cursor is a bit different from the context menu placement. Here are examples.

Screen Shot 2022-09-14 at 7 38 51 PM

maxagin: > We have two sizes of tooltip hints in the table. Do we want the tooltips for the indicators in the top left-hand corner to be this size?

I see now what @mattseddon talking about -font size. I would make all the tooltips the same as what we have on the table now if possible + adjust the placement.

PR #2391, opened 2022-09-13
Update selected row foreground

julieg18:
* updates selected row foreground, matching how VSCode trees work

## Demo

https://user-images.githubusercontent.com/43496356/189943322-88c885ff-5b3c-4cec-a0c1-81471e317a64.mov

Followup to [#2367 comment](https://github.com/iterative/vscode-dvc/pull/2367#discussion_r967442568)

### DIFF
--- a/webview/src/shared/variables.scss
+++ b/webview/src/shared/variables.scss
@@ -15,6 +15,7 @@ $row-bg-color: $bg-color;
 $header-fg-color: $fg-color;
 $row-bg-alt-color: var(--vscode-sideBar-background);
 $row-bg-selected-color: var(--vscode-list-activeSelectionBackground);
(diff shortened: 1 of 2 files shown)


webview/src/shared/variables.scss
```diff
@@ -15,6 +15,7 @@ $row-bg-color: $bg-color;
 $header-fg-color: $fg-color;
 $row-bg-alt-color: var(--vscode-sideBar-background);
 $row-bg-selected-color: var(--vscode-list-activeSelectionBackground);
+$row-fg-selected-color: var(--vscode-list-activeSelectionForeground);
```
julieg18, line 18: From what I can tell, this variable is always set in all themes so we don't need to a fallback like we did with the icon color.

sroy3 (approved):

mattseddon (approved):

Issue #1966, opened 2022-07-01
Story: Compare other branches and commits in plots and the experiments table


dberenbaum:
In the Plots and Experiments views, is it possible to compare the workspace or experiments to past commits or branches? Right now, it seems limited to experiments based on the current commit. For example, in plots it would be nice to be able to enter any git ref here:

<img width="759" alt="Screen Shot 2022-07-01 at 11 03 22 AM" src="https://user-images.githubusercontent.com/2308172/176920455-c97c683c-4e80-4ab5-8f7d-9a17debc52aa.png">

- [x] First step https://github.com/iterative/vscode-dvc/issues/2458
- [x] Next implement https://github.com/iterative/vscode-dvc/issues/1966#issuecomment-1179177339

DavidGOrtega:
Im preparing also an issue regarding to branches 😋

mattseddon:
Close to a duplicate of #456. I think we only need one of these issues.

dberenbaum:
I don't think we should be constrained by existing CLI flags here like `--all-branches` etc. VS Code might be able to provide way more flexible to pick and choose specific refs from the git tree within the UI. I think this type of functionality could be broadly useful not just in this feature, but in other features like removing unwanted experiments or data.

sroy3:
#2392 Displays the previous two commits. Next step could be have a "show more commits" at the bottom of the table. It could add two (we can easily fine tune that to the number we want) more commits until there are no more to show.

dberenbaum:
If we can make it sufficiently performant, can we default to a table like `dvc exp show -A` and collapse other branches and old commits?

daveraghav:
Is there a plan to show more than 2 previous commits to vsc-extension experiments table anytime soon?

shcheklein:
@daveraghav yes! just as a matter of getting feedback - how many commits you would prefer to see? do you need to see branches?

daveraghav:
Thanks for the update @shcheklein. Ideally the more the better. Either using 'Load More Commits' or something similar as mentioned @sroy3 or by making it configurable in extension settings. Also showing experiments from other branches would be great as multiple data scientists often work across parallel branches on the same project.

shcheklein:
> Also showing experiments from other branches would be great as multiple data scientists often work across parallel branches on the same project.

[...]

PR #2392, opened 2022-09-13
Multiple commits in the experiments table

sroy3:
Part of #1966 
Screen Shot 2022-09-13 at 2 23 18 PM
 Quite straight forward. Props to @mattseddon for thinking ahead, because this just works.


sroy3:
https://user-images.githubusercontent.com/3683420/190235297-4c0c98d2-1832-42df-af44-66dfe774622c.mov

mattseddon (approved):
Initially, I was very confused by the chevron placement of the second commit but it does seem like everything is working as expected:

Screen Shot 2022-09-15 at 4 35 13 pm (2)

I do think we now need to nest experiments underneath their respective commit in the experiments tree:

Screen Shot 2022-09-15 at 4 38 51 pm (2)

The top of the tree is flat and fairly confusing. We should follow up and make that change ASAP.

There is at least one function (`collectWorkspaceRunningCheckpoint`) which is (probably) checking previously commits when it doesn't need to be.

We should also add a small multi-commit test fixture :pray:

mattseddon:
Unable to apply an experiment from a previous commit: 

https://user-images.githubusercontent.com/37993418/190335647-3348ba64-3498-4b40-b70f-8844e5bb2f65.mov

This doesn't work for either the commit/short sha or the name. It looks like `exp apply` will only accept names from the current commit. That means we need to raise a bug in DVC and suppress the option from the tree/palette/table for previous commits (for the time being).

extension/src/cli/dvc/constants.ts
```diff
@@ -28,9 +28,11 @@ export enum SubCommand {
 }
```
julieg18, line 28: Looks like the row borders are now a bit buggy which I believe is due to [this line in our styles](https://github.com/iterative/vscode-dvc/blob/84a9883203a063265cad7aab8d56220634d8498c/webview/src/experiments/components/table/styles.module.scss#L239):

Screen Shot 2022-09-15 at 7 43 29 AM

julieg18: If I remember correctly, the CSS lines were added to make the last row on the table not have a border 🤔

julieg18 (approved):
Great work! Agreed with Matt comments. I was also confused about the placement of the commit rows since they're at the same level as the `main` branch.

wolmir (approved):

PR #2396, opened 2022-09-14
Move vega title truncation into the extension

mattseddon:
This PR moves all of the title truncation logic back into the extension where we already manipulate the templates. The reason for doing this is to try and keep the logic centralised (avoid later surprises).

Doing this work made me realise (again) how much of a tangled mess the current test fixtures are. I will block out some time to improve things this week.


webview/src/plots/components/util.ts
```diff
@@ -47,76 +44,3 @@ export const getNbItemsPerRow = (size: PlotSize) => {
 
 export const shouldUseVirtualizedGrid = (nbItems: number, size: PlotSize) =>
   nbItems > MaxItemsBeforeVirtualization[size]
-
-const truncateTitlePart = (title: string, size: number) =>
```
mattseddon, line 51: [F] Moved this to `extension/src/plots/vega/util.ts` and took the tests from `App.test.tsx` and put them into the associated test file.

julieg18 (approved):
Looks good! Though Storybook appears to be seeing a difference that I can't see... Probably unrelated to the pr 🤔

sroy3 (approved):

PR #2410, opened 2022-09-16
Exclude queued experiments from select experiments for plots quick pick

mattseddon:
### Demo

https://user-images.githubusercontent.com/37993418/190561301-68ac36fe-ef4f-43f1-a696-e3d84ac7c314.mov


julieg18 (approved):
Looks good!

sroy3 (approved):

extension/src/experiments/model/index.ts
```diff
@@ -324,13 +324,15 @@ export class ExperimentsModel extends ModelWithPersistence {
...
-    })
+    return this.getExperiments()
+      .filter(({ queued }) => !queued)
+      .map(experiment => {
+        const checkpoints = this.getCheckpointsWithType(experiment.id)
+        if (!definedAndNonEmpty(checkpoints)) {
+          return experiment
+        }
+        return { ...experiment, checkpoints }
+      })
```
wolmir, line 335: What about the following, to get rid of the filter and map?
```
  public getExperimentsWithCheckpoints(): ExperimentWithCheckpoints[] {
    const experimentsWithCheckpoints: ExperimentWithCheckpoints[] = []
    const experiments = this.getExperiments()

    for (const experiment of experiments) {
      const { queued, id } = experiment
      const checkpoints = !queued && this.getCheckpointsWithType(id)

      if (definedAndNonEmpty(checkpoints)) {
        experimentsWithCheckpoints.push({
          ...experiment,
          checkpoints
        })
      }
    }

    return experimentsWithCheckpoints
  }
```

And in `array.ts`:
```
export const definedAndNonEmpty = (
  maybeArray: readonly unknown[] | undefined | false
): maybeArray is unknown[] => {
  return !!(maybeArray && maybeArray.length > 0)
}
```

mattseddon: I think 
```
      const checkpoints = !queued && this.getCheckpointsWithType(id)

      if (definedAndNonEmpty(checkpoints)) {
```
needs to be 

```      
      if (queued){
        continue
      }
      const checkpoints = this.getCheckpointsWithType(id)

      if (!definedAndNonEmpty(checkpoints)) {
        experimentsWithCheckpoints.push(experiment)
        continue
      }

      ...
```

mattseddon: I'll update to simplify 👍🏻.

Issue #2390, opened 2022-09-13
Broken yaml hierarchy in the default output


tibor-mach:
cc @shcheklein 
Continuing the discussion on [Discord](https://discord.com/channels/485586884165107732/1018917508746588192) 

Hierarchy of parameters seems to be split into multiple columns in the VS code extension.

A simple example where this can be reproduced:
https://github.com/tibor-mach/vscodedvc-example

daavoo:
Was about to report similar.

This is the simplest structure I have found that triggers the bug:

```yaml
one:
  two:
    three: 
      works: true
      four: 
        breaks: true
```

tibor-mach:
@daavoo  In my example it breaks even earlier though (example.steps should be under steps)
![image](https://user-images.githubusercontent.com/56956998/189886110-7831f38c-e93f-4d68-91fc-489a736ab593.png)

tibor-mach:
> Adding more nested levels cause the top-level row to be expended (one.two.three, etc.)

@daavoo Ah, ok, that could be it ;-)

mattseddon:
It's related to this: https://github.com/iterative/vscode-dvc/issues/950

tibor-mach:
Maybe a good solution would be to add a configuration option to the extension so that users could limit/not limit nesting as they wish?

mattseddon:
👍🏻

I can't remember the reason behind the approach right now. All that I can remember is that there was one. I'll check with the team in the morning and get back to you.

shcheklein:
@daavoo @tibor-mach folks, to clarify. We do limit number of levels we go down. Do you consider this limitation itself as a problem (and a solution would be to configure it for example), or the problem is that split itself into columns is not properly handled (we have `example.steps` and `example` -> `steps` simultaneously in the table). Second is potentially a bug with the current implementation. I need to check this.

julieg18:
I don't think the logic that limits the number of head layers has a bug inside of it since Studio gives you the same column layout:

<img width="1376" alt="image" src="https://user-images.githubusercontent.com/43496356/190691948-1230afc3-23ec-47b9-a62e-dd89f3489504.png">

<img width="1178" alt="image" src="https://user-images.githubusercontent.com/43496356/190692025-2ec8dfc7-fb1a-45c0-8fa4-6b837d598f16.png">

Since there's no bug to fix, the only options that come to mind on what to do next are to add some kind of configuration option or leave as is 🤔

shcheklein:
[...]

PR #2436, opened 2022-09-20
Add config for updating max amount of table head layers

julieg18:
* adds a config option for adjusting the max amount of table head rows

## Demo

~https://user-images.githubusercontent.com/43496356/191370315-9c752b6c-72a7-4610-ae98-89d3b218ddc7.mov~

https://user-images.githubusercontent.com/43496356/191857765-db49d952-5189-465c-b61c-39598491af4e.mov

Part of #2390


extension/src/experiments/columns/model.ts
```diff
@@ -63,6 +73,14 @@ export class ColumnsModel extends PathSelectionModel<Column> {
     )
   }
 
+  public setColumnHeadLayersMax(max: number) {
+    this.columnHeadLayersMaxLength = max
+    this.persist(
```
mattseddon, line 78: Rather than store this in the memento you might want to set a config option. Take a look at `extension/src/vscode/config.ts` as an entry point for the extension's existing config options.

extension/package.nls.json
```diff
@@ -84,6 +84,8 @@
   "config.doNotShowUnableToFilter.title": "Do not warn before disabling auto-apply filters to experiment selection.",
   "config.dvcPath.description": "Path or shell command to the DVC binary. Required unless Microsoft's Python extension is installed and the `dvc` package found in its environment.",
   "config.dvcPath.title": "DVC Path (or shell command)",
+  "config.expTableHeadMaxLayers.title": "Maximum amount of Experiment table head rows",
```
julieg18, line 87: Is there a better way to phrase the title/description?

extension/src/experiments/columns/model.ts
```diff
@@ -22,6 +25,9 @@ export class ColumnsModel extends PathSelectionModel<Column> {
       PersistenceKey.METRICS_AND_PARAMS_COLUMN_WIDTHS,
       {}
     )
+    this.columnHeadLayersMaxLength =
```
julieg18, line 28: Decided to store the config value in the Model vs in a lower level function so that the config is only collected once (vs every time a column is created for example), but this means that the variable is passed over several different functions to get to `limitAncestorDepth`.

extension/src/experiments/columns/collect/util.ts
```diff
@@ -15,7 +15,7 @@ export const limitAncestorDepth = (
     concatenated layer itself counts as one; because of this, we must subtract 3
     from what we want the final layer count to be.
   */
-  const convertedLimit = limit - 3
+  const convertedLimit = (limit >= 3 ? limit : 3) - 3
```
julieg18, line 18: Currently, this function only works correctly if the limit is at least 3. There's probably a way to rewrite the logic to allow for 1 or 2 layers though. Should this be done in a followup?

extension/package.json
```diff
@@ -554,6 +554,12 @@
           "type": "string",
           "default": null
         },
+        "dvc.expTableHeadMaxLayers": {
```
julieg18, line 557: If you update the config, you need to restart the window to see the change to the table. Is it supposed to change once you update the config without restarting?

extension/src/experiments/columns/model.ts
mattseddon: `getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS)` is not an expensive call. I don't think we need to store the value. Making the call also means that we don't need to add a listener.

See https://github.com/iterative/vscode-dvc/blob/998d3f3496ac451c7c32f1a59be4477286b0f68e/extension/src/test/suite/util.ts#L47-L48 for an example of a configuration change event listener.

(also fixes [this](https://github.com/iterative/vscode-dvc/pull/2436/files#r977045140)).

maxagin:
1. Can we update the max amount of table head layers without closing and reopening the extension?
2. Again. It feels very much like hiding table columns, which are controlled in the sidebar. Do you see this suggestion as an option in the future?

extension/src/experiments/columns/model.ts
mattseddon: Will still need a listener to fire the tree/table updated event emitter(s) whenever the config changes. See https://github.com/iterative/vscode-dvc/blob/bb10bcce4cd85e30e3b53aa0825e9b3273625d98/extension/src/experiments/workspace.ts#L24

extension/package.json
julieg18: Table updates on config change now!

extension/src/experiments/columns/model.ts
julieg18: Thanks for the info!

extension/src/experiments/index.ts
```diff
@@ -123,6 +130,14 @@ export class Experiments extends BaseRepository<TableData> {
       })
     )
 
+    this.dispose.track(
+      workspace.onDidChangeConfiguration((event: ConfigurationChangeEvent) => {
+        if (event.affectsConfiguration(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS)) {
+          this.cliData.update()
```
julieg18, line 136: Is there a better way to update the table when the configuration is updated 🤔

julieg18 (commented):
> Can we update the max amount of table head layers without closing and reopening the extension?

Fixed now! The table head layers will now update once you change the config without needing to restart :)

extension/src/experiments/index.ts
mattseddon: No, I think this is the only way as we don't keep a copy of the exp show data.

extension/src/experiments/columns/collect/util.ts
```diff
@@ -1,21 +1,23 @@
...
   /*
     The depth is only limited for the middle of the path array.
     The first and final layer are excluded, and the
     concatenated layer itself counts as one; because of this, we must subtract 3
     from what we want the final layer count to be.
   */
-  const convertedLimit = limit - 3
+  const collectedLimit =
+    Number(getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS)) ||
+    INITIAL_TABLE_HEAD_MAX_LAYERS
```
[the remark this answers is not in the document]
mattseddon: The tests should 

1. Reflect the default if no value is provided.
2. Test that changing the config value changes the header.
3. Confirm that values of 0,1,2 do not break anything.


See https://github.com/iterative/vscode-dvc/blob/f5748d1e6248f487716f3fa691cce1641a28673d/extension/src/setup.test.ts#L296 for example of mocking config value.

When you follow up to add ability to concatenate to headers to 1 (i.e just leafs as per experiment table) 0 should be the setting for unlimited.

mattseddon (approved):
👍🏻

sroy3 (approved):

extension/src/experiments/columns/collect/util.ts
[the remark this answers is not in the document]
julieg18: > The tests should
> 1. Reflect the default if no value is provided.
> 2. Test that changing the config value changes the header.
> 3. Confirm that values of 0,1,2 do not break anything.

Completely forgot about updating the tests! Will update in a followup!

PR #2439, opened 2022-09-21
Update `view/title` commands to account for VS Code change

mattseddon:
This PR accounts for this VS Code change: https://github.com/microsoft/vscode/issues/42903.

A tree's `view/title` commands will now be called with the selected item for the context object. This means that if the user is working with a multi-project workspace they will be able to bypass project selection in a tree's `view/title` command by selecting the project's root before clicking on the title's command. We previously had one or two of these commands inline for the root but we can now drop that behaviour.

### Demo

default behaviour (quick pick shown):

https://user-images.githubusercontent.com/37993418/191484823-95346a44-2c86-4300-9f2b-3247e0517cc8.mov

with root selected (project auto-selected for command):

https://user-images.githubusercontent.com/37993418/191483895-e6ae5818-4597-4c50-bcea-646c96e66996.mov


julieg18 (approved):
Looks good!

Probably an issue on my end, but when I select a workspace, then open one of the views, I still get a quickpick pop-up: 


https://user-images.githubusercontent.com/43496356/191499176-ef3e56e1-1a20-463b-ba94-01214dc47723.mov

sroy3 (approved):

mattseddon:
> Looks good!
> 
> Probably an issue on my end, but when I select a workspace, then open one of the views, I still get a quickpick pop-up:
> 
>  Screen.Recording.2022-09-21.at.7.02.06.AM.mov

🤦🏻 this change is only available in Insiders until the next VS Code release... probably the most important thing to mention in the descriptions, sorry!

PR #2445, opened 2022-09-22
Send initial YAML and JSON files right after language client starts (2/3)

wolmir:
The standard documents manager provided by the VSCode SDK is only aware of text documents that the user has already opened in the editor area.

This is annoying because we would expect a feature like Go to Definitions to be immediately aware of the params.yaml file for example.

To solve this problem this PR will create a custom request from the Client to the Server with an initial batch of interesting documents. For now they are any yaml or json files that are not inside a hidden folder.

### DIFF
--- a/extension/src/lspClient/languageClient.ts
+++ b/extension/src/lspClient/languageClient.ts
@@ -1,4 +1,4 @@
-import { workspace } from 'vscode'
+import { Uri, workspace } from 'vscode'
 import {
   LanguageClient,
   LanguageClientOptions,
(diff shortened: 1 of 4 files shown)


mattseddon (approved):

mattseddon:
Suggestion: When we can open a file, i.e it is a folder we could use `@command:revealInExplorer` to show the folder in the explorer tree.

Some example usage in`vscode`:

https://github.com/microsoft/vscode/blob/6d62e3b80b950cd38294bf09083c93334aa5ec15/extensions/git/src/commands.ts#L2868

https://github.com/microsoft/vscode/blob/6d62e3b80b950cd38294bf09083c93334aa5ec15/extensions/markdown-language-features/src/util/openDocumentLink.ts#L25

https://github.com/microsoft/vscode/blob/6d62e3b80b950cd38294bf09083c93334aa5ec15/src/vs/workbench/contrib/terminal/browser/links/terminalLinkOpeners.ts#L116

julieg18 (approved):

PR #2471, opened 2022-09-26
Add tests for column depth configuration

julieg18:
Followup to [#2436 comment](https://github.com/iterative/vscode-dvc/pull/2436#discussion_r978172645)

### DIFF
--- a/extension/src/experiments/columns/collect/index.test.ts
+++ b/extension/src/experiments/columns/collect/index.test.ts
@@ -8,8 +8,12 @@ import columnsFixture from '../../../test/fixtures/expShow/columns'
 import workspaceChangesFixture from '../../../test/fixtures/expShow/workspaceChanges'
 import uncommittedDepsFixture from '../../../test/fixtures/expShow/uncommittedDeps'
 import { ExperimentsOutput } from '../../../cli/dvc/reader'
+import { getConfigValue } from '../../../vscode/config'
 
 jest.mock('../../../vscode/config')
+
+const mockedGetConfigValue = jest.mocked(getConfigValue)
+mockedGetConfigValue.mockImplementation(() => 5)
 
 describe('collectColumns', () => {
   it('should return a value equal to the columns fixture when given the output fixture', () => {
--- a/extension/src/experiments/columns/collect/util.ts
+++ b/extension/src/experiments/columns/collect/util.ts
@@ -16,8 +16,9 @@ export const limitAncestorDepth = (
     concatenated layer itself counts as one; because of this, we must subtract 3
     from what we want the final layer count to be.
   */
-  const collectedLimit =
-    Number(getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS)) || limit
+  const collectedLimit = Number(
+    getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS, limit)
+  )
   const convertedLimit = (collectedLimit >= 3 ? collectedLimit : 3) - 3
   if (rest.length <= convertedLimit) {
     return ancestors
--- a/extension/src/experiments/columns/collect/util.ts
+++ b/extension/src/experiments/columns/collect/util.ts
@@ -1,11 +1,14 @@
 import { Column, ColumnType } from '../../webview/contract'
 import { Value } from '../../../cli/dvc/reader'
 import { ConfigKey, getConfigValue } from '../../../vscode/config'
-import { INITIAL_TABLE_HEAD_MAX_LAYERS } from '../consts'
 
 export type ColumnAccumulator = Record<string, Column>
 
-export const limitAncestorDepth = (ancestors: string[], sep: string) => {
+export const limitAncestorDepth = (
+  ancestors: string[],
+  sep: string,
+  limit = 5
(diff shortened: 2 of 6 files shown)


extension/src/experiments/columns/collect/util.ts
```diff
@@ -14,9 +17,7 @@ export const limitAncestorDepth = (ancestors: string[], sep: string) => {
     from what we want the final layer count to be.
   */
   const collectedLimit =
-    Number(getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS)) ||
-    INITIAL_TABLE_HEAD_MAX_LAYERS
-
+    Number(getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS)) || limit
```
julieg18, line 20: `getConfigValue` always defaults to 5 but since the function could return undefined according to typescript, I couldn't take the `|| limit` part off without receiving a TS error. 

Trying to adjust `getConfigValue` also just lead to other different kinds of errors 🤔

sroy3 (approved):

extension/src/experiments/columns/collect/util.ts
wolmir: @julieg18 In my version the getConfigValue doesn't return 5 by default. I couldn't think of a way to remove the || limit part either.

But this worked for me:

In config.ts:
```
export const getConfigValue = <T = string, D = string>(
  key: ConfigKey,
  defaultValue?: D | T
): T =>
  workspace.getConfiguration().get(key, defaultValue ?? '') as unknown as T
```

And in util.ts:
```
  const collectedLimit = Number(
    getConfigValue(
      ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS,
      limit
    )
  )
```

julieg18: > @julieg18 In my version the getConfigValue doesn't return 5 by default.

When you're running the tests (it should be returning `undefined` in the tests since it's just a mock)? Or when you're in production/development?

wolmir: No, sorry, I meant in general based only on the compiler results. I didn't run the tests specifically

julieg18: > But this worked for me:

Works great, thanks!

extension/src/experiments/columns/collect/index.test.ts
```diff
@@ -8,9 +8,13 @@ import columnsFixture from '../../../test/fixtures/expShow/columns'
 import workspaceChangesFixture from '../../../test/fixtures/expShow/workspaceChanges'
 import uncommittedDepsFixture from '../../../test/fixtures/expShow/uncommittedDeps'
 import { ExperimentsOutput } from '../../../cli/dvc/reader'
+import { getConfigValue } from '../../../vscode/config'
 
 jest.mock('../../../vscode/config')
 
+const mockedGetConfigValue = jest.mocked(getConfigValue)
+mockedGetConfigValue.mockImplementation(() => 5)
```
mattseddon, line 16: [I] The usual pattern is to use `beforeEach` with `resetAllMocks` to keep tests independent

```
beforeEach(() => {
  jest.resetAllMocks()
})
```
I thin you could use `mockReturnValueOnce` instead of `mockImplementation` as well. Final code would probably be 

```
beforeEach(() => {
  jest.resetAllMocks()
  mockedGetConfigValue.mockReturnValueOnce(5)
})
```

for this particular file as we are not varying the option.

extension/src/vscode/config.ts
```diff
@@ -10,8 +10,11 @@ export enum ConfigKey {
   PYTHON_PATH = 'dvc.pythonPath'
 }
 
-export const getConfigValue = <T = string>(key: ConfigKey): T =>
-  workspace.getConfiguration().get(key, '') as unknown as T
+export const getConfigValue = <T = string, D = string>(
+  key: ConfigKey,
+  defaultValue?: D | T
```
mattseddon, line 15: does this translates to `defaultValue?: string` ... 😕

mattseddon: I think this could be 
```
export const getConfigValue = <T = string>(
  key: ConfigKey,
  defaultValue: T | undefined
): T =>
  workspace.getConfiguration().get(key, defaultValue ?? '') as unknown as T
```

extension/src/experiments/columns/model.test.ts
```diff
@@ -9,30 +9,62 @@ import outputFixture from '../../test/fixtures/expShow/output'
...
 import {
   dataTypesOutput,
   columns as dataTypesColumns
 } from '../../test/fixtures/expShow/dataTypes'
+import { getConfigValue } from '../../vscode/config'
 
 jest.mock('../../vscode/config')
 
+const mockedGetConfigValue = jest.mocked(getConfigValue)
+mockedGetConfigValue.mockImplementation(() => 5)
```
mattseddon, line 25: [I] Can also use a `beforeEach` this time I would but the `mockReturnValueOnce` calls inside of the tests as the value is being varied throughout the file

mattseddon (commented):
This was meant to be an approval

PR #2481, opened 2022-09-27
Cleanup after drag and drop

### LINKED ISSUE #2423: Images table DnD: header stuck in the "dnd" state
https://user-images.githubusercontent.com/3659196/190881374-b774fe63-5ec7-4c37-9b6f-6a9e6dc38dd3.mov

I would also expect for ESC to cancel the operation and return everything into its initial state

sroy3:
Closes #2423 

https://user-images.githubusercontent.com/3683420/192616319-6fbf955d-df67-47f9-bb9e-72575b0c12ae.mov

Since `onDragEnd` is being called only half the time, there is now a failsafe on mouse move. If the mouse moves with no mouse button pressed and the drag and drop state is not clean, it'll call the cleanup function to reset the state.

### DIFF
--- a/webview/src/shared/components/dragDrop/DragDropContainer.tsx
+++ b/webview/src/shared/components/dragDrop/DragDropContainer.tsx
@@ -109,6 +109,7 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
     (state: PlotsState) => state.dragAndDrop
   )
   const draggedOverIdTimeout = useRef<number>(0)
+  const pickedUp = useRef<boolean>(false)
   const dispatch = useDispatch()
 
   const cleanup = useCallback(() => {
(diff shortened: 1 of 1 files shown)


webview/src/shared/components/dragDrop/DragDropContainer.tsx
```diff
@@ -189,6 +201,7 @@ export const DragDropContainer: React.FC<DragDropContainerProps> = ({
     dispatch(changeRef(undefined))
 
     onDrop?.(oldDraggedId, draggedRef?.group || '', group, droppedIndex)
+    cleanup()
```
sroy3, line 204: I don't think this should be there. I need to look into why I added this and if it was intentional.

mattseddon (approved):

PR #2482, opened 2022-09-27
Allow experiment table column depth below 3

### LINKED ISSUE #2390: Broken yaml hierarchy in the default output

julieg18:
## To Do

- [ ] Add Tests

Fixes #2390

### DIFF
--- a/extension/src/experiments/columns/collect/deps.ts
+++ b/extension/src/experiments/columns/collect/deps.ts
@@ -37,7 +37,7 @@ export const collectDeps = (acc: ColumnAccumulator, data: ExperimentFields) => {
       path,
       buildDepPath(...limitedDepthAncestors),
       [ColumnType.DEPS, file],
-      label,
+      limitedDepthAncestors.length === 0 ? file : label,
       shortenForLabel(hash)
     )
   }
--- a/extension/src/experiments/columns/collect/metricsAndParams.ts
+++ b/extension/src/experiments/columns/collect/metricsAndParams.ts
@@ -38,10 +38,16 @@ const collectMetricOrParam = (
 
   mergeValueColumn(
     acc,
-    buildMetricOrParamPath(type, ...limitedDepthAncestors, label),
+    buildMetricOrParamPath(
+      type,
+      ...(limitedDepthAncestors || pathArray),
+      label
+    ),
     buildMetricOrParamPath(type, ...limitedDepthAncestors),
     [type, ...pathArray, label],
-    label,
+    limitedDepthAncestors.length === 0
+      ? buildMetricOrParamPath(type, ...pathArray, label).slice(type.length + 1)
+      : label,
     value
   )
 }
--- a/extension/src/experiments/columns/collect/util.ts
+++ b/extension/src/experiments/columns/collect/util.ts
@@ -10,15 +10,24 @@ export const limitAncestorDepth = (
   limit = 5
 ) => {
   const [path, ...rest] = ancestors
-  /*
-    The depth is only limited for the middle of the path array.
+  const collectedLimit = Number(
+    getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS, limit)
+  )
+
+  switch (collectedLimit) {
+    case 0:
+      return ancestors
+    case 1:
+      return []
+    case 2:
+      return [ancestors.join(sep)]
+  }
+
+  /* The depth is only limited for the middle of the path array.
     The first and final layer are excluded, and the
     concatenated layer itself counts as one; because of this, we must subtract 3
     from what we want the final layer count to be.
   */
-  const collectedLimit = Number(
-    getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS, limit)
-  )
   const convertedLimit = (collectedLimit >= 3 ? collectedLimit : 3) - 3
   if (rest.length <= convertedLimit) {
     return ancestors


extension/src/experiments/columns/collect/util.ts
```diff
@@ -10,16 +11,29 @@ export const limitAncestorDepth = (
...
+  )
+
+  switch (collectedLimit) {
+    case 0:
+      return ancestors
+    case 1:
+      return []
+    case 2:
+      return [
+        buildMetricOrParamPath(ColumnType.PARAMS, ...ancestors).slice(
```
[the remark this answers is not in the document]
sroy3: I'd bring everything below in a default case here to make it obvious that 0, 1 and 2 aren't the only options

maxagin:
Q: why `:` and not `\`. Example `C:\home\sally\statusReport`
Thanks

extension/src/experiments/columns/collect/util.ts
```diff
@@ -10,16 +11,29 @@ export const limitAncestorDepth = (
...
+  /* The depth is only limited for the middle of the path array.
     The first and final layer are excluded, and the
     concatenated layer itself counts as one; because of this, we must subtract 3
     from what we want the final layer count to be.
   */
-  const collectedLimit = Number(
-    getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_LAYERS, limit)
-  )
-  const convertedLimit = (collectedLimit >= 3 ? collectedLimit : 3) - 3
+  const convertedLimit = (collectedLimit >= 3 ? collectedLimit : 5) - 3
```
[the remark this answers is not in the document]
sroy3: I'd expect it to be infinite (even if it's mentioned 0 in infinite)

sroy3 (approved):
Like @mattseddon mentioned yesterday, I feel like column names should be truncated from the left instead. Otherwise it kind of looks like the same column is duplicated

Screen Shot 2022-09-28 at 1 33 53 PM

extension/src/experiments/columns/collect/deps.ts
```diff
@@ -37,7 +37,7 @@ export const collectDeps = (acc: ColumnAccumulator, data: ExperimentFields) => {
       path,
       buildDepPath(...limitedDepthAncestors),
       [ColumnType.DEPS, file],
-      label,
+      limitedDepthAncestors.length === 0 ? file : label,
```
mattseddon, line 40: [Q] How does this work for 1/2? I don't understand what is happening here.

julieg18: When it comes a config of `2`, all of the ancestors are joined in `limitedDepthAncestors` and the value column label stays as is. 
But when it comes to a config of `1`, there are no `limitedDepthAncestors` since there is only one cell so `limitAncestorDepth` returns an empty array and the value column `label` needs to hold all ancestors plus the label aka `file`. 

Hopefully, that answers your question! Maybe there is a better way to go about this though 🤔

julieg18:
> Q: why `:` and not `\`. Example `C:\home\sally\statusReport` 
> Thanks

@maxagin, Apologies, I'm not sure what you're referring to. Could you give me some more context?

extension/src/experiments/columns/collect/util.ts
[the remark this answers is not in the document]
julieg18: Sounds good! Set negative numbers to infinite.

extension/src/experiments/columns/collect/util.ts
```diff
@@ -10,22 +11,35 @@ export const limitAncestorDepth = (
...
-  const convertedLimit = (collectedLimit >= 3 ? collectedLimit : 3) - 3
-  if (rest.length <= convertedLimit) {
-    return ancestors
+
+  switch (collectedLimit) {
+    case 1:
+      return []
+    case 2:
+      return [
+        buildMetricOrParamPath(ColumnType.PARAMS, ...ancestors).slice(
```
mattseddon, line 23: [Q] Should `limitAncestorDepth` function handle all `ColumnTypes`? If I pass a metric in here with the limit set to 2 I'll get a Param back out?

extension/src/experiments/columns/collect/deps.ts
mattseddon: I would pull this inline value out, give it an explanatory name and probably even access the file config value instead of doing this here....

After reading further:

All of the `limitedDepthAncestors.length === 0 ?` ternary logic should live together in the same place / function. Can it be consolidated somehow? It feels like the `limitAncestorDepth` function should return both the `label` and `limitedDepthAncestors` if possible.  That would mean that all of this logic is encapsulated and we don't have to repeat the transformations in a few places.

extension/src/experiments/columns/collect/metricsAndParams.ts
```diff
@@ -38,10 +38,16 @@ const collectMetricOrParam = (
...
+      type,
+      ...(limitedDepthAncestors || pathArray),
+      label
+    ),
     buildMetricOrParamPath(type, ...limitedDepthAncestors),
     [type, ...pathArray, label],
-    label,
+    limitedDepthAncestors.length === 0
+      ? buildMetricOrParamPath(type, ...pathArray, label).slice(type.length + 1)
+      : label,
```
mattseddon, line 50: [C] This call to `mergeValueColumn` is way too verbose for my tiny pea brain.

julieg18: Moved label creation to `limitAncestorDepth`!

extension/src/experiments/columns/collect/util.ts
julieg18: Created a function for creating the path that takes in separator arguments instead of using `buildMetricOrParamPath`.

extension/src/experiments/columns/collect/deps.ts
julieg18: Makes sense, thanks! Moved label creation to limitAncestorDepth.

julieg18:
Just noticed that there's an issue with single file columns on `2` layers:

Screen Shot 2022-09-29 at 2 52 15 PM

Will debug and open a second pr!

julieg18:
Ok, I think I fixed all the bugs that were spotted (and did my best to improve the code)! Are we good to merge this?

sroy3:
> Ok, I think I fixed all the bugs that were spotted (and did my best to improve the code)! Are we good to merge this?

LGTM!

PR #2452, opened 2022-09-23
Add flexible plots legends to plots tree

mattseddon:
# 2/2 `main` <- #2403 <- this

Related to #1757

This PR adds shape and stroke dash information into the plots tree.

### Screenshots

Screen Shot 2022-09-23 at 4 07 41 pm

image

**Note:** The hacky way that we add shape + detail to the plot template causes the on-hover behaviour to break. I will change the approach in the next PR.

image

PR #2486, opened 2022-09-28
Remove erroneous shape from vertical on hover line when shape dimension is added

mattseddon:
Follow up from #2452 

I tried a lot of things to make this a general solution. 

1. Updating the encoding in the correct place makes it impossible to turn the legend back on for zoomed-in plots. 
2. Regex-based solutions are risky as they could update the spec in unexpected ways.
3. Considered sending two versions of the template along with the details on how to fill it to the webview but I believe that is overkill because we are fixing an edge case.

I ended up going with a specific solution to patch the template only for shapes. I think this is good enough to get us started and to the next iteration from the DVC team.

### Screenshot

Screen Shot 2022-09-28 at 7 01 26 pm

If users run into this issue then we can revisit but I think this is good enough for this iteration. I do not envisage many people needing/seeing shapes.


mattseddon:
Note: Should be able to test this with `2.28.0` of DVC.

extension/src/plots/vega/util.ts
```diff
@@ -294,23 +307,30 @@ export const extendVegaSpec = (
...
+          disable: false,
+          symbolFillColor: 'grey'
+        }
+      },
       strokeDash: {
         legend: {
-          disable: false
+          disable: false,
+          symbolFillColor: 'transparent',
+          symbolStrokeColor: 'grey'
```
mattseddon, line 333: [F] Moved these as per the ask in #2452. Thought that the approach that I was going to take was going to be significantly different to this one.

mattseddon:
Closes https://github.com/iterative/vscode-dvc/issues/1757

PR #2487, opened 2022-09-28
Do not walk unnecessary keys in truncate titles

mattseddon:
Small optimisation that I noticed whilst working on #2486.


extension/src/plots/vega/util.ts
```diff
@@ -255,6 +255,11 @@ export const truncateTitles = (
     const specCopy: Record<string, unknown> = {}
 
     for (const [key, value] of Object.entries(spec)) {
+      if (['data', 'color', 'strokeDash', 'shape', 'detail'].includes(key)) {
```
mattseddon, line 258: [F] There should be not titles in any of these keys.

julieg18 (approved):
Looks good to me!

sroy3 (approved):

PR #2488, opened 2022-09-28
Bump min tested version of DVC to 2.28.0

mattseddon:


julieg18 (approved):
Looks good!

sroy3 (approved):

mattseddon:
This failed a lot of times because there were experiments on a previous commit that `plots diff` was trying to find (not possible because they were never pushed to the remote). I will update the e2e tests to remove experiments from previous commits.

Issue #2450, opened 2022-09-23
Images table: fix titles UX (on codespaces and beyond)


shcheklein:
Here is the behavior:

https://user-images.githubusercontent.com/3659196/191877366-a500671c-25aa-4275-ad65-0873d33a06d6.mov

- Button disappears if you try to click it
- No tooltip for the title that is shortened
- Button appears behind the "pinned" item delimiter
- If we have multiple images - does it make sense to show the full (longer) title?

PR #2490, opened 2022-09-28
Add tooltip to comparison table row path

sroy3:
Part of #2450 

https://user-images.githubusercontent.com/3683420/192807522-9cbb4d78-5533-46ec-b289-c0e82bfa67c1.mov

### DIFF
--- a/webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
+++ b/webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
@@ -11,6 +11,7 @@ import { ChevronDown, ChevronRight } from '../../../shared/components/icons'
 import { PlotsState } from '../../store'
 import { CopyButton } from '../../../shared/components/copyButton/CopyButton'
 import { isSelecting } from '../../../util/strings'
+import Tooltip from '../../../shared/components/tooltip/Tooltip'
 
 export interface ComparisonTableRowProps {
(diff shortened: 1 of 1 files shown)


mattseddon (approved):

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -44,7 +45,13 @@ export const ComparisonTableRow: React.FC<ComparisonTableRowProps> = ({
           <div className={styles.rowPath}>
             <button className={styles.rowToggler} onClick={toggleIsShownState}>
               <Icon icon={isShown ? ChevronDown : ChevronRight} />
-              <span className={styles.pathText}>{path}</span>
+              <Tooltip
+                content={path}
+                placement="bottom-start"
+                delay={[1000, 0]}
```
julieg18, line 51: Do we want this delay in some kind of variable?

julieg18 (approved):
Looks good!

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
sroy3: You're right. I lifted it from the experiments table headers and thought the same but then forgot about it. I'll do a follow up for this.

PR #2493, opened 2022-09-28
Stabilize e2e tests

mattseddon:
This PR stabilises our e2e test suite.

Note: We may be kicking the can down the road as insiders will become the stable version shortly.

### DIFF
--- a/extension/src/test/e2e/extension.test.ts
+++ b/extension/src/test/e2e/extension.test.ts
@@ -17,6 +17,9 @@ suite('DVC Extension For Visual Studio Code', () => {
   before('should finish loading the extension', async function () {
     this.timeout(240000)
     await waitForViewContainerToLoad()
+    const workbench = await browser.getWorkbench()
+    await workbench.executeCommand('DVC: Garbage Collect Experiments')
+    await browser.keys('Enter')
     return dismissAllNotifications()
   })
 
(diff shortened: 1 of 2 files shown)


extension/src/test/e2e/wdio.conf.ts
```diff
@@ -33,7 +33,7 @@ export const config: Options.Testrunner = {
   capabilities: [
     {
       browserName: 'vscode',
-      browserVersion: 'insiders',
+      browserVersion: 'stable',
```
mattseddon, line 36: Something has happened in insiders which means you cannot focus the plots webview after the experiments webview has been opened. I am still looking into it but this should get everything moving again.

mattseddon: The problem is only for the test suite and `webdriverio`, there is no loss of functionality with insiders.

extension/src/test/e2e/extension.test.ts
```diff
@@ -17,6 +17,9 @@ suite('DVC Extension For Visual Studio Code', () => {
   before('should finish loading the extension', async function () {
     this.timeout(240000)
     await waitForViewContainerToLoad()
+    const workbench = await browser.getWorkbench()
+    await workbench.executeCommand('DVC: Garbage Collect Experiments')
+    await browser.keys('Enter')
```
mattseddon, line 22: [F] This removes experiments that are found against previous commits. It is very likely that they will be missing from the remote so `plots diff` will fail to load the data.

shcheklein (approved):

PR #2504, opened 2022-09-29
LSP: New settings option to Disable server and Better Goto Definition for property and file paths (1/2)

wolmir:
This PR will allow the Goto Definition for feature for properties inside parameter files, and also for files referenced in the dvc.yaml.

It will also introduce an extension setting to disable the language server when needed.

Demo:

https://user-images.githubusercontent.com/1231848/193884449-272037d5-6aa5-4f6c-82bc-61f29cd643ad.mov


mattseddon:
Needs #2530 to be ready before merging.

languageServer/src/languageHelpers/baseLanguageHelper.ts
```diff
@@ -0,0 +1,105 @@
...
+    return this.textDocument.offsetAt(position)
+  }
+
+  protected positionAt(offset: number) {
+    return this.textDocument.positionAt(offset)
+  }
+
+  private findLocationsForPropertySymbol(symbol: DocumentSymbol) {
+    const propertyPath = symbol.detail
+    const itIsHere = propertyPath && this.hasProperty(propertyPath)
```
sroy3, line 60: Nitpick, but is it possible to rename `isHere` as it follows our boolean naming conventions more.

languageServer/src/languageHelpers/yamlHelper.ts
```diff
@@ -0,0 +1,220 @@
...
+  isPair,
+  isScalar,
+  Pair,
+  parseDocument,
+  Scalar,
+  Node,
+  visit
+} from 'yaml'
+import { BaseLanguageHelper } from './baseLanguageHelper'
+import * as RegExes from './regexes'
```
sroy3, line 19: Since you can still import the regexes individually (ex.: `filePaths`), I think it'd be better to rename the exported constants in `regexes.ts` to identify them as regexes and simply import them individually. The other solution would be to wrap them all in a `regexes` object and export that instead, but I think the first solution is better.

languageServer/src/languageHelpers/yamlHelper.ts
```diff
@@ -0,0 +1,220 @@
...
+      const end = this.positionAt(nodeEnd)
+      const range = Range.create(start, end)
+      return Location.create(this.textDocument.uri, range)
+    }
+
+    return null
+  }
+
+  protected toJSON(): unknown {
+    return this.rootNode.toJS()
```
sroy3, line 64: Why `toJS()` instead of `toJSON()`? Might be worth a comment.

languageServer/src/regexes.ts
```diff
@@ -1,3 +1,4 @@
 export const variableTemplates = /\${([^}]+)}/g
 export const filePaths = /[\d/A-Za-z]+\.[A-Za-z]+/g
 export const alphadecimalWords = /[\dA-Za-z]+/g
+export const propertyPathLike = /[\d.A-Z[\]a-z]+/g
```
sroy3, line 4: Any differences between this file and languageServer/src/languageHelpers/regexes.ts? They should be merged if so. Either way we should only have one file only for regexes.

extension/src/lspClient/languageClient.ts
```diff
@@ -35,31 +28,24 @@ export class LanguageClientWrapper extends Disposable {
...
-    await this.client.sendRequest('initialTextDocuments', {
-      textDocuments
-    })
+      workspace.onDidChangeConfiguration((event: ConfigurationChangeEvent) => {
+        if (
+          event.affectsConfiguration(ConfigKey.DISABLE_LANGUAGE_SERVER) &&
+          this.isRunning() &&
+          this.languageServerDisabled()
+        ) {
+          this.stop()
```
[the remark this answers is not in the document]
wolmir: Thank you!

languageServer/src/languageHelpers/baseLanguageHelper.ts
wolmir: Yes, thanks!

languageServer/src/languageHelpers/yamlHelper.ts
wolmir: Yes, that's what I tried first, but I worried if anyone would be annoyed about typing a prefix every time they create/use the constant.
But now that I'm writing this I realised that typing `Regexes.something` is exactly the same thing and now I'm feeling really dumb.
I'll change it, thanks!

languageServer/src/languageHelpers/yamlHelper.ts
wolmir: Makes more sense, thanks!

languageServer/src/regexes.ts
wolmir: Forgot about it, will delete it 👍

extension/src/lspClient/languageClient.ts
```diff
@@ -35,31 +28,24 @@ export class LanguageClientWrapper extends Disposable {
...
 
   async start() {
-    await this.client.start()
-
-    const files = await findFiles('**/*.{yaml,json,py,toml}', '.??*')
-
-    const textDocuments = files.map(filePath => {
-      const uri = Uri.file(filePath).toString()
-      const languageId = filePath.endsWith('yaml') ? 'yaml' : 'json'
-      const text = readFileSync(filePath, 'utf8')
```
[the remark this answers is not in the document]
mattseddon: The watcher would not have caused this issue. We already have at least one watcher which watches the entire workspace (and more).

languageServer/src/LanguageServer.ts
```diff
@@ -10,16 +10,14 @@ import {
   Location,
```
julieg18, line 10: File definitions inside of `cmd`s don't seem to work correctly on my end: 

https://user-images.githubusercontent.com/43496356/194176392-50d8b562-08e6-46bd-b3bd-ddada358f3b1.mov

languageServer/src/LanguageServer.ts
```diff
@@ -10,16 +10,14 @@ import {
...
   Position,
```
julieg18, line 11: What should be shown if a file hasn't been opened yet? Currently, it appears to take us somewhere that doesn't seem to make sense: 

https://user-images.githubusercontent.com/43496356/194176758-b776d5b4-f84e-4d95-a8a1-f1156b5aee34.mov

languageServer/src/languageHelpers/baseLanguageHelper.ts
```diff
@@ -0,0 +1,105 @@
...
+  Location,
+  SymbolKind
+} from 'vscode-languageserver/node'
+
+export interface LanguageHelper {
+  findSymbolAtPosition(position: Position): DocumentSymbol | undefined
+  findLocationsFor(symbol: DocumentSymbol): Location[]
+}
+
+export abstract class BaseLanguageHelper<RootNode> implements LanguageHelper {
```
mattseddon, line 15: It's usual practice in the codebase to have this a `<T>`. I was looking for where `<RootNode>` is defined for a couple of minutes.

languageServer/src/languageHelpers/baseLanguageHelper.ts
```diff
@@ -0,0 +1,105 @@
+import { has } from 'lodash'
+import { TextDocument } from 'vscode-languageserver-textdocument'
+import {
+  DocumentSymbol,
+  Position,
+  Location,
+  SymbolKind
+} from 'vscode-languageserver/node'
+
+export interface LanguageHelper {
```
mattseddon, line 10: [N] Probably don't need this interface on top of the abstract class.

extension/src/lspClient/languageClient.ts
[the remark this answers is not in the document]
mattseddon: The more I think about this the more I feel like the approach was backwards. We should be parsing each `dvc.yaml` file when it is opened and checking to see which symbols are files. From there we read those files so that we can jump into them if needed.

languageServer/src/TextDocumentWrapper.ts
```diff
@@ -45,132 +33,11 @@ export class TextDocumentWrapper implements ITextDocumentWrapper {
     return parseDocument(this.getText())
```
mattseddon, line 33: `getYamlDocument` is no longer used anywhere (seems like it has been moved into the `languageHelper`) it should be removed from this class and the interface.

extension/src/lspClient/languageClient.ts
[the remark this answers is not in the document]
mattseddon: Relevant: https://code.visualstudio.com/updates/v1_72#_improved-startup-performance

See "NO MORE SYNC FS CALLS"

Issue #2501, opened 2022-09-29
Extension crashing vscode extension host


bobertlo:
I can't really troubleshoot this right now, but I am getting an error "Extension host terminated unexpectedly 3 times within the last 5 minutes." that stops when I disable the extension. This first happened on a linux ssh remote host when I logged in this morning at work, and is happening also on my MacBook which was working fine at midnight last night on the same project.

Another user reported the issue in discord and said it is effecting some projects and not others. In the project I am using I only have data tracked for ETL and no pipelines enabled at all, just some the `.dvc` tracked datasets and nothing else going on.

I am at work and can't troubleshoot this, but it sounds like it should be reproducible.

shcheklein:
@bobertlo we are looking in this.

mattseddon:
@bobertlo. 

Firstly, sorry for the terrible experience. Secondly, I've just queued up a release with a possible fix. If you have time can you do a few things for me:

1. When `0.4.10` is released please check to see if it works.
2. If it doesn't (and if possible) can you please share a project with me that the extension crashes?
3. Can you please confirm that `0.4.8` also crashes (just want to double check)?

🙏🏻 and sorry again.

mattseddon:
`0.4.10` is available now.

bobertlo:
It's resolved. Thanks!

mattseddon:
@wolmir need a postmortem for this one.

PR #2505, opened 2022-09-29
Turn off language server again

mattseddon:
Based on 

image

Possible cause of #2501

We should turn off and ship an update to see if this fixes the issue.

If the issue was only for `0.4.9` then I would guess that it is an issue with flexible plots. See as the crash was introduced in `0.4.8` this is (IMO) the most likely cause.

### DIFF
--- a/demo/dvc.yaml
+++ b/demo/dvc.yaml
@@ -2,21 +2,21 @@ stages:
   train:
     cmd: python train.py
     deps:
-    - data/MNIST
-    - train.py
+      - data/MNIST
+      - train.py
     params:
-    - params.yaml:
+      - params.yaml:
     outs:
-    - model.pt:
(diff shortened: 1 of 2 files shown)


demo/dvc.yaml
```diff
@@ -2,21 +2,21 @@ stages:
...
-    - predictions.json:
-        template: confusion
-        x: actual
-        y: predicted
+      - training_metrics
+      - misclassified.jpg
+      - predictions.json:
+          template: confusion
+          x: actual
+          y: predicted
```
mattseddon, line 22: [F] Formatting only. Given the time-sensitive nature of the change, I think we can ship as is.

julieg18 (approved):

wolmir (approved):

PR #2517, opened 2022-10-03
Update table header context menu 

julieg18:
* Adds "Update Header Depth" item to header 

## Demo

https://user-images.githubusercontent.com/43496356/193711009-db465cf0-ae59-4955-aadd-c6f14fb25eb2.mov

Related to #2390


extension/src/experiments/webview/messages.ts
```diff
@@ -136,6 +136,9 @@ export class WebviewMessages {
           this.showPlots()
         ])
 
+      case MessageFromWebviewType.UPDATE_HEADER_DEPTH_SETTINGS:
+        return commands.executeCommand('workbench.action.openWorkspaceSettings')
```
julieg18, line 140: The closest thing I could find to linking to the dvc config settings was to just link to workspace settings in general... 

The only doable solution I can think of is creating a command that allows the user to update the config from the command palette 🤔

mattseddon: All you need to do is use [setConfigValue](https://github.com/iterative/vscode-dvc/blob/1efac498e4589de18877530c7210290de51a0254/extension/src/vscode/config.ts#L19). You can set the value through any means you want.

mattseddon: You will need to use a `quickInput` to get the required number.

webview/src/experiments/components/App.test.tsx
```diff
@@ -702,6 +702,7 @@ describe('App', () => {
       const itemLabels = menuitems.map(item => item.textContent)
       expect(itemLabels).toStrictEqual([
         'Open to the Side',
+        'Update Header Depth',
```
mattseddon, line 705: [I] Should be `Set Max Header Depth`.

mattseddon (approved):
Should consolidate the use of layers and depth. Would be good to see a test too.

sroy3 (approved):

julieg18:
> Should consolidate the use of layers and depth. Would be good to see a test too.

Updated the text/variables, using depth instead of layers! And apologies, forgot tests again 😅 Will open them up in a followup!

extension/src/vscode/title.ts
```diff
@@ -17,7 +17,8 @@ export enum Title {
   SELECT_PLOTS = 'Select Plots to Display',
   SELECT_SORT_DIRECTION = 'Select Sort Direction',
   SELECT_SORTS_TO_REMOVE = 'Select Sort(s) to Remove',
-  SETUP_WORKSPACE = 'Setup the Workspace'
+  SETUP_WORKSPACE = 'Setup the Workspace',
+  SET_EXPERIMENTS_HEADER_DEPTH = 'Set Maximum Amount of Experiment Table Depth'
```
mattseddon, line 21: Don't need `Amount of`.

mattseddon: Probably want to add `Header`

shcheklein:
Thanks @julieg18 for the demo and the feature! Quick follow up questions / suggestions:

- Why there is a delay after applying the setting? Or at least it seems so ...
- In the quick pick menu let's add the description - e.g. that 0 means unlimited
- Minor: Depth probably is not the right term (should it be just height for simplicity?)

julieg18:
> Why there is a delay after applying the setting? Or at least it seems so ...

Yes, there is a delay! It takes a sec for the table to rebuild. 

> In the quick pick menu let's add the description - e.g. that 0 means unlimited

Sure! I'll add it in a followup.

> Minor: Depth probably is not the right term (should it be just height for simplicity?)

"height" sounds good to me! Will update in a followup :)

shcheklein:
Thanks!

> Yes, there is a delay. It takes a sec for the table to rebuild.

Can is be faster? Does it run a CLI command inside?

PR #2520, opened 2022-10-04
Move types from reader into new contract file

mattseddon:
# 1/2 `main` <- this <- #2521 

Prerequisite of moving from the current boolean `running` and `queued` fields in `exp show` to a new consolidated `status` field. 

We will need to import the new `ExperimentStatus` enum into the table webview. We cannot do this from the reader because the reader imports `join` from `path` (which isn't available). Makes sense (IMO) to re-order the types.

We now have another `contract` file. This time for the contract (shared types) between the extension and the CLI.


julieg18 (approved):
Looks good!

sroy3 (approved):

Issue #2516, opened 2022-10-03
Showing wrong info for non-linear flexible plots


dberenbaum:
Now that https://github.com/iterative/vscode-dvc/issues/1757 has been closed, I can see linear flexible plots with nice formatting 🙌 :

<img width="1121" alt="Screen Shot 2022-10-03 at 12 13 59 PM" src="https://user-images.githubusercontent.com/2308172/193627458-31d244ed-eabf-476e-9de6-9335baa0af01.png">

However, the confusion matrix above is not being shown as expected since it is supposed to be comparing train and test for each revision. Here's what it looks like outside of VS Code:

<img width="1383" alt="Screen Shot 2022-10-03 at 12 12 17 PM" src="https://user-images.githubusercontent.com/2308172/193627629-7ab2c2dc-a2f5-4b48-a757-ebab675887c7.png">

VS Code seems to be combining train and test into one confusion matrix.

dberenbaum:
cc @pared

mattseddon:
Can you share the project please @dberenbaum.

dberenbaum:
🤦 Sorry about that: https://github.com/dberenbaum/example-get-started/tree/flexible_plots

mattseddon:
I have to concatenate the required data into the rev field in order to render the confusion matrix as expected: 

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/193736883-ec684669-cb2d-477d-9a93-ce48e5037147.png">

This has raised another issue:

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/193736952-49b172fb-1d79-42ca-9ff7-b9a02469d9f3.png">

Even regular-sized confusion matrices are too small once we get to 4 revisions (cc @sroy3).

PR #2523, opened 2022-10-04
Render flexible confusion matrices as expected

mattseddon:
# 1/2 `main` <- this <- #2531

Closes https://github.com/iterative/vscode-dvc/issues/2516

The problem was that the revision field was not being updated by us before trying to render the plot as expected. In order to do that we have to combine the rev field with the filename in the underlying data (no change to the template for these ones).

Successfully rendering the plot then causes a separate issue which is addressed in #2531.

### Screenshot:

image

Test project: https://github.com/dberenbaum/example-get-started/tree/flexible_plots (output copied as a new test fixture).

**Note**: `+337,296` of the diff comes from `extension/src/test/fixtures/plotsDiff/multiSource.ts`


extension/src/test/fixtures/plotsDiff/output/multiSource.ts
```diff
@@ -0,0 +1,5 @@
+import { getMultiSourceOutput } from '..'
+
+const data = getMultiSourceOutput()
+
+export default data
```
mattseddon, line 5: [F] This uses the regular pattern if we want to include the multi-source plots in a new story then I should be able to wire them up pretty quickly.

sroy3 (approved):

julieg18 (approved):
Great work!

PR #2521, opened 2022-10-04
Bump min DVC version to 2.30.0 (Use status from exp show)

mattseddon:
# 2/3 `main` <- #2520 <- this <- #2535

Accommodates the breaking change in https://github.com/iterative/dvc/pull/8318. We are moving from 2 boolean fields to an enum which will now include:

```
export enum ExperimentStatus {
  FAILED = 'Failed',
  QUEUED = 'Queued',
  RUNNING = 'Running',
  SUCCESS = 'Success'
}
```

We cannot merge until the above PR is merged. I will also need to increase the `MIN_CLI_VERSION`. I will raise follow-up PRs to indicate failed experiments in the table (another behaviour change).

Branch for testing: `git+https://github.com/karajan1001/dvc.git@fix7986`.

PR #2535, opened 2022-10-06
Display failed experiments

mattseddon:
# 3/3 `main` <- #2520 <- #2521 <- this

Relates to https://github.com/iterative/dvc/pull/8318

This PR adds code to display failed experiments as experiments with errors. Failed experiments will be displayed in the exact same way as other experiments with errors.

### Screenshot

Screen Shot 2022-10-06 at 1 36 38 pm

Steps to recreate state: 
1. Using `git+https://github.com/karajan1001/dvc.git@fix7986` (update min value of DVC)
2. Queue several experiments
3. `dvc queue start -j 3`

**Note**: I have added a new record to the base test fixture which relates to this situation.

### DIFF
--- a/extension/src/cli/dvc/contract.ts
+++ b/extension/src/cli/dvc/contract.ts
@@ -1,6 +1,8 @@
 import { Plot } from '../../plots/webview/contract'
 
-export type DvcError = { error: { type: string; msg: string } }
+type ErrorContents = { type: string; msg: string }
+
+export type DvcError = { error: ErrorContents }
 
 export type Changes = {
   added?: string[]
@@ -22,7 +24,7 @@ export type Value = SingleValue | SingleValue[]
 
 export interface ValueTreeOrError {
   data?: ValueTree
-  error?: { type: string; msg: string }
+  error?: ErrorContents
 }
 
 type RelPathObject<T> = {
@@ -63,11 +65,12 @@ export interface ExperimentFields extends BaseExperimentFields {
   metrics?: ValueTreeRoot
   deps?: Deps
   outs?: RelPathObject<Out>
+  error?: ErrorContents
 }
 
 export interface ExperimentFieldsOrError {
   data?: ExperimentFields
-  error?: { type: string; msg: string }
+  error?: ErrorContents
 }
 
 export interface ExperimentsBranchOutput {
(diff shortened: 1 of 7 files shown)


sroy3:
> 1. Using `git+https://github.com/karajan1001/dvc.git@fix7986` (update min value of DVC)

This was merged and the branch deleted. `git+https://github.com/iterative/dvc` until it's released.

sroy3 (approved):

julieg18 (commented):
Possibly due to me using the wrong DVC version, but shouldn't these queued experiments have a clock icon and not be selectable for plots? 

image

mattseddon:
> Possibly due to me using the wrong DVC version, but shouldn't these queued experiments have a clock icon and not be selectable for plots?
> 
> image

This is the unexpected behaviour that the breaking change causes. 

If the branch gets merged you should be able to use upstream to test. In this case that you be `git+https://github.com/iterative/dvc` because it went into `main` which is the default branch.

sroy3:
> This is the unexpected behaviour that the breaking change causes.
> 
> If the branch gets merged you should be able to use upstream to test. In this case that would be `git+https://github.com/iterative/dvc` because it went into `main` which is the default branch.

I was getting the same when using `git+https://github.com/iterative/dvc``

mattseddon:
> I was getting the same when using git+https://github.com/iterative/dvc

I have just double-checked:

image

The checklist would be:

1. `vscode-dvc` is on `handle-failed-exp` branch. 
2. dev-server is started
3. `git+https://github.com/iterative/dvc` is installed in the demo project virtual environment (`pip install -r requirements.txt -U`)
4. extension debugger is open with demo project as the root
5. DVC is provided to the extension from the virtual environment

sroy3:
> The checklist would be:
> 
> 1. `vscode-dvc` is on `handle-failed-exp` branch.
> 2. dev-server is started
> 3. `git+https://github.com/iterative/dvc` is installed in the demo project virtual environment (`pip install -r requirements.txt -U`)
> 4. extension debugger is open with demo project as the root
> 5. DVC is provided to the extension from the virtual environment

These are the exact steps I was running. I think I'm having problems with my python extension for some reason. The virtual environment does not seem to get activated automatically anymore and manual setup seems to not work as expected. I'll try again.

julieg18:
> If the branch gets merged you should be able to use upstream to test. In this case that would be git+https://github.com/iterative/dvc because it went into main which is the default branch.

Thanks for the checklist! Everything looking good from my end: 

image

PR #2544, opened 2022-10-07
Fix UX of extension using fallback global CLI when Python extension is active

mattseddon:
# 1/2 `main` <- this <- #2552

This PR fixes the UX of using a global CLI fallback when the Python extension is active. Previously the extension would give a less than helpful warning if the CLI could not be found in the selected virtual environment and the wrong version was found globally. This PR digs into the implementation and moves things around in order to give more helpful directions for this situation.

#### not found with Python, not found globally

https://user-images.githubusercontent.com/37993418/194786089-84c5b071-59c6-4566-b177-31b593874c69.mov

### not found with Python, install globally

https://user-images.githubusercontent.com/37993418/194786263-1ab3c8f1-0f5b-4874-a7cb-ff60c4b2a9fd.mov

### not found with Python, global behind

https://user-images.githubusercontent.com/37993418/194786307-59ef3781-6d6f-4306-ae04-015123fd2da7.mov

#### behind with Python (no global check)

https://user-images.githubusercontent.com/37993418/194786333-1fd31f1f-cb10-4a5c-a798-cd90a9607b24.mov

### correct version from Python

https://user-images.githubusercontent.com/37993418/194786386-f4cee864-d82a-4f64-a00a-5e245c92aa0e.mov

### DIFF
--- a/extension/src/cli/dvc/version.test.ts
+++ b/extension/src/cli/dvc/version.test.ts
@@ -1,6 +1,10 @@
-import { isVersionCompatible, extractSemver, ParsedSemver } from './version'
+import {
+  isVersionCompatible,
+  extractSemver,
+  ParsedSemver,
+  CliCompatible
+} from './version'
 import { MIN_CLI_VERSION, LATEST_TESTED_CLI_VERSION } from './constants'
-import { Toast } from '../../vscode/toast'
 
 jest.mock('./constants', () => ({
   ...jest.requireActual('./constants'),
@@ -8,11 +12,6 @@ jest.mock('./constants', () => ({
   MIN_CLI_VERSION: '2.9.4'
 }))
 jest.mock('../../vscode/config')
-jest.mock('../../vscode/toast')
-
-const mockedToast = jest.mocked(Toast)
-const mockedWarnWithOptions = jest.fn()
-mockedToast.warnWithOptions = mockedWarnWithOptions
 
 beforeEach(() => {
   jest.resetAllMocks()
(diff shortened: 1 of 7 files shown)


extension/src/cli/dvc/version.test.ts
```diff
@@ -57,125 +56,114 @@ describe('isVersionCompatible', () => {
     patch: latestTestedPatch
   } = extractSemver(LATEST_TESTED_CLI_VERSION) as ParsedSemver
 
-  it('should be compatible and not send a toast message if the provided version matches the min version', () => {
```
mattseddon, line 60: [F] The message has been hoisted out of this function so that it can be changed for different situations. The tests have been moved to `extension/setup.test.ts`

extension/src/cli/dvc/version.ts
```diff
@@ -5,6 +5,27 @@ import {
 } from './constants'
 import { Toast } from '../../vscode/toast'
 
+export enum CliCompatible {
```
mattseddon, line 8: [F] The return value from `isVersionCompatible` needs to provide more than a boolean value as we need more information on what message to give to the user. These are all of the possible situations that we can have when checking the CLI's version.

extension/src/setup.ts
```diff
@@ -163,37 +171,21 @@ export const setupWorkspace = async (): Promise<boolean> => {
...
-  }
-  const binPath = await getPythonBinPath()
-
-  return (
-    text +
-    ` For auto Python environment activation ensure the correct interpreter is set. Active Python interpreter: ${binPath}. `
-  )
-}
-
 const getToastOptions = (isPythonExtensionInstalled: boolean): Response[] => {
```
mattseddon, line 174: [F] The bulk of the logic will be moved into a new file in the follow-up PR.

extension/src/setup.ts
```diff
@@ -209,38 +201,149 @@ const warnUserCLIInaccessible = async (
...
+  globalDvcVersion: string | undefined
+): Promise<void> => {
+  const binPath = await getPythonBinPath()
+
+  return warnUserCLIInaccessible(
+    extension,
+    true,
+    `The extension is unable to access an appropriate version of the CLI. The CLI was not located using the interpreter provided by the Python extension. ${
+      globalDvcVersion ? globalDvcVersion + ' is' : 'The CLI is also not'
+    } installed globally. For auto Python environment activation ensure the correct interpreter is set. Active Python interpreter: ${binPath}.`
```
julieg18, line 215: Maybe I'm wrong, but should there be a comma in this sentence?

```suggestion
    } installed globally. For auto Python environment activation, ensure the correct interpreter is set. Active Python interpreter: ${binPath}.`
```

This message is also rather lengthy 🤔 Should we try to simplify the message? First idea that   comes to mind is combining the 2nd and 3rd sentences 🤷

julieg18 (approved):
Great work!

extension/src/setup.ts
mattseddon: I have dropped a few words but I am concerned that if we change the structure of the message that it will be less clear. This message/edge case is the whole reason this PR was created. It should not be seen often but it should get users out of trouble.

Issue #2373, opened 2022-09-10
Plots: no clear connection between "circles" and plots view


maxagin:
It is evident that users are confused about what the circle beside the EXPs is for. Also if the user opens Plots, but no EXPs are selected, the Plots view is empty and the user does not know how to add EXPs from the Table view. This issue is partially resolved with the ability to add experiments from the Command Palette. However, the education regarding how both views should accompany each other is missing.

References:

- https://www.notion.so/iterative/Dan-Martinec-feedback-for-VS-code-689d8945e5f54bdc8f6e064a06fccb28#c0b8c0d19ee943aea59af390cd169c71
- https://www.notion.so/iterative/Mikhail-Rozhkov-VSCode-DVC-Ext-106392d8bb9d4b69b02755419b8fe823#de2532d6020b43668de533f66eab07f1

I would like to propose a solution for both cases when the views are opened for the first time.

&nbsp;

## 👉 Show a short explanation tooltip beside the EXP's circle when the user first time opens the EXPs table view.

To display the EXP in the Plots view click the circle beside the EXP name.
Button text example: Got it

<img width="500" alt="Screen Shot 2022-09-12 at 4 53 23 PM" src="https://user-images.githubusercontent.com/98249521/189756979-6ef8b82e-9592-455e-88d7-eb114c92381e.png">

Do you have in mind any other functionality that we would like to show upfront?

&nbsp;

## 👉 If no EXPs are selected and the user opens Plots, show information (use illustration for better understanding) that explains how the user can add EXPs to the Plots view from the EXPs view along with the button that adds EXPs from the Command Palette.

To display EXP in the Plots view activate the circle beside the EXP title in the Table view.
You can add max 7 EXPs to the Plots.

Alternatively, click the button below to add EXPs from the Command Palette.

<img width="500" alt="Screen Shot 2022-09-10 at 3 19 28 PM" src="https://user-images.githubusercontent.com/98249521/189498847-0aa85598-43af-4046-b692-8663f7b0b7d7.png">

[Figma](https://www.figma.com/file/cOlfG6r9dFa8OvWkRRQFuK/Point-to-critical-functionality-upfront?node-id=1404%3A5882)

mattseddon:
Related to #2041, #1958, #1383

shcheklein:
We've introduced the ribbon to mitigate this. Users now should see it and be able to pick experiments there. (so, circles in the table become secondary or ternary after the side panel and the ribbon).

[...]

PR #2545, opened 2022-10-08
Plotting walkthrough prototype

mattseddon:
(Rough) example of what is possible using the available APIs

Still has discoverability issues.

### Demo

https://user-images.githubusercontent.com/37993418/194687070-ed3441d7-fe90-4089-8f86-cccca98935a4.mov

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -89,6 +89,12 @@
     ],
     "commands": [
       {
+        "title": "Do Plots Walkthrough",
+        "command": "dvc.plotsWalkthrough",
+        "category": "DVC",
+        "icon": "$()"
+      },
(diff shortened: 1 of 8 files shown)


shcheklein:
Yep, I'm not sure how can it solve the problem (I'm still not sure that that is a very big problem though).

mattseddon:
> Yep, I'm not sure how can it solve the problem (I'm still not sure that that is a very big problem though).

If the connection between circles and plots is not a big problem, why do we have https://github.com/iterative/vscode-dvc/issues/2373 as p1?

mattseddon:
I agree it is not a great idea.

PR #2563, opened 2022-10-11
Make the plot sizes use numbers underneath

sroy3:
Part of #2585 

https://user-images.githubusercontent.com/3683420/195160042-4216a646-bf95-4462-b474-2839c9e478ce.mov


webview/src/plots/components/comparisonTable/ComparisonTableWrapper.tsx
```diff
@@ -11,7 +11,7 @@ export const ComparisonTableWrapper: React.FC = () => {
   const { size, isCollapsed } = useSelector(
     (state: PlotsState) => state.comparison
   )
-  const handleResize = (size: PlotSize) => {
+  const handleResize = (size: number) => {
```
mattseddon, line 14: [C] Seems like there are three versions of the same `handleResize` wrapper function.

mattseddon (approved):

webview/src/plots/components/comparisonTable/ComparisonTableWrapper.tsx
sroy3: It's because they are dispatching 3 different actions (they share the same name, but are imported from different files).

sroy3:
Will wait to merge until more work for the grid is lined up

webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -63,22 +69,15 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
   currentSize,
   menu
 }) => {
-  const [size, setSize] = useState<PlotSize>(currentSize)
+  const [size, setSize] = useState<number>(currentSize || 400)
```
julieg18, line 72: Maybe I'm wrong, but isn't it recommended that when it comes to updating state based off of props, you should update with a `useEffect` hook? Like this:

```
const [size, setSize] = useState<number>(400)

  useEffect(() => {
    setCurrentSize(currentSize)
  }, [currentSize]);
```

sroy3: We're already setting the size inside the in the `changeSize` function, we don't need to wait for the size to come back from the extension. We only care about the initial one.

Issue #2395, opened 2022-09-14
Exp info when selecting exps to display in Plots


alex000kim:
When adding experiments to Plots, I'd like to see at least some info about the experiments I am adding because experiment names don't tell me anything useful.
An idea:
Maybe we can display metrics next to the experiment names, and display more info (e.g. params) on hover?

![](https://user-images.githubusercontent.com/17241029/189727169-a2fd56fb-18db-4af0-998d-1435f18b575f.png)

mattseddon:
How about adding the first 3 metrics that are found in the table (by leftmost position) to the description in the quick pick:

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/194985557-e733fcc1-ef32-447e-b998-acab9339231c.png">

shcheklein:
Params and data information are equally important. We should pick first three columns in the table (any columns). Ideally we should be able to pin those columns we are interested in.

I think we should also include date (and ideally how long did it take to run, but that's a separate ticket). We can show those under the exp name for example in quick pick?

mattseddon:
Suggested format for a new tooltip in the experiments tree: 

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/195001146-339a615b-c8d1-4864-958a-ed1c9ad2bd59.png">

PR #2567, opened 2022-10-12
Add hints into experiment table row action tooltips

mattseddon:
# 2/2 `main` <- #2553 <- this

This PR adds hints to the experiment table row action tooltips. Initially, I wanted to highlight how to open the plots' view but seems like a good idea to try and put something useful into the starred tooltip as well.

### Demo

https://user-images.githubusercontent.com/37993418/195232467-1308ee92-c450-4371-b16a-e3d225ba6310.mov

### DIFF
--- a/webview/src/experiments/components/table/CellRowActions.tsx
+++ b/webview/src/experiments/components/table/CellRowActions.tsx
@@ -63,8 +63,14 @@ export const CellRowAction: React.FC<CellRowActionProps> = ({
   )
 }
 
-const getTooltipContent = (determiner: boolean, text: string): string =>
-  determiner ? `Un${text.toLowerCase()}` : text
+const getTooltipContent = (
+  determiner: boolean,
+  action: string,
+  helperText?: string
+): string =>
+  'Click to ' +
+  (determiner ? `un${action}` : action) +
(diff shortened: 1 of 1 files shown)


sroy3 (approved):

julieg18 (approved):
Good work!

shcheklein:
Can we put an action on the tooltip (make it clickable)? 

Unplot --> "hide plots" ? 

Can we do the same tooltip in the side panel?

Overall, looks good to me.

mattseddon:
@shcheklein 

> Can we do the same tooltip in the side panel?

Would you prioritize this over adding extra data into the tooltip as per https://github.com/iterative/vscode-dvc/issues/2395#issuecomment-1274092202

I can try to combine the two but I don't think it will look very good/be as useful there.

shcheklein:
> Would you prioritize this over adding extra data into the tooltip as per https://github.com/iterative/vscode-dvc/issues/2395#issuecomment-1274092202

No, definitely actual data take priority like in your comment. I was thinking about the circle tooltip only, but may be we can include in the tooltip a link at the bottom

mattseddon:
Will follow up to make tooltip actionable

PR #2572, opened 2022-10-12
Screenshot for "Select up to 7 experiments in the table view to visua…

maxagin:
Image for the Get Started


maxagin:
Also testing a new screenshot style - no shadow, but it has a border.

shcheklein:
Let's not change the style for now (seems minor, and there was a reason to keep the shadow - looked better on a dark background, more contrast). It would be great to avoid the full update of all the images. Or do we need / plan to update all the images? (Closing for now, please feel free to reopen it also).

Please also consolidate it with the text update in a single PR.

PR #2573, opened 2022-10-12
Update the DVC: Get Started / Plots Dashboard

maxagin:
Select (click the circle beside the experiment name) one or more up to 7 experiments in the table view to visualize in the Plots Dashboard. screenshot
Use DVC: Show Plots from the command palette to open it or open it using the table's row context menu.

https://user-images.githubusercontent.com/98249521/199637051-566d183e-e38d-4f67-acd0-eccbcc0308a6.mov

---

~~@shcheklein : We need to wait for the https://github.com/iterative/vscode-dvc/pull/2553 to land~~


shcheklein:
Thanks @maxagin . The image should be also part of this PR please.

Also, please see in the file and wrap text at 80 symbols.

maxagin:
Sorry forgot to wrap the text. Will create an update

extension/resources/walkthrough/plots.md
```diff
@@ -1,9 +1,17 @@
 # Plots Dashboard
 
-Select one or more experiments to visualize in the
-[**Plots Dashboard**](command:dvc.showPlots). Use
+Select (click the circle beside the 🧪 experiment name) up to 7 experiments in
```
shcheklein, line 3: Let's please remove the emoji (we don't use them in the walkthrough this way).

shcheklein (changes requested):
- We need to wait for the https://github.com/iterative/vscode-dvc/pull/2553 to land
- Please use https://tinypng.com/ to make sure that image is optimized

shcheklein:
The problem with this approach is that people don't see now the plots dashboard itself when they open the page (they would need to scroll). Should we change the order of things?

maxagin:
> The problem with this approach is that people don't see now the plots dashboard itself when they open the page (they would need to scroll). Should we change the order of things?

Maybe we can rename the section titles e.g. 

- Experiments View ~~Experiments Table~~
- Plots View ~~Plots Dashboard~~

Screen Shot 2022-10-20 at 6 21 50 PM

WDYT?

Otherwise, we need to change the textual content I think.

---

> that people don't see now the plots dashboard

Before it also was not at the top!

Screen Shot 2022-10-20 at 6 27 10 PM

shcheklein:
@maxagin re renaming - let's do one thing at at a time please ;)

Was it at least somewhat visible? My point is that even if it was bad we are making it worse. Should we reorder things a bit?

shcheklein:
@maxagin feel too much and more images to maintain.

shcheklein:
@maxagin so, how about changing the order and finally merging this and the second PR (we can release everything today / tomorrow then)?

maxagin:
> more images to maintain.

1. Three more images won't make any difference if all the images will have to be changed.
2. Personally I think this is a very clever idea - to show right away the main extension views, thus we can educate the user about what he can expect and hopefully reduce the number of leaving users.

> changing the order

I am not sure this is good, as the content of this section is clearly organized to explain what you need to do step by step: 
1. What to do to show EXPs in Plots view
2. Types of plots
3. Explaining the types

@shcheklein, what is your take on this?

shcheklein:
@maxagin any updates on this? :)

maxagin:
Completely went out of my radar! I will deal with it ASAP @shcheklein

shcheklein:
@maxagin could you update / or upload the video please?

maxagin:
@shcheklein the description video is updated.

shcheklein (approved):

PR #2574, opened 2022-10-12
Match title from the left section

maxagin:
Would expect to see the same title (correct for all the sections).


shcheklein (changes requested):
I think the idea was to keep both - person clicks on the Available Commands and there is Command Palette that gives a sense how to access them. Not sure if it's beneficial tbh. If we decide to rename after all (I'm fine with that) we should rename the file also, and update the package.json

Issue #1958, opened 2022-06-30
Walkthroughs: feedback and suggestions (Get Started)


maxagin:
- [x]  👉 11. New screenshot style
Let's not change the style for now (seems minor, and there was a reason to keep the shadow - looked better on a dark background, more contrast). It would be great to avoid the full update of all the images. Or do we need / plan to update all the images? (Closing for now, please feel free to reopen it also).

Please also consolidate it with the text update in a single PR.

- [x]  👉 9. Update the DVC: Get Started / Plots Dashboard

First of all, we can update the DVC: Get Started / Plots Dashboard, ~~I think this will be already helpful.~~
We can explicitly mention and show the screenshot of how to select the EXP for Plots.

### The draft

Select `(click the circle beside the experiment name)` ~~one or more~~ `up to 7` experiments `in the table view` to visualize in the Plots Dashboard. 
`screenshot`
Use DVC: Show Plots from the command palette to open it `or open it using the table's row context menu`.

<img width="500" alt="Screen Shot 2022-10-06 at 9 59 17 PM" src="https://user-images.githubusercontent.com/98249521/194450541-6e9291c7-ed21-4991-b0ca-227b902b5c79.png">

Screenshot example (do not use this one):

<img width="500" alt="Screen Shot 2022-10-06 at 9 57 35 PM" src="https://user-images.githubusercontent.com/98249521/194450399-0895976c-081d-4ca9-83f1-751b9198de19.png">

Related to https://github.com/iterative/vscode-dvc/issues/2373

---

- [ ]  👉 8. Opportunity for cross-promotion which can actually be meaningful in terms of user workflow. One of the many possible options. For example, in GetStarted, after Troubleshoot & Support add Collaborate, including information on how you collaborate in DVC Studio.

Message examples:
a.Now you are all set up, please do not forget to learn how our DVC Studio can help you and your team to collaborate on your ML projects…
b.Now you are all set up, please do not forget to create your account in DVC Studio where the collaboration on ML projects is happening

See the discussion here:
https://iterativeai.slack.com/archives/C01APS0FHDM/p1658437149849249

---

- [ ] 👉 2. Would expect to see the same title (correct for all the sections)

<img width="400" alt="Screen Shot 2022-06-29 at 11 11 33 PM" src="https://user-images.githubusercontent.com/98249521/176584832-d635e755-622e-4bd6-9771-6424e34a20c1.png">

[...]

PR #2575, opened 2022-10-12
New title "Helpful Resources"

maxagin:


shcheklein (commented):
@maxagin I don't see this as a substantial change tbh. Please, let's not micro-optimize things.

maxagin:
@shcheklein I just followed the Walkthroughs: feedback and suggestions (Get Started) #1958 ticket description.

shcheklein:
Yep, it makes sense. I would skip certain things in that feedback. And even if we do them, let's do a single PR? It becomes too granular otherwise :)

PR #2576, opened 2022-10-12
Changing one of the sections title

maxagin:


shcheklein (changes requested):
Same here, not a substantial change. If we do this, we should be updating this in a single PR

PR #2577, opened 2022-10-12
Testing table depth test

julieg18:


mattseddon:
One of the problems here is that the test leaves the workspace in an altered state. The config value is left as 0 after the test runs which means subsequent runs will always fail:

image

mattseddon:
In the extension integration tests we reset the value after each call: https://github.com/iterative/vscode-dvc/blob/9b458fb52e24eff0b8dace471c13c2db3ed69d05/extension/src/test/suite/extension.test.ts#L47-L48

But could probably use `after` to remove the config value in this instance:

```
  after(() => {
    workspace
      .getConfiguration()
      .update(ConfigKey.EXP_TABLE_HEAD_MAX_DEPTH, undefined, false)
  })
```

julieg18:
> One of the problems here is that the test leaves the workspace in an altered state. The config value is left as 0 after the test runs which means subsequent runs will always fail:
> In the extension integration tests we reset the value after each call:
> But could probably use after to remove the config value in this instance:

Nice catch! Thanks for the idea on using the `after` hook!  Could this problem be the reason behind the `telemetry` event being called twice in this test during some of the CI calls? I don't see how, but maybe I'm missing something 🤔

mattseddon:
> > One of the problems here is that the test leaves the workspace in an altered state. The config value is left as 0 after the test runs which means subsequent runs will always fail:
> > In the extension integration tests we reset the value after each call:
> > But could probably use after to remove the config value in this instance:
> 
> Nice catch! Thanks for the idea on using the `after` hook! Could this problem be the reason behind the `telemetry` event being called twice in this test during some of the CI calls? I don't see how, but maybe I'm missing something 🤔

Sometimes when using Mocha the calls can run into each other when the tests are async. I would just test that one of the calls was what you expected it to be.

mattseddon:
Looks like this has been replaced by #2621

PR #2587, opened 2022-10-13
Use error foreground for status bar item color if CLI unavailable

mattseddon:
From https://iterativeai.slack.com/archives/C01APS0FHDM/p1665690104423039?thread_ts=1665680827.468069&cid=C01APS0FHDM

"Extension is not initialized (should we be highlighting this better ( in red op yellow with a warning symbol?)"

### Demo

https://user-images.githubusercontent.com/37993418/195705585-b233996f-0ba9-4c82-bc18-85359729660f.mov


shcheklein:
Cool, thanks Matt. Can we do the same on the side panel (where we put a message that DVC is incompatible). Also may be give a hint there about the current version of DVC, put a link to docs (get started walkthough)?

mattseddon:
> Can we do the same on the side panel (where we put a message that DVC is incompatible). 

I will check the API but I don't think so.

> Also may be give a hint there about the current version of DVC, put a link to docs (get started walkthrough)?

There is already a toast popup with this information.

shcheklein:
> There is already a toast popup with this information

Yes, but it's easy to dismiss it and then you are lost.

Another, out of scope, but a bit related suggestion to improve the awareness is to change the quick pick to always show three options (gray out the Python one if extension is not installed, show the message that it needs to be installed if you want to pick the env automatically)

mattseddon:
> Yes, but it's easy to dismiss it and then you are lost.

Fair point but the text in the sidebar has to be static we cannot inject values into it.

> Another, out of scope, but a bit related suggestion to improve the awareness is to change the quick pick to always show three options (gray out the Python one if extension is not installed, show the message that it needs to be installed if you want to pick the env automatically)

The API does not allow us to do this either.

sroy3 (approved):

julieg18 (approved):

shcheklein:
Thanks, Matt! It's a good improvement. We still need to think on what to do on the side panel / webviews to make it 100% clear what's going on and how to solve the issue.

Issue #2256, opened 2022-08-24
Table header improvements


shcheklein:
- [x] Remove sort menu option from the parent params column drop down. Check for metrics. (https://github.com/iterative/vscode-dvc/pull/2583)

<img width="313" alt="Screen Shot 2022-08-24 at 1 41 17 PM" src="https://user-images.githubusercontent.com/3659196/186519509-0ea912c2-d10e-4e1b-9629-a6618654f7f0.png">

- [x] Add hide columns / section there (and in every other menu) (https://github.com/iterative/vscode-dvc/pull/2593)
- [x] ESC must work to close context menu
- [x] The context menu in the header cells should appear only on the right click. Currently also works with the left click (https://github.com/iterative/vscode-dvc/pull/2603)
- [x] Context menu "Change height" should be available everywhere. Now it doesn't show on metric file column.
- [x] Add filter by to context menus
- [ ] Add hover styles to header cells
- [x] The sort arrow in the header cell does need to have BG on hover but needs to have a cursor (pointer)
- [ ] Click on the Sort arrow - the context menu appears - if you hover on the arrow again the tooltip will be shown on top of the context menu. Do not show a tooltip when the menu is opened
- [x] ~The D&D cursor appears when the header block has hovered. Blocks like "Create" have no active space on top. Let's make sure that all the space of the block can be used to D&D~
- [x] Context menu doesn't show on an empty stub on top of a column (e.g. `train.py` dependecy).

maxagin:
> There must be a way to copy file names / section names. Full name and partial name.

Please see: Text selection in the table's header and body cells #2260

maxagin:
> There must be a way to copy file names / section names. Full name and partial name.

It will be handled here https://github.com/iterative/vscode-dvc/issues/2260

maxagin:
Some of the todo items were moved to #2241

maxagin:
- The context menu in the header cells should appear only on the right click. Currently also works with the left click

The description was updated. The new task was included.

maxagin:
Can “Open to the Side” params, but not training_metrics. Why? If I am not mistaken this should work for any file that can be opened on the side.

wolmir:
> Add hide columns / section there (and in every other menu)

This one is deceptively tricky, I implemented a solution, and now I'm adding/fixing some tests. Still, it will need some manual testing during review to catch side-effects.

PR #2593, opened 2022-10-14
Add missing "Hide Column" option to the other headers

wolmir:
This PR will remove the leaf-only restriction for the "Hide Column" context menu option in the experiment table's headers.

Part of https://github.com/iterative/vscode-dvc/issues/2256

I'm having problems with my demo and get-started projects, so I would appreciate some manual testing here, if possible.
Specifically, I would like to know if the Hide Column option is working well for all headers.
Thanks!


mattseddon:
> I'm having problems with my demo and get-started projects, so I would appreciate some manual testing here, if possible.

Do you have the latest version of DVC installed for those projects? There was a breaking change that meant we had to bump the min version of DVC required.

wolmir:
Yes, I updated and downgraded the DVC version, but It still has some problems. It could be some side-effect of me messing around with the DVC cli. I will try with a fresh demo repository later, and see if it works there.

mattseddon:
Unable to hide `Created` column:

Screen Shot 2022-10-17 at 9 52 32 am (2)

PR title should mention that the change is only for "Hide Column" and not all options e.g "Open to the Side"

Please add the appropriate label.

sroy3 (changes requested):
I don't think this should apply for placeholders.

https://user-images.githubusercontent.com/3683420/196191413-b15f5537-e290-4d05-bf0d-27f8a8cc425a.mov

wolmir:
> I don't think this should apply for placeholders.

@sroy3 Do you mean only the Hide Column option, or the entire context menu?

shcheklein:
@wolmir @sroy3 

> I don't think this should apply for placeholders.

I actually added an item to the #2256 yesterday to make it this way (the way on the recording). When I was using the product last time it was counterintuitive that there are some placeholders, for me the train.py was a single column header and it felt like a bug that I had to click on a very specific part of the header to make it work.

sroy3:
> > I don't think this should apply for placeholders.
> 
> @sroy3 Do you mean only the Hide Column option, or the entire context menu?

The context menu should be the same as when clicking the item (we shouldn't feel the split between item and placeholder).

wolmir:
> > > I don't think this should apply for placeholders.
> > 
> > 
> > @sroy3 Do you mean only the Hide Column option, or the entire context menu?
> 
> The context menu should be the same as when clicking the item (we shouldn't feel the split between item and placeholder).

Got it, thanks!

sroy3 (approved):

webview/src/experiments/components/table/header/TableHeader.tsx
```diff
@@ -27,7 +27,7 @@ const possibleOrders = {
 } as const
```
julieg18, line 27: The experiment column placeholder is showing a "Hide column" option: 

image

webview/src/experiments/components/table/header/TableHeader.tsx
```diff
@@ -27,7 +27,7 @@ const possibleOrders = {
...
```
julieg18, line 28: The column placeholders have a  "Hide Column" option, but it doesn't seem to work: 

https://user-images.githubusercontent.com/43496356/197221199-2d08ba46-5680-49ed-928e-7221079d76ef.mov

webview/src/experiments/components/table/header/TableHeader.tsx
```diff
@@ -27,7 +27,7 @@ const possibleOrders = {
...
 
 interface TableHeaderProps {
```
julieg18, line 29: If I attempt to to hide a parent column that already had hidden children, it adds the missing children instead of hiding the entire column:

https://user-images.githubusercontent.com/43496356/197222423-89094ddc-b94c-41ab-a9cc-35936614a5c5.mov

julieg18 (changes requested):

webview/src/experiments/components/table/header/TableHeader.tsx
wolmir: I addressed this specific issue in #2643 , if that's ok

webview/src/experiments/components/table/header/TableHeader.tsx
wolmir: Sorry for the delay here, this is a tricky one

wolmir: I found the issue and added a (currently) failing test for it here: https://github.com/iterative/vscode-dvc/pull/2643/files#r1004679812

wolmir: I'll start fixing the issues there today

wolmir:
@sroy3 @julieg18 Could you take another look here, please?

Mostly I fixed the issues in #2643 which will be merged here first, and then this one

mattseddon:
Can come back to this at a later date

sroy3:
I'll fix what's missing. With what's been merged back from the other PR, there shouldn't be much left.

sroy3:
Looks like everything is complete now. I'll dismiss Julie's review since she won't be able to re-review anytime soon, but I'll wait for another approval to merge.

Issue #2569, opened 2022-10-12
DVC tab endlessly loading


gcoter:
Hello 🙂 I would like to use the DVC's VS Code extension (I have version v0.5.0 installed currently) on one of my projects. I upgraded DVC (because my current version was not compatible), here are the details about my environment (note that I am using WSL):

```
dvc version
DVC version: 2.30.0 (pip)
---------------------------------
Platform: Python 3.8.10 on Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.29
Subprojects:
        dvc_data = 0.17.1
        dvc_objects = 0.7.0
        dvc_render = 0.0.12
        dvc_task = 0.1.3
        dvclive = 0.12.0
        scmrepo = 0.1.1
Supports:
        http (aiohttp = 3.8.3, aiohttp-retry = 2.8.3),
        https (aiohttp = 3.8.3, aiohttp-retry = 2.8.3),
        ssh (sshfs = 2022.6.0)
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/sdc
Caches: local
Remotes: ssh
Workspace directory: ext4 on /dev/sdc
Repo: dvc, git
```

With this setup, I am not able to use the extension because the view in the tab keeps on loading endlessly:

![image](https://user-images.githubusercontent.com/17814966/195285800-571274b7-2fc4-49e4-8d42-c9b43912c4b0.png)

Could you help me?

gcoter:
Actually, I might have a beginning of explanation. When I run `dvc exp show` in a terminal, it asks me for the SSH password of the remote. I guess this is why the experiment view gets stuck like this?

Still, I don't understand why `dvc exp show` needs to connect to the remote. Is it not supposed to show me the experiments which are present locally?

mattseddon:
Can you please share the output from `exp show --json`.

gcoter:
Hi @mattseddon, thanks for your answer and your help 🙂 Unfortunately I cannot share the output of `dvc exp show --json` because it contains information related to a project that I work on for my company 😅 Is there something specific that you would like to look for?

I didn't know that a `train` stage was expected, why is it the case?

gcoter:
Hi @mattseddon, I saw that a new version was released with your fix (`v0.5.1`), I got the update in my VS Code so I was able to test it. Unfortunately, I still have the same issue 😕

julieg18:
[...]

PR #2598, opened 2022-10-15
Fix experiment table changes collection bug

mattseddon:


extension/src/experiments/columns/collect/index.ts
```diff
@@ -49,22 +50,16 @@ export const collectColumns = (data: ExperimentsOutput): Column[] => {
...
+  baseline?: ExperimentFieldsOrError
+}): ExperimentFields | undefined => value?.baseline?.data
 
 export const collectChanges = (data: ExperimentsOutput): string[] => {
   const changes: string[] = []
 
-  let workspace
-  let currentCommit
-
-  for (const [key, value] of Object.entries(data)) {
```
mattseddon, line 61: Noticed this issue whilst staring at the code for https://github.com/iterative/vscode-dvc/issues/2569#issuecomment-1279053085.

This would not return the correct data now there are > 2 entries in the `exp show` data. It would actually return the oldest commit

extension/src/experiments/columns/collect/metricsAndParams.ts
```diff
@@ -109,7 +110,7 @@ const collectChange = (
   commitData: ExperimentFields,
   ancestors: string[] = []
 ) => {
-  if (typeof value === 'object') {
+  if (!Array.isArray(value) && typeof value === 'object') {
```
mattseddon, line 113: [F] If we get to an array we want to compare the array, not try to walk it.

sroy3 (approved):

Issue #2540, opened 2022-10-07
Hard to distinguish stared and un-stared exps


maxagin:
>Also, it is hard to distinguish stared and un-stared exps (un-stared icons look too solid 🙂). Some design change is needed. - Dmitry https://www.notion.so/iterative/Dmitry-s-Codespace-talk-feedback-1e1cfb4f0ac34ba7a1daaca20c369b08#a4d2b5008b8b4fdf863b0d5d794d437b

>This happened after one of the latest updates. Here is how it is supposed to look. Basically what is important, is clearly see the difference between stared and un-stared EXPs. 
For some reason, I am unable to share a screenshot, so here is a link with an example https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1807%3A124133 - Maxim

Related - https://github.com/iterative/vscode-dvc/issues/2241#issuecomment-1244479064

mattseddon:
Feedback has been cherry-picked. This is the more general feedback on stars:

> Stars - it’s a smart idea limit “labels” (in wandb and mlflow) by stars only (a single label). But I have not find it useful at all 😟 What is the expected flow? Should I use it in filters? Filters are complicated experience (in zoom-in mode, not Studio mode) - see above.

maxagin:
Here is the answer I have posted on the Notion:

> Stars - it’s a smart idea limit “labels” (in wandb and mlflow) by stars only (a single label). But I have not find it useful at all 😟 What is the expected flow?...

>> The main purpose is to be able to mark interesting experiments so you can find them faster.

shcheklein:
I think it's a fair feedback and includes multiple things from simple (this ticket - bad style decision, hard to distinguish) to more complicated UX questions (toggle / filter only starred experiments). 

My take - both things can and should be addressed. I would probably make star stand out (yellow) and grey out inactive stars. See GMail experience.

maxagin:
My two cents. The screenshot below is what we had before. The yellow - maybe (not sure we need more accent than on the screenshot)

<img width="316" alt="Screen Shot 2022-10-08 at 10 01 34 PM" src="https://user-images.githubusercontent.com/98249521/194734146-85a7f147-586f-4c55-a341-fd8e78991772.png">

shcheklein:
Closed by #2604

PR #2601, opened 2022-10-16
Improve star icon contrast

shcheklein:
Closes https://github.com/iterative/vscode-dvc/issues/2540

The best (to my mind) stable among different themes combination that I think better highlights the action. It's the same color as icon in the header:

https://user-images.githubusercontent.com/3659196/196015855-2b31d3a0-486b-4da1-bf64-acc221fac521.mov

Open to trying other things. I've tried @maxagin suggestion, but it's now becomes too light to my mind, though blends nicely with the checkbox and button:

Screen Shot 2022-10-15 at 8 02 00 PM
Screen Shot 2022-10-15 at 8 02 16 PM

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -510,10 +510,6 @@ $bullet-size: calc(var(--design-unit) * 4px);
           justify-content: center;
           width: 14px;
           cursor: pointer;
-
-          &[aria-checked='true'] {
-            opacity: 1;
-          }
         }
 
         .indicatorIcon {
@@ -559,11 +555,12 @@ $bullet-size: calc(var(--design-unit) * 4px);
 }
 
 .experimentCell .rowActions .starSwitch svg {
-  fill: $row-action-icon-color;
+  fill: $row-action-star;
+}
 
-  .rowSelected & {
-    fill: $row-action-selected-icon-color;
-  }
+.experimentCell .rowActions .starSwitch[aria-checked='true'] svg {
+  fill: $row-action-star-checked;
+  opacity: 0.8;
 }
 
 .cellHintTooltip {
--- a/webview/src/shared/variables.scss
+++ b/webview/src/shared/variables.scss
@@ -36,6 +36,9 @@ $row-action-selected-icon-color: var(
   var(--vscode-icon-foreground)
 );
 
+$row-action-star: var(--vscode-descriptionForeground);
+$row-action-star-checked: var(--vscode-editorLightBulb-foreground);
+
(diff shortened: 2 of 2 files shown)


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -559,11 +555,12 @@ $bullet-size: calc(var(--design-unit) * 4px);
 }
 
 .experimentCell .rowActions .starSwitch svg {
```
mattseddon, line 557: How about leaving the filled icon color as is and using the checkbox-border for the unfilled one?

```
.experimentCell .rowActions .starSwitch svg {
  fill: $checkbox-border;
}

.experimentCell .rowActions .starSwitch[aria-checked='true'] svg {
  fill: $row-action-icon-color;
}
```

https://user-images.githubusercontent.com/37993418/196055811-3ba77c37-29a2-4f72-a4da-eec211f5d148.mov

IMO the yellow looks out of place wrt the rest of the UI.

shcheklein: Yep, I was trying that alternative also.

I think the star action itself is not as visible enough then. I admit that I like it visually more, but I think we are sacrificing usability to aesthetics in that case.

> IMO the yellow looks out of place wrt the rest of the UI.

Yes, but it's super common UX pattern + if we keep star color a bit more aggressive, we have to do fill color more aggressive as well.

mattseddon: > I admit that I like it visually more, but I think we are sacrificing usability to aesthetics in that case.

My opinion is that both are just as usable as one another but one looks better. 

I am not prepared to fight to the death about it.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -559,11 +555,12 @@ $bullet-size: calc(var(--design-unit) * 4px);
...
-  fill: $row-action-icon-color;
+  fill: $row-action-star;
+}
 
-  .rowSelected & {
-    fill: $row-action-selected-icon-color;
-  }
+.experimentCell .rowActions .starSwitch[aria-checked='true'] svg {
```
sroy3, line 561: Does it need that specificity to apply the rule? I'd simply change it to `.starSwitch[aria-checked='true'] svg {` otherwise so that if we ever get a variant, we can override the rules without adding to much specificity to it.

sroy3: I just saw `.experimentCell .rowActions .starSwitch svg {` and that's why this selector is that long. If it's possible to cut them both. It's easier to change the markup afterward.

sroy3 (approved):

PR #2603, opened 2022-10-16
Disable Left Click trigger for Context Menu

wolmir:
This PR will disable the left-click trigger for the context menu, as part of https://github.com/iterative/vscode-dvc/issues/2256

It also moves the header components to a sub-folder to help navigation a bit.


webview/package.json
```diff
@@ -7,6 +7,7 @@
     "lint:prettier": "prettier -c '**/*.{js,json,jsx,md,scss,ts,tsx,yaml,yml}'",
     "lint:eslint": "eslint --cache '**/*.{js,ts,jsx,tsx}'",
     "lint:tsc": "tsc --noEmit",
+    "lint:webpack": "webpack --mode development",
```
sroy3, line 10: What is the use of this new task?

sroy3 (approved):

webview/package.json
wolmir: Forgot to mention it in the description.

The first time I moved the files VSCode didn't automatically fix the styles.scss imports, and the error wasn't caught by the commit hook because it relied only on the typescript compiler which was ok with it.

So I decided to add this to the linting tasks to catch errors like those locally.

sroy3: Cool. Can you link it to the root of the project (add a task in the root package.json that calls this task)? It'll be easier to reuse it that way if we don't have to `cd` into the webview folder.

mattseddon: Please move this unrelated change out, we need to be able to come back and find the reason for the change from the PR.

wolmir: Will do, thanks!

sroy3:
I'll fix what's missing. With what's been merged back from the other PR, there shouldn't be much left.

sroy3:
Re-opened the wrong PR

PR #2604, opened 2022-10-16
Update experiment table icons to use --vscode-descriptionForeground and use --vscode-editorLightBulb-foreground for selected stars

mattseddon:
### Screenshot 

Screen Shot 2022-10-18 at 8 46 38 am (2)

See inline comments for discussion.

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -18,7 +18,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
   left: 0;
   display: inline-block;
   position: absolute;
-  border: 2px solid $row-action-icon-color;
+  border: 2px solid $icon-color;
   border-top: unset;
   border-left: unset;
   transition: transform 0.2s, right 0.2s, bottom 0.2s;
@@ -101,7 +101,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
   background: $checkbox-background;
   border-radius: $bullet-size;
   border: calc(var(--border-width) * 1px) solid $checkbox-border;
-  color: $row-action-icon-color;
+  color: $icon-color;
   cursor: pointer;
   display: flex;
   height: $bullet-size;
@@ -134,7 +134,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
       width: 4px;
       height: 4px;
       vertical-align: middle;
-      border: 1px solid $row-action-icon-color;
+      border: 1px solid $icon-color;
       background-color: $checkbox-background;
     }
 
@@ -162,7 +162,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
 }
 
 .queued {
-  color: $row-action-icon-color;
+  color: $icon-color;
   display: flex;
   height: 16px;
   left: 2px;
@@ -331,7 +331,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
 
       .expandedRowArrow,
       .contractedRowArrow {
-        border-color: $row-action-selected-icon-color;
+        border-color: $selected-icon-color;
       }
     }
   }
@@ -560,10 +560,10 @@ $bullet-size: calc(var(--design-unit) * 4px);
 }
 
 .experimentCell .rowActions .starSwitch svg {
-  fill: $row-action-icon-color;
+  fill: $icon-color;
 
   .rowSelected & {
-    fill: $row-action-selected-icon-color;
+    fill: $selected-icon-color;
   }
 }
 
@@ -766,7 +766,7 @@ $badge-size: 0.85rem;
     fill: $watermark-color;
   }
   &.indicatorWithCount svg {
-    fill: $fg-color;
+    fill: $icon-color;
   }
 }
 
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -505,18 +505,6 @@ $bullet-size: calc(var(--design-unit) * 4px);
           margin-right: 2px;
         }
 
-        .starSwitch {
-          display: inline-flex;
-          align-items: center;
-          justify-content: center;
-          width: 14px;
-          cursor: pointer;
-
-          &[aria-checked='true'] {
-            opacity: 1;
-          }
-        }
-
         .indicatorIcon {
           display: inline-flex;
           align-items: center;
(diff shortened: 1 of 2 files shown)


demo/dvc.lock
```diff
@@ -26,7 +26,7 @@ stages:
       md5: 0aa0e279e904572c28c1203ebfa41b29
       size: 310000
     - path: training_metrics
-      md5: 5abc8d18d6b57bd513acf95e55e2755a.dir
+      md5: 461c1325295e6288ae78cb45d329dbc0.dir
```
shcheklein, line 29: change is not needed?

mattseddon: sorry

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -101,7 +101,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
   background: $checkbox-background;
   border-radius: $bullet-size;
   border: calc(var(--border-width) * 1px) solid $checkbox-border;
-  color: $fg-color;
+  color: $row-action-icon-color;
```
shcheklein, line 104: Q: what does this change?

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -162,6 +162,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
 }
 
 .queued {
+  color: $row-action-icon-color;
```
shcheklein, line 165: it's not an action though?

Also, should it be aligned with the plots action and should we use `--checkbox-background`?

webview/src/experiments/components/table/styles.module.scss
mattseddon: The color of the bullets for unselected experiments. Makes them match unselected experiments in the experiments tree.

webview/src/experiments/components/table/styles.module.scss
mattseddon: The clock is an icon inside of the row action block. 

If we use `--checkbox-background` it will be almost invisible on most themes. Example: 

https://user-images.githubusercontent.com/37993418/196083307-f8211702-9cf7-4edb-b701-c2d1f7a5510b.mov

This color matches this icon to the icons shown in the sidebar. I would be happy to change to $checkbox-border if we used the same color for empty stars in https://github.com/iterative/vscode-dvc/pull/2601

shcheklein: `--vscode-descriptionForeground` for action icons in the header and for action icon in #2601
`--checkbox-border` + `--checkbox-background` for bullet and checkboxes
`--vscode-icon-foreground` - for chevron action and now for queue icons

Both `--vscode-icon-foreground` and `--vscode-descriptionForeground` are quite stable and more or less high contrast as far as I remember. We can probably use one of those only for all icons (stars and the header and chevron). Probably I picked the `--vscode-descriptionForeground` for stars since it had less contrast compared to the surrounding checkboxes and bullets. Can we use it for chevron - don't know, need to try.

For queue, I still make sense to use checkboxes (the border one if background is not visible). Otherwise it'll might stand out too much from bullets and checkboxes and stars also. `--vscode-icon-foreground` had the highest contrast of three options as far as I remember.

webview/src/experiments/components/table/styles.module.scss
shcheklein: okay, I see. Could you do the recording across all themes please?

To me it felt okay for them to "blend" with checkboxes in the table tbh.

shcheklein: btw, does it apply only to the internal circle?

mattseddon: > btw, does it apply only to the internal circle?

yes

mattseddon: https://user-images.githubusercontent.com/37993418/196088476-b195342e-c3f7-48b4-8a8e-59fcb5a8829f.mov

webview/src/experiments/components/table/styles.module.scss
mattseddon: Using `--vscode-icon-color` for all icons:

https://user-images.githubusercontent.com/37993418/196090395-67983a1e-a2bb-4abe-8ebd-a2d1605569b0.mov

Using `--checkbox-border`

https://user-images.githubusercontent.com/37993418/196091286-85463add-69d2-447e-8b2f-f97b55ab0d46.mov

shcheklein: How about `--vscode-descriptionForeground`, the current one for actions in the header ? (clearly `--checkbox-border` is not visible enough, that's why I didn't pick if for stars).

mattseddon: The thing that bothers me is the discrepancy with the tree. We could change the circle color to match quite easily but the clock would be more difficult:

https://user-images.githubusercontent.com/37993418/196102189-233697cc-fec1-4c60-8538-4effff2d565f.mov

sroy3 (approved):

webview/src/experiments/components/table/styles.module.scss
shcheklein: Thanks, Matt. Look reasonable to me. Checkboxes, stars, bullets blend nicely. WDYT?

I'm less worried about the tree tbh. And to be fair, we have a very large discrepancy with the checkboxes, with the action icons, etc (those things in the table that drive us towards _not_ using the `--vscode-icon-color` in the first place). We should then update those? We'll face some other issues - e.g. in some themes webview has a different background / style compared to the side panel.

Btw, on this screenshot:

Screen Shot 2022-10-17 at 9 58 40 AM

there are a few issues (I can find and fix them later on my own if it takes time) - plots icon color is wrong and chevron ideally should also have the same color as the stars / checkboxes / other icons (so that we can get rid of extra styles at least).

mattseddon: > Thanks, Matt. Look reasonable to me. Checkboxes, stars, bullets blend nicely. WDYT?

I keep coming back to [this comment](https://github.com/iterative/vscode-dvc/pull/2601#discussion_r996490558):

> I think the star action itself is not as visible enough then. I admit that I like it visually more, but I think we are sacrificing usability to aesthetics in that case.

I think we should use the icon color for all the icons. We can use that horrible yellow for the star to make it more obvious when it is selected.

shcheklein: > I think we should use the icon color for all the icons. We can use that horrible yellow for the star to make it more obvious when it is selected.

Just to clarify that I understand you correctly. so, in your green theme example - all of the object will become bright white - bullets, actions in the header, chevron, queue icons ... it feels a bit aggressive for the table to me (since it's not a small set of icons, but rather a repeating pattern per row) + checkboxes will in the complete shadow because of this.

I'm fine with the the way it looks with the `--vscode-descriptionForeground` . Let's proceed with this please then.

mattseddon: Have updated to this:

Screen Shot 2022-10-18 at 8 46 38 am (2)


Will update the PR description, merge this and close #2601 as this contains that change as well.

shcheklein (approved):

PR #2617, opened 2022-10-18
Account for null being an object in workspace change data collection

mattseddon:
Closes https://github.com/iterative/vscode-dvc/issues/2569

See comment inline.

Notes: 

I had included a new test fixture as part of this change. I think that https://dagshub.com/kingabzpro/kaggle-titanic-dvc will be handy in the future (real-world project). It was the first project from https://github.com/iterative/awesome-iterative-projects#real-world-projects that gave the same error as shown in #2569. 
There were a few takeaways from looking at these projects that we can talk about tomorrow.

### DIFF
--- a/extension/src/experiments/columns/collect/metricsAndParams.ts
+++ b/extension/src/experiments/columns/collect/metricsAndParams.ts
@@ -110,7 +110,7 @@ const collectChange = (
   commitData: ExperimentFields,
   ancestors: string[] = []
 ) => {
-  if (!Array.isArray(value) && typeof value === 'object') {
+  if (value && !Array.isArray(value) && typeof value === 'object') {
     for (const [childKey, childValue] of Object.entries(value as ValueTree)) {
       collectChange(changes, type, file, childKey, childValue, commitData, [
         ...ancestors,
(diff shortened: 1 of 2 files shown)


extension/src/experiments/columns/collect/metricsAndParams.ts
```diff
@@ -110,7 +110,7 @@ const collectChange = (
   commitData: ExperimentFields,
   ancestors: string[] = []
 ) => {
-  if (!Array.isArray(value) && typeof value === 'object') {
+  if (value && !Array.isArray(value) && typeof value === 'object') {
```
mattseddon, line 113: Yep, `typeof null === 'object'`

image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#difference_between_null_and_undefined

extension/src/test/fixtures/expShow/kaggleDvcTitanic.ts
```diff
@@ -0,0 +1,1227 @@
+import { ExperimentStatus } from '../../../cli/dvc/contract'
```
mattseddon, line 1: I think we will want to make a story out of this:

Screen Shot 2022-10-18 at 2 42 47 pm

I will rearrange the fixtures.

shcheklein (approved):

PR #2621, opened 2022-10-18
Improve max table depth test

julieg18:
* use ConfigKey variable for config name
* reset workspace configuration with an after hook
* delete unneeded await

### DIFF
--- a/extension/src/test/suite/experiments/index.test.ts
+++ b/extension/src/test/suite/experiments/index.test.ts
@@ -67,6 +67,7 @@ import { shortenForLabel } from '../../../util/string'
 import { GitExecutor } from '../../../cli/git/executor'
 import { WorkspacePlots } from '../../../plots/workspace'
 import { RegisteredCommands } from '../../../commands/external'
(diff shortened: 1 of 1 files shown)


mattseddon (approved):

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -172,6 +173,12 @@ suite('Experiments Test Suite', () => {
   })
 
   describe('handleMessageFromWebview', () => {
+    afterEach(() => {
```
julieg18, line 176: I'm not sure why but using `after` instead of `afterEach` [threw an error on testing](https://github.com/iterative/vscode-dvc/actions/runs/3276627871/jobs/5393409165#step:8:211)

mattseddon: probably not imported

julieg18: > probably not imported

Whoops, forgot to import 😅

PR #2624, opened 2022-10-18
Demo project as a submodule

### LINKED ISSUE #2584: Change demo project to a submodule
We would like the demo repo to be more independent from the main repo. We would like to be able to commit inside of it without it affecting the rest of the project.

A good solution for this would be to convert it to a submodule.

We could also add multiple submodules if needed.

We previously moved away from submodules (#82) so we need to make sure the side effects are minimal.

sroy3:
# 1/2 `main` <- this <- #2627

Closes #2584


extension/src/cli/git/constants.ts
```diff
@@ -1,11 +1,42 @@
...
+
+export const getGitPath = (gitRoot: string, path: gitPath | string) => {
+  const gitDir = gitRootGitDir[gitRoot] || getGitDirPath(gitRoot)
+  gitRootGitDir[gitRoot] = gitDir
+
+  if (path === gitPath.DOT_GIT) {
+    return gitDir
+  }
+
+  return join(gitDir, ...path.split('/'))
```
mattseddon, line 38: should be able to use `resolve` and not have to split the path (`'/'` won't work on windows).

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vscode-dvc-demo"]
+	path = vscode-dvc-demo
```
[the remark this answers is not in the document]
sroy3: You're right, it's not because the repo is called vscode-dvc-demo that the path needs to be called the same thing. That'll definitely simplify the diff. I'll make the change.

extension/src/experiments/model/filterBy/tree.test.ts
```diff
@@ -59,7 +59,7 @@ describe('ExperimentsFilterByTree', () => {
       mockedExperiments,
       mockedInternalCommands
     )
-    mockedGetDvcRoots.mockReturnValueOnce(['demo'])
+    mockedGetDvcRoots.mockReturnValueOnce(['vsocde-dvc-demo'])
```
mattseddon, line 62: vsocde-dvc-demo is not correct

extension/src/cli/git/constants.ts
sroy3: `split` is used only on (`GIT_LOGS_REFS` specifically):
```
export enum gitPath {
  DOT_GIT = '.git',
  DOT_GIT_HEAD = 'HEAD',
  DOT_GIT_INDEX = 'index',
  GIT_REFS = 'refs',
  GIT_LOGS_REFS = 'logs/ref',
  HEADS_GIT_REFS = 'heads'
}
```

extension/src/fileSystem/index.ts
```diff
@@ -63,6 +64,35 @@ export const findAbsoluteDvcRootPath = async (
...
+      .split(/\r?\n/)
+      .find(line => line.indexOf(gitDirPrefix) === 0)
+    return resolve(gitRoot, ...(gitDirLine?.slice(8).split('/') || []))
+  }
+  return dotGitPath
+}
+
+const gitRootGitDir: { [key: string]: string } = {}
+
+export const getGitPath = (gitRoot: string, path: string) => {
```
mattseddon, line 85: [F] Moved this out of constants as it is reading the file system. We could probably even afford to drop `gitRootGitDir` if we wanted to.

extension/src/fileSystem/index.ts
```diff
@@ -63,6 +64,35 @@ export const findAbsoluteDvcRootPath = async (
...
+  const dotGitPath = join(gitRoot, gitPath.DOT_GIT)
+
+  const isSubmodule = lstatSync(dotGitPath).isFile()
+  if (isSubmodule) {
+    const dotGitAsFileContent = readFileSync(dotGitPath, 'utf8')
+    const gitDirPrefix = 'gitdir: '
+    const gitDirLine = dotGitAsFileContent
+      .split(/\r?\n/)
+      .find(line => line.indexOf(gitDirPrefix) === 0)
+    return resolve(gitRoot, ...(gitDirLine?.slice(8).split('/') || []))
```
mattseddon, line 78: [F] I was slightly concerned about this on Windows so wrote a test. It seems to behave as expected on all platforms

mattseddon: The reason that I switched from `join` to `resolve` is that I don't want any `..` in the path. It could end giving the watcher the wrong type of `RelativePath`

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "demo"]
+	path = demo
+	url = https://github.com/iterative/vscode-dvc-demo
```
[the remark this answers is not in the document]
mattseddon: I had to move the already created folder with git mv and then change the .git file and move the module under the root .git

extension/src/cli/git/executor.test.ts
```diff
@@ -38,9 +39,10 @@ describe('GitExecutor', () => {
     } as unknown as EventEmitter<CliStarted>
   })
 
+  const cwd = standardizePath(__dirname) as string
```
mattseddon, line 42: [F] Followed up and split this function `standardizePath(something) as string` is everywhere and dumb.

mattseddon:
@sroy3  please review, and feel free to update too.

mattseddon (approved):

package.json
```diff
@@ -25,7 +25,7 @@
     "build": "yarn turbo run package",
     "install-frozen-lockfile": "./scripts/install-frozen-lockfile.sh",
     "dev-server": "yarn turbo run dev --parallel",
-    "postinstall": "husky install",
+    "postinstall": "husky install && git submodule init && git submodule update",
```
mattseddon, line 28: maybe we should do yarn setup:venv in here too. WDYT?

sroy3: Good idea. Just one step to get up and running.

sroy3 (commented):
Works great. Thanks! (Can't approve because I'm owner of the PR, but ✅ )

julieg18 (approved):
Great work!

package.json
mattseddon: I'll have more of a think about it and follow up.

PR #2626, opened 2022-10-18
Add webpack dummy build as a lint step

wolmir:
This PR will add webpack --mode development as a linting step.
The objective is to catch bundling errors that tsc --no-emit ignores, like styles.scss imports, for example.


webview/package.json
```diff
@@ -7,6 +7,7 @@
     "lint:prettier": "prettier -c '**/*.{js,json,jsx,md,scss,ts,tsx,yaml,yml}'",
     "lint:eslint": "eslint --cache '**/*.{js,ts,jsx,tsx}'",
     "lint:tsc": "tsc --noEmit",
+    "lint:webpack": "webpack --mode development",
```
mattseddon, line 10: [Q] Do we need `lint:tsc` if we have this? Could we change to `lint:compile` for all sub-projects? Leave `extension` as is but change the two sub-projects bundled by webpack to use this?

wolmir: That's a good point, thanks! I'll add lint:compile to the two webpacked projects and remove lint:tsc from them

sroy3 (approved):

julieg18 (approved):
Looks good to me!

PR #2628, opened 2022-10-19
Fix integration test suite after hook

mattseddon:
Comments inline


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -173,11 +173,11 @@ suite('Experiments Test Suite', () => {
   })
 
   describe('handleMessageFromWebview', () => {
-    after(() => {
+    after(() =>
       workspace
-        .getConfiguration(ConfigKey.EXP_TABLE_HEAD_MAX_DEPTH)
```
mattseddon, line 178: [F] This was not behaving as expected. It kept leaving behind the config option after the suite had run. If you take a look at the other usage that we have of `workspace.getConfiguration()` it always looks like the new usage.

extension/src/vscode/config.ts
```diff
@@ -5,7 +5,7 @@ export enum ConfigKey {
   DO_NOT_SHOW_CLI_UNAVAILABLE = 'dvc.doNotShowCliUnavailable',
   DO_NOT_SHOW_WALKTHROUGH_AFTER_INSTALL = 'dvc.doNotShowWalkthroughAfterInstall',
   DO_NOT_SHOW_UNABLE_TO_FILTER = 'dvc.doNotShowUnableToFilter',
-  EXP_TABLE_HEAD_MAX_DEPTH = 'dvc.experimentsTableHeadMaxHeight',
+  EXP_TABLE_HEAD_MAX_HEIGHT = 'dvc.experimentsTableHeadMaxHeight',
```
mattseddon, line 8: [F] Removed the old reference to depth

sroy3 (approved):

julieg18 (approved):
Looks good!

Issue #2581, opened 2022-10-12
Plot updates are slow and are fetching information that is not needed


shcheklein:
(video is too large to embed, sorry about that)

https://www.dropbox.com/s/1nhipdskbnzchyn/Screen%20Recording%202022-10-12%20at%202.19.43%20PM.mov?dl=0

Two things:

- When we start a new experiment it keeps showing old plots for a while. Can we grey out / force remove if we know that update is coming? (need to do some research there, e.g. resume an experiment mode means that we should just continue a line in the same color)
- Plots update command takes too much time
- Plots update command runs with extra arguments for no reason. When workspace is selected it runs with:

```bash
dvc plots diff workspace 7a43153 46bd8b3 9a83545 -o .dvc/tmp/plots --split --json - COMPLETED (100245ms)
```

where `7a43153`, `46bd8b3`, `9a83545` the HEAD of the branch and two previous commits.

shcheklein:
The last item can be narrowed down to this:

https://github.com/iterative/vscode-dvc/blob/c83e76367f7ea3159acd9adfc5474b33b7269a01/extension/src/plots/model/index.ts#L204-L206

and this PR:

https://github.com/iterative/vscode-dvc/pull/1949/files

One the issue is fixed now. 

Btw, using the `workspace` revision alone gives me an error (I would expect this, since `workspace` is not an actual Git revision):

```
plots,diff,workspace,-o,.dvc/tmp/plots,--split,--json failed with ERROR: unknown Git revision 'workspace' retrying...
```

mattseddon:
> When we start a new experiment it keeps showing old plots for a while. Can we grey out / force remove if we know that update is coming? (need to do some research there, e.g. resume an experiment mode means that we should just continue a line in the same color)
> Plots update command takes too much time

This should all be fixed when we come back around to do errors.

shcheklein:
> This should all be fixed when we come back around to do errors.

@mattseddon could you clarify please? Not sure how this is related to the errors tbh.

mattseddon:
> > This should all be fixed when we come back around to do errors.
> 
> @mattseddon could you clarify please? Not sure how this is related to the errors tbh.

We need to redesign the way that we are interacting with the CLI to get the plots data in order to get errors. When we do this we can start to optimise the code.

shcheklein:
I think those things are related, but probably we can reset things on our end also forcefully when new experiments starts in the workspace to mitigate this?

[...]

PR #2629, opened 2022-10-19
Only patch workspace only calls to plots diff

mattseddon:
Address one of the points from https://github.com/iterative/vscode-dvc/issues/2581

> Plots update command runs with extra arguments for no reason. 

From https://github.com/iterative/vscode-dvc/issues/2581#issuecomment-1276939733:

> The last item can be narrowed down to this:
> 
> https://github.com/iterative/vscode-dvc/blob/c83e76367f7ea3159acd9adfc5474b33b7269a01/extension/src/plots/model/index.ts#L204-L206
> 
> and this PR:
> 
> https://github.com/iterative/vscode-dvc/pull/1949/files
> 
> One the issue is fixed now.
> 
> Btw, using the `workspace` revision alone gives me an error (I would expect this, since `workspace` is not an actual Git revision):
> 
> ```
> plots,diff,workspace,-o,.dvc/tmp/plots,--split,--json failed with ERROR: unknown Git revision 'workspace' retrying...
> ```

I am reluctant to unwind any of the changes made in https://github.com/iterative/vscode-dvc/pull/1949 right now and they are unrelated to this particular change.

In order to get significant performance increases we will need to re-do this sub-system and we will need help from the CLI.


sroy3 (approved):

julieg18 (approved):
Looks good to me!

PR #2630, opened 2022-10-19
Switch integration tests back to insiders

mattseddon:
I think I understand what happened now:

- VS Code started logging the entire `cpuprofile` whenever it gets run.
- Integration coverage causes slow startup (especially on the CI)
- Slow startup trigger `cpuprofile`
- Full `cpuprofile` gets logged
- Parent process crashes with OOM
- Sad face

Insiders build should be working again tomorrow.


mattseddon:
I think that this: https://github.com/microsoft/vscode/pull/163987 is what is causing the crash.

sroy3 (approved):

julieg18 (approved):
Great work on debugging the issue!

mattseddon:
Seems broken again. Will leave on stable for the time being.

PR #2632, opened 2022-10-19
Do not toggle plot section when clicking a link

sroy3:
https://user-images.githubusercontent.com/3683420/196721711-fcef4026-4d91-4360-8396-17129f787224.mov


sroy3:
You were right @shcheklein, `stopPropagation` had some weird effect. Either it was stringified or changed the context of the link.

julieg18 (approved):
Good job!

shcheklein (commented):
Thanks @sroy3 for improving the experience! From what I can see we probably need to test that element doesn't belong the tooltip container (?) to completely fix this (now it's still reacts if you click on the tooltip (not link)).

sroy3:
> Thanks @sroy3 for improving the experience! From what I can see we probably need to test that element doesn't belong the tooltip container (?) to completely fix this (now it's still reacts if you click on the tooltip (not link)).

Will try doing it

Issue #2539, opened 2022-10-07
Cleanup `Plot` & `Plot and Show` in the context menu


maxagin:
## Improvements

- Remove `Plot`
- Rename `Plot and Show` to `Plot`
- The context menu in case of a single EXP should have a `Plot` option.
- This also needs to work if the user just right-clicks on the EXP (without selecting it). 
- The right-click on EXP does not need to remove the row hover style (if possible).
-  If more than 7 EXPs are selected grey out `Plot` (always grey out the inactive options. For education the menu can show what is available).
- Make behavior additive
- Context menu on ESC don't remove checkboxes

## Consider

- Add unplot option (was #2448)

mattseddon:
> Should we remove Plot/Plot and show from the context menu if more than 7 EXPs are selected? Need a good solution here.

The current behaviour is to only plot the multi-selected rows.

shcheklein:
> Should we remove Plot/Plot and show from the context menu if more than 7 EXPs are selected? 

No, I think we need to grey out inactive options always. Menu can show what is available (education).

> Right-clicking on EXP shows the context menu with options: Plot / Plot and Show. I suggest removing the Plot option, as not essential.

Make sense. We can even rename it to `Plot` and make it default behavior to active the Plot webview.

maxagin:
I have updated the description.

math-alpha:
Hey @maxagin I would love to tackle this one

shcheklein:
@math-alpha please give a try, let us know if you need a hand.

math-alpha:
I am unsure what the "unplot" menu is supposed to do.
Is that unselecting from the experiments?

shcheklein:
@math-alpha it should be doing the opposite from what the Plot option is doing (enables plots for an experiment(s) on the Plots view).

math-alpha:
Update

- [x] Remove Plot
- [x] Rename Plots and Show to Plot
- [x] The context menu in case of a single EXP should have a Plot option.
- [x] This also needs to work if the user just right-clicks on the EXP (without selecting it).
- [ ] The right-click on EXP does not need to remove the row hover style (if possible).
- [ ] If more than 7 EXPs are selected grey out Plot (always grey out the inactive options. For education the menu can show what is available).
- [x] Add unplot option (was https://github.com/iterative/vscode-dvc/issues/2448)

Issue #2448, opened 2022-09-23
Add "unplot" context menu


shcheklein:
I would expect to have it here:

<img width="1159" alt="Screen Shot 2022-09-22 at 6 13 40 PM" src="https://user-images.githubusercontent.com/3659196/191876298-4f0530b5-2070-480a-91d3-b210796c0eb8.png">

mattseddon:
Merged with #2539

PR #2633, opened 2022-10-19
Ngadou 2539 plot and show in the context menu

math-alpha:
Code for #2539 

- [x] Remove Plot
- [x] Rename Plots and Show to Plot
- [x] The context menu in case of a single EXP should have a Plot option.
- [x] This also needs to work if the user just right-clicks on the EXP (without selecting it).
- [x] Add unplot option (was https://github.com/iterative/vscode-dvc/issues/2448)

To do
- [ ] The right-click on EXP does not need to remove the row hover style (if possible).
- [ ] If more than 7 EXPs are selected grey out Plot (always grey out the inactive options. For education the menu can show what is available).


mattseddon:
Hi @math-alpha,

Thanks for the contribution. 

You're our first-ever open-source contributor 🎉!

Please give us a little bit of time to set up the repository correctly so that you can run the CI.

Thanks

math-alpha:
@mattseddon any other recommended issue I can focus on?

webview/src/experiments/components/App.test.tsx
```diff
@@ -791,7 +791,12 @@ describe('App', () => {
       jest.advanceTimersByTime(100)
       const menuitems = screen.getAllByRole('menuitem')
       const itemLabels = menuitems.map(item => item.textContent)
-      expect(itemLabels).toStrictEqual(['Modify and Run', 'Modify and Queue'])
+      expect(itemLabels).toStrictEqual([
+        'Modify and Run',
+        'Modify and Queue',
+        'Plot',
+        'Unplot'
```
mattseddon, line 798: [I] We shouldn't show both of these at the same time. Especially when one of them will not have any effect. If the experiment is selected for plotting we should show "unplot" and visa versa.

webview/src/experiments/components/table/RowContextMenu.tsx
```diff
@@ -202,6 +212,20 @@ const getSingleSelectMenuOptions = (
       isWorkspace,
       !hasRunningExperiment
     ),
+    experimentMenuOption(
+      [id],
+      'Plot',
+      MessageFromWebviewType.SET_EXPERIMENTS_AND_OPEN_PLOTS,
+      false,
+      false
```
mattseddon, line 220: [I] The UX here is a bit broken. If I already have experiments selected and I select 'plot' for a single experiment I would expect for the action to add this experiment to the existing selections not replace them:

https://user-images.githubusercontent.com/37993418/197085608-6efac9c6-2736-4e40-ac9a-f9737bf4fed5.mov

This will also mean that if there are already 7 experiments selected that the option should be greyed out/unavailable.

webview/src/experiments/components/table/RowContextMenu.tsx
```diff
@@ -154,6 +162,8 @@ const getSingleSelectMenuOptions = (
   const isNotExperimentOrCheckpoint =
     isQueued(status) || isWorkspace || depth <= 0
 
+  status = isNotExperimentOrCheckpoint ? undefined : status
```
mattseddon, line 165: [Q] Why do we need to make this change?

webview/src/experiments/components/table/RowContextMenu.tsx
```diff
@@ -78,15 +86,15 @@ const getMultiSelectMenuOptions = (
     experimentMenuOption(
       selectedIds,
       'Plot',
-      MessageFromWebviewType.SET_EXPERIMENTS_FOR_PLOTS,
+      MessageFromWebviewType.SET_EXPERIMENTS_AND_OPEN_PLOTS,
```
mattseddon, line 89: [I] Following on from my previous comment about the 'Plot' action for a single experiment I think this should also be additive. If there are 3 experiments already selected then the multi-select option should add to those three experiments (not replace them) and be available up until it will select > 7 experiments. At which point it should be greyed out.

webview/src/shared/components/messagesMenu/styles.module.scss
```diff
@@ -17,6 +17,16 @@
   }
 }
 
+.disabledItem {
```
mattseddon, line 20: [C] Based on the previous comments you will need this but right now it is not required and should not have been included in the PR.

mattseddon (changes requested):
Thanks for your contribution!

> @mattseddon any other recommended issue I can focus on?

I would definitely keep your focus on this right now. There are a few behaviour changes needed before we'll be able to merge this. I've left some comments on the PR. We'll also need some new tests for the changes. 

Please come back to me with any questions that you have 🙏🏻.

math-alpha:
Hey, @mattseddon I am on the fixes. Thanks for raising them.

shcheklein:
hey, @math-alpha any updates on this by chance?

mattseddon:
Hey @math-alpha, it would be good to get this contribution fixed up and merged. 

Please LMK if there is anything at all that I can do to help you. Please also LMK if you not longer intend to work on this. 

🙏🏻

math-alpha:
> LMK

I have this issue with styling the "greying out" button when more than 7 experiments are selected

mattseddon:
> > LMK
> 
> I have this issue with styling the "greying out" button when more than 7 experiments are selected

To get started why not remove the option from the context menu under those circumstances?

math-alpha:
> > > LMK
> > 
> > 
> > I have this issue with styling the "greying out" button when more than 7 experiments are selected
> 
> To get started why not remove the option from the context menu under those circumstances?

Okay I will try that.

PR #2634, opened 2022-10-19
Do not run build twice in CI

mattseddon:
Follow up from #2626

The build was being run twice in the CI for each commit. I've moved things around so that turborepo takes care of caching it for us. 

It should also be noted that `lint:compile` is not run by the recommit hook. See

1. https://github.com/iterative/vscode-dvc/blob/8374d66258031349219d7923de97baf0cbcab6a5/.husky/pre-commit#L4
2.  https://github.com/iterative/vscode-dvc/blob/8374d66258031349219d7923de97baf0cbcab6a5/.lintstagedrc.js#L1


sroy3 (approved):

julieg18 (commented):
Looks good!

julieg18 (approved):

PR #2635, opened 2022-10-20
Rearrange exp show fixtures

mattseddon:
# 1/2 `main` <- this <- #2638 

This PR rearranges the `exp show` test fixtures so that we can insert a new "real world" fixture. The main reason for the new fixture is so that we can add a new story based on it.


julieg18 (approved):
Good work!

PR #2642, opened 2022-10-20
Do not toggle plot section on tooltip click

sroy3:
Update on #2632 (this [comment](https://github.com/iterative/vscode-dvc/pull/2632#pullrequestreview-1148374552))

https://user-images.githubusercontent.com/3683420/197005678-38ea0f9c-0c52-478b-879c-cb5406c28d9e.mov


webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -137,11 +136,25 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
     </div>
   )
 
+  const isTooltip = (el: Element) => {
+    let currentNode = el
+    while (currentNode.nodeName !== 'SUMMARY') {
```
shcheklein, line 141: just to be safe - should we check here for null/undefined (whatever it is in JS :) ). I mean, let's break the cycle if we got to the very top and still don't see summary.

sroy3: This function is called for the `onClick` inside of the `<summary />` element. The `<summary />` element is always the top node of this event. I'll add the `BODY` node name in the unexplained case it would go pass the `SUMMARY` one as it doesn't hurt, but I'm totally sure this can't happen.

webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -137,11 +136,25 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
     </div>
   )
 
+  const isTooltip = (el: Element) => {
+    let currentNode = el
+    while (!['SUMMARY', 'BODY'].includes(currentNode.nodeName)) {
+      if (
+        !!currentNode.className.toLocaleLowerCase &&
+        currentNode.className.toLocaleLowerCase().includes('tooltip')
```
julieg18, line 144: With the function checking for just the text `tooltip`, the tooltip icon also doesn't cause the section to toggle: 

https://user-images.githubusercontent.com/43496356/197218952-db72b2c5-ddb9-420c-9af8-ad4ba019c90c.mov

Was this change intended? If so, should we stop showing the mouse pointer when hovering over the icon?

julieg18 (approved):
Good work!

webview/src/plots/components/PlotsContainer.tsx
sroy3: It was intended, but I'll remove the pointer.

PR #2643, opened 2022-10-20
Refactor context-menu options in table header

wolmir:
This is a follow-up of #2593 which is the target branch for this

It will extend the sorting options to some column placeholders that have a single child column.
The objective is to prevent an empty placeholder to have different options than the column whose place it is holding.

This will also move the header menu content to its own component in an effort to simplify the context-menu logic a bit.


sroy3:
The actions do not work it triggered from the placeholder:

https://user-images.githubusercontent.com/3683420/197207133-83a54bc5-eb71-432a-b642-84c0cb59a0a8.mov

wolmir:
> The actions do not work it triggered from the placeholder

Thanks! I'll try to add a test for the placeholder case

extension/src/experiments/columns/model.test.ts
```diff
@@ -249,4 +249,62 @@ describe('ColumnsModel', () => {
...
+      await model.transformAndSet(outputFixture)
+      expect(model.getSelected()).not.toStrictEqual(columnsFixture)
+
+      model.toggleStatus(parentPath)
+
+      const children = model.getChildren(parentPath)
+      expectStatus(children, Status.SELECTED)
+    })
+
+    it('should unselect the toggled children of an unselected column', async () => {
```
wolmir, line 294: This is the failing test for https://github.com/iterative/vscode-dvc/pull/2593#discussion_r1001869678

It took some time to find the issue. I will start fixing them today

sroy3:
> The actions do not work it triggered from the placeholder:
> 
>  Screen.Recording.2022-10-21.at.9.27.23.AM.mov

I'm getting the same thing. Is that normal?

wolmir:
> > The actions do not work it triggered from the placeholder:
> > Screen.Recording.2022-10-21.at.9.27.23.AM.mov
> 
> I'm getting the same thing. Is that normal?

No, not normal, it's sending the placeholder's synthetic id instead of the actual column id as the message payload. I'll fix it, thanks!

wolmir:
@sroy3 @julieg18 Can you take another look here please?

julieg18 (commented):
Looks like `yarn lint` is failing. Can you correct the problem?

sroy3:
> > > The actions do not work it triggered from the placeholder:
> > > Screen.Recording.2022-10-21.at.9.27.23.AM.mov
> > 
> > 
> > I'm getting the same thing. Is that normal?
> 
> No, not normal, it's sending the placeholder's synthetic id instead of the actual column id as the message payload. I'll fix it, thanks!

It's still the same.

sroy3 (changes requested):

sroy3:
Fixed sorting on placeholders


https://user-images.githubusercontent.com/3683420/201367956-2bfca67b-b095-4326-a69e-3226e3478d16.mov

What's missing is hiding only columns underneath a parent, but I think we should do this as a follow-up

https://user-images.githubusercontent.com/3683420/201368426-c3527166-d799-4441-ac1e-8343bed82341.mov

sroy3 (approved):

PR #2653, opened 2022-10-21
Update packages to fix got vulnerability

sroy3:
Fixes got [vulnerability](https://github.com/iterative/vscode-dvc/security/dependabot/30)


sroy3:
Doesn't work. But this is a dev dependency and a moderate vulnerability.

PR #2677, opened 2022-10-26
bump latest DVC tested version to 2.31.0

shcheklein:
Before the release better to bump it if there are no objections.

### DIFF
--- a/extension/src/cli/dvc/constants.ts
+++ b/extension/src/cli/dvc/constants.ts
@@ -1,7 +1,7 @@
 import { join } from 'path'
 
 export const MIN_CLI_VERSION = '2.30.0'
-export const LATEST_TESTED_CLI_VERSION = '2.30.0'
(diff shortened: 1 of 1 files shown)


extension/src/cli/dvc/constants.ts
```diff
@@ -1,7 +1,7 @@
 import { join } from 'path'
 
 export const MIN_CLI_VERSION = '2.30.0'
-export const LATEST_TESTED_CLI_VERSION = '2.30.0'
+export const LATEST_TESTED_CLI_VERSION = '2.31.0'
```
julieg18, line 4: Along with updating the `LATEST_TESTED_CLI_VERSION`, we need to update our demo project `requirements.txt`. I'll go ahead and open a pr!

julieg18 (approved):

sroy3 (approved):

Issue #2585, opened 2022-10-13
Plots grid with plot resize


sroy3:
We want to redefine the plots grid and allow resizing of the plots (proposal can be found [here](https://www.notion.so/iterative/Resize-responsiveness-grid-for-Plots-WebView-39c16e0e53a149f09ffdd23b33950ec8)). 

We can split this in a couple of steps:

- [X] Change the internal structure to use numbers instead of the defined `PlotSize` enum. (#2563)
- [X] Allow resize of plot horizontally
- [X] Make the default height of plot 280px (since we have a 4:3 ratio for plots, that means 375px wide)
- [X] Remove the size picker
- [X] Fix resizing of last item(s) 
- [x] Allow resize of plot vertically
- [x] Resize of comparison table
- ~~Add a button to reset the size~~
- [x] Make scrollbars local to a section
- [X] Make sliders sticky
- ~~Calculate the default number of items per row from the screen size~~
- [x] Make section headers sticky

sroy3:
> * [ ]  Fix resizing of last item(s)

If you're dragging the corner of the last element and it resizes at the same time, you'll soon find yourself running out of space to resize while if you were resizing the first element instead, the other elements would just start moving to the next line.

@mattseddon proposed using a slider instead of dragging a corner of the plot to fix this issue. This way we can quickly see how the resize is affecting all plots, we could set a clear minimum and maximum, and steps would also be easier to get right.

shcheklein:
> @mattseddon proposed using a slider instead of dragging a corner of the plot to fix this issue. This way we can quickly see how the resize is affecting all plots, we could set a clear minimum and maximum, and steps would also be easier to get right.

we would need to keep a slider nearby each plot then though?

I feel it's still better to have a regular resize grip tbh. It's a thing that is common and expected vs something that is custom.

sroy3:
> > @mattseddon proposed using a slider instead of dragging a corner of the plot to fix this issue. This way we can quickly see how the resize is affecting all plots, we could set a clear minimum and maximum, and steps would also be easier to get right.
> 
> we would need to keep a slider nearby each plot then though?
> 
> I feel it's still better to have a regular resize grip tbh. It's a thing that is common and expected vs something that is custom.

I'd keep it in the top bar (a resize icon button + slider dropdown).

I agree it feels more natural to grip and resize, but we have to keep in mind that there are multiple edge cases regarding this solution.

[...]

PR #2680, opened 2022-10-26
Plot resize prototype

sroy3:
Should have been part of #2585 but as mentioned in the planning meeting yesterday, the plan was to change the number of items per row, not the exact size of the plot. Opening a PR for history and discussion purposes.

The prototype currently only works horizontally and still has a lot of bugs.

## Demo

https://user-images.githubusercontent.com/3683420/198063295-a2498f39-2d95-4a38-a31f-756e3e1068e2.mov


shcheklein:
@sroy3 could you please clarify, what is the difference with the https://github.com/iterative/vscode-dvc/issues/2585 - is it the same?

sroy3:
> Member

This prototype is the one I showed in the planning meeting. It resizes the plots as a grid to the exact plot size you moved the line to. This does not fill the line completely because an exact size does not always equal an exact number of item (see #2681).

The fact that I hadn't implemented[ snapping points ](https://github.com/iterative/vscode-dvc/pull/2681#issuecomment-1293558419)yet confused us everyone I think. I guess I'll just continue on the path I was on before the meeting.

PR #2694, opened 2022-10-28
Fix missing exp selection quick pick details on inital extension render

julieg18:
* has `getFirstThreeColumnOrder` collect column order from data if `columnOrder` is an empty array

Currently, our `columnOrder` variable is an empty array until a user moves a column. If a user attempts to select experiments for plots, but hasn't moved columns in the table, they won't see any experiment details.

**Current:**

image

**PR:**

image

### DIFF
--- a/extension/src/experiments/columns/model.ts
+++ b/extension/src/experiments/columns/model.ts
@@ -29,7 +29,9 @@ export class ColumnsModel extends PathSelectionModel<Column> {
   }
 
   public getFirstThreeColumnOrder(): string[] {
-    return this.columnOrderState.slice(1, 4)
+    return this.columnOrderState.length === 0
+      ? this.getFirstThreeColumnOrderFromData()
+      : this.columnOrderState.slice(1, 4)
   }
 
   public getColumnWidths(): Record<string, number> {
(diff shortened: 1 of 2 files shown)


extension/src/experiments/columns/model.ts
```diff
@@ -82,6 +84,13 @@ export class ColumnsModel extends PathSelectionModel<Column> {
     return this.data.length > 1
   }
 
+  private getFirstThreeColumnOrderFromData(): string[] {
+    return this.data
```
julieg18, line 88: This seemed like the easiest way to collect the missing `columnOrder` but is there a better way to fix this bug?

sroy3: Maybe we should just add a default `columnOrderState` if it's empty instead. We could do that on line 112. That way, we'll never have to check whether it's empty or not.

sroy3 (approved):

extension/src/experiments/columns/model.ts
julieg18: Good idea! From a little testing I did, it will take a bit of time to make that change so I'll update in a follow-up at a later date!

PR #2711, opened 2022-10-31
Improve exp selection quick pick description

julieg18:
* truncate key strings to a max of 15 length 
* add "," between list items

## Demo

https://user-images.githubusercontent.com/43496356/199125204-88a124c0-9b6f-4e50-bc50-1302f5a79e2d.mov

### DIFF
--- a/extension/src/experiments/model/quickPicks.ts
+++ b/extension/src/experiments/model/quickPicks.ts
@@ -13,6 +13,7 @@ import { Experiment } from '../webview/contract'
 import { Title } from '../../vscode/title'
 import { splitColumnPath } from '../columns/paths'
 import { formatDate } from '../../util/date'
+import { truncateFromLeft } from '../../util/string'
 
 type QuickPickItemAccumulator = {
   items: QuickPickItemWithValue<Experiment | undefined>[]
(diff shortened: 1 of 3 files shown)


extension/src/experiments/model/quickPicks.ts
```diff
@@ -38,10 +39,13 @@ const getItem = (experiment: Experiment, firstThreeColumnOrder: string[]) => ({
           ? formatDate(collectedVal)
           : collectedVal?.value || collectedVal
 
-      return `${splitUpName[splitUpName.length - 1]}:${value}`
+      return `${truncateFromLeft(
+        splitUpName[splitUpName.length - 1],
```
julieg18, line 43: While this does keep a user seeing all three values, the default way (truncating at the very end) allows a user to read the entire output on hover with a tooltip: 

image

Personally, I prefer the default vscode way, but what does everyone else think?

shcheklein:
@julieg18 looks great. One improvement is to also shorten float values, similar to the table. Otherwise it take quite a lot of space.

extension/src/experiments/model/quickPicks.ts
maxagin: Having all the values in one line is quite not readable :(

shcheklein: I think it's better to optimize for more values vs a tooltip. Means that we'll have to cut names. 

Also, is it possible to highlight names vs values in some way? (e.g. make the bold, or different color for values, etc?)

maxagin: An example

Screen Shot 2022-10-31 at 10 39 28 PM

sroy3 (approved):

extension/src/experiments/model/quickPicks.ts
julieg18: > Also, is it possible to highlight names vs values in some way? (e.g. make the bold, or different color for values, etc?)
> Having all the values in one line is quite not readable :(

I don't think it's possible to apply any markdown or highlighting to the details. We also can't add new lines to any of the text.

julieg18:
> @julieg18 looks great. One improvement is to also shorten float values, similar to the table. Otherwise it take quite a lot of space.

Good idea! Truncated float values and updated the demo!

PR #2738, opened 2022-11-07
Check hidden status when getting first three exp table column order

julieg18:
* has `ColumnModel`'s `getFirstThreeColumnOrder` check for the columns' hidden status, which stops our experiment tree tooltips and our exp selection quick pick from showing hidden column data. 

## Demo

https://user-images.githubusercontent.com/43496356/200365613-7f6af54f-5257-4ff0-b905-31adb0b550a6.mov

Followup to [#2706 comment ](https://github.com/iterative/vscode-dvc/pull/2706#pullrequestreview-1163374758)


extension/src/experiments/columns/model.ts
```diff
@@ -108,6 +110,40 @@ export class ColumnsModel extends PathSelectionModel<Column> {
...
+  }
+
+  private getColumnsFromType(type: string): string[] {
+    const childrenColumns: string[] = []
+    const dataWithType = this.data.filter(({ path }) => path.startsWith(type))
+    this.findChildrenColumns(type, dataWithType, childrenColumns)
+    return childrenColumns
+  }
+
+  private getColumnOrderFromData() {
```
julieg18, line 137: Originally, I just went through the columns data and collected all columns that don't have children, but that causes some discrepancies.

To make sure that the columns collected from the data were in the correct order, I ended up needing to write a couple of different functions. But maybe there is a less complex way 🤔

extension/src/experiments/workspace.ts
```diff
@@ -314,7 +314,7 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
     )
 
     experiments.dispose.track(
-      experiments.onDidChangeColumnOrder(() => {
+      experiments.onDidChangeColumnOrderOrStatus(() => {
```
julieg18, line 317: Since an `onDidChangeColumnOrder` and `onDidChangeColumnStatus` event does the same thing, I combined the two events into one. Though, we could split them into two events if preferred :)

extension/src/path/selection/model.ts
```diff
@@ -17,17 +17,22 @@ export abstract class PathSelectionModel<
 
   protected data: T[] = []
 
+  protected statusChanged?: EventEmitter<void>
```
julieg18, line 20: Made the `statusChanged` event optional since it's not used in the plots model.

julieg18 (commented):
~Looks like the CI tests are failing in different prs. Will mark this ready for review for now and can look into later today (if it's still not working, that is).~

extension/src/experiments/columns/model.ts
sroy3: So this is the original `columnOrderState` right? Why not set it to the once so that it doesn't need to get recalculated every time?

sroy3 (commented):
https://user-images.githubusercontent.com/3683420/200593316-988d8832-3143-47a5-aeca-d9c2f5881b21.mov

The empty tooltip looks a little weird

extension/src/experiments/columns/model.ts
julieg18: Done!

julieg18:
> The empty tooltip looks a little weird

Nice catch! Fixed :)

sroy3 (approved):

PR #2745, opened 2022-11-08
Fix exp tree tooltips and quick pick selection ignoring columns with falsy values

julieg18:
Currently our code will not show a column if its collected value is "falsy", which means it ignores columns with falsy values.

**Current:**

image

**PR:**

image

### DIFF
--- a/extension/src/experiments/model/util.ts
+++ b/extension/src/experiments/model/util.ts
@@ -1,7 +1,43 @@
 import get from 'lodash.get'
 import { formatDate } from '../../util/date'
+import { truncate } from '../../util/string'
 import { splitColumnPath } from '../columns/paths'
 import { Experiment } from '../webview/contract'
+
+type Value = undefined | null | [] | string | number
+
+const getValueType = (value: Value) => {
+  if (Number.isNaN(value)) {
+    return 'NaN'
+  }
+  if (typeof value === 'string' && Date.parse(value)) {
+    return 'date'
+  }
+  if (Array.isArray(value)) {
+    return 'array'
+  }
+  if (value === '') {
+    return 'empty'
+  }
+  return typeof value
+}
+
+const getStringifiedValue = (value: Value) => {
+  const type = getValueType(value)
+  switch (type) {
+    case 'date':
+      return typeof value === 'string' ? formatDate(value) : ''
+    case 'array':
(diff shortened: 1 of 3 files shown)


extension/src/experiments/model/util.ts
```diff
@@ -13,8 +13,9 @@ export const getDataFromColumnPath = (
     columnPath === 'Created' && collectedVal
       ? formatDate(collectedVal)
       : collectedVal?.value || collectedVal
+
   return {
     splitUpPath,
-    value: value || null
+    value: value === 0 || value ? value : null
```
julieg18, line 19: Falsy values are `undefined`, `null`, `NaN`, `0`, `""`, and `false`. With this change, we will show the column if it has a value that is 0 or it's none falsy value.

sroy3: Yes, it is possible to have boolean values. Might be best to just show everything so:
```
return {
  splitUpPath,
  value,
}
```

sroy3 (approved):

extension/src/experiments/model/util.ts
sroy3: I think `undefined` would still be useful to see. If it's not set for a row and it is in all others, it's actually what makes it jump out. Same for the empty string (for that one, I think I'd replace it with a literal `empty` just to make it clear.

julieg18: Makes sense! Updated the code to hopefully properly stringify any possible values!

extension/src/experiments/model/util.ts
```diff
@@ -1,20 +1,57 @@
...
+  if (typeof value === 'string' && Date.parse(value)) {
+    return 'date'
+  }
+  if (Array.isArray(value)) {
+    return 'array'
+  }
+  if (value === '') {
+    return 'empty'
+  }
+  return typeof value
```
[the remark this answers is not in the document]
sroy3: This is a util function, I think it's fine to keep as is. It's quite easy to read and understand what it does. The many return statements aren't making things complicated.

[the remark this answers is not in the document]
julieg18: Personally, I thought it looked better with the else if statements so I updated it :). But if you think it looks better, I can undo it!

PR #2747, opened 2022-11-09
Resizing plots horizontally

sroy3:
A start on #2585 

https://user-images.githubusercontent.com/3683420/201690172-d3671a68-ec24-4915-be11-956d679e35f2.mov


sroy3:
@shcheklein , @julieg18 and @mattseddon , you can review now even if it's not quite ready. You can try the branch or the Storybook, so that when I go back to working on this, we have a good idea of what we want to change.

extension/src/plots/vega/util.test.ts
```diff
@@ -139,7 +139,7 @@ describe('extendVegaSpec', () => {
 
   it('should truncate all titles from the left to 50 characters for large plots', () => {
     const spec = withLongTemplatePlotTitle()
-    const updatedSpec = extendVegaSpec(spec, 500)
+    const updatedSpec = extendVegaSpec(spec, PlotSizeNumber.LARGE)
```
sroy3, line 142: This was the source of my weird failing tests. Shows the importance of always using variables, because when they change, that makes it so much harder to find where they are used.

sroy3:
I noticed something is still wrong with resizing other than the first item. On it.

webview/src/plots/components/App.test.tsx
```diff
@@ -580,55 +576,21 @@ describe('App', () => {
...
-    fireEvent.click(largeButton)
-    wrapper = await getWrapper()
-    expect(
-      JSON.stringify(wrapper.style).includes(`${PlotSizeNumber.LARGE}`)
-    ).toBe(true)
-  })
-
   it('should send a message to the extension with the selected size when changing the size of plots', () => {
-    renderAppWithOptionalData({
+    // eslint-disable-next-line testing-library/render-result-naming-convention
```
mattseddon, line 580: [Q] Should we turn off this rule?

webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -88,38 +85,16 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
...
-  const menuItems: IconMenuItemProps[] = [
-    {
-      icon: Dots,
-      onClickNode: (
-        <SizePicker currentSize={size} setSelectedSize={changeSize} />
-      ),
-      tooltip: 'Resize'
-    }
-  ]
+  const menuItems: IconMenuItemProps[] = []
```
mattseddon, line 97: [Q] Can we remove the menu altogether for template plots:

image

mattseddon: Same for images:

image

What is the plan there? Are we dropping resizing from images altogether?

extension/src/plots/webview/contract.ts
```diff
@@ -2,9 +2,10 @@ import { VisualizationSpec } from 'react-vega'
 import { Color } from '../../experiments/model/status/colors'
 
 export const PlotSizeNumber = {
-  LARGE: 500,
-  REGULAR: 400,
-  SMALL: 300
+  LARGE: 1,
```
mattseddon, line 5: [C] Just an observation that these large plots generally don't fit on a single screen if the sidebar is hidden: 

image

Could set a max-height?

extension/src/plots/webview/contract.ts
```diff
@@ -2,9 +2,10 @@ import { VisualizationSpec } from 'react-vega'
...
+  MEDIUM: 3,
+  REGULAR: 2,
+  SMALL: 4
```
mattseddon, line 8: [N] Seems strange to have medium and regular, could have large and x-large instead.

webview/src/plots/components/styles.module.scss
```diff
@@ -168,13 +231,13 @@ $gap: 20px;
 .plot.multiViewPlot {
   aspect-ratio: calc(0.8 * var(--scale) + 0.2);
   grid-column: span var(--scale);
-  width: calc(var(--size) * var(--scale));
+  width: calc(300px * var(--scale));
```
mattseddon, line 234: [Q] Is this intended behaviour: 

https://user-images.githubusercontent.com/37993418/203680633-346f9205-c3ce-483e-9f8b-f61a42e986cd.mov

Should it be possible to resize a multiview plot? Should the option be removed from multiview plots?

extension/src/plots/webview/contract.ts
mattseddon: Another observation: D'n'D a plot across sections (multiview vs single) when plots are large is nearly impossible:


https://user-images.githubusercontent.com/37993418/203681208-47f65bc7-119f-4807-a9c7-1c3e25e308a7.mov

webview/src/plots/components/ZoomablePlot.tsx
```diff
@@ -45,14 +60,48 @@ export const ZoomablePlot: React.FC<ZoomablePlotProps> = ({
...
+    return () => {
+      window.clearTimeout(enableClickTimeout.current)
+    }
+  }, [])
 
+  const handleOnClick = () =>
+    !clickDisabled.current && dispatch(setZoomedInPlot({ id, plot: plotProps }))
+
+  const commonResizerProps = {
+    onGrab: () => {
```
mattseddon, line 73: [C] Once I've grabbed the resizer there is no way to bail out, I have to resize the plots before returning them to their original size. Could add the ability to escape as a follow-up?

mattseddon (approved):

webview/src/plots/components/PlotsContainer.tsx
sroy3: > [Q] Can we remove the menu altogether for template plots:
> 
> image

There was supposed to be a reset size button, but this PR was getting a little too big. I'll remove it until the new button it in place.

sroy3: > What is the plan there? Are we dropping resizing from images altogether?
The resizing will have to work differently as it's a table and stretches indefinitely. Right now resizing is more like changing the number of items per row from 1 to 4 and the comparison table is always one line. Resizing is disabled for now.

extension/src/plots/webview/contract.ts
sroy3: Ideally, resizing plots vertically would solve that, but I'll set a max-height until then.

extension/src/plots/webview/contract.ts
sroy3: Thanks, I was searching hard for how to call it and that's the best I could come up with.

webview/src/plots/components/styles.module.scss
sroy3: No it shouldn't be possible (yet). Forgot to remove it.

extension/src/plots/webview/contract.ts
sroy3: `max-height` conflicts with `aspect-ratio`. So I changed the ratio to be `9 / 5` instead of `4 / 3`. Looks good on my screen, until we get to 4 elements per row with the sidebar (this isn't the default number per row though). IMO, it's fine if it doesn't alway
Screenshot 2022-11-24 at 10 00 38 AM
s look perfect (there will always be small screens), that is why we allow resizing.
Screenshot 2022-11-24 at 10 00 54 AM
Screenshot 2022-11-24 at 10 01 14 AM

webview/src/plots/components/ZoomablePlot.tsx
sroy3: Done!

extension/src/plots/webview/contract.ts
sroy3: Went with `SMALL` and `SMALLER` instead as it wouldn't change what `REGULAR` is and it worked better for key order as well.

PR #2748, opened 2022-11-09
Fix unit tests for non en-US timezones

mattseddon:
Unit tests were failing locally for me. The cause was the difference in formatting between `dateStyle: 'medium'` in our different timezones. => `DD MMM YYYY` vs `MMM DD, YYYY`


julieg18 (approved):
Looks good!

sroy3 (approved):

PR #2759, opened 2022-11-12
Make plots ribbon sticky on scroll

### LINKED ISSUE #2731: Pin plots ribbon
The experiments legend ribbon at the top of the plots view needs to be pinned so it’s visible even when scrolling down to see more plots.

shcheklein:
Closes #2731

https://user-images.githubusercontent.com/3659196/201453111-936dea0b-2dd7-43c0-a86e-d0aace3f2fdc.mov

(writing it here primarily for the sake of future generation wondering how to use intersection observer API / `react-intersection-observer` to handle top nav bar / menu that changes its size):

- `rootMargin` can be used to offset the root / parent element to the height of the menu. It works well if menu / nav bar height is stable. But would require creating a new observer when / if size changes. It breaks the point of using it and complicates code, etc, etc. See also https://github.com/w3c/IntersectionObserver/issues/428
- Idea here is simple - let's offset the main parent element by a small margin (1-10px) and require full intersection with the menu (`threshold` is 1 or close to that)

I was not able to make 1.0 threshold work (some pixels are not visible in the sticky mode?), but it work well with 0.95-0.99, etc.

### DIFF
--- a/webview/src/plots/components/ribbon/styles.module.scss
+++ b/webview/src/plots/components/ribbon/styles.module.scss
@@ -73,12 +73,22 @@
 
 .list {
   display: flex;
+  position: sticky;
   align-items: center;
   gap: 6px;
   border-bottom: 1px solid $border-color;
   padding: 10px 15px;
   margin: 0;
   flex-wrap: wrap;
+  top: 0;
+  width: 100%;
+  z-index: 100;
+  background-color: var(--vscode-editor-background);
+  transition: box-shadow 0.25s;
+
+  &.withShadow {
+    box-shadow: 0 5px 8px -2px $shadow;
+  }
 }
 
 .buttonWrapper {
--- a/webview/src/plots/components/ribbon/styles.module.scss
+++ b/webview/src/plots/components/ribbon/styles.module.scss
@@ -83,7 +83,7 @@
   top: 0;
   width: 100%;
   z-index: 100;
-  background-color: var(--vscode-editor-background);
+  background-color: $bg-color;
   transition: box-shadow 0.25s;
 
   &.withShadow {
--- a/webview/src/shared/components/webviewWrapper/styles.module.scss
+++ b/webview/src/shared/components/webviewWrapper/styles.module.scss
@@ -1,4 +1,5 @@
 .webviewWrapper {
   width: 100%;
-  height: 100%;
+  height: auto;
+  min-height: 100%;
 }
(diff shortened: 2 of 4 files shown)


webview/src/shared/components/webviewWrapper/styles.module.scss
```diff
@@ -1,4 +1,5 @@
 .webviewWrapper {
   width: 100%;
```
sroy3, line 2: Changing this to `min-content` would solve the following bug:


https://user-images.githubusercontent.com/3683420/201679946-a7cfc21a-6bc1-4be2-8c88-3dc6754e0eac.mov

webview/src/plots/components/ribbon/styles.module.scss
```diff
@@ -73,12 +73,22 @@
...
   align-items: center;
   gap: 6px;
   border-bottom: 1px solid $border-color;
   padding: 10px 15px;
   margin: 0;
   flex-wrap: wrap;
+  top: 0;
+  width: 100%;
+  z-index: 100;
+  background-color: var(--vscode-editor-background);
```
sroy3, line 86: variable `$bg-color` maps to `var(--vscode-editor-background)`. We try to only use scss variables (easier to use and understand).

sroy3 (approved):

webview/src/shared/components/webviewWrapper/styles.module.scss
julieg18: Technically, this bug is in main and not part of this pr. Could be fixed separately :)

julieg18 (approved):
Great work!

webview/src/shared/components/webviewWrapper/styles.module.scss
shcheklein: thanks @sroy3, good catch. The fix proposed changes the whole UX in a way that we probably don't want. Instead of the ribbon taking always size of the viewport only and overflowing items, it's now taking the whole space (defined by the most wide plot). I'm not sure this is a good idea. Also, plots size control will move to the edge if we do it this way

I can find a way to fill the space with something, but the whole layout might change after the #2747 and further changes?

sroy3: You are right, it is not as simple as that a change. What I'd do is make the ribbon `100%` and create a div inside (ribbon-content or something else) and give it a width of `100vw` (might have to calculate padding in there).

shcheklein: let's skip it for now and get back to it after #2585 is done then

PR #2760, opened 2022-11-12
Add "views" section to sidebar

### LINKED ISSUE #2683: Add controls to the sidebar to help users understand what views we are offering
The problem we are trying to solve here is that the new user is not aware of two main views we have: EXPs and Plots. That means we need explicitly to point the user to this information. Along with the Get Started improvements we hopefully can have a better result.

---

Make the section first on top for better accessibility.

It should contain:

- Open experiments and plots in the split view
- Open experiments 
- Open plots

### V1.
Users can open and close the views from the sidebar (not sure it is possible. To check!)

Screen Shot 2022-10-26 at 4 57 48 PM

### V2.
Users can open and close the views from the sidebar.

Screen Shot 2022-10-26 at 4 57 56 PM

### V3.
User can only open views (links).

Screen Shot 2022-10-26 at 4 58 03 PM

### V4.
User can only open views (buttons).

Screen Shot 2022-10-26 at 4 58 09 PM

julieg18:
* adds a "views" section to sidebar
* adds a "Show Plots and Experiments" command to command palette 
* Replaces the view buttons, "Add Sort" and "Add Filter" with links to stop the sidebar from looking crowded
* moves "plots" view to be above "sort" and "filter" views since it's of higher importance 

## To Do

- [x] Fix e2e tests

## Demo

https://user-images.githubusercontent.com/43496356/202029624-a57fa747-c1cd-43b3-a61a-f54fe0cf3ab5.mov

Fixes #2683


extension/src/extension.ts
```diff
@@ -227,6 +238,19 @@ export class Extension extends Disposable implements IExtension {
 
     registerExperimentCommands(this.experiments, this.internalCommands)
     registerPlotsCommands(this.plots, this.internalCommands)
+    this.internalCommands.registerExternalCommand(
```
julieg18, line 241: Since this command touches both plots and experiments, I wasn't sure where to put it. For now I left it in `extension.ts` but we could move it into one of the above functions or its own file 🤔

extension/src/extension.ts
```diff
@@ -227,6 +238,19 @@ export class Extension extends Disposable implements IExtension {
...
+      RegisteredCommands.EXPERIMENT_AND_PLOTS_SHOW,
+      async (context: VsCodeContext) => {
+        await this.experiments.showWebview(
```
julieg18, line 244: Does this need additional testing? Both of these show commands have testing already... I also wasn't sure where testing would even be located since it touches both plots and experiments.

extension/src/webview/workspace.ts
```diff
@@ -27,14 +27,14 @@ export abstract class BaseWorkspaceWebviews<
     }
   }
 
-  public async showWebview(overrideRoot?: string) {
+  public async showWebview(overrideRoot: string, viewColumn?: ViewColumn) {
```
julieg18, line 30: We couldn't have both `overrideRoot` and `viewColumn` be optional so I decided to make the `overrideRoot` argument be an empty string if undefined.

extension/src/extension.ts
sroy3: Maybe at least one to make sure both views are open.

sroy3 (approved):

mattseddon:
[Q] Should these actions be prefixed with "show" to match the actions in the command palette? Alternatively, should we change the command palette actions to use open instead of show

I.e in the view we could use:

- show experiments
- show plots 
- show both views

or leave the view and change the command palette actions to 

- Open Experiments
- Open Plots

The only problem with the second option is that users would have to find the moved commands. 

Last question: Should we add a "Show Experiments and Plots" command to the palette?

PR #2785, opened 2022-11-17
Fix map on undefined is path doesn't exist in rev

shcheklein:
Fixes critical part of the https://github.com/iterative/vscode-dvc-demo/pull/7#issuecomment-1315854742 .

I'm not sure how did plots work tbh with this, unless I'm missing something.

At least I'm getting now this:

Screen Shot 2022-11-16 at 10 08 03 PM

@sroy3 layout for matrices is pretty bad here :(

As a followup I'll try to find what's wrong with flexible plots and add tests if it's a reasonable effort.

### DIFF
--- a/extension/src/plots/model/collect.ts
+++ b/extension/src/plots/model/collect.ts
@@ -508,15 +508,16 @@ const updateDatapoints = (
   fields: string[]
 ): unknown[] =>
   selectedRevisions
-    .flatMap(revision =>
-      revisionData?.[revision]?.[path].map(data => {
+    .flatMap(revision => {
+      const datapoints = revisionData?.[revision]?.[path] || []
+      datapoints.map(data => {
         const obj = data as Record<string, unknown>
         return {
           ...obj,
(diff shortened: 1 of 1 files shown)


mattseddon (approved):

shcheklein:
Test is failing :(, can't take a look now ... but I'll try to figure out this tomorrow ...

extension/src/plots/model/collect.ts
```diff
@@ -508,15 +508,16 @@ const updateDatapoints = (
   fields: string[]
 ): unknown[] =>
   selectedRevisions
-    .flatMap(revision =>
-      revisionData?.[revision]?.[path].map(data => {
+    .flatMap(revision => {
+      const datapoints = revisionData?.[revision]?.[path] || []
+      datapoints.map(data => {
```
mattseddon, line 513: Test is failing because this needs to be returned...

mattseddon: Fixes the issue:

image

mattseddon: image

Issue #2673, opened 2022-10-25
Multi-root-workspaces do not work


alkatar21:
dvc: 2.30.1
dvc VS-Code extension: v0.5.7

I am using a multi-root workspace and dvc works as expected on the console.

But the tracked data is not shown in the Explorer under DVC Tracked. There are also no changes shown in the Source Control tab and the extension tab is loading all the time after visiting the tab once:
![grafik](https://user-images.githubusercontent.com/61387986/197807032-ebd9a8f6-d475-462c-90d5-6859808532d2.png)

The output window shows the following:
```bash
[version: 0.5.7, ...] > /bin/python3 -m dvc root - INITIALIZED
[version: 0.5.7, ...] > /bin/python3 -m dvc --version - INITIALIZED
[version: 0.5.7, ...] > /bin/python3 -m dvc --version - COMPLETED (312ms)
[version: 0.5.7, ...] > /bin/python3 -m dvc root - FAILED with code 253 (443ms)
[31mERROR[39m: you are not inside of a DVC repository (checked up to mount point '/workspaces/<project>')
```

If I remove all other folders and only keep the one wich uses DVC and reloed the window everything works as expected.
If I add the folders again now, the extension still works as expected and simply adds the Project folder as additional hierarchy to the DVC Tracked window.
If I know reload the window everything is again like in the beginning and the extension does not work.

sroy3:
@alkatar21 just to make sure I can reproduce things on my side correctly, how is you workspace built at the moment? Are the folders in the workspace all under the same folder and one of these folders is a DVC project? Or is it a custom workspace (saved as a file) with multiple folders added to it (folders that aren't all necessarily under the same folder)? Thanks.

alkatar21:
@sroy3 I'm not sure I understand it correctly. It's a workspace that VSCode calls [Multi-root-workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces). So yes, there is an extra VS-Code file that organizes which folders are present. Does that help?

And here is the official source for extension authors with another link with a link to more docs about the API: https://code.visualstudio.com/docs/editor/multi-root-workspaces#_extensions

sroy3:
@alkatar21 thanks it does help. Sorry I didn't know VS Code had an official name for this.

alkatar21:
Another observation I just made:
[...]

PR #2791, opened 2022-11-17
Load projects inside multi-root workspaces

sroy3:
There are other problems further ahead (eg. running experiments), but this makes the project load. Progress in #2673 

https://user-images.githubusercontent.com/3683420/202545162-b5067b2f-35c5-459b-b864-3cbcf50bd2da.mov


extension/src/setup.ts
```diff
@@ -159,7 +159,7 @@ export const setup = async (extension: IExtension) => {
     return
   }
 
-  extension.setRoots()
+  await extension.setRoots()
```
sroy3, line 162: `dvcRoots` are used by the `initialize` called on line 170. If there are multiple workspace folders, I guess it took too long to set before it was used. There must have been some race conditions happening in a basic project as well.

julieg18 (commented):
Looks good!

julieg18 (approved):

mattseddon (approved):

PR #2806, opened 2022-11-21
Fix webview test warnings

mattseddon:
This PR gets rid of all of these warnings currently shown in the logs:

image

E.g: https://github.com/iterative/vscode-dvc/actions/runs/3510362072/jobs/5880080200#step:8:713

### DIFF
--- a/webview/src/test/experimentsTable.tsx
+++ b/webview/src/test/experimentsTable.tsx
@@ -1,5 +1,6 @@
 import { configureStore } from '@reduxjs/toolkit'
 import {
+  act,
   fireEvent,
   render,
   within,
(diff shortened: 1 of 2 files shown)


webview/src/test/experimentsTable.tsx
```diff
@@ -104,3 +105,8 @@ export const expandRow = (label: string) => {
 
 export const selectedRows = () =>
   screen.queryAllByRole('row', { selected: true })
+
+export const advanceTimersByTime = (ms: number) =>
+  act(() => {
+    jest.advanceTimersByTime(ms)
```
sroy3, line 111: `jest` has been removed everywhere in the other file. Should it be the same here?

sroy3 (approved):

webview/src/test/experimentsTable.tsx
mattseddon: I replaced the call in the other file with this function  so that jest.advaceTimersByTime is always wrapped in act. I kept the name to like the behaviour. Should I change it to make it more obvious why we need it?

sroy3: Sorry, I somehow thought that some of them were imported from jest anyway. (First PR to review on a Monday morning). Just nevermind the comment.

PR #2813, opened 2022-11-22
Match the way that undefined/numeric experiment column values are shown in the table in tooltips and quick picks

mattseddon:
This PR changes the way that we display experiment column values that are undefined in the experiments view tooltip and experiments quick pick. We now match the way that undefined is displayed in the table. I have also removed the truncation marks from truncated values as I didn't feel they added much value to the user.

Main reason for doing this is that `undefined` means nothing to the average Python user.

LMK what you think.

### Screenshots

Updated after feedback: 

image

image

image

### DIFF
--- a/extension/src/experiments/model/util.ts
+++ b/extension/src/experiments/model/util.ts
@@ -1,40 +1,38 @@
 import get from 'lodash.get'
 import { formatDate } from '../../util/date'
-import { truncate } from '../../util/string'
 import { splitColumnPath } from '../columns/paths'
 import { Experiment } from '../webview/contract'
 
 type Value = undefined | null | [] | string | number
 
-const getValueType = (value: Value): string => {
-  let type: string = typeof value
-  if (Number.isNaN(value)) {
-    type = 'NaN'
-  } else if (typeof value === 'string' && Date.parse(value)) {
-    type = 'date'
-  } else if (Array.isArray(value)) {
-    type = 'array'
-  } else if (value === '') {
-    type = 'empty'
-  }
-  return type
-}
+const isDate = (value: Value): boolean =>
+  !!(typeof value === 'string' && Date.parse(value))
+
+const isLongNumber = (value: Value): boolean =>
+  typeof value === 'number' && value.toString().length > 7
 
 const getStringifiedValue = (value: Value): string => {
-  const type = getValueType(value)
-  switch (type) {
-    case 'date':
-      return typeof value === 'string' ? formatDate(value) : ''
-    case 'array':
-      return `[${value?.toString()}]`
-    case 'empty':
-    case 'NaN':
-      return type
-    case 'number':
-      return truncate(String(value), 7)
-    default:
-      return String(value)
+  if (Number.isNaN(value)) {
+    return 'NaN'
   }
+
+  if (isDate(value)) {
+    return formatDate(value as string)
+  }
+
+  if (Array.isArray(value)) {
+    return `[${value?.toString()}]`
+  }
+
+  if (value === undefined) {
+    return '-'
+  }
+
+  if (isLongNumber(value)) {
+    return (value as number).toPrecision(5)
+  }
+
+  return String(value)
 }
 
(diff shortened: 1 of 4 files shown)


extension/src/experiments/model/util.ts
```diff
@@ -1,40 +1,38 @@
...
 
 type Value = undefined | null | [] | string | number
 
-const getValueType = (value: Value): string => {
-  let type: string = typeof value
+const isDate = (value: Value): boolean =>
+  !!(typeof value === 'string' && Date.parse(value))
+
+const isLongNumber = (value: Value): boolean =>
+  typeof value === 'number' && value.toString().length > 7
```
mattseddon, line 12: [F] Matches table formatting (for now)

shcheklein:
LGTM ... I would change it in the future though. On tooltips, `...` looks like we cut the text (it was too long or something), not that it's an undefined value.

mattseddon:
> LGTM ... I would change it in the future though. On tooltips, `...` looks like we cut the text (it was too long or something), not that it's an undefined value.

@shcheklein I can change all of the '...' to '-' as per the CLI (table included): 
image

WDYT?

shcheklein:
@mattseddon yep, that look better to me (and consistent with CLI).

mattseddon:
Updated after feedback: 

image

image

extension/src/experiments/model/util.ts
```diff
@@ -1,40 +1,38 @@
...
+  if (Array.isArray(value)) {
+    return `[${value?.toString()}]`
+  }
+
+  if (value === undefined) {
+    return '-'
   }
+
+  if (isLongNumber(value)) {
+    return (value as number).toPrecision(5)
```
shcheklein, line 32: I think it might have some implications:

just looking briefly into this - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision

it's doing a lot of stuff under the hood?

also, I would check how does it behave with just a regular large number - `12345678900`? Will it transform it into scientific notation? (may be it's even better, btw vs just throwing part of the number away).

Suggestion- let's do this as a separate PR? do a proper research what do we want to show for different types of numbers? wdyt?

extension/src/experiments/model/util.ts
```diff
@@ -1,40 +1,38 @@
...
+  }
+
+  return String(value)
```
shcheklein, line 35: Q: double check what happens with type == empty and NaN - are we getting the same result?

mattseddon: `NaN` is the same we are not changing empty string anymore.

extension/src/experiments/model/util.ts
mattseddon: For now I copied the code that we use to display numbers in the table: 

```
const defaultPrecision = 5 // for when we can't calculate real precision yet

export const formatFloat = (value: number): string => {
  const automatic = value.toString()
  if (automatic.length > 7) {
    return value.toPrecision(defaultPrecision)
  }
  return automatic
}
```

Whilst not perfect I thought this was better than truncating the number.

mattseddon: It will transform large numbers into scientific notation. I also spent time researching alternatively yesterday and could not come up with anything better, which is why I proposed this.

shcheklein: sounds good, Matt, thanks

extension/src/experiments/model/util.ts
shcheklein: > we are not changing empty string anymore.

can it break anything downstream by chance?

shcheklein (approved):

extension/src/experiments/model/util.ts
mattseddon: I have double-checked. The only clients of this function are the experiments view tooltip and experiments for plotting quick-pick. Both display the value as expected 👍🏻.

PR #2815, opened 2022-11-22
Tests speedup

wolmir:
This will move some tests from App.test.tsx to other test files as an attempt to speed up testing time.
It shaved off ~20 secs of running time when executing locally.
Hopefully it will do the same in CI.


mattseddon:
Appears to slow the build pipeline down. From recent runs of the webview jest suite in `build`:

- run time on main: `180.516s`
- run time on this branch: `249.354s`

mattseddon (changes requested):
Please post multiple run times of the test suite from both CI pipelines so we can make a call.

Issue #2424, opened 2022-09-18
Refine new/running experiment for plots


shcheklein:
When a new experiment is running there are a few entries in the table that affect what is being show in the plot and change their state dynamically. It's important for this logic to be stable (predictable) and not duplicate entities on the plots dashboard, properly pick which experiment is plotted above which, and what colors they have during the run and afterwards.

One of the bugs:

<img width="583" alt="Screen Shot 2022-09-17 at 7 08 42 PM" src="https://user-images.githubusercontent.com/3659196/190882638-309d7f83-2b9e-49b3-bf0b-77759f9151f1.png">

It's not clear why it's stuck (keeps one of the experiments in the middle). 

Besides that, on this example:

- we have 4 items selected by running a single experiment (workspace, branch, checkpoint, experiment HEAD). Can we reduce this number?
- we need to check what people see during the run and make sure that at the end there is no jump (e.g. color doesn't change)
- what happens if there are no enough slots (6 selected and I need to run an extra experiments that would enable at least 3 plots)? We should handle this gracefully (Always keep and extra slot for a running experiment in the workspace, don't show it at all?)

mattseddon:
> One of the bugs:

Caused by https://github.com/iterative/dvc/issues/7693, only a problem with checkpoints.

> * we have 4 items selected by running a single experiment (workspace, branch, checkpoint, experiment HEAD). Can we reduce this number?
> * we need to check what people see during the run and make sure that at the end there is no jump (e.g. color doesn't change)
> * what happens if there are no enough slots (6 selected and I need to run an extra experiments that would enable at least 3 plots)? We should handle this gracefully (Always keep and extra slot for a running experiment in the workspace, don't show it at all?)

The simplest thing to do would be to not auto-select new experiments.

shcheklein:
Thanks @mattseddon . 

> The simplest thing to do would be to not auto-select new experiments.

I don't think this is good from the product perspective unfortunately. We can (and probably we already do?) disable it for the queue.

For the workspace we need to have a clear stable semantics. It should be a single line / images that corresponds to a running experiment (or to the latest checkpoint), we don't want to have 3 or 4 selected things in this case. Plus we need to check for the edge case that I mentioned (experiment's color doesn't jump, what happens if number of selected things > 7 already, etc)

mattseddon:
[...]

PR #2831, opened 2022-11-25
Use linear path to update plots data when experiments update

mattseddon:
WIP

### DIFF
--- a/extension/src/plots/index.ts
+++ b/extension/src/plots/index.ts
@@ -142,7 +142,6 @@ export class Plots extends BaseRepository<TPlotsData> {
     this.fetchMissingAndSendPlots()
   }
 
-  // eslint-disable-next-line sonarjs/cognitive-complexity
   private async fetchMissingAndSendPlots(
     overrideRevs?: SelectedExperimentWithColor[]
   ) {
--- a/extension/src/plots/index.ts
+++ b/extension/src/plots/index.ts
@@ -157,11 +157,7 @@ export class Plots extends BaseRepository<TPlotsData> {
       const data = await this.data.fetch(missingRevs)
 
       await Promise.all([
-        this.plots.transformAndSetPlots(
-          data,
-          missingRevs,
-          this.experiments.getSelectedRevisions()
-        ),
+        this.plots.transformAndSetPlots(data, missingRevs, selectedRevs),
         this.paths.transformAndSet(data)
       ])
     }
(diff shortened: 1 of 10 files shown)


extension/src/plots/model/index.ts
```diff
@@ -105,6 +106,10 @@ export class PlotsModel extends ModelWithPersistence {
   }
 
   public async transformAndSetPlots(data: PlotsOutput, revs: string[]) {
+    if (data.error) {
```
mattseddon, line 109: [F] Related to this comment: https://github.com/iterative/vscode-dvc/issues/1649#issuecomment-1295969178. Will fix separately.

extension/src/plots/index.ts
```diff
@@ -48,7 +49,7 @@ export class Plots extends BaseRepository<TPlotsData> {
     this.experiments = experiments
 
     this.plots = this.dispose.track(
-      new PlotsModel(this.dvcRoot, experiments, workspaceState)
```
mattseddon, line 51: [F] Removing experiments from here so that we can't get into a race condition where the selected revisions change mid-update.

extension/src/plots/model/index.ts
```diff
@@ -101,15 +102,23 @@ export class PlotsModel extends ModelWithPersistence {
...
+      return
+    }
+
+    const cliIdToLabel = this.getCLIIdToLabel(selectedRevs)
 
     this.fetchedRevs = new Set([
       ...this.fetchedRevs,
-      ...revs.map(rev => cliIdToLabel[rev])
+      ...newRevs.map(rev => cliIdToLabel[rev])
     ])
```
mattseddon, line 122: [F] I think we can expand `fetchedRevs` to include state information about each revision. We can then display this information on the ribbon. The 4 states would be:

- loading
- loaded with data
- loaded no data
- loaded with errors

We could potentially combine 3 & 4.

mattseddon: WDYT?

mattseddon:
I've been chasing shadows in this PR. The specific issue that I was trying to solve is checkpoints being updated too quickly to ever be shown inside of the plots webview. This happens when there is not much data in the experiments table + checkpoints are created in a rapid manner. A better solution for that particular issue has been proposed in #2424.

I am going to take a different approach to speed up the ribbon being loaded.

PR #2832, opened 2022-11-25
Delay creation of plots to remove optional logic

mattseddon:
This is a precursor to the work being done in #2831. We can simplify some of the logic in `Plots` and `PlotsData` by delaying the creation of plots repositories until after `Experiments` have been created for each dvc root.

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -393,11 +393,14 @@ export class Extension extends Disposable implements IExtension {
         this.dvcRoots,
         this.updatesPaused,
         this.resourceLocator
-      ),
-      this.plots.create(this.dvcRoots, this.updatesPaused, this.resourceLocator)
+      )
     ])
-
-    this.experiments.linkRepositories(this.plots)
+    this.plots.create(
+      this.dvcRoots,
+      this.updatesPaused,
+      this.resourceLocator,
+      this.experiments
+    )
 
     return Promise.all([
       this.repositories.isReady(),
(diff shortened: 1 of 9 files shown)


extension/src/extension.ts
```diff
@@ -393,11 +393,14 @@ export class Extension extends Disposable implements IExtension {
...
-      this.plots.create(this.dvcRoots, this.updatesPaused, this.resourceLocator)
+      )
     ])
-
-    this.experiments.linkRepositories(this.plots)
+    this.plots.create(
+      this.dvcRoots,
+      this.updatesPaused,
+      this.resourceLocator,
+      this.experiments
```
mattseddon, line 402: [F] By creating this after we have created the experiments repositories we can simply/remove a lot of the optional chaining logic from `Plots` and `PlotsData` (model is passed in on instantiation instead of being set afterwards).

sroy3 (approved):
On my end, loading plots can take a little while (mainly because that's usually the first thing I open when launching the extension). Wondering if the message should say more than just "Loading Plots..." If we are delaying their creation + already long loading time.

mattseddon:
> On my end, loading plots can take a little while (mainly because that's usually the first thing I open when launching the extension). Wondering if the message should say more than just "Loading Plots..." If we are delaying their creation + already long loading time.

Technically this change shouldn't hurt performance as we have to wait for experiments to be built before we know which revisions to request. This is just a simpler/smarter way to structure the code.

PR #2836, opened 2022-11-25
Fix the number of available revisions for multi view plots

### LINKED ISSUE #2834: Multi view plots do not always show all revisions
Because multi view plots use the number of revisions to display correctly (it uses the number of revision to calculate the width), having some revisions missing inside it can result in things like the following

Screenshot 2022-11-25 at 9 48 11 AM

sroy3:
Fixes #2834 

**Before**:
Screenshot 2022-11-25 at 9 41 59 AM

**Now**:
Screenshot 2022-11-25 at 11 04 54 AM

### DIFF
--- a/extension/src/test/suite/plots/index.test.ts
+++ b/extension/src/test/suite/plots/index.test.ts
@@ -39,6 +39,7 @@ import * as Telemetry from '../../../telemetry'
 import { EventName } from '../../../telemetry/constants'
 import { ExperimentStatus } from '../../../cli/dvc/contract'
 import { SelectedExperimentWithColor } from '../../../experiments/model'
+import { Experiments } from '../../../experiments'
 
 suite('Plots Test Suite', () => {
   const disposable = Disposable.fn()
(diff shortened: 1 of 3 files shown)


sroy3:
I think the failing test `should send the correct data to the webview for flexible plots` is simply not fully set up and that the `multiSourceEncoding` is missing. I'm looking into adding it, but I'm not quite sure how to. @mattseddon if you know how to do this quickly, help would be appreciated.

mattseddon:
> I think the failing test `should send the correct data to the webview for flexible plots` is simply not fully set up and that the `multiSourceEncoding` is missing. I'm looking into adding it, but I'm not quite sure how to. @mattseddon if you know how to do this quickly, help would be appreciated.

I'll take a look.

extension/src/test/suite/plots/index.test.ts
```diff
@@ -763,14 +764,18 @@ suite('Plots Test Suite', () => {
...
-      const { plots, messageSpy, mockPlotsDiff, experiments } =
-        await buildPlots(disposable, multiSourcePlotsDiffFixture)
-
-      stub(experiments, 'getSelectedRevisions').returns([
+      stub(Experiments.prototype, 'getSelectedRevisions').returns([
         { label: 'workspace' },
         { label: 'main' }
       ] as SelectedExperimentWithColor[])
 
+      stub(Experiments.prototype, 'getBranchRevisions').returns([])
```
mattseddon, line 772: [F] Multi-source test fixture doesn't match the default exp show fixture. The change required would blow this PR out so I have just patched it for now. Will follow-up once this is merged.

mattseddon (approved):

extension/src/test/suite/plots/index.test.ts
sroy3: Thank you!

PR #2837, opened 2022-11-26
Update multi source test fixture to contain the expected revisions (branch revision not name)

mattseddon:
# 2/2 `main` <- #2836 <- this

This PR correctly fixes the issue with the flexible plots test fixture discovered in #2836 (and undoes the patch).

`plots diff` is currently called with the branch revision and not the name.


sroy3 (approved):
Thanks for this. Wouldn't have figured it on my own quickly.

PR #2841, opened 2022-11-28
Add progress ring to plots ribbon while data is loading

### LINKED ISSUE #2835: Revisions on the ribbon can be very slow to appear
This can result in a confusing screen for some time (I got stuck like this for a good 20 seconds).

Screenshot 2022-11-25 at 9 48 11 AM

mattseddon:
# 1/2 `main` <- this <- #2859

Closes #2835. 

This PR adds a progress ring to revisions in the plots ribbon whilst data is being loaded. This would have been a simple change had it not been for a race condition in running checkpoint experiments where the revision was never actually loaded until the experiment finished.

This change comes with a caveat (see inline comments). I will follow up straight away.

### Demo

https://user-images.githubusercontent.com/37993418/204709714-e01c386c-78f3-4d1f-999c-1a06b433084a.mov


sroy3:
This looks great! Makes things a little more dynamic.

webview/src/plots/components/ribbon/Ribbon.tsx
```diff
@@ -23,12 +22,6 @@ export const Ribbon: React.FC = () => {
   const revisions = useSelector(
     (state: PlotsState) => state.webview.selectedRevisions
   )
-  const [order, setOrder] = useState<string[]>([])
-  const reorderId = 'id'
-
-  useEffect(() => {
-    setOrder(pastOrder => performOrderedUpdate(pastOrder, revisions, reorderId))
-  }, [revisions])
```
mattseddon, line 31: [F] There was a race condition between the data update and `useEffect` that would cause `reorderObjectList(order, revisions, reorderId)` to drop selected checkpoint experiments from the ribbon. Deleting this means that we need to update the message passed for comparison plots and split the order of the comparison table columns from `selectedRevisions` (to reinstate => `'should not reorder the ribbon when comparison plots are reordered'`).

webview/src/plots/components/App.test.tsx
```diff
@@ -1818,37 +1818,6 @@ describe('App', () => {
         type: MessageFromWebviewType.REFRESH_REVISIONS
       })
     })
-
-    it('should not reorder the ribbon when comparison plots are reordered', () => {
```
mattseddon, line 1822: see https://github.com/iterative/vscode-dvc/pull/2841/files#r1034211672

webview/src/plots/components/ribbon/Ribbon.tsx
mattseddon: I will do this as an immediate follow up

webview/src/plots/components/ribbon/styles.module.scss
```diff
@@ -27,6 +26,11 @@
     display: inline;
     font-size: 0.8125rem;
   }
+
+  .fetching {
```
mattseddon, line 30: [F] Gives us the space to display an icon against the experiment (e.g `<Error/>` if we have an issue with it)

webview/src/stories/Ribbon.stories.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
```
mattseddon, line 1: [F] Add a story to demonstrate the styling of the ribbon with and without progress ring(s)

mattseddon: I'm sure we'll need to iterate on the style.

PR #2842, opened 2022-11-28
Fix table header sticky on scroll

### LINKED ISSUE #2840: Sticky header broken in experiments table view
image

After scroll: 

image

This is strange because we have a story for scrolling which doesn't appear to be broken.

shcheklein:
Fixes #2840 

Sticky doesn't work if ancestors don't have height set and overflow is set to auto (e.g. https://uxdesign.cc/position-stuck-96c9f55d9526). I don't think we need `overflow: auto` anymore since table now takes the whole page (@rogermparent probably can confirm, he introduced it initially here https://github.com/iterative/vscode-dvc/commit/11333007ff939d409e12a63da4625d327ebed570)

Rebase and merge after https://github.com/iterative/vscode-dvc/pull/2850


shcheklein:
Looking into storybook ...

mattseddon (approved):
Works fine in the app. No idea what is going on in SB

mattseddon:
Attempted to fix storybook with #2843 & #2844

shcheklein:
Thanks, Matt! Let me try to fix it. I'll figure out something. No need to distract you on it I think :)

mattseddon:
> Thanks, Matt! Let me try to fix it. I'll figure out something. No need to distract you on it I think :)

Storybook should be back to normal in #2850. Won't do any more.

shcheklein:
> Won't do any more

that's it probably though? story book was the only thing ... I was wondering if decorator is the right fix at all (or decorate plots instead in a similar way we had for the table) and wanted to take a look, but since decorator is done we can merge and be done with that?

PR #2850, opened 2022-11-28
Add styled divs to maintain table story snapshot parity

mattseddon:


webview/src/stories/Table.stories.tsx
```diff
@@ -102,6 +102,13 @@ export default {
     tableData
   },
   component: Experiments,
+  decorators: [
+    Story => (
+      <div style={{ height: '900px', overflow: 'auto' }}>
```
shcheklein, line 107: q: why 900? I see in chromatic it's complaining about it being longer? (it was 600 or something)

mattseddon: I'm going off what I read here: https://www.chromatic.com/test?appId=5ffc93f46e9344002155f496&id=63848c156632934f84497266

image

Also, 900px is the default height as specified [here](https://www.chromatic.com/docs/snapshots#why-is-my-content-being-cut-off-vertically-in-my-snapshots).

shcheklein (approved):

PR #2854, opened 2022-11-29
Fix plots file watchers (diff output key is not always a file)

### LINKED ISSUE #2851: Top-level plots are not updated while the experiment is running
> From memory we get the files to watch from the keys of the plots data. This means that top-level plot keys like dvclive/dvc.yaml::dvclive/plots/metrics/foo/bar.tsv will break the current process. Needs an update.

_ @mattseddon

mattseddon:
# 2/2 `main` <- #2852 <- this

Closes #2851.

image

☝🏻 I actually remembered correctly.

Unfortunately, I realised that the key can actually be the plot's title which means we need to interrogate the data 😢.

### DIFF
--- a/extension/src/data/index.ts
+++ b/extension/src/data/index.ts
@@ -108,7 +108,7 @@ export abstract class BaseData<
     )
   }
 
-  abstract collectFiles(data: T): string[]
-
   abstract managedUpdate(path?: string): Promise<unknown>
+
+  protected abstract collectFiles(data: T): string[]
 }
--- a/extension/src/plots/data/index.ts
+++ b/extension/src/plots/data/index.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'vscode'
+import { collectFiles } from './collect'
 import { PlotsOutputOrError } from '../../cli/dvc/contract'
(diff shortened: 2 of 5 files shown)


extension/src/data/index.ts
```diff
@@ -108,7 +108,7 @@ export abstract class BaseData<
     )
   }
 
-  abstract collectFiles(data: T): string[]
-
   abstract managedUpdate(path?: string): Promise<unknown>
+
+  protected abstract collectFiles(data: T): string[]
```
mattseddon, line 113: [F] TIL this is possible.

extension/src/plots/data/index.ts
```diff
@@ -59,22 +59,19 @@ export class PlotsData extends BaseData<{
       ...args
     )
 
+    this.notifyChanged({ data, revs })
```
mattseddon, line 62: [F] We don't want to delay updating the rest of the app whilst we mess around looking for files to watch.

mattseddon:
I had to update the test fixture because the Windows tests were failing.

sroy3 (approved):

PR #2859, opened 2022-11-30
Separate comparison table and plot ribbon revision orders

mattseddon:
# 2/2 `main` <- #2841 <- this

Immediate follow-up from the previous PR. The aim here is to separate the plots ribbon and comparison table states (order) and simplify the code in the process.

### Demo

https://user-images.githubusercontent.com/37993418/204763584-71579d04-040b-4959-ba73-6bf54350a0bc.mov

The last thing that I want to do before moving on is add some enhanced empty states for when we only have loading revisions present in the webview.

### DIFF
--- a/extension/src/plots/model/collect.ts
+++ b/extension/src/plots/model/collect.ts
@@ -11,7 +11,8 @@ import {
   Plot,
   TemplatePlotEntry,
   TemplatePlotSection,
-  PlotsType
+  PlotsType,
+  Revision
 } from '../webview/contract'
 import {
   ExperimentFieldsOrError,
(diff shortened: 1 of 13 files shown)


extension/src/plots/model/collect.ts
```diff
@@ -818,9 +835,11 @@ export const collectOverrideRevisionDetails = (
   }
 
   return {
-    overrideOrder: comparisonOrder
-      .map(label => orderMapping[label])
-      .filter(Boolean),
+    overrideComparison: reorderObjectList(
```
mattseddon, line 838: [F] Providing both sets of revisions here potentially in different orders means we don't have to worry about reordering later and simplifies the code outside of this collection process

sroy3 (approved):

Issue #2776, opened 2022-11-16
Keep "views" sidebar section on welcome screen


julieg18:
The views section should stay on the sidebar even before DVC is initialized for a better-unified experience, opening the views with placeholder text. 

<img width="427" alt="image" src="https://user-images.githubusercontent.com/43496356/202049817-b9bce22b-1c6f-4ca4-87ec-9b31fccfcbc1.png">

Followup to #2760

PR #2861, opened 2022-11-30
Add webview for when dvc is not available or not initialized

sroy3:
A start on #2776 

https://user-images.githubusercontent.com/3683420/204931601-e09f795c-3095-4479-b6f9-f8e6b49650f1.mov


sroy3:
The structure should change a lot in the following PRs. Not sure if it's worth adding tests for now. I'll add them if you think otherwise.

I'll also follow-up for the name and icons of the webview and change the order of the views in the sidebar.

extension/src/webview/workspace.ts
```diff
@@ -48,5 +66,15 @@ export abstract class BaseWorkspaceWebviews<
     return overrideRoot || (await this.getFocusedOrOnlyOrPickProject())
   }
 
+  protected async showEmptyWebview(viewColumn?: ViewColumn) {
+    await createWebview(
```
sroy3, line 70: Creating a simple webview for now, but I've already started creating a class for it to add messages.

extension/package.json
```diff
@@ -1351,7 +1351,7 @@
       {
         "view": "dvc.views.webviews",
         "contents": "[Show Experiments](command:dvc.showExperiments)\n[Show Plots](command:dvc.showPlots)\n[Show Experiments and Plots](command:dvc.showExperimentsAndPlots)",
-        "when": "dvc.commands.available && dvc.project.available"
+        "when": "true"
```
mattseddon, line 1354: [Q] Would it be easier to show the same view but with different underlying commands under different circumstances? Would that simplify the code elsewhere?

i.e 
```
      {
        "view": "dvc.views.webviews",
        "contents": "[Show Experiments](command:dvc.showExperiments)\n[Show Plots](command:dvc.showPlots)\n[Show Experiments and Plots](command:dvc.showExperimentsAndPlots)",
        "when": "dvc.commands.available && dvc.project.available"
},
      {
        "view": "dvc.views.webviews",
        "contents": "[Show Experiments](command:dvc.showEmptyWebview)\n[Show Plots](command:dvc.showEmptyWebview)\n[Show Experiments and Plots](command:dvc.showEmptyWebview)",
        "when": "!dvc.commands.available || !dvc.project.available"
}
```

mattseddon (approved):
Left one [Q] that might help to simplify code paths. Just need to check that the new webview is correctly bundled with the extension.

extension/package.json
sroy3: I think you're right. I'll try that. Thanks!

PR #2864, opened 2022-11-30
Open plots in split view when request sent from table

### LINKED ISSUE #2078: Toggle native split views when clicking on the plots icon in the EXPs view
We want to try to educate the user to use both views when working on an extension. Or at least to understand how we notify about the cross changes in both views. 

The same will be done for the EXPs icon in the Plots view (it's coming and will replace the current ribbon. Check here for more details: https://github.com/iterative/design/issues/29)

Screen Shot 2022-07-21 at 2 11 17 PM

Screen Shot 2022-07-21 at 2 12 29 PM

mattseddon:
Closes #2078

Noticed this issue yesterday whilst we were doing grooming. Knew it would only take ~5 minutes so decided to knock it over.

### Demo

https://user-images.githubusercontent.com/37993418/204922342-ee7a7afb-fd82-44e1-b56b-d56730b8b916.mov

### DIFF
--- a/extension/src/experiments/webview/messages.ts
+++ b/extension/src/experiments/webview/messages.ts
@@ -124,7 +124,7 @@ export class WebviewMessages {
         return this.focusSortsTree()
 
       case MessageFromWebviewType.OPEN_PLOTS_WEBVIEW:
-        return this.showPlots()
+        return this.showPlotsToSide()
 
       case MessageFromWebviewType.SHARE_EXPERIMENT_AS_BRANCH:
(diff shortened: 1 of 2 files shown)


shcheklein (approved):

shcheklein:
I'm still not 100% we should be doing this tbh. But okay, let's give it a try.

mattseddon:
> I'm still not 100% we should be doing this tbh. But okay, let's give it a try.

Need to close issues that we don't want to do/aren't sure about.

PR #2865, opened 2022-12-01
Add loading state for sections and images to plots webview

mattseddon:
This PR makes the plots webview show a loading state as a revision is fetched.

#### Demo

https://user-images.githubusercontent.com/37993418/204957739-33fe5475-b093-4fd3-b75e-b61eea03722b.mov

https://user-images.githubusercontent.com/37993418/204957808-43e40438-9661-404f-b153-1121ff2d191d.mov

https://user-images.githubusercontent.com/37993418/205024750-c863a24f-1f19-46fb-9bc3-f9e9ebc9b177.mov


extension/src/plots/webview/contract.ts
```diff
@@ -119,7 +119,7 @@ export interface TemplatePlotsData {
 }
 
 export type ComparisonPlot = {
-  url: string | undefined
+  url?: string
```
mattseddon, line 122: [F] Type was incompatible with the data I dumped for a test.

webview/src/plots/components/LoadingSection.tsx
```diff
@@ -0,0 +1,10 @@
+import { Revision } from 'dvc/src/plots/webview/contract'
```
mattseddon, line 1: [Q] Better place for this component?

webview/src/plots/components/comparisonTable/styles.module.scss
```diff
@@ -88,13 +88,18 @@ $gap: 4px;
...
+.noImage {
   background-color: $bg-color;
   border-style: solid;
   border-width: thin;
   border-color: $watermark-color;
 }
 
+.noImageContent {
+  padding-top: 25%;
+  padding-bottom: 25%;
```
mattseddon, line 100: [F] Spent a couple of hours trying to find a more elegant way to do this. Please LMK if there is a better solution.

sroy3: Setting a height on the `td` would act in the same way (if other cells are bigger, the cell will get bigger as height on cells act as a min-height). I think that the only downside to your solution is that it is not intuitive why it is needed, that is until you have only cells without image content. IMO you can leave as is, you can add a comment if you want.

sroy3 (approved):

Issue #2698, opened 2022-10-29
DnD on the header is broken


shcheklein:
You can see below how it sometimes doesn't want to move columns:

https://user-images.githubusercontent.com/3659196/198849323-4f850db4-dd71-49cc-9c65-9e293ac4e8b6.mov

https://user-images.githubusercontent.com/3659196/198849325-8ebecf73-cec5-41bd-b421-8c0fa749581f.mov

We also need to improve the thresholds. You have to drag for too long to see any reaction at all. It feels unstable and unpredictable at the moment.

shcheklein:
Thresholds were improved, drop zone highlight is explicit now and disappears if target can't be dropped . This can be closed.

Issue #2262, opened 2022-08-25
Drag–and–Drop improvements


maxagin:
## Motivation

The goal is to have a better user experience and performance. 

- UX: It's in general preferable to have as less as possible unnecessary movements of the UI elements
- Performance: Fewer elements will change their positions, the faster will be the processing

## Principles

1. Do not change the composition before the user undersend where he wants to locate the object he is moving
2. Move only the reflection of the dragged object 
3. Display the indicator that shows the exact place where the dragged object can be placed
4. Only after the new place is found initiate the changes in the composition
5. Be able to reset the workspace (will be added to the bar on top of the table #2075)

## Apply these principles to following

- [x] Table's header  https://github.com/iterative/vscode-dvc/issues/2262#issuecomment-1228670872
- [ ] Table's body - EXPs (if we will like to have D&D for the EXPs)
- [ ] Plots sections
- [ ] Plots sections content 
- [x] Other? - where the Drag–and–Drop is used

## Related

[D&D interaction patterns study](https://www.notion.so/iterative/D-D-interaction-patterns-study-e6b840cf45e74bcbaef5157218dae2a0)

shcheklein:
Let's simplify and make a single checkbox with other improvements. The whole ticket can one checkbox please with performance as motivation.

There are also some bugs in DnD that we can do small tickets about or one with checkboxes - feedback is pretty bad still, no way to cancel DnD, almost no way to get back to the initial state while doing DnD - let's focus on these things first please.

maxagin:
For the "Table's header sections"

1. We do not want to change the table content before the user undersend where he wants to locate the section. Only after the user is sure where he like to place the section we can initiate the changes in the table. 
2. We move only the _reflection of the dragged cell_ (just a cell not the column!) + display the indicator that shows the exact place where the cell can be located. After the place is found we initiate the changes in the table including the column relocation. 
This also will help to have a _better performance_, while dragging the object, as we will be dealing just with a cell instead of an entire column without changing the layout of the view all the time the object changes its position. and moving just one object instead of changing the column's positions.
3. The style of the indicator can be the same one as we are planning to use when resizing the header cell (the new style is in [Figma](https://www.figma.com/file/kTErWSdnP5PVeU6UKdx0gR/Experiments?node-id=1671%3A118298))

sroy3:
[...]

PR #2876, opened 2022-12-04
Improve table header DnD

shcheklein:
Partially addresses #2698 #2262

Changes semantics of DnD to be similar to VS Code tabs (and Notion, and other table headers). It's more responsive and clear where column lands. In it's simplest form one can do quick swap now:

https://user-images.githubusercontent.com/3659196/205471216-ce8cde30-c7ee-4140-b848-dbe4df6f795a.mov

It inserts either to the left or to the right of the target depending on the DnD direction:

https://user-images.githubusercontent.com/3659196/205471286-139dd592-7efa-464e-81e1-fdd49eb40fac.mov

Contrast theme:

https://user-images.githubusercontent.com/3659196/205472608-c81c5605-e558-4f02-a9df-3b2cd25aaf7b.mov

## Minor things:

- drop some unused styles (can it be detected and linted?)
- fix Icons (sort, filter) placeholder so that it's not visible on DnD when there are no icons (before there was a small 4px X 7px box)
- launch action to run a test file under debugger (is there any other alternative way to do? I was not able to setup it w/o this change)

Clearly, there will be more improvements to DnD. It's quite broken atm. Next major thing is to fix work with placeholders and when it lands to icons. It should be working always w/o these edge cases.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -35,6 +35,20 @@
           "VSC_DEBUG": "true"
         }
       }
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Jest Current File (webview)",
+      "program": "${workspaceFolder}/node_modules/.bin/jest",
+      "args": ["${fileBasenameNoExtension}", "--config", "jest.config.js"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": true,
+      "windows": {
+        "program": "${workspaceFolder}/node_modules/jest/bin/jest"
+      },
+      "cwd": "${workspaceFolder}/webview"
     }
   ]
 }
--- a/webview/src/shared/components/iconMenu/IconMenu.tsx
+++ b/webview/src/shared/components/iconMenu/IconMenu.tsx
@@ -26,6 +26,8 @@ export const IconMenu: React.FC<IconMenuProps> = ({ items }) => {
     disabled: tooltipDisabled
   })
 
+  const visibleItems = items.filter(({ hidden }) => !hidden)
+
   return (
     <Tooltip
       singleton={tooltipSource}
@@ -46,10 +48,9 @@ export const IconMenu: React.FC<IconMenuProps> = ({ items }) => {
           setTooltipDisabled(false)
         }}
       >
-        <ul className={styles.menu} role="menu">
-          {items
-            .filter(({ hidden }) => !hidden)
-            .map(item => (
+        {visibleItems.length > 0 ? (
+          <ul className={styles.menu} role="menu">
+            {visibleItems.map(item => (
               <IconMenuItem
                 {...item}
                 key={item.tooltip}
@@ -57,7 +58,10 @@ export const IconMenu: React.FC<IconMenuProps> = ({ items }) => {
                 menuTarget={menuTarget}
               />
             ))}
-        </ul>
+          </ul>
+        ) : (
+          <></>
+        )}
       </Tooltip>
     </Tooltip>
   )
--- a/webview/src/experiments/components/table/Table.test.tsx
+++ b/webview/src/experiments/components/table/Table.test.tsx
@@ -445,7 +445,7 @@ describe('Table', () => {
         headers.indexOf('threshold')
       )
 
-      const [, startingNode] = screen.getAllByText('summary.json')
+      const [startingNode] = screen.getAllByText('summary.json')
       dragAndDrop(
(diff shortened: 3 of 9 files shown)


.vscode/launch.json
```diff
@@ -35,6 +35,20 @@
...
+      "name": "Jest Current File (webview)",
+      "program": "${workspaceFolder}/node_modules/.bin/jest",
+      "args": ["${fileBasenameNoExtension}", "--config", "jest.config.js"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": true,
+      "windows": {
+        "program": "${workspaceFolder}/node_modules/jest/bin/jest"
+      },
+      "cwd": "${workspaceFolder}/webview"
```
mattseddon, line 51: [I] It would be better to add this as a task to tasks.json. Something like:

```
    {
      "args": [
        "workspace",
        "dvc-vscode-webview",
        "jest",
        "${fileBasenameNoExtension}"
      ],
      "type": "process",
      "isBackground": false,
      "label": "Jest Current File (webview)",
      "command": "yarn",
      "options": { "cwd": "${workspaceFolder}" }
    }

```

mattseddon: Generally we run these commands in our terminals.

mattseddon: From running the tests it appears that this change has caused a lot of noise in the logs:

https://github.com/iterative/vscode-dvc/actions/runs/3611484435/jobs/6085998133#step:8:697

webview/src/shared/components/iconMenu/IconMenu.tsx
```diff
@@ -46,18 +48,18 @@ export const IconMenu: React.FC<IconMenuProps> = ({ items }) => {
           setTooltipDisabled(false)
         }}
       >
-        <ul className={styles.menu} role="menu">
-          {items
-            .filter(({ hidden }) => !hidden)
-            .map(item => (
+        {visibleItems.length > 0 ? (
```
mattseddon, line 51: Change upsetting the tests is in here

mattseddon: Changing to an empty JSX element seems to do the trick:

```
        {visibleItems.length > 0 ? (
          <ul className={styles.menu} role="menu">
            {visibleItems.map(item => (
              <IconMenuItem
                {...item}
                key={item.tooltip}
                tooltipTarget={tooltipTarget}
                menuTarget={menuTarget}
              />
            ))}
          </ul>
        ) : (
          <></>
        )}
```

mattseddon (approved):
Solid improvement on current behaviour. Left some comments.

Observations:


https://user-images.githubusercontent.com/37993418/205520110-aa1f0072-47ca-4676-a65a-44fe547b2790.mov

My current understanding is that CSS cleanup has to be done manually 😢

.vscode/launch.json
shcheklein: > It would be better to add this as a task to tasks.json

any specific reason for this?

> From running the tests it appears that this change has caused a lot of noise in the logs:

looking into this ... I skipped `ul` consciously. To avoid an empty invisible block when there are no items. I'll look how to silent of skip `ul` in a different way

webview/src/shared/components/iconMenu/IconMenu.tsx
shcheklein: thanks, trying it out

shcheklein: done! let's see if that resolves it

.vscode/launch.json
mattseddon: > any specific reason for this?

Yes, the task API is generally what you would use for running dev processes in a more "VS Code way":

```
Tasks in VS Code can be configured to run scripts and start processes so that many of these existing tools can be used from within VS Code without having to enter a command line or write new code. Workspace or folder specific tasks are configured from the tasks.json file in the .vscode folder for a workspace.

```
Docs: https://code.visualstudio.com/docs/editor/tasks

If you are actually using breakpoints with the tests. Please disregard.

shcheklein: > If you are actually using breakpoints with the tests. Please disregard.

Yep, that was the only reason to add this. I'm trying to run pretty everything under debugger - makes it easier to see things right away.

shcheklein:
Thanks @mattseddon for the review and a fixing the noise, looks like it's fine now. Merging ...

PR #2877, opened 2022-12-05
Stabilise colors of running experiments

mattseddon:
# 1/5 `main` <- this <- #2888 <- #2882 <-#2889 <- #2890

Will close https://github.com/iterative/vscode-dvc/issues/2424

This PR changes the behaviour of how we auto-select experiments for plotting. 

We do not select "new" experiments unless:
1. A new experiment is being run in the workspace (we select the workspace)
2. An experiment is being run in the queue (we select the experiment)

If the workspace is selected when an experiment finishes running in the workspace then we move the color from the workspace onto the newly created experiment.

This change means that when the user moves between commits previously unseen experiments will not be automatically selected.

Current behaviour for checkpoints:

https://user-images.githubusercontent.com/37993418/205598332-c1223dfd-5ccd-4e98-bc49-767f116ae742.mov


extension/src/experiments/model/status/collect.ts
```diff
@@ -22,11 +46,11 @@ const collectStatus = (
   unassignColors?: Color[]
 ) => {
   const { id, status } = experiment
-  if (!id || isQueued(status) || hasKey(acc, id)) {
+  if (!id || isQueued(status) || (id !== 'workspace' && hasKey(acc, id))) {
```
mattseddon, line 49: careful with this, need to make sure that workspace can be unselected and not magically selected again

extension/src/experiments/model/status/collect.test.ts
```diff
@@ -14,28 +18,31 @@ describe('collectColoredStatus', () => {
     return mockExperiments
   }
 
-  const collectedColoredStatus = (experiments: Experiment[]) =>
-    collectColoredStatus(
+  it('should set unseen checkpoints and experiments to unselected', () => {
```
mattseddon, line 21: [F] This is a big behaviour change which affects a lot of tests. I have left it included in this PR because breaking it out would have led to changing a lot of the same tests twice.

extension/src/test/suite/plots/index.test.ts
```diff
@@ -169,98 +168,6 @@ suite('Plots Test Suite', () => {
       )
     })
 
-    it('should re-fetch data when moving between branches', async () => {
```
mattseddon, line 172: [F] This is N/A because the experiments will no longer be selected.

extension/src/test/suite/plots/util.ts
```diff
@@ -61,6 +63,17 @@ export const buildPlots = async (
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const experimentsModel: ExperimentsModel = (experiments as any).experiments
+  experimentsModel.setSelected([
+    { id: 'workspace' },
+    { id: 'main' },
+    { id: 'exp-e7a67' },
+    { id: 'test-branch' },
+    { id: 'exp-83425' },
+    { id: 'exp-f13bca' }
+  ] as Experiment[])
```
mattseddon, line 75: [F] Had to do this to keep the test fixture looking the same way for all of the plots tests. This should be updated later.

extension/src/experiments/model/status/collect.ts
```diff
@@ -169,3 +211,100 @@ export const collectSelected = (
...
+        stillRunning
+      )
+    ) {
+      continue
+    }
+
+    if (previouslyRunningExecutor === 'workspace') {
+      collectFinishedWorkspaceExperiment(
+        acc,
+        getMostRecentExperiment(experiments, coloredStatus)
```
mattseddon, line 303: [F] This is an assumption that the most recently created "unseen" experiment is the new one.

extension/src/plots/model/collect.test.ts
```diff
@@ -367,74 +367,71 @@ describe('collectOverrideRevisionDetails', () => {
...
-              id: 'e',
-              label: 'e',
-              logicalGroupName: runningGroup,
-              sha: 'e',
-              status: ExperimentStatus.SUCCESS
-            }
-          ] as Experiment[]
-        }[id])
-    )
+    const { overrideComparison, overrideRevisions } =
```
mattseddon, line 370: [F] Change here is that `unfinishedRunningExperiments` has been moved into `Experiments`

mattseddon: otherwise the tests should be unchanged.

extension/src/test/suite/experiments/model/tree.test.ts
```diff
@@ -358,9 +358,9 @@ suite('Experiments Tree Test Suite', () => {
         {
           displayColor: colors[1],
           fetched: true,
-          group: '[exp-83425]',
-          id: '23250b33e3d6dd0e136262d1d26a2face031cb03',
-          revision: '23250b3'
+          group: '[exp-e7a67]',
```
mattseddon, line 361: [F] Order has changed because I changed status of an experiment

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1106,52 +1106,50 @@ suite('Experiments Test Suite', () => {
       await experiments.isReady()
```
mattseddon, line 1106: [F] All changes due to not selecting experiments by default

shcheklein:
LGTM! we can merge and move forward (I didn't do technical review).

Q: should we show both workspace and experiment as "running"?

Q: should we show an experiment name in the ribbon not workspace?

BTW, even on this demo it visible that there is a delay (it starts showing something after a few epochs).

sroy3 (approved):

julieg18 (approved):
Great work!

PR #2882, opened 2022-12-05
Remove running checkpoint experiment workspace race condition code from plots

mattseddon:


sroy3 (approved):

sroy3:
The number on the plots icon in the experiments table changing is normal right (seen in the UI tests)?

julieg18 (approved):

PR #2888, opened 2022-12-07
Prevent toggle selection of checkpoint experiment running in the workspace

mattseddon:
# 2/5 `main` <- #2877 <- this <- https://github.com/iterative/vscode-dvc/pull/2882 <- #2889 <- #2890

This PR prevents the selection of checkpoint experiments running in the workspace via toggling.

### Demo

https://user-images.githubusercontent.com/37993418/206063679-1534a478-457b-49ee-9348-9f6d1660c12c.mov

The only other path I can think of that would need patching is `Select up to 7 Experiments to Display in Plots` via a quick pick. At this point, I don't think that it is worth the effort and the behaviour could be confusing.

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -16,7 +16,8 @@ import {
   pickFilterToAdd,
   pickFiltersToRemove
 } from './model/filterBy/quickPick'
-import { tooManySelected } from './model/status'
+import { Color } from './model/status/colors'
+import { tooManySelected, UNSELECTED } from './model/status'
 import { starredSort } from './model/sortBy/constants'
 import { pickSortsToRemove, pickSortToAdd } from './model/sortBy/quickPick'
 import { ColumnsModel } from './columns/model'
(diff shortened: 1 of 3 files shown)


shcheklein:
Q: to clarify - does it prevent manual toggle as well? Or the one that happens automatically?

(For manual - why would be the reason to prevent this?)

mattseddon:
> Q: to clarify - does it prevent manual toggle as well? Or the one that happens automatically?
> 
> (For manual - why would be the reason to prevent this?)

It is the manual toggle that I am suppressing. I want to do this because I want to remove the code needed to patch this race condition:

https://user-images.githubusercontent.com/37993418/206080775-9cccad2b-cf97-4e33-94fd-e29be601281a.mov

The workspace (triggered by watchers) will always update quicker than the path of experiment table updates => get new revisions => request revisions

sroy3 (approved):

julieg18 (approved):

PR #2889, opened 2022-12-07
Extract workspace constant

mattseddon:
# 4/5 `main` <-#2877 <- #2888 <- #2882 <- this <- #2890

This PR creates a constant for the string 'workspace' (I have seen + typed that word so many times in the last week).

### DIFF
--- a/extension/src/cli/dvc/contract.ts
+++ b/extension/src/cli/dvc/contract.ts
@@ -51,6 +51,8 @@ export enum ExperimentStatus {
   SUCCESS = 'Success'
 }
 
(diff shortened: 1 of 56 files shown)


extension/src/cli/dvc/contract.ts
```diff
@@ -51,6 +51,8 @@ export enum ExperimentStatus {
   SUCCESS = 'Success'
 }
 
+export const EXPERIMENT_WORKSPACE_ID = 'workspace'
```
mattseddon, line 54: [F] This is the crux of the change

sroy3 (approved):

PR #2892, opened 2022-12-07
Update data when an event is fired for a parent directory (events grouped in Codespaces)

mattseddon:


extension/src/data/index.ts
```diff
@@ -68,9 +70,12 @@ export abstract class BaseData<
...
     return this.dispose.track(
       createFileSystemWatcher(getRelativePattern(this.dvcRoot, '**'), path => {
+        const relPath = relative(this.dvcRoot, path)
         if (
-          this.getWatchedFiles().some(watchedRelPath =>
-            path.endsWith(watchedRelPath)
+          this.getWatchedFiles().some(
+            watchedRelPath =>
+              path.endsWith(watchedRelPath) ||
+              isSameOrChild(relPath, watchedRelPath)
```
mattseddon, line 78: [F] If the parent directory has changed then we can assume that the watched path has changed.

Here is the example from `magnetic-tiles-defect`:

Plots folder structure looks like this:

```
└── training
    ├── metrics.json
    └── plots
        └── metrics
            ├── dice.tsv
            ├── eval
            │   └── loss.tsv
            ├── jaccard_coeff.tsv
            └── train
                └── loss.tsv
```

We get only the following event during an experiment:

```
[trace] [File Watcher (parcel)] [ADDED] /workspaces/magnetic-tiles-defect/training/metrics.json
[trace] [File Watcher (parcel)] [ADDED] /workspaces/magnetic-tiles-defect/training/plots
```

That meant that we missed the event because none of the "watched paths" had been changed.

extension/src/test/suite/plots/data/index.test.ts
```diff
@@ -211,13 +214,38 @@ suite('Plots Data Test Suite', () => {
...
+      bypassProcessManagerDebounce(mockNow, 2)
+
+      const secondDataUpdatedEvent = new Promise(resolve =>
+        disposable.track(data.onDidUpdate(() => resolve(undefined)))
+      )
+
+      const absParentDirectory = join(dvcDemoPath, parentDirectory)
+
+      if (getProcessPlatform() === 'win32') {
+        removeDir(absParentDirectory)
```
mattseddon, line 238: `utimes` will not update a directory on Windows (assumption from the test failing). I have gone with this destructive action for now. If there are ever any issues we can come back and skip this test for Windows instead.

sroy3 (approved):

Issue #2793, opened 2022-11-17
Story: Improve Onboarding


julieg18:
We need to make what the product has to offer more clear when it's first installed and help users who don't have a DVC project set up already to get started with experiments/plots.

### Steps

- [x] Improve walkthrough (#2573, #2588)
- [x] #2683
- [x] #2776
- [x] https://github.com/iterative/vscode-dvc/issues/2765
- [x] Look into assisting users with installing DVC. How Jupiter notebooks install packages for users is one example to look into. https://github.com/iterative/vscode-dvc/issues/2829
- [x] Help users get started with their own experiments or plots We could use some kind of template for projects or use the upcoming changes to DVCLive to generate experiments. (See discussion in #2608)
- [x] Review Get Started - https://github.com/iterative/vscode-dvc/issues/2725
- [x] #2947 - setup onboarding for users who do not have project
- [ ] https://github.com/iterative/vscode-dvc/issues/3021
- [ ] Review, including the messaging (e.g. Global - can users understand?, we don't need details like `exp_save`, etc) #3057

mattseddon:
We need to take into consideration the details here: https://github.com/iterative/dvclive-exp-tracking for the way the onboarding flow needs to look. 

I think there need to be two separate options when there is no project:
1. I want to init a new project
2. I want to instrument my code

cc @sroy3

shcheklein:
2 does require 1 though? (it requires `dvc init`)

mattseddon:
I am wrong

mattseddon:
When we detect a user has a global install of the CLI and no project or data we can put a screen in front of the CLI unavailable screen. This screen will look like this:

### No Project

![image](https://user-images.githubusercontent.com/37993418/208808978-a0b1109c-f244-4a1b-8544-74b5fbdb4761.png)

### With Project 

![image](https://user-images.githubusercontent.com/37993418/208808934-f3698e8e-9fa7-4e10-b935-6557504035f4.png)

Both of the buttons with the "secondary appearance" (one of which is highlighted below) will result in onboarding being dismissed for the remainder of the session (or we could have the state for future sessions).

![image](https://user-images.githubusercontent.com/37993418/208809149-ebd6b29f-2380-45a4-ab41-ded2ac1a4359.png)

We also need to add the above Dismiss button + Don't Show Again to the project has no data screen:

[...]

PR #2894, opened 2022-12-07
Dynamic Get Started webview

sroy3:
Part of #2793 

https://user-images.githubusercontent.com/3683420/206229995-b9ad4892-9020-4c84-810b-111657d95ab3.mov


webview/src/getStarted/components/ProjectUninitialized.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const ProjectUninitialized: React.FC<ProjectUninitializedProps> = ({
+  initializeProject
+}) => {
+  return (
+    <EmptyState>
+      <div>
+        <h1>DVC is not initialized</h1>
+        <p>
+          The currently open folder does not contain a DVC project. You can
```
mattseddon, line 17: [C] Could swap `The currently open folder` for `The current workspace`

extension/src/extension.ts
```diff
@@ -443,9 +442,26 @@ export class Extension extends Disposable implements IExtension {
...
   }
 
+  public getAvailable() {
+    return this.cliAccessible
+  }
+
+  public async initProject() {
+    const root = getFirstWorkspaceFolder()
+    if (root) {
+      await this.dvcExecutor.init(root)
```
mattseddon, line 456: [Q] I assume the reason for `willRecheck` is that we may need to wait a little bit after initialization before checking to see if the workspace is ready to go. Could we move the delay out here with an explanatory variable? Would that make the code more obvious? Or are you trying to make the UX seamless in that the user doesn't necessarily have to wait?

I think it would be ok to make everyone wait as long as we displayed a progress notification (see `Toast.showProgress`) that overall lastest for ~5 seconds. 

E.g => "Creating Project" => "Verifying Contents" => "Notifying Extension". After writing this I am not thinking that maybe we could actually check for the existence of the `.dvc` folder every 500ms. Once it is in place we could then fire `this.workspaceChanged.fire()`. That might actually simplify everything.

LMK what you think, feel free to disagree, ask clarifying questions, change no or follow up. Up to you.

mattseddon (approved):

extension/src/extension.ts
sroy3: `willRecheck` is not for the after initialization. It's to check whether DVC is installed or not. If the project has a virtual environment, the watchers are going to catch when DVC gets installed later. If on setup DVC is not installed and you globally install it later, the extension won't be aware of it, unless we check every 5 seconds for it. It's inside it's own function so that it's easily mockable for the tests.

webview/src/getStarted/components/ProjectUninitialized.tsx
sroy3: Will change it in both places (sidebar and webview)

PR #2899, opened 2022-12-08
Update demo project

mattseddon:
Depends on https://github.com/iterative/vscode-dvc-demo/pull/27

Wasted the best part of a day on this. Changing `.gitignore`s in the submodule led to some strange ~caching~ issue on checkout and we ended up with a broken-looking repo from an [SCM point of view](https://github.com/iterative/vscode-dvc-demo/pull/27).

The repo should look like this after an experiment:

image

But I was seeing this from the e2e tests (remote only):

Source Control View - should show the expected changes after running an experiment

~The fix is to manually scrub the git cache for the submodule in the CI.~ Looks like something in the code coverage test suite (unit or integration) is messing with the repo state in the CI. I've split the end-to-end test out into its own environment. This will give quicker run times and avoid situations like this in the future.

### DIFF
--- a/.github/workflows/end-to-end-test.yml
+++ b/.github/workflows/end-to-end-test.yml
@@ -1,4 +1,4 @@
-name: Continuous Integration
+name: End To End Test
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


sroy3 (approved):

julieg18 (approved):
Looks good!

PR #2911, opened 2022-12-09
Make the size of the welcome screen in the sidebar grow

sroy3:


sroy3:
Nevermind, this isn't consistent/doesn't work. My welcome view was always really small, I thought I'd fixed it with this code, but then when I reverted to take a screenshot of before, I couldn't reproduce anymore. I guess I had minimized the view and it then stuck.

PR #2915, opened 2022-12-12
Only show plots which are available for selected revisions

mattseddon:
Currently, the plots path tree works additively. Every time that a new revision is collected we check to see which plots have been returned and add newly found paths to the existing ones. This can become an issue where we have unstable plots between revisions. E.g in the hackathon project where each experiment generates a list of misclassified images. These images will rarely match between experiments. 

The solution in this PR is to collect which plots are associated with which revisions. We can then display only the available paths for each of the selected revisions.

### Demo

https://user-images.githubusercontent.com/37993418/206943556-34ee6aa4-2065-4627-8d4d-a416eaa67054.mov

https://user-images.githubusercontent.com/37993418/207228331-46f489a9-1d1d-4ba4-a6c7-2cf25bec3b72.mov


extension/src/path/selection/model.ts
```diff
@@ -49,33 +49,35 @@ export abstract class PathSelectionModel<
       .map(element => ({ ...element, selected: !!this.status[element.path] }))
   }
 
-  public toggleStatus(path: string) {
+  public toggleStatus(path: string, ...args: unknown[]) {
```
mattseddon, line 52: Wrap `getTerminalNodes`, `getTerminalNodeStatuses`, etc in `PathsModel`  so that we don’t have to mess with the base class

mattseddon: Can't maintain this class without type safety.

extension/src/experiments/columns/model.ts
```diff
@@ -91,6 +91,20 @@ export class ColumnsModel extends PathSelectionModel<Column> {
     })
   }
 
+  public getTerminalNodes(): (Column & { selected: boolean })[] {
```
mattseddon, line 94: [F] Both these methods have diverged

webview/src/plots/components/GetStarted.tsx
```diff
@@ -17,7 +17,7 @@ export const AddPlots: React.FC<AddPlotsProps> = ({
   <div>
     <NoPlotsText />
     <div>
-      {!hasSelectedPlots && (
+      {hasSelectedRevisions && !hasSelectedPlots && (
```
mattseddon, line 20: [F] If there are no selected revisions then there cannot be any plots

sroy3 (approved):

julieg18:
Looks like Storybook created an incorrect screenshot:

image

Is it just a one-time thing or was it caused by the pr 🤔

mattseddon:
> Looks like Storybook created an incorrect snapshot:
> 
> image
> 
> Is it just a one-time thing or was it caused by the pr 🤔

This reflects the new behaviour

PR #2923, opened 2022-12-12
Use DVCLive signal file to indicate that an experiment is running in the workspace

mattseddon:
As part of the onboarding drive, the DVC team has made it possible to run experiments using only DVCLive. In order for a new user (with a non-DVC ML project) to see the table and associated live plots, all they will need to do is:

1. Install DVC and DVCLive as Python packages. 
1. Run `dvc init`.
1. Instrument their training script with DVCLive calls/callbacks.
1. Run their training script.

This means that we will have a new way to onboard people into DVC. However, due to the changes made in #2877 the newly created experiments were no longer being auto-selected for plotting.

In order to rectify this I raised https://github.com/iterative/dvclive/pull/392 which has DVCLive create a signal file whenever one of these special DVCLive-only experiments is being run.

This PR uses the above changes to auto-select experiments run by DVCLive only (outside of a normal experiment) for plotting.

### Demo

https://user-images.githubusercontent.com/37993418/207185448-6a3fa798-73e1-4b5e-b0a1-1dcee2c791ca.mov

[Original slack discussion](https://iterativeai.slack.com/archives/C01APS0FHDM/p1670530599513299)

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -1614,6 +1614,7 @@
     "lint-staged": "13.1.0",
     "mocha": "10.0.0",
     "mock-require": "3.0.3",
+    "process-exists": "4.1.0",
     "shx": "0.3.4",
     "sinon": "15.0.0",
     "sinon-chai": "3.7.0",
--- a/extension/src/experiments/model/index.test.ts
+++ b/extension/src/experiments/model/index.test.ts
@@ -76,15 +76,73 @@ describe('ExperimentsModel', () => {
 
   it('should return the expected rows when given the output fixture', () => {
     const model = new ExperimentsModel('', buildMockMemento())
-    model.transformAndSet(outputFixture)
+    model.transformAndSet(outputFixture, false)
     expect(model.getRowData()).toStrictEqual(rowsFixture)
   })
 
   it('should return the expected rows when given the survival fixture', () => {
     const model = new ExperimentsModel('', buildMockMemento())
-    model.transformAndSet(survivalOutputFixture)
+    model.transformAndSet(survivalOutputFixture, false)
 
     expect(model.getRowData()).toStrictEqual(survivalRowsFixture)
+  })
+
+  it('should set the workspace to running if a signal file for a DVCLive only experiment has been found', () => {
+    const model = new ExperimentsModel('', buildMockMemento())
+    const dvcLiveOnly = {
+      [EXPERIMENT_WORKSPACE_ID]: {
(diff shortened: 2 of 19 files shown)


extension/package.json
```diff
@@ -1614,6 +1614,7 @@
     "lint-staged": "13.1.0",
     "mocha": "10.0.0",
     "mock-require": "3.0.3",
+    "process-exists": "4.1.0",
```
mattseddon, line 1617: This is the last non-esm only version. Renovate will complain, we will ignore it.

extension/src/experiments/model/index.test.ts
```diff
@@ -76,17 +76,75 @@ describe('ExperimentsModel', () => {
...
 
   it('should return the expected rows when given the survival fixture', () => {
     const model = new ExperimentsModel('', buildMockMemento())
-    model.transformAndSet(survivalOutputFixture)
+    model.transformAndSet(survivalOutputFixture, false)
 
     expect(model.getRowData()).toStrictEqual(survivalRowsFixture)
   })
 
+  it('should set the workspace to running if a signal file for a DVCLive only experiment has been found', () => {
```
mattseddon, line 90: [F] This test is the addition in this file. The diff has gone a bit wacky further down.

extension/src/processExecution.ts
```diff
@@ -58,3 +59,6 @@ export const executeProcess = async (
   const { stdout } = await createProcess(options)
   return stdout
 }
+
+export const processExists = (pid: number): Promise<boolean> =>
```
mattseddon, line 63: [F] Gives type safety that we pass a number to the original `processExists`. This is important because (at least on M1) giving the string version of a valid and running numeric pid returns `false` 😱

extension/src/test/suite/util.ts
```diff
@@ -172,6 +173,10 @@ export const buildDependencies = (
     'createFileSystemWatcher'
   ).returns(mockDisposable)
 
+  const mockCheckSignalFile = stub(FileSystem, 'checkSignalFile').resolves(
+    false
+  )
```
mattseddon, line 178: [F] We don't gain much by reading the filesystem for each of these tests so I've mocked out the call.

sroy3 (approved):

mattseddon:
> Thank you @mattseddon! The blog post and release are planned for 12/15 FYI.

@dberenbaum will get this released today.

Issue #2457, opened 2022-09-24
Improve experiments naming


shcheklein:
One of the common feedbacks we are getting is that experiment names are not meaningful w/o metrics in the side panel and in the plots webview.

We need to introduce:

- columns that we can pin (metrics, params, data) and get this information into experiment names, or replaces hash
- names - let's try to generate auto names? similar to how codespace does this. Some short adj + short noun
- tooltips with the full list of metrics, params, full hash, etc

shcheklein:
Related https://github.com/iterative/vscode-dvc/issues/2395

shcheklein:
Related https://github.com/iterative/vscode-dvc/issues/1820#issuecomment-1253154206

aschuh-hf:
Apart from improving the display of auto-generated names, would it make sense to allow the user to name an experiment themselves when creating it? E.g., using `Modify & Queue`. Right now, the UI goes through a sequence of questions to ask for the different parameters values one indicated wanting to modify for the experiment. Why not also ask (maybe as final step) what to name the experiment, or to keep the auto-generated default otherwise?

daavoo:
> names - let's try to generate auto names? similar to how codespace does this. Some short adj + short noun

I added a random name generator in DVCLive, mostly out of necessity for the Python-only experiments scenario (it was simpler than changing how DVC currently generates the exp hashes):

https://github.com/iterative/dvclive/pull/366/commits/53df56748b0144bfd5c10390cc83a3b3085078a7

julieg18:
With tooltips added to the plots ribbon blocks (#2924) and DVC updating experiment names, are we ready to close this issue or is there more to be done?

shcheklein:
@julieg18 yes, let's close this when the final part is merged!

PR #2924, opened 2022-12-13
Add columns tooltip to plots ribbon

julieg18:
* adds tooltips to each experiment ribbon block, showing the first three columns

## Demo

https://user-images.githubusercontent.com/43496356/207654132-bafc499c-1e47-4d93-b8ea-b0c2af22807a.mov

Part of #2457


julieg18:
image

Weird, code climate seems to be having connection issues 🤔

extension/src/plots/model/util.ts
```diff
@@ -0,0 +1,20 @@
+import { getDataFromColumnPath } from '../../experiments/model/util'
+import { Experiment } from '../../experiments/webview/contract'
+
+export const getRevisionFirstThreeColumns = (
```
sroy3, line 4: In `experiments/model/tree.ts` `getTooltipTable` you have something very similar. Should these two function share some code?

webview/src/plots/components/ribbon/RibbonBlock.tsx
```diff
@@ -23,7 +23,20 @@ export const RibbonBlock: React.FC<RibbonBlockProps> = ({
 }) => {
   const exp = revision.group?.replace(/[[\]]/g, '') || revision.revision
 
-  return (
+  const tooltipContent = (
+    <table className={styles.columnsTable}>
+      <tbody>
+        {revision.firstThreeColumns.map(({ path, value }, ind) => (
+          <tr key={ind}>
```
sroy3, line 30: Might be better to use the path as the key as the order can change, but the path won't. It's not that expensive to let the row re-render, but it's still good practice to avoid using the order as keys.

sroy3 (commented):
Screenshot 2022-12-13 at 3 47 32 PM

There is some kind of weird indentation happening with each row

webview/src/plots/components/ribbon/RibbonBlock.tsx
julieg18: Done!

extension/src/plots/model/util.ts
julieg18: Looking at the two functions, they do similar things but I'm not seeing a way to share code since they take the results from `getDataFromColumn` and do different things to the returned values.

julieg18:
> There is some kind of weird indentation happening with each row

Thanks for spotting that! Fixed!

sroy3 (approved):

Issue #2765, opened 2022-11-13
When there is no pipeline, no params, deps, metrics we should show empty table screen


shcheklein:
Related #2725 #2608 

Can be tested on DVC repo itself.

<img width="817" alt="Screen Shot 2022-11-13 at 12 48 22 PM" src="https://user-images.githubusercontent.com/3659196/201543879-b59f4258-a52b-49e5-895a-5e314e3cf33c.png">

If we don't show the table in this case it will allow us to show some message on how to get started.

Probably should be also fixed on the DVC side cc @dberenbaum

dberenbaum:
Not sure how important it is in CLI unless/until we have some more useful hint to show here. For VS Code, I guess having this detected by CLI would be the easiest way to know whether to show the message?

shcheklein:
I think we are fine on VS Code side - we can detect that it's empty. But yes, it would be more beneficial for DVC to show some onboarding message vs an empty table that is not very meaningful.

sroy3:
@dberenbaum @daavoo are there any code snippets we could show the user in that case? Currently, we are showing a blank screen saying the project has no data, but we'd like to help the user get setup as quickly as possible.

daavoo:
> @dberenbaum @daavoo are there any code snippets we could show the user in that case? Currently, we are showing a blank screen saying the project has no data, but we'd like to help the user get setup as quickly as possible.

Maybe show something like the tabs in Section 3 of [the WIP blog post](https://iterative-ai-blog-exp-t-qktxw1.herokuapp.com/blog/exp-tracking-dvc-python?tab=DVC-extension-for-VS-Code#how-to-start):


https://user-images.githubusercontent.com/12677733/207657642-eb3112fc-e2f9-4768-9751-085160970a5c.mov

dberenbaum:
> @dberenbaum @daavoo are there any code snippets we could show the user in that case? Currently, we are showing a blank screen saying the project has no data, but we'd like to help the user get setup as quickly as possible.

When is it shown? I still see a the table as soon as I make a commit

<img width="417" alt="Screenshot 2022-12-15 at 9 40 25 AM" src="https://user-images.githubusercontent.com/2308172/207889190-5ba60b92-c14a-44b9-9d6d-5b8206644e93.png">

What are the options for what we could embed here? Text? Video? Can there be multiple choices?

[...]

PR #2927, opened 2022-12-13
Show a screen when there is no data in the project

sroy3:
Part of #2765

https://user-images.githubusercontent.com/3683420/207441620-108daa67-39d8-4fcc-87e1-33097426f51b.mov

The page now shows more information:

https://user-images.githubusercontent.com/3683420/208148327-fe3bccae-c03b-4ea4-b72c-e16991419ec1.mov


webview/src/getStarted/components/NoData.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { CodeSlider } from '../../shared/components/codeSlider/CodeSlider'
+import { EmptyState } from '../../shared/components/emptyState/EmptyState'
+
+export const NoData: React.FC = () => {
+  return (
+    <EmptyState>
+      <div>
+        <h1>Your project contains no data</h1>
+        <div>
+          Enable DVC experiment tracking using{' '}
```
julieg18, line 16: Should we insert a sentence or two about running DVC experiments the usual way? Or would that make things look complicated 🤔

webview/src/getStarted/components/NoData.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { CodeBlock } from '../../shared/components/codeBlock/CodeBlock'
+import { CodeSlider } from '../../shared/components/codeSlider/CodeSlider'
+import { EmptyState } from '../../shared/components/emptyState/EmptyState'
+
+export const NoData: React.FC = () => {
+  return (
+    <EmptyState>
+      <div>
+        <h1>Your project contains no data</h1>
+        <div>
```
julieg18, line 15: For accessibility purposes, we should probably wrap the text in a `p` tag.

webview/src/shared/components/codeSlider/CodeSlider.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+interface CodeBlockWithTitle extends CodeBlockProps {
+  title: string
+}
+
+interface CodeSliderProps {
+  codeBlocks: CodeBlockWithTitle[]
+}
+
+export const CodeSlider: React.FC<CodeSliderProps> = ({ codeBlocks }) => {
```
julieg18, line 14: Is there an issue with using [`VSCodePanels`](https://github.com/microsoft/vscode-webview-ui-toolkit/blob/main/src/panels/README.md)? I'd assume we'd want to use them to make our web views as "vscode-like" as possible.

webview/src/shared/components/codeSlider/CodeSlider.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const [active, setActive] = useState(0)
```
julieg18, line 15: Can we change the tabs on hover (cursor, highlight text, add border on hover for example)? 

Probably minor, but there's also this weird kind of jumping that happens when switching tabs:

https://user-images.githubusercontent.com/43496356/208214297-ef4fd517-8e8e-4986-b522-2db9ea457157.mov

webview/src/shared/components/codeSlider/CodeSlider.tsx
julieg18: Would probably resolve the comment about tabs improvements as well.

julieg18 (approved):
Have some minor comments and questions, but I think it's definitely good enough for a first iteration. Great work!

extension/src/experiments/index.ts
```diff
@@ -517,6 +522,7 @@ export class Experiments extends BaseRepository<TableData> {
 
   private notifyColumnsChanged() {
     this.columnsChanged.fire()
+    setContextValue('dvc.project.hasData', this.getHasData())
```
mattseddon, line 525: This should copy the pattern we use for `hasCheckpoints` in `WorkspaceExperiments`:

```
      onDidChangeCheckpoints(() => {
        const workspaceHasCheckpoints = Object.values(this.repositories).some(
          experiments => experiments.hasCheckpoints()
        )

        setContextValue('dvc.experiment.checkpoints', workspaceHasCheckpoints)
      })
```

extension/src/experiments/workspace.ts
```diff
@@ -329,6 +329,15 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
     )
   }
 
+  public getHasData(dvcRoots: string[]) {
```
mattseddon, line 332: This shouldn't need `dvcRoots` as it has the `repositories` object available. Could use this:
```
  public hasData() {
    return Object.values(this.repositories).some(repository =>
      repository.hasData()
    )
  }

```
or 

```
    for (const repository of Object.values(this.repositories)) {
      if (repository.getHasData()) {
        return true
      }
    }
    return false
  }
```

mattseddon (approved):

webview/src/getStarted/components/NoData.tsx
sroy3: I copied this off the blog post. I also think it might be confusing the steps together as this would come after. From there I think it'd be easier to show than write about it.

webview/src/getStarted/components/NoData.tsx
sroy3: I know, it's only because of `<CodeBlock language="python" inline>save_dvc_exp=True</CodeBlock>` which uses `<div>` inside.

webview/src/shared/components/codeSlider/CodeSlider.tsx
sroy3: The data is centered that's why there is some jumping. I was thinking about it this weekend. I can probably make the code block the same height (through a wrapper). That might look weirder as it will be off-centered, but maybe it won't.

webview/src/shared/components/codeSlider/CodeSlider.tsx
sroy3: Didn't know about `VSCodePanels` will try it.

webview/src/getStarted/components/NoData.tsx
julieg18: The `div` on line 27 of `CodeBlock.tsx`? Why not replace that with a `span` and apply any needed styles to it with CSS?

sroy3: Because styling is done by the lib.

extension/src/experiments/index.ts
sroy3: As in using a variable for the second parameter? Or do you mean that the whole thing is not where it should be?

webview/src/shared/components/codeSlider/CodeSlider.tsx
sroy3: Done

extension/src/experiments/index.ts
sroy3: I figured it out, will be in the follow up

PR #2928, opened 2022-12-13
Recheck for global CLI becoming available

mattseddon:
Paired on this solution this morning with @sroy3 

Follow up from [#2894 (comment)](https://github.com/iterative/vscode-dvc/pull/2894/files#r1045301489)

If the CLI cannot be located when the extension runs setup for the first time then we need to recheck periodically if the extension has become available globally. The previous implementation would recursively call `setup` every 5 seconds if the CLI could not be found on the first pass. This had a couple of issues:

- There are 4 different things that trigger setup and each of these checked every 5 seconds if they don't find the correct thing (including manual setup the workspace). This could have led to multiple redundant calls.
- The code inside `setup` was unnecessarily complicated

The solution is to split out the global recheck from the rest of the setup code and only call it once at the extension's instantiation.

### Demo

https://user-images.githubusercontent.com/37993418/207483094-af9313c5-99ee-4792-99cc-71038a8abd67.mov

### DIFF
--- a/extension/src/setup.test.ts
+++ b/extension/src/setup.test.ts
@@ -1,6 +1,6 @@
 import { resolve } from 'path'
 import { extensions, Extension, commands } from 'vscode'
-import { setup, setupWorkspace } from './setup'
+import { setup, setupWithGlobalRecheck, setupWorkspace } from './setup'
 import { flushPromises } from './test/util/jest'
 import {
   getConfigValue,
@@ -24,6 +24,7 @@ import {
   MIN_CLI_VERSION
 } from './cli/dvc/constants'
 import { extractSemver, ParsedSemver } from './cli/dvc/version'
+import { delay } from './util/time'
 
 jest.mock('vscode')
 jest.mock('./vscode/config')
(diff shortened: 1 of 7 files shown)


sroy3 (approved):

extension/src/setup.test.ts
```diff
@@ -618,3 +620,102 @@ describe('setup', () => {
...
+    isPythonExtensionUsed: mockedIsPythonExtensionUsed,
+    resetMembers: mockedResetMembers,
+    setAvailable: mockedSetAvailable,
+    setCliCompatible: mockedSetCliCompatible,
+    setRoots: mockedSetRoots,
+    setupWorkspace: mockedSetupWorkspace,
+    unsetPythonBinPath: mockedUnsetPythonBinPath
+  }
+
+  it('should periodically check to see if the CLI is available globally if the extension fails to initialize', async () => {
```
mattseddon, line 640: Jest's fake timers do not work with recursive calls to setTimeout 🤸🏻 (spent a couple of hours on this).

PR #2934, opened 2022-12-14
Prevent unnecessary CLI calls in the integration test suite

mattseddon:
After seeing #2929 fail numerous times on 🪟s I went down the rabbit hole of trying to completely clean our integration test logs (and thus clean up unexpected behaviour). Found some very weird behaviour and dead code but think I got to the bottom of it all. Comments inline.

### DIFF
--- a/extension/src/experiments/data/index.ts
+++ b/extension/src/experiments/data/index.ts
@@ -66,10 +66,6 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
     return this.notifyChanged(data)
   }
 
-  public forceUpdate() {
-    return this.processManager.forceRunQueued()
-  }
-
   protected collectFiles(data: ExperimentsOutput) {
     this.collectedFiles = collectFiles(data, this.collectedFiles)
   }
--- a/extension/src/processManager.ts
+++ b/extension/src/processManager.ts
@@ -55,17 +55,6 @@ export class ProcessManager extends Disposable {
     return this.runQueued()
   }
 
-  public async forceRunQueued(): Promise<void> {
-    const next = this.getNextFromQueue()
-    if (!next) {
-      return
-    }
-
-    const { process } = this.processes[next]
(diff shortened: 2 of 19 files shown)


extension/src/experiments/data/index.ts
```diff
@@ -66,10 +66,6 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
     return this.notifyChanged(data)
   }
 
-  public forceUpdate() {
```
mattseddon, line 69: [F] All the force update stuff is was dead.

extension/src/processManager.ts
```diff
@@ -55,17 +55,6 @@ export class ProcessManager extends Disposable {
     return this.runQueued()
   }
 
-  public async forceRunQueued(): Promise<void> {
```
mattseddon, line 58: [F] Only place this was called was a test. I was in this class in the first place to try and stop the queue from being run after it had been disposed of.

extension/src/workspace/index.ts
```diff
@@ -38,7 +39,9 @@ export abstract class BaseWorkspace<
   }
 
   public reset(): void {
-    this.repositories = reset<T>(this.repositories, this.dispose)
```
mattseddon, line 41: [F] I wanted to be sure that `this.dispose` was not unbounded

extension/src/test/suite/extension.test.ts
```diff
@@ -447,6 +449,7 @@ suite('Extension Test Suite', () => {
     })
 
     it('should set the dvc.cli.incompatible context value', async () => {
+      stub(Watcher, 'createFileSystemWatcher').returns(mockDisposable)
```
mattseddon, line 452: [F] Don't want to accidentally setup all the watchers here.

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #2939, opened 2022-12-14
Do not rely on number of call to execute process stub in integration tests

mattseddon:
We can no longer rely on a stub of `executeProcess` not being called inadvertently. This is because we have a process checking if the CLI has become available globally running in the background when the first setup fails.

This was the root cause of Windows failure here: https://github.com/iterative/vscode-dvc/actions/runs/3699529541/jobs/6267003237


sroy3 (approved):

julieg18 (approved):
Looks good!

Issue #2925, opened 2022-12-13
Add signal about which DVC is being used


daavoo:
The extension supports multiple options for `Setup the Workspace`:  `Auto`, `Manual`, `Global`.

However, there is no signal to identify which option is being currently being used by the extension (beyond inspecting the `Output` of the extension where each call to dvc might include the full path to the venv).

I think the DVC extension could show a signal in the current status bar (i.e. extending display name from the current `DVC` to `DVC (Auto)`).

It might also be useful for the `Auto` and `Manual` options to display the name of the actual venv used similar ,
to how the Python extension does it:

<img width="269" alt="Captura de Pantalla 2022-12-13 a las 18 22 49" src="https://user-images.githubusercontent.com/12677733/207401634-ad3ae9ae-234c-47c9-831f-4f35c99aff2a.png">

SoyGema:
as food for thought, this **signal** might have several forms:

1 ) A **downright popup** that includes the python path and version used by the extension and the virtual environment. 
2 ) The python version next to the **circle UI** in bottom left next to DVC icon .  

This could have an impact on DVCLive experiments integration. Python version is showed in python notebook and it´s recommendable to have an indication that is correctly synced with DVC.

Thanks for the time dedicated to this 👍

mattseddon:
Semi-related.

As part of improving the onboarding experience, we are working on auto-installing DVC / DVCLive as Python packages if they cannot be located. 

@daavoo @SoyGema what are your initial thoughts on this screen:

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/208002954-7e445116-2b83-42ed-803f-308084b39c8c.png">

SoyGema:
Looks great in my opinion! 

As a user , and taking into account left to right reading scenario, I would first **setup the workspace** and then **Select the python interpreter.**  Unclear to me if this is the happy path or it is the [other way around ](https://iterativeai.slack.com/archives/C01APS0FHDM/p1670842909689889?thread_ts=1670842047.540029&cid=C01APS0FHDM) as @daavoo mentioned . 



Regarding what happens once you setup both the **WorkSpace** and the Python Interpreter, It might be great to be shown something like this! 
<img width="1728" alt="HOW could BE" src="https://user-images.githubusercontent.com/24204714/208080036-080a8c95-135f-4e10-9813-97fad1464983.png">



**Regarding impact in DVCLive experiments :** 
[...]

PR #2944, opened 2022-12-15
Provide option to auto-install DVC if unavailable

mattseddon:
# 1/2 `main` <- this <- #2990 

This PR adds the option to auto-install DVC from the Setup webview if a Python interpreter can be found on the system.

Before shipping this change I will need to integrate the Setup webview with the current process of showing toast pop-ups when initialization fails.

#### Old

https://user-images.githubusercontent.com/37993418/208381006-502b619c-9b25-4433-b035-809d322863fa.mov

#### Current

https://user-images.githubusercontent.com/37993418/209023548-8413ded9-772e-4e8e-b206-1abfbbc6bd7c.mov

~Asked for some early feedback in #2925~


extension/src/experiments/workspace.ts
```diff
@@ -329,9 +329,19 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
...
+
+    if (repositories.some(repository => repository.getHasData())) {
+      return true
+    }
+
+    if (repositories.some(repository => repository.getHasData() === false)) {
+      return false
+    }
+
+    return allLoading
```
mattseddon, line 344: [F] Needed for the edge case that the current project has data but CLI needs to be installed. I.e [this](https://github.com/iterative/vscode-dvc/pull/2944/files#diff-f1285e681b370f15ccca1f8455d2009a0dcde397464d0e5092282a932ad4962cR78)

extension/src/extension.ts
```diff
@@ -241,6 +241,7 @@ export class Extension extends Disposable implements IExtension {
       this.config.onDidChangeExecutionDetails(async () => {
         const stopWatch = new StopWatch()
         try {
+          this.changeSetupStep()
```
mattseddon, line 244: [F] Will update the details displayed on the Setup screen for where the CLI will be installed.

extension/src/setup/index.ts
```diff
@@ -60,17 +60,21 @@ export class Setup extends BaseRepository<TSetupData> {
       return
     }
 
+    const pythonBinPath = await findPythonBinForInstall()
+
     this.webviewMessages.sendWebviewMessage(
       cliAccessible,
       projectInitialized,
+      isPythonExtensionInstalled(),
+      getPythonBinDisplayText(pythonBinPath),
```
mattseddon, line 69: [I] Move this

extension/src/setup/webview/messages.ts
```diff
@@ -21,19 +27,53 @@ export class WebviewMessages {
...
-      projectInitialized
+      isPythonExtensionInstalled,
+      projectInitialized,
+      pythonBinPath
     })
   }
 
   public handleMessageFromWebview(message: MessageFromWebview) {
     if (message.type === MessageFromWebviewType.INITIALIZE_PROJECT) {
       return this.initializeProject()
```
mattseddon, line 45: [I] This still needs analytics attached. Probably should be done in the extension.

webview/src/setup/components/CliUnavailable.tsx
```diff
@@ -1,16 +1,89 @@
...
+}
+
+const OfferToInstall: React.FC<{
+  children: React.ReactNode
+  pythonBinPath: string
+  installDvc: () => void
+}> = ({ installDvc, pythonBinPath, children }) => (
+  <div>
+    <p>
+      {`DVC & DVCLive can be auto-installed as packages with ${pythonBinPath}`}
```
julieg18, line 20: Technically, I don't think we have to wrap the entire string with curly braces: 
```suggestion
      DVC & DVCLive can be auto-installed as packages with {pythonBinPath}
```

extension/src/setup/autoInstall.ts
```diff
@@ -0,0 +1,64 @@
+import { getPythonExecutionDetails } from '../extensions/python'
```
julieg18, line 1: When I install DVC but the project isn't initialized, it takes me straight to the "Your project has no data" page. Shouldn't we be showing a message saying you need to initialize DVC?

julieg18: Ah, wait, I'm seeing this behavior on main as well.

julieg18 (approved):
Great work!

extension/src/setup/autoInstall.ts
mattseddon: 😱

mattseddon: @julieg18 I cannot recreate this:

https://user-images.githubusercontent.com/37993418/209230803-d7414ba9-6928-4f2f-8575-9be905d56438.mov

Are you sure the workspace that you were testing against didn't contain a DVC project?

julieg18: > @julieg18 I cannot recreate this:

Probably an issue on my end. 

> Are you sure the workspace that you were testing against didn't contain a DVC project?

Yes, pretty sure. It shows no file when I try `find . -name .dvc`. I'm getting this error when the extension or I try running a dvc command:

```
ERROR: /Users/juliegalvan is not a git repository
```

mattseddon: that scenario is covered by #2947

mattseddon: for anyone that arrives here: debugged on a call with @julieg18 

there was a `.dvc` folder under `/Users/juliegalvan` which was causing some very unexpected/confusing behaviour. After deleting that folder everything was resolved. An issue will be raised in the DVC repo to discuss

julieg18: Issue created! https://github.com/iterative/dvc/issues/8731

PR #2948, opened 2022-12-15
Ensure that all watcher's listeners are disposed

mattseddon:
See comment inline


extension/src/fileSystem/watcher.ts
```diff
@@ -34,18 +34,18 @@ export const fireWatcher = (path: string): Promise<void> => {
...
   const fileSystemWatcher = workspace.createFileSystemWatcher(glob)
-  fileSystemWatcher.onDidCreate(uri => listener(uri.fsPath))
-  fileSystemWatcher.onDidChange(uri => listener(uri.fsPath))
-  fileSystemWatcher.onDidDelete(uri => listener(uri.fsPath))
-
-  return fileSystemWatcher
+  track(fileSystemWatcher)
+  track(fileSystemWatcher.onDidCreate(uri => listener(uri.fsPath)))
+  track(fileSystemWatcher.onDidChange(uri => listener(uri.fsPath)))
+  track(fileSystemWatcher.onDidDelete(uri => listener(uri.fsPath)))
```
mattseddon, line 50: [F] The difference here is that previously I expected that disposing the `fileSystemWatcher` meant that all of the listeners were disposed as well. I now think that this might not actually be the case. Just to be sure I have started tracking the listeners.

sroy3 (approved):

julieg18 (approved):
Great work!

PR #2951, opened 2022-12-16
Add step into setup the workspace which reveals the current interpreter

mattseddon:
Related to #2944 & #2925

This PR adds a step in "Setup the Workspace" (which we might remove shortly) that shows the user which environment is currently selected when they choose to use the Python extension for auto-location of the virtual environment. They can then decide whether to stick with that environment or select another one.

### Demo

https://user-images.githubusercontent.com/37993418/208025159-615498d4-3948-4e28-9141-c29efba18607.mov

### DIFF
--- a/extension/src/extensions/python.ts
+++ b/extension/src/extensions/python.ts
@@ -1,6 +1,9 @@
+import { relative } from 'path'
 import { commands, Event, Uri } from 'vscode'
 import { executeProcess } from '../processExecution'
 import { getExtensionAPI, isInstalled } from '../vscode/extensions'
+import { getFirstWorkspaceFolder } from '../vscode/workspaceFolders'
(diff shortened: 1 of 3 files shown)


mattseddon:
> Would it cause even more confusion? In case it's already auto and people don't even know / care which one specifically an extra question might lead to extra doubts?

You might think the Python interpreter is set to something it is not for the workspace. That would be more confusing.

shcheklein:
> You might think the Python interpreter is set to something it is not for the workspace. 

Could you clarify please? I'm still not sure what problem does this solve tbh.

Even if we want to show this additional info, can we do this on the first screen (part of the Auto ... line?)

It still feels to me that it's rather complicating the happy path.

PR #2956, opened 2022-12-16
Improve plots ribbon block tooltips

julieg18:
* highlight table keys based off type
* add copy button to table values

### DIFF
--- a/extension/src/plots/model/util.ts
+++ b/extension/src/plots/model/util.ts
@@ -1,17 +1,24 @@
 import { getDataFromColumnPath } from '../../experiments/model/util'
 import { Experiment } from '../../experiments/webview/contract'
+import { RevisionFirstThreeColumns } from '../webview/contract'
 
 export const getRevisionFirstThreeColumns = (
   firstThreeColumns: string[],
   experiment: Experiment
 ) => {
-  const columns: Array<{ path: string; value: string }> = []
+  const columns: RevisionFirstThreeColumns = []
   for (const path of firstThreeColumns) {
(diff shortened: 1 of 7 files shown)


shcheklein:
Amazing improvement! 

QQ - when it a number, does the copy button copy an actual value? Just to double check?

julieg18:
> when it a number, does the copy button copy an actual value? Just to double check? And same with data hash ...

When you say the "actual value", are you referring to the value that isn't truncated at all? If so, yes!

shcheklein (approved):

extension/src/plots/model/util.ts
```diff
@@ -1,17 +1,24 @@
...
+      experiment,
+      path
+    )
+    const type = splitUpPath[0]
+
     if (value) {
       columns.push({
+        fullValue,
         path: path.slice(type.length + 1) || path,
+        type,
```
mattseddon, line 21: [Q] Why not send only the raw data from the model and have the client(s) format it? See `webview/src/experiments/util/buildDynamicColumns.tsx` for how this is handled by the experiments table.

mattseddon: If not there is some commonality with `ExperimentsTree.getTooltipTable`. The duplication should be removed/combined.

mattseddon (approved):

extension/src/plots/model/util.ts
julieg18: It definitely would simplify the code a lot! Will do it in a followup :)

PR #2965, opened 2022-12-18
Table DnD cleanup, reset drop zone on leave

shcheklein:
Cont https://github.com/iterative/vscode-dvc/pull/2876

Partially addresses https://github.com/iterative/vscode-dvc/issues/2698 https://github.com/iterative/vscode-dvc/issues/2262

- Drops some unused code
- Makes target disappear when it is not actually active

https://user-images.githubusercontent.com/3659196/208282132-9078dfbf-d753-4b20-ae6a-b70cec8b562f.mov

### DIFF
--- a/webview/src/experiments/components/table/header/TableHead.tsx
+++ b/webview/src/experiments/components/table/header/TableHead.tsx
@@ -82,7 +82,7 @@ export const TableHead = ({
     displacedHeader && cb(displacedHeader)
   }
 
-  const onDragUpdate = (e: DragEvent<HTMLElement>) => {
+  const onDragEnter = (e: DragEvent<HTMLElement>) => {
     findDisplacedHeader(e.currentTarget.id, displacedHeader => {
       if (!isExperimentColumn(displacedHeader.id)) {
         dispatch(setDropTarget(displacedHeader.id))
@@ -93,6 +93,12 @@ export const TableHead = ({
   const onDragEnd = () => {
     fullColumnOrder.current = undefined
     draggingIds.current = undefined
+    dispatch(setDropTarget(''))
+  }
+
+  const onDragLeave = () => {
+    // note: for this to work it's important to have `pointer-events: none;`
+    // on text children to avoid duplicate dragEnter and dragLeave events fired
     dispatch(setDropTarget(''))
   }
 
@@ -130,9 +136,10 @@ export const TableHead = ({
           headerGroup={headerGroup}
           columns={allHeaders}
           onDragStart={onDragStart}
-          onDragUpdate={onDragUpdate}
+          onDragEnter={onDragEnter}
           onDragEnd={onDragEnd}
           onDrop={onDrop}
+          onDragLeave={onDragLeave}
           root={root}
           setExpColumnNeedsShadow={setExpColumnNeedsShadow}
         />
--- a/webview/src/shared/components/dragDrop/Draggable.tsx
+++ b/webview/src/shared/components/dragDrop/Draggable.tsx
@@ -1,7 +1,6 @@
 import React, { DragEvent } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { setGroup } from './dragDropSlice'
-import { DropTarget } from './DropTarget'
 import { ExperimentsState } from '../../../experiments/store'
 
 export type DragFunction = (e: DragEvent<HTMLElement>) => void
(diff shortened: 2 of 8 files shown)


webview/src/experiments/components/table/header/TableHead.tsx
```diff
@@ -44,6 +44,7 @@ export const TableHead = ({
   )
   const dispatch = useDispatch()
   const orderedColumns = useColumnOrder(columns, columnOrder)
+  let leave_enter_counter = 0
```
shcheklein, line 47: Here is the main change - logic around this counter, and `onDragLeave` below

mattseddon (approved):
Regression test(s) would be good

webview/src/experiments/components/table/header/TableHead.tsx
shcheklein: yes, good catch .. using multiple languages simultaneously hurt me

shcheklein:
> Regression test(s) would be good

yes, will try to add (tbh, not critical thing for us, but if test is easy, fast and compact it won't hurt - I'll check)

julieg18 (approved):

webview/src/experiments/components/table/header/TableHead.tsx
sroy3: There is already the same functionality is `useDeferedDragLeave` with the boolean `hoveringSomething`. Also we try to avoid using `let` variables inside of React. We use `useState` or `useRef` instead. In this case since we don't need a re-render we'd use a ref. That will avoid side effects when mouting/unmounting the component.

webview/src/shared/components/dragDrop/Draggable.tsx
```diff
@@ -11,29 +10,29 @@ export interface DraggableProps {
...
   onDragStart,
-  onDragEnd
+  onDragEnd,
+  onDragLeave
 }) => {
   const groupState = useSelector(
     (state: ExperimentsState) => state.dragAndDrop.groups[group] || {}
   )
   const dispatch = useDispatch()
-  const { draggedOverId, draggedId } = groupState
```
sroy3, line 36: Do you think it's possible to also remove it from `dragDropSlice`? And even simplify the whole thing?

sroy3 (approved):

webview/src/shared/components/dragDrop/Draggable.tsx
shcheklein: Yes! Just didn't want to make it even bigger and even more unfocused. I'll try to do this as a followup. We can keep simplifying it.

webview/src/experiments/components/table/header/TableHead.tsx
shcheklein: Thanks @sroy3 . This is useful. `useDeferedDragLeave` looks good (this case is a bit different, but I hope it'll work the same w/o any flickering). Two questions - I see that it has 500 as a delay, can we make it smaller? Any specific reason for this?

Also, I would need to call `dispatch` on an actual, final leave. What would be the best way? Parametrize and pass a callback or do this on the component level - some hook that depends on `hoveringSomething`? What would be the best?

sroy3: You can use `immediateDragLeave` and `immediateDragEnter` instead (inside the same file). If I recall correctly the delay is used because of entering/leaving groups vs entering/leaving another plot.

shcheklein: sorry, to clarify - why is it 500, not 50 or 10 for example (I think delay is used for a few reasons there - to. prevent flickering - e.g. left one plot -> entered another - in between those actions we don't want to react, also in some browsers events are generated in a different order). So, I'm fine with defered logic - that's exactly what I need I think. I want to make it faster though.

Let me do this, and I'll ask you to review.

sroy3: Lowering the timeout does not seem to have any negative effect. I can't say with 100% certitude, but I think 500 might have been chosen practically at random.

shcheklein:
@sroy3 PTAL, managed to solve it with CSS only. `deferredLeave` is not what is needed here. The sequence of events here:

... ChildEnter -> ParentLeave ... (when it comes to hovering the internal span with the column name on it). Counter with `useRef` would work but would also require additional stuff (defer Leave on FireFox so that the sequence of events is the same everywhere, etc).

shcheklein:
@mattseddon I think it's good to go, CC findings are not new.

mattseddon (approved):

PR #2186, opened 2022-08-15
Add select python interpreter option to setup workspace toast message

mattseddon:
Closes #2176

Users can get into a situation where they currently have the wrong Python interpreter selected and expect the environment to auto-activate. This prompt should remind them that they need to set the correct interpreter. 

### Demo

https://user-images.githubusercontent.com/37993418/184568267-5f22caf2-580b-4d83-885c-de5bbd4ffd17.mov

**Edit**: Toast updated to 

image

PR #2967, opened 2022-12-19
Simplify Setup the Workspace by removing manual option when Python extension is installed

mattseddon:
Related to #2944, #2793 & #2925

This PR simplifies "Setup the Workspace" by removing the "Manual" option when `ms-python.python` is installed. It also diverts the user to `Python: Select Interpreter` when they select "Auto". Showing the select interpreter dialog does two things:

1. Confirms the environment that they are using.
2. Allows them to choose another if they want to.

### Demo

https://user-images.githubusercontent.com/37993418/208332027-ccc2f249-0dc0-4293-8ea4-7a17b595e753.mov

There are no longer further options after "Auto". If the user wishes to select a "Global" DVC then they should choose that option.

https://user-images.githubusercontent.com/37993418/208332783-0f79a9ea-9086-4aac-b245-bf0c36c6c8bc.mov

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,15 +146,15 @@ These are the VS Code [settings] available for the Extension:
 
 [settings]: https://code.visualstudio.com/docs/getstarted/settings
 
-| **Option**                             | **Description**                                                                                                                                      |
-| -------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `dvc.dvcPath`                          | Path or shell command to the DVC binary. Required unless Microsoft's [Python extension] is installed and the `dvc` package found in its environment. |
-| `dvc.pythonPath`                       | Path to the desired Python interpreter to use with DVC. Required when using a virtual environment.                                                   |
-| `dvc.experimentsTableHeadMaxHeight`    | Maximum height of experiment table head rows.                                                                                                        |
-| `dvc.doNotShowWalkthroughAfterInstall` | Do not prompt to show the Get Started page after installing. Useful for pre-configured development environments                                      |
-| `dvc.doNotRecommendRedHatExtension`    | Do not prompt to install the Red Hat YAML extension, which helps with DVC YAML schema validation (`dvc.yaml` and `.dvc` files).                      |
-| `dvc.doNotShowCliUnavailable`          | Do not warn when the workspace contains a DVC project but the DVC binary is unavailable.                                                             |
-| `dvc.doNotShowUnableToFilter`          | Do not warn before disabling auto-apply filters when these would result in too many experiments being selected.                                      |
+| **Option**                             | **Description**                                                                                                                                          |
+| -------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `dvc.dvcPath`                          | Path or shell command to the DVC binary. Required unless Microsoft's [Python extension] is installed and the `dvc` package found in its environment.     |
+| `dvc.pythonPath`                       | Path to the desired Python interpreter to use with DVC. Should only be utilized when using a virtual environment without Microsoft's [Python extension]. |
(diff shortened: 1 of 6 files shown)


shcheklein:
Hmm, I'm not sure I understand what does it solve tbh, Matt could you clarify please?

mattseddon:
> Hmm, I'm not sure I understand what does it solve tbh, Matt could you clarify please?

New users with the Python extension installed are presented with an overwhelming amount of choices/paths. This cuts down the number of choices that they are presented with/have to make.

There is no good reason to have one virtual environment selected using the Python extension and then another completely separate one selected using our extension.

The old path of `Auto => Needs access to DVC outside of the environment => *All possible paths` is the exact same path as `Global => *All possible paths`. They behave the same under the hood.

IMO for #2944 we have to present the "happiest path" possible with the smallest amount of options. If the Python extension is installed we should present the path to the interpreter that is selected by that extension and/or give the option to change the interpreter. If the extension is not installed then we can give the user's base python environment and the option to change that path by setting `dvc.pythonPath`.

shcheklein:
I see. Yes! Manual virtual env doesn't make sense. I would argue it doesn't make much sense even if python extension is not installed. It should be manual path to DVC in case something goes wrong, in both cases. + we can show a greyed out Auto option and link to install MS Python extension or to docs.

>  It also diverts the user to Python: Select Interpreter when they select "Auto". Showing the select interpreter dialog does two things:

This part - I think we discussed that it's not clear if gives any value? I still don't think it's needed. We can show in the first menu current path, show link to ms Python extension config, show on hover additional info in the status bar. But actively showing menu would confuse ppl even more.

mattseddon:
> + we can show a greyed out Auto option and link to install MS Python extension or to docs.
> show link to ms Python extension config,

Cannot do this with VS Code quickPick API.

> We can show in the first menu current path,

sure

> show on hover additional info in the status bar. 

Can show "environment set by Python extension" otherwise we are duplicating information.

> But actively showing menu would confuse ppl even more.

I disagree, I think we would be adding friction by asking users to go elsewhere.

mattseddon:
> > But actively showing menu would confuse ppl even more.

> I disagree, I think we would be adding friction by asking users to go elsewhere.

Also, if the workspace is setup and working correctly users should not be running "Setup the Workspace" at all. If DVC is currently unavailable then I would want to see the list of available environment and the one that is currently selected.

shcheklein:
> Can show "environment set by Python extension" otherwise we are duplicating information.

This should be visible (Auto, Manual, Global) in the status bar itself I think.

On hover it's fine to duplicate.

> I disagree, I think we would be adding friction by asking users to go elsewhere.

My logic - in most cases it will work fine out of the box and ppl are familiar with Python ext would recognize its path and will know how to find it. Providing a step that in most cases is not needed is a bit strange. Or do you expect that ppl will actively need to use it?

> If DVC is currently unavailable then I would want to see the list of available environment and the one that is currently selected and be sure that I have selected the right one now.

I'm fine to show it, in multiple places. Redirecting ppl into ms python extension is a bit too much. 

Btw, do I understand it correctly that it will change not only DVC's binary path, but it actually changes Python extension settings?

mattseddon:
> My logic - in most cases it will work fine out of the box and ppl are familiar with Python ext would recognize its path and will know how to find it. Providing a step that in most cases is not needed is a bit strange. Or do you expect that ppl will actively need to use it?

It should definitely work out of the box. If it does then there will be no need for a user to even run the setup wizard. I expect that if the CLI cannot be found and the user is using the Python extension then the number one cause of confusion will be the wrong Python environment being selected/active.

> I'm fine to show it, in multiple places. Redirecting ppl into ms python extension is a bit too much.
> Btw, do I understand it correctly that it will change not only DVC's binary path, but it actually changes Python extension settings?

Yes and we already do this from a Toast message when the CLI cannot be found:

image

https://github.com/iterative/vscode-dvc/pull/2186

sroy3 (approved):

PR #2971, opened 2022-12-19
Refactor `getDataFromColumnPath` logic 

julieg18:
* simplify code and reduce duplication
* ~fix bug with long integers~

Followup to comments in #2956

### DIFF
--- a/extension/src/experiments/model/quickPick.test.ts
+++ b/extension/src/experiments/model/quickPick.test.ts
@@ -147,7 +147,7 @@ describe('pickExperiments', () => {
           description: '[exp-456]',
           detail: `Created:${formatDate(
             mockedExperiments[1].Created as string
-          )}, split:2.2000e+9, data/data.xml:22a1a29`,
+          )}, split:2200043556, data/data.xml:22a1a29`,
           label: '456fsf4',
           value: mockedExperiments[1]
         },
--- a/extension/src/experiments/model/quickPick.test.ts
+++ b/extension/src/experiments/model/quickPick.test.ts
@@ -147,7 +147,7 @@ describe('pickExperiments', () => {
           description: '[exp-456]',
           detail: `Created:${formatDate(
             mockedExperiments[1].Created as string
-          )}, split:2200043556, data/data.xml:22a1a29`,
+          )}, split:2.2000e+9, data/data.xml:22a1a29`,
           label: '456fsf4',
           value: mockedExperiments[1]
         },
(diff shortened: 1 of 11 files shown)


extension/src/experiments/model/quickPick.test.ts
```diff
@@ -147,7 +147,7 @@ describe('pickExperiments', () => {
           description: '[exp-456]',
           detail: `Created:${formatDate(
             mockedExperiments[1].Created as string
-          )}, split:2.2000e+9, data/data.xml:22a1a29`,
```
julieg18, line 150: Spotted this bug that was happening with long integers while working on the refactoring.

mattseddon: This was actually intended (see https://github.com/iterative/vscode-dvc/pull/2813#discussion_r1029869455). 

There is a long tail of edge cases here. For example, why would we treat `2200043556.1` differently? What if we had `10000000000000000`? The JS numbers rabbit hole is deep.

extension/src/experiments/model/util.ts
```diff
@@ -8,8 +8,10 @@ type Value = undefined | null | [] | string | number
 const isDate = (value: Value): boolean =>
   !!(typeof value === 'string' && Date.parse(value))
 
-const isLongNumber = (value: Value): boolean =>
-  typeof value === 'number' && value.toString().length > 7
+const isLongFloatNumber = (value: Value): boolean =>
+  typeof value === 'number' &&
+  !Number.isInteger(value as number) &&
+  value.toString().length > 7
```
mattseddon, line 14: personally, I would revert this change.

sroy3 (approved):

extension/src/experiments/model/quickPick.test.ts
julieg18: Oh, whoops! Must have missed that comment! The table has the full number so I assumed it was a bug. Will revert this change :)

mattseddon: Do you think we should update the webview?

julieg18: > Do you think we should update the webview?

Sure! Apologies if that's what you wanted originally, I misunderstood. When you say webview, are you referring to just the tooltips or the table as well? I'll open a pr that does both for now.

mattseddon: I didn't even think about that until you replied. No need to apologise. It was probably me that missed something back in the original PR.

PR #2973, opened 2022-12-19
Open experiments automatically once setup is done

sroy3:
Part of #2793 

(demo is faked to show what happens when all conditions are met)

https://user-images.githubusercontent.com/3683420/208518078-e7e920c5-7411-4455-9017-b25105d1676d.mov

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -41,6 +41,7 @@ import { createTypedAccumulator } from '../util/object'
 import { pickPaths } from '../path/selection/quickPick'
 import { Toast } from '../vscode/toast'
 import { ConfigKey } from '../vscode/config'
+import { setContextValue } from '../vscode/context'
 import { checkSignalFile } from '../fileSystem'
 import { DVCLIVE_ONLY_RUNNING_SIGNAL_FILE } from '../cli/dvc/constants'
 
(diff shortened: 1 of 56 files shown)


extension/src/experiments/index.ts
```diff
@@ -497,6 +502,9 @@ export class Experiments extends BaseRepository<TableData> {
   }
 
   public getHasData() {
```
mattseddon, line 504: you might be able to use this here:
```
if (this.deferred.state === 'none'){
    return false
}
```

extension/src/experiments/workspace.ts
```diff
@@ -28,13 +29,15 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
   public readonly updatesPaused: EventEmitter<boolean>
 
   private readonly checkpointsChanged: EventEmitter<void>
+  private readonly hasDataChanged: EventEmitter<void>
```
mattseddon, line 32: Might make things easier if you listen to `WorkspaceExperiments.columnsChanged.event` in the Extension instead of injecting this in. (or create `onDidChangeExperimentColumns` and listen to that)

extension/src/extension.ts
```diff
@@ -152,6 +152,14 @@ export class Extension extends Disposable implements IExtension {
       )
     )
 
+    const onHasData = this.experiments.columnsChanged.event
```
mattseddon, line 155: [N] completely undocumented naming convention is `onDidChangeHasData`

mattseddon (approved):
Dawned on me last night that we need to consider what happens when the CLI is installed but below the minimum expected version 😢.

PR #2990, opened 2022-12-22
Direct users to setup webview from warning popups

mattseddon:
# 2/2 `main` <- #2944 <- this

This is the first step in integrating the Setup webview with existing user flows. We are now: 

1. Suppressing warning toast messages whenever the webview is open.
3. Directing users to the webview from the toast messages.

### Demo

https://user-images.githubusercontent.com/37993418/209039258-6347c813-6d05-4011-85b9-3ec4c2a497f9.mov

(no stray toast)

https://user-images.githubusercontent.com/37993418/209039397-25976d88-2168-4319-bc42-65d48a3f4eab.mov

PR #2992, opened 2022-12-22
Patch CLI incompatible path for onboarding

mattseddon:
# 3/3 `main` <- #2944 <- #2990 <- this

This PR patches an edge case where the onboarding process can be opened under the circumstances of the CLI being below the minimum version accepted by the extension. Under these circumstances, we direct the user to a new screen (see new story).

### Demo

https://user-images.githubusercontent.com/37993418/209054132-dc2590a3-a3fd-4a86-ae16-ad303ddb2c96.mov

### DIFF
--- a/extension/src/cli/dvc/constants.ts
+++ b/extension/src/cli/dvc/constants.ts
@@ -1,9 +1,5 @@
 import { join } from 'path'
 
-export const MIN_CLI_VERSION = '2.30.0'
-export const LATEST_TESTED_CLI_VERSION = '2.38.0'
-export const MAX_CLI_VERSION = '3'
-
 export const UNEXPECTED_ERROR_CODE = 255
 
 export const TEMP_PLOTS_DIR = join('.dvc', 'tmp', 'plots')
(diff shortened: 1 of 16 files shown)


extension/src/cli/dvc/constants.ts
```diff
@@ -1,9 +1,5 @@
 import { join } from 'path'
 
-export const MIN_CLI_VERSION = '2.30.0'
```
mattseddon, line 3: [F] Could not import this into the webview from here because path is not available and it brought the whole thing down

webview/src/setup/components/CliIncompatible.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { MIN_CLI_VERSION } from 'dvc/src/cli/dvc/contract'
+import { EmptyState } from '../../shared/components/emptyState/EmptyState'
+
+export const CliIncompatible: React.FC = () => (
+  <EmptyState>
+    <div>
+      <h1>DVC is incompatible</h1>
+      <p>The located CLI is incompatible with the extension</p>
+      <p>The minimum version is {MIN_CLI_VERSION}</p>
+      <p>Please update your install and try again</p>
```
julieg18, line 11: Would it be a good idea to add a button here that installs the latest version of DVC? Like we have in `CliUnavailable`?

julieg18 (approved):
Looks good!

webview/src/setup/components/CliIncompatible.tsx
mattseddon: The idea here is that this person should not be going through onboarding if they already have the CLI installed and the version is below the minimum required version. There are too many combinations that we would need to cater for (e.g what if CLI is installed globally). This patch at least avoids us showing something confusing.

Issue #2947, opened 2022-12-15
Onboarding to Git


dberenbaum:
If I try to initialize DVC through the extension but I'm not in a Git repo, it fails and gives ugly error messages. Can it instead explain that this is a prerequisite and either initialize Git or instruct the users on how to do so?

<img width="555" alt="Screenshot 2022-12-15 at 9 32 11 AM" src="https://user-images.githubusercontent.com/2308172/207887278-1a1c9331-b157-44c1-863e-7f8738a9cb7b.png">

<img width="1066" alt="Screenshot 2022-12-15 at 9 32 22 AM" src="https://user-images.githubusercontent.com/2308172/207887286-f96e2e98-697a-451b-8935-3936da2b3a11.png">

dberenbaum:
Also, after then initializing DVC, it might be helpful to guide the user to committing that setup, both as a general good practice and because the experiments onboarding experience will fail in a repo with no commits.

shcheklein:
This is heavily depends on DVC (ugly message it's a bug in DVC I think cc @skshetry ). Being able to show something more reasonable - DVC should be returning it probably in a bit more structured way (error code?)

dberenbaum:
DVC could structure the error more if that's the best way for the extension to detect if the user is in a Git repo, but can the extension detect the lack of `.git` earlier and suggest `git init` before `dvc init` (or do them in tandem when needed)?

shcheklein:
> but can the extension detect the lack of .git earlier and suggest git init before dvc init (or do them in tandem when needed)?

Yes, that's a good idea. And yes I think we can and should do this

mattseddon:
We can add a screen with a button that directs the user to the Source Control view using `workbench.scm.focus`.

https://user-images.githubusercontent.com/37993418/208325686-a9d81f56-9060-4e27-a77f-d0e6b9c2f7c7.mov

shcheklein:
Can we do this w/o SCM support, on our end? Any additional step is additional friction ...

PR #3002, opened 2022-12-23
Offer to Git init when there is no DVC project or Git repository

mattseddon:
# 1/3 `main` <- this <- #3017 <- #3018 

Closes https://github.com/iterative/vscode-dvc/issues/2947

### Demo

https://user-images.githubusercontent.com/37993418/209270262-6a0edc24-1342-477c-a87d-df0a16f7a0c1.mov

https://user-images.githubusercontent.com/37993418/209270874-4d022246-ac78-4dc6-8d41-93516c53cb4e.mov

**Note:** Text has been adjusted slightly since the demos were made. See Storybook for details.


webview/src/setup/components/App.test.tsx
```diff
@@ -14,21 +14,33 @@ jest.mock('../../shared/components/codeSlider/CodeSlider')
...
+}: {
+  canGitInitialize: boolean | undefined
+  cliCompatible: boolean | undefined
+  hasData: boolean | undefined
+  isPythonExtensionInstalled: boolean
+  needsGitInitialized: boolean | undefined
+  projectInitialized: boolean
   pythonBinPath: string | undefined
-) => {
+}) => {
```
mattseddon, line 33: [F] I changed the signature to a destructured object because it was getting out of control.

extension/src/extension.ts
```diff
@@ -168,9 +173,12 @@ export class Extension extends Disposable implements IExtension {
       new Setup(
```
mattseddon, line 173: I am going to move a big chunk of the logic from the extension into `Setup` in a follow-up.

mattseddon: I will increase test coverage when I make that change.

extension/src/extension.ts
```diff
@@ -548,6 +556,44 @@ export class Extension extends Disposable implements IExtension {
...
 
+  private initializeGit() {
+    const cwd = getFirstWorkspaceFolder()
+    if (cwd) {
+      this.gitExecutor.init(cwd)
+      this.workspaceChanged.fire()
+    }
+  }
+
+  private async canGitInitialize(needsGitInit: boolean) {
```
mattseddon, line 567: [I] Both of these need to set context values so we can show Initialize Git in a welcome view (follow-up)

webview/src/setup/components/ProjectUninitialized.tsx
```diff
@@ -1,28 +1,83 @@
...
+
+const DvcUninitialized: React.FC<{ initializeDvc: () => void }> = ({
+  initializeDvc
+}) => (
+  <EmptyState>
+    <Header />
+    <p>
+      The current workspace does not contain a DVC project. You can initialize a
+      project which will enable features powered by DVC. To learn more about how
+      to use DVC please read <a href="https://dvc.org/doc">our docs</a>
```
julieg18, line 54: I'm no grammar expert, but should this be: 

```suggestion
      to use DVC, please read <a href="https://dvc.org/doc">our docs</a>.
```

webview/src/setup/components/ProjectUninitialized.tsx
```diff
@@ -1,28 +1,83 @@
-import { VSCodeButton } from '@vscode/webview-ui-toolkit/react'
 import React from 'react'
```
julieg18, line 1: Are we planning on adjusting our welcome tree view? Showing a link to setup instead of just showing the "Initialize Project" button for example?

webview/src/setup/components/ProjectUninitialized.tsx
julieg18: Now that I look at it, a lot of the setup statements don't end with a period. Out of curiosity, is there a particular reason why?

webview/src/setup/components/ProjectUninitialized.tsx
```diff
@@ -1,28 +1,83 @@
...
+
+export const ProjectUninitialized: React.FC<ProjectUninitializedProps> = ({
+  initializeDvc,
+  needsGitInitialized,
+  canGitInitialize,
+  initializeGit
+}) => {
+  if (needsGitInitialized) {
+    return (
+      <GitUninitialized
```
julieg18, line 75: Is there a way to detect if a git repo has been initialized via the CLI so the webview updates automatically: 

https://user-images.githubusercontent.com/43496356/209836990-c35cdb32-f818-4573-a4a9-86eaf09fd115.mov

julieg18 (approved):
Great work!

webview/src/setup/components/ProjectUninitialized.tsx
mattseddon: Yes, we need another watcher. I'll put it at the end of the chain as a follow-up.

webview/src/setup/components/ProjectUninitialized.tsx
mattseddon: My plan was to show the same buttons in the welcome view but we can simplify and drop all that logic instead.

webview/src/setup/components/ProjectUninitialized.tsx
mattseddon: I don't have a good answer. Guess I'm going to spend some time today adding full stops to text.

mattseddon: I think it is a spill-over from the titles

PR #3005, opened 2022-12-24
Upgrade e2e test dependencies and Mocha

mattseddon:
🎄 🎅🏻 💩 
----

This PR gets us past the breaking change in `wdio-vscode-service`. In order to do this the end to end test suite has to become and ESM module (sub-package). This means that it cannot call any of the rest of the codebase directly (which is probably a good thing).


.husky/pre-commit
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 . "$(dirname $0)/_/husky.sh"
 
-yarn run tsc -p tsconfig.json --noEmit
```
mattseddon, line 4: [F] This now throws some wacky errors

mattseddon: Can safely replace with the command on the right or drop altogether. Makes me think we no longer need the `tsconfig.json` at the root of the monorepo.

extension/src/test/e2e/extension.test.ts
```diff
@@ -1,5 +1,4 @@
 import { join } from 'path'
```
mattseddon, line 1: [F] Most of the changes are indentation based (dropped the suite statement as it no longer worked). Will mark the actual changes and give reasons.

extension/src/test/e2e/util.ts
```diff
@@ -1,10 +1,10 @@
 import {
   ChainablePromiseArray,
   ChainablePromiseElement,
-  ElementArray
+  ElementArray,
+  Key
 } from 'webdriverio'
 import { ViewControl } from 'wdio-vscode-service'
-import { delay } from '../../util/time'
```
mattseddon, line 7: [F] Good thing the new v8 API has it's own delay as we can't import anything from our own repo here (not ESM).

extension/src/test/e2e/extension.test.ts
```diff
@@ -11,184 +10,190 @@ import {
   runModifiedExperiment,
   waitForDvcToFinish,
   waitForViewContainerToLoad
-} from './util'
-import { ExperimentsWebview } from './pageObjects/experimentsWebview'
-import { PlotsWebview } from './pageObjects/plotsWebview'
-
-suite('DVC Extension For Visual Studio Code', () => {
```
mattseddon, line 18: [F] Can no longer use suite.

extension/src/test/e2e/extension.test.ts
```diff
@@ -11,184 +10,190 @@ import {
...
-
-      expect(finalRows.length).toStrictEqual(initialRows + epochs)
-      await webview.unfocus()
-      await waitForDvcToFinish()
-      const workbench = await browser.getWorkbench()
-      await workbench.executeCommand('Terminal: Kill All Terminals')
-    }).timeout(180000)
+    expect(await webview.table$$).toHaveLength(1)
+
+    return webview.unfocus()
```
mattseddon, line 63: [F] This test has to be split in half or the framework cannot find the command palette.

extension/src/test/e2e/extension.test.ts
```diff
@@ -11,184 +10,190 @@ import {
...
   })
 
-  describe('Plots Webview', () => {
-    before(async () => {
-      await closeAllEditors()
-    })
+  it('should show the correct number of records before running an experiment', async function () {
```
mattseddon, line 66: [F] This test has to be split in half or the framework cannot find the command palette.

julieg18 (approved):
> 🎄 🎅🏻 💩

Ah yes... Christmas tree santa poop..

Code looks good though! Great work!

PR #3007, opened 2022-12-25
Cleanup table dnd code: remove unused slice/group logic

shcheklein:
Followup #3006 

Simplify and cleanup unused code.


mattseddon (approved):

mattseddon:
Is the plots webview unaffected?

shcheklein:
> Is the plots webview unaffected?

No, they should be file. `Draggable` is not used in plots. I think as another followup I may remove the whole Draggable, I'm not sure we need it now.

PR #3016, opened 2022-12-27
Add commit messages to experiment views 

### LINKED ISSUE #2458: Multiple commits: show commit information, other improvements
After implementing multiple commits logic we can start improving the UX a bit.

Hash as a name is not informative. It's not even clear if that is a commit or an experiment.

For commits we have more information - author, commit name, tags, etc.

I think if we have tags, we should show them. If not, we should show the commit message (any part of the message is better vs hash).Tooltip should include everything else - author names, etc. Eventually going into a direction of tooltips that GitLens has.

julieg18:
# `main` <= this <= #3029 <= #3035

I'm still working on researching points made in the last planning meeting, but here's how things look if we place the commit message as secondary text (similar to what we do with experiment names). 

## Demo

https://user-images.githubusercontent.com/43496356/209989576-cc7e6d4a-f121-475a-87ce-b31f79f44943.mov

Fixes #2458


extension/src/experiments/model/index.test.ts
```diff
@@ -426,33 +477,49 @@ describe('ExperimentsModel', () => {
...
-  it("should fetch an empty array if the experiment's params cannot be found", () => {
-    const model = new ExperimentsModel('', buildMockMemento())
-    model.transformAndSet(outputFixture, false)
+  it("should fetch an empty array if the experiment's params cannot be found", async () => {
+    const model = new ExperimentsModel(
+      '',
+      buildMockMemento(),
+      mockedInternalCommands
+    )
+    await model.transformAndSet(outputFixture, false)
```
julieg18, line 522: Adding a promise to the model's `transformAndSet` caused me to add more `await`s to multiple tests.

extension/src/cli/git/constants.ts
```diff
@@ -44,4 +47,4 @@ export enum Commit {
 
 export const DEFAULT_REMOTE = 'origin'
 
-export type Args = (Command | Flag | Commit | typeof DEFAULT_REMOTE)[]
+export type Args = (Command | Flag | Commit | typeof DEFAULT_REMOTE | string)[]
```
julieg18, line 50: I needed to be able to add unique `sha` ids as an argument but is there a better way to fix the type besides just allowing strings in general?

shcheklein:
@julieg18 thanks!

Can we try as an experiment to remove square brackets, make hashes white, make exp names and commit messages a bit smaller in size + keep spaced from the hash when they are on the same line.

julieg18:
> Can we try as an experiment to remove square brackets, make hashes white, make exp names and commit messages a bit smaller in size + keep spaced from the hash when they are on the same line.

Sure! Here's a screenshot (I can also open up a branch to demo if preferred). Technically, hashes are already white in the VSCode dark theme, but we could look into an alternative variable that highlights text more on different themes 🤔 

image

shcheklein:
Perfect! I like it better, I think it's good for this iteration!

Let's merge (if code looks better and @mattseddon and @sroy3 approve). And let's next think what about some experiments on how we can make the split between nested items and commits more obvious.

shcheklein:
Btw on the theme you had initially, in the tree there is a visible difference in colors (exp name and description), but in the table there is no difference - do we know the reason?

shcheklein:
As a followup:

- We need to make sure that we see the full information about the commit in the tooltip
- We need to make sure we see the commit message in the Quick pick and in the ribbon for plots
- Test with Git tags

julieg18:
> Perfect! I like it better, I think it's good for this iteration!

Great, will update the branch once I clear a small CSS bug I spotted with this change!

> Btw on the theme you had initially, in the tree there is a visible difference in colors (exp name and description), but in the table there is no difference - do we know the reason?

Yes, we're using the `--vscode-editor-foreground` variable for the table cell text while I believe the tree view uses a variable called `--vscode-sideBar-foreground`.

shcheklein:
> Yes, we're using the --vscode-editor-foreground variable for the table cell text while I believe the tree view uses a variable called --vscode-sideBar-foreground.

it makes sense, and we have to stick with it ... we can research if there is a color on the *editor* group that would look white or at least more contrast

shcheklein:
As a followup - we should try to use this icon (at least in the table, if VS Code tree doesn't allow that, or create a colored version of it for the tree)

Screen Shot 2022-12-29 at 12 35 11 PM


Screen Shot 2022-12-29 at 12 34 53 PM

Screen Shot 2022-12-29 at 11 50 06 AM

Screen Shot 2022-12-29 at 12 37 13 PM

shcheklein (approved):
Thanks @julieg18 ! let's summarize and create a followup research / improvement ticket based on the discussion.

extension/src/experiments/index.ts
```diff
@@ -114,7 +114,7 @@ export class Experiments extends BaseRepository<TableData> {
     this.onDidChangeCheckpoints = this.checkpointsChanged.event
 
     this.experiments = this.dispose.track(
-      new ExperimentsModel(dvcRoot, workspaceState)
+      new ExperimentsModel(dvcRoot, workspaceState, internalCommands)
```
mattseddon, line 117: [I] We shouldn't be calling `internalCommands` from inside of the model. Can we call `git log` with `git log --pretty=format:%B -n 3` or `git log HEAD~2..HEAD --pretty=format:%B` or equivalent and pass that into `transformAndSet`? That would do a few things: 

1. limit the number of times we call Git
2. stop us from having to mock the function/internal commands inside of the tests
3. not add to the responsibilities of the model

mattseddon: Should be able to use `-z` to separate the messages with `null`s instead of new lines.

mattseddon: Three calls every time we call `exp show` will clog up the output channel too: 
image

extension/src/experiments/model/index.test.ts
mattseddon: another reason to get the data outside.

extension/src/experiments/index.ts
mattseddon: `git log --pretty="format:%H %B" -n 3 -z ` could be the way to go, that way you can collect an object that is keyed by the `sha`.

mattseddon (approved):
Can we please restructure before merging this one? LMK if you have any questions.

extension/src/experiments/index.ts
julieg18: > Should be able to use -z to separate the messages with nulls instead of new lines.

Thanks, I couldn't figure out a way to split the commit messages if they all came in one output and that flag works great!

PR #3025, opened 2022-12-29
Add DVC and Git initialization watcher

mattseddon:
Follow up from #3002 ([this](https://github.com/iterative/vscode-dvc/pull/3002#discussion_r1058433592)). 

### Demo

https://user-images.githubusercontent.com/37993418/209887331-17ad1ccf-ac27-40cf-b163-f5c2181baf49.mov

Note: Turns out `dvc init` from the command line also didn't work.

### DIFF
--- a/extension/src/setup/index.ts
+++ b/extension/src/setup/index.ts
@@ -1,4 +1,5 @@
 import { Event, EventEmitter, ViewColumn, commands } from 'vscode'
+import { Disposable, Disposer } from '@hediet/std/disposable'
 import isEmpty from 'lodash.isempty'
 import { SetupData as TSetupData } from './webview/contract'
 import { WebviewMessages } from './webview/messages'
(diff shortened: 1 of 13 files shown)


extension/src/setup/index.ts
```diff
@@ -455,6 +466,42 @@ export class Setup
...
+    if (
+      !(
+        (path && (path.endsWith(gitPath.DOT_GIT) || path.includes(DOT_DVC))) ||
+        disposer.disposed
+      )
+    ) {
+      return
+    }
+
+    if (path.includes(DOT_DVC)) {
```
mattseddon, line 497: [F] If the path includes `.dvc` then assume there is a project and dispose of the listener.

extension/src/setup/index.ts
```diff
@@ -280,13 +289,16 @@ export class Setup
   private setProjectAvailability() {
     const available = this.hasRoots()
     setContextValue('dvc.project.available', available)
+    if (available && this.dotFolderWatcher && !this.dotFolderWatcher.disposed) {
```
mattseddon, line 292: [F] If there is a project available and the watcher hasn't been disposed yet then get rid of it

julieg18 (approved):
Great work!

Issue #3021, opened 2022-12-28
Make sure that experiment buttons are properly handled in DVCLive-only case


shcheklein:
We need to think what do we do when pipeline is not available (DVCLive scenario).

- [x] Run button should be able to detect that pipeline is not available
- [x] Quick actions should be able to detect this also
- [x] Use Run to create pipelines - generate a template. Similar to creating launch config in VS Code
- [x] Stop button, etc - should be handled properly (#3027 )

What else I'm missing here?

sroy3:
Tested and everything in this ticket is done.

PR #3027, opened 2022-12-29
Fix stop button

mattseddon:
# 1/2 `main` <- this <- #3041

Part of #3021

This PR fixes the stop experiment button by: 

 - Allowing users to stop experiments that are running in the DVCLive-only context. 
 - Hitting the button when we cannot stop the experiment (we don't have any reference to the process).

### Demo

https://user-images.githubusercontent.com/37993418/209922611-1d7d6ed1-e639-472f-bed8-970e99955a9a.mov

https://user-images.githubusercontent.com/37993418/210297560-e69f3563-d55c-4775-a072-2a04f3dfad41.mov

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -47,6 +47,8 @@ import { collectWorkspaceScale } from './telemetry/collect'
 import { GitExecutor } from './cli/git/executor'
 import { GitReader } from './cli/git/reader'
 import { Setup } from './setup'
+import { definedAndNonEmpty } from './util/array'
+import { stopProcesses } from './processExecution'
 
 export class Extension extends Disposable {
   protected readonly internalCommands: InternalCommands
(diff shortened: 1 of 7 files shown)


extension/src/extension.ts
```diff
@@ -191,9 +193,17 @@ export class Extension extends Disposable {
...
-        const wasRunning = this.dvcRunner.isExperimentRunning()
+        const dvcLiveOnlyPids = await this.experiments.getDvcLiveOnlyPids()
+        const wasRunning =
+          this.dvcRunner.isExperimentRunning() ||
+          definedAndNonEmpty(dvcLiveOnlyPids)
         try {
-          const stopped = await this.dvcRunner.stop()
+          const [dvcLiveOnlyStopped, runnerStopped] = await Promise.all([
+            stopProcesses(dvcLiveOnlyPids),
+            this.dvcRunner.stop()
```
mattseddon, line 203: [I] We can only stop experiments under these circumstances. Need to set another context value and hide `dvc.stopRunningExperiment` if we cannot stop the process.

mattseddon: added

julieg18 (approved):
Great work!

PR #3028, opened 2022-12-29
Simplify run and reset action name

shcheklein:
I think we don't need to have an explicit Reset. It raises questions (do I actually need it? what does it mean, etc).


shcheklein:
Can't approve UI tests now (saying I need to be a reviewer) ... don't remember this is happening before

mattseddon (approved):

shcheklein:
Minor: why rebase is disabled btw?

mattseddon:
> Minor: why rebase is disabled btw?

Make sure multiple commits aren't merged by mistake.

PR #3029, opened 2022-12-29
Update experiment secondary name

julieg18:
# `main` <= #3016 <= this <= #3035

* adjusts styling and takes off brackets of commit data strings
* Updates table stories with commit message

## Demo

~https://user-images.githubusercontent.com/43496356/210012467-578968d1-42f7-496c-aca2-4dd95a226683.mov~

https://user-images.githubusercontent.com/43496356/210022501-14425aeb-e1ff-43f5-9267-8f2a813b9f14.mov

Resolves [comments](https://github.com/iterative/vscode-dvc/pull/3016#issuecomment-1367523274) in #3016

### DIFF
--- a/extension/src/experiments/model/collect.ts
+++ b/extension/src/experiments/model/collect.ts
@@ -86,7 +86,7 @@ const getDisplayNameOrParent = (
     return `(${shortenForLabel(checkpointParent)})`
   }
   if (name) {
-    return name
+    return `[${name}]`
   }
 }
 
--- a/extension/src/experiments/model/collect.ts
+++ b/extension/src/experiments/model/collect.ts
@@ -86,7 +86,7 @@ const getDisplayNameOrParent = (
     return `(${shortenForLabel(checkpointParent)})`
   }
   if (name) {
-    return `[${name}]`
+    return `${name}`
   }
 }
 
@@ -313,7 +313,7 @@ const collectFromBranchesObject = (
 
 const formatCommitMessage = (commit: string) => {
   const lines = commit.split('\n').filter(Boolean)
-  return `[${lines[0]}${lines.length > 1 ? ' ...' : ''}]`
+  return `${lines[0]}${lines.length > 1 ? ' ...' : ''}`
 }
 
 const addCommitDataToBranches = async (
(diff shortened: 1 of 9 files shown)


extension/src/experiments/model/collect.ts
```diff
@@ -86,7 +86,7 @@ const getDisplayNameOrParent = (
     return `(${shortenForLabel(checkpointParent)})`
   }
   if (name) {
-    return `[${name}]`
```
julieg18, line 89: Did we want to take off brackets in all secondary names (quick pick, tree, table) or just take off the brackets in the table? Currently this pr takes off them in all places but we could easily just do the table if preferred :)

julieg18, line 89: Is there any possible issues with taking off the brackets of the experiment `displayNameOrParent` that I missed?

julieg18:
@shcheklein, did we want to take brackets off the experiment names and the commit messages or just the commit messages?

shcheklein:
@julieg18 we can try to keep them- hard to tell, w/o looking into the result.

Not related: why don't we see the experiments in the drop down on the demo and we see only commits ... (cc @mattseddon )?

julieg18:
> @julieg18 we can try to keep them- hard to tell, w/o looking into the result.

Ok, will go ahead and add the brackets to the experiment names! Converting this to a draft for now! 

> Not related: why don't we see the experiments in the drop down on the demo and we see only commits ... (cc @mattseddon )?

Originally, this command ("Modify Experiment Param(s) and Resume") shows just experiment hashes, though after a discussion with Matt, we'd probably want to add the first three columns like we do with the "Select Experiments to Display in Plots" command. We can also look into adding the first three columns and the commit... Can research and update in a followup!

shcheklein:
> Ok, will go ahead and add the brackets to the experiment names! Converting this to a draft for now!

can you do a few screenshots with and w/o them when there 10+ experiments and a few commits? that would help to decide

> Originally, this command ("Modify Experiment Param(s) and Resume") shows just experiment hashes

my question was not about additional info, sorry for the confusion ... it's more about why do we see only commits there, not experiments

julieg18:
> can you do a few screenshots with and w/o them when there 10+ experiments and a few commits? that would help to decide

With brackets on experiment names:

image

Without brackets on experiment names:

image

shcheklein:
I like it with brackets :) thanks @julieg18 .. wdyt?

julieg18:
> I like it with brackets :) thanks @julieg18 .. wdyt?

Agreed, the brackets on exp names look better to me!

mattseddon (approved):

PR #3030, opened 2022-12-30
Enable sub-project selection in monorepos

### LINKED ISSUE #3011: Change running directory
Hi, is there a way to change the default running dir?

I have a mono-repo with the following structure:

```
.dvc
--train
----.dvc
```

where the global dvc only stores final models and datasets, while the ``train`` subfolder is used for experiments. How can I tell the extension it should run dvc in the ``train`` subdirectory, so that I can observe experiment logs?

I'm fine with setting this option once and for all, so it should be simpler than multi-project selection.

(Related to https://github.com/iterative/vscode-dvc/issues/1931#issue-1276755970)

mattseddon:
Closes #3011

This PR enables auto-discovery of sub-projects within DVC monorepos. It also provides a `dvc.focusedProjects` config option which users can set to override the auto-discovered projects within the workspace.

### Demo

https://user-images.githubusercontent.com/37993418/210022818-4205c707-fc36-4f97-8c6a-01e401b0da06.mov


extension/package.json
```diff
@@ -529,6 +529,16 @@
     "configuration": {
       "title": "DVC",
       "properties": {
+        "dvc.focusedProjects": {
```
shcheklein, line 532: call it currentWorkingProject or just currentProject?

mattseddon: the option allows multiple projects to be set

shcheklein: what is the semantics of this then?

shcheklein: I see from the demo, thanks. Is there a name / term in VS Code when you open multiple projects / workspaces (what is the right term?)?

mattseddon: You can open multiple folders inside of a workspace which is known as a multi-root workspace.
https://code.visualstudio.com/docs/editor/multi-root-workspaces.

In terms of the Git extension I cannot find an option to limit/set the repositories to a certain list in the workspace.

shcheklein: How do we call this feature in the extension (ability to have multiple roots)? Btw also, how do we determine in the plots, experiments table which one to use?

mattseddon: > How do we call this feature in the extension (ability to have multiple roots)?

We do not have a user-facing term for this. They are auto-discovered when the extension is started.

> Btw also, how do we determine in the plots, experiments table which one to use?

We use a quick pick to let the user choose which one they want to open: 

https://user-images.githubusercontent.com/37993418/210023764-ce86651f-e3c0-4de9-b196-9131747ad98a.mov

julieg18 (approved):
Looks good!

extension/package.json
shcheklein: thanks, Matt, a few questions:

- what would happen by default in that repo from the issue? would I see three projects? (one top level and 2 nested?)
- effectively this new focused option limits the number of auto-detected projects? or it also does something else?
- in the current version, if I open experiments and pick the project - does it affect plots automatically? Run commands, etc - or it will keep asking me for every command?

extension/package.nls.json
```diff
@@ -75,6 +75,8 @@
   "command.views.experimentsTree.selectExperiments": "Select Experiments to Display in Plots",
   "command.views.plotsPathsTree.selectPlots": "Select Plots to Display",
   "command.views.plotsPathsTree.refreshPlots": "Refresh Plots for Selected Experiments",
+  "config.focusedProjects.description": "A subset of paths to the workspace's available DVC projects. Using this option will override project auto-discovery. Only absolute paths inside of the workspace will be included. All other DVC projects in the workspace will be ignored.",
```
shcheklein, line 78: does it have to be an absolute path?

README.md
```diff
@@ -151,6 +151,7 @@ These are the VS Code [settings] available for the Extension:
 | `dvc.dvcPath`                          | Path or shell command to the DVC binary. Required unless Microsoft's [Python extension] is installed and the `dvc` package found in its environment.     |
 | `dvc.pythonPath`                       | Path to the desired Python interpreter to use with DVC. Should only be utilized when using a virtual environment without Microsoft's [Python extension]. |
 | `dvc.experimentsTableHeadMaxHeight`    | Maximum height of experiment table head rows.                                                                                                            |
+| `dvc.focusedProjects`                  | A subset of paths to the workspace's available DVC projects. Using this option will override project auto-discovery.                                     |
```
shcheklein, line 154: how about `dvc.projects`? I'm still not sure about this "focused" term ...

extension/package.json
mattseddon: > thanks, Matt, a few questions:
> 
> * what would happen by default in that repo from the issue? would I see three projects? (one top level and 2 nested?)

You would see all three projects (they are not related).

> * effectively this new focused option limits the number of auto-detected projects? or it also does something else?

Essentially yes but if for any reason there was one that wasn't found. E,g it is nested more than 1 level deep it could be used to add the full list.

> * in the current version, if I open experiments and pick the project - does it affect plots automatically? Run commands, etc - or it will keep asking me for every command?

If the experiments table is focused for a particular project (i.e is in the active editor) then all top-level experiment-related commands will be run against that project automatically. Anything selected through the UI for that particular project (i.e context menus in the webview or tree) will be automatically run against that project. Same for plots. In terms of opening plots from experiments, if you do it from the experiments webview then it will open the correct project's plots. If not then you will be prompted.

extension/package.nls.json
mattseddon: Yes, all of the paths that we accept currently have to be absolute.

README.md
mattseddon: IMO projects needs some kind of determiner to signal that it is only needed under some special condition. I considered override, included and current. I also worked through lots of different combinations of this text. I tried to consider the end user.

extension/package.nls.json
shcheklein: any specific reason for this? can we resolve from the project / workspace root?

extension/package.json
shcheklein: > If the experiments table is focused for a particular project (i.e is in the active editor)

To clarify, it means I can open multiple exp tables? 

> Anything selected through the UI

what happens with any other commands (quick pick, play button in params.yaml, etc)?

extension/package.nls.json
mattseddon: One reason is that there can be collisions when a user is using a mutli-root workspace. I.e a relative path can exist in both.

extension/package.json
mattseddon: > To clarify, it means I can open multiple exp tables?

Yes

> what happens with any other commands (quick pick, play button in params.yaml, etc)?

If there is a choice between projects and a webview is in the active editor then the webview's project should be automatically chosen. This will work for quick picks. It won't work for the play button in params.yaml because that file is not linked in to the aforementioned behaviour. The extension will also ask you every time for every command if no webviews are open. That's why if you are working with a single project I'd recommend opening it in its own workspace.

mattseddon: We can add a quick pick to set this option if we think people will use it.

PR #3032, opened 2022-12-30
GitHub actions update

dacbd:
Use the new output files for actions
https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


julieg18 (approved):
Looks good to me!

PR #3035, opened 2022-12-30
Restructure experiment commit message logic

julieg18:
# `main` <= #3016 <= #3029 <= this

Resolves [review comments](https://github.com/iterative/vscode-dvc/pull/3016#pullrequestreview-1232853927) in #3016

### DIFF
--- a/extension/src/cli/git/reader.ts
+++ b/extension/src/cli/git/reader.ts
@@ -39,7 +39,7 @@ export class GitReader extends GitCli {
       Command.LOG,
       Flag.PRETTY_FORMAT_COMMIT_MESSAGE,
       Flag.NUMBER,
-      '3',
+      '6',
       Flag.SEPARATE_WITH_NULL
     )
     try {
@@ -48,7 +48,6 @@ export class GitReader extends GitCli {
         const [sha, ...splitMessage] = commit.split(' ')
         return [sha, splitMessage.join(' ')]
       })
-
       return Object.fromEntries(commits)
     } catch {
       return {}
--- a/extension/src/cli/git/reader.ts
+++ b/extension/src/cli/git/reader.ts
@@ -7,7 +7,7 @@ import { trimAndSplit } from '../../util/stdout'
 import { isDirectory } from '../../fileSystem'
 
 export const autoRegisteredCommands = {
-  GIT_GET_COMMIT_MESSAGE: 'getCommitMessage',
+  GIT_GET_LAST_THREE_COMMIT_MESSAGES: 'getLastThreeCommitMessages',
   GIT_GET_REPOSITORY_ROOT: 'getGitRepositoryRoot',
   GIT_HAS_CHANGES: 'hasChanges',
   GIT_LIST_UNTRACKED: 'listUntracked'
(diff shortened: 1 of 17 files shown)


extension/src/test/suite/extension.test.ts
```diff
@@ -19,19 +19,14 @@ import { DvcReader } from '../../cli/dvc/reader'
 import expShowFixture from '../fixtures/expShow/base/output'
 import plotsDiffFixture from '../fixtures/plotsDiff/output'
 import * as Disposer from '../../util/disposable'
-import {
-  RegisteredCliCommands,
-  RegisteredCommands
-} from '../../commands/external'
+import { RegisteredCommands } from '../../commands/external'
```
julieg18, line 22: Saw these warnings when I ran `yarn lint`. I think they're leftover from #3035

extension/src/cli/git/reader.ts
```diff
@@ -31,18 +31,28 @@ export class GitReader extends GitCli {
...
+    [sha: string]: string
+  }> {
     const options = getOptions(
       cwd,
       Command.LOG,
       Flag.PRETTY_FORMAT_COMMIT_MESSAGE,
       Flag.NUMBER,
-      '1',
-      sha
+      '3',
```
julieg18, line 42: Not sure why but `dvc exp show -n 3` doesn't actually show the last three commits for our current demo project, but the 1st, 2nd, and 4th commits 🤔 

image

mattseddon: looks like `exp show` is only showing merges and possibly the first parent:

```
git log --pretty="format:%H %B" HEAD~5..HEAD -z --merges --first-parent
```

This causes an issue where we cannot reliably grab the required commits by just going X commits back in time. We will need to pass the oldest commit into `git log` along with the required options. You should be able to get the commit by taking the last entry from `Object.keys` of `exp show` 😢.

extension/src/test/suite/extension.test.ts
mattseddon: I followed up in #3039

julieg18:
@mattseddon Can you look over the changes when you have a chance?

extension/src/cli/git/reader.ts
```diff
@@ -31,18 +31,28 @@ export class GitReader extends GitCli {
...
+      Flag.SEPARATE_WITH_NULL
     )
     try {
-      return await this.executeProcess(options)
-    } catch {}
+      const output = await this.executeProcess(options)
```
mattseddon, line 46: I would keep this code as dumb as possible. Return a string and let `collectExperiments` deal with it. We want to try and keep all of the parsing logic in one place.

mattseddon: Parsing the string somewhere in `addCommitDataToBranches` would make sense to me

mattseddon: Would be good to add a test for the parsing logic after it gets moved too.

mattseddon (approved):
> @mattseddon Can you look over these changes when you have a chance?

Looks good. 

Good thing you checked the demo project as the current logic needs to be updated. Please take a look the comments that I've left and LMK if you need any clarification.

extension/src/cli/git/reader.ts
julieg18: > You should be able to get the commit by taking the last entry from Object.keys of exp show 😢.

Good idea! I'll go ahead and increase the `-n` from 3 to 6 so there is less of a chance of it happening, merge, and fix it properly in a followup.

PR #3040, opened 2023-01-02
Add project selection quick pick

mattseddon:
Follow up from #3030

This PR adds a quick pick that can be used to focus on particular projects in the current workspace. This quick pick is only made available when there is > 1 project in the workspace.

### Demo

https://user-images.githubusercontent.com/37993418/210293160-01e5f040-217e-498a-835a-f3326def5054.mov


julieg18 (approved):
Works great!

PR #3045, opened 2023-01-04
Use process polling to ensure DVCLive only PIDs are still running

### LINKED ISSUE #2991: Auto-selecting experiments
The onboarding experience looks way better now! I mostly only found minor issues that I don't think are worth prioritizing right now.

The one place where I still see some friction is seeing the results of my experiment (skip to 0:10, something's wrong with my video editing).

https://user-images.githubusercontent.com/2308172/209039833-be278d6d-aaea-4c51-8a45-12b6cba8d893.mp4

Some ways to better highlight the experiment:

* In the "experiments" section in the sidebar, expand to show the new experiment and highlight it by default.
* In the "plots" section in the sidebar and the plots view, auto-select the experiment and show the available plots, or at least don't say "no plots to display" when there are plots that just need to be selected.

These are mostly changes that I think were introduced in #2877. No need to revisit that whole discussion again, but can we bring back some of the auto-select behavior that got removed there? I'm still finding there is much more friction in finding my experiment results now than there was a few weeks ago.

Old experience for comparison:

https://user-images.githubusercontent.com/2308172/209039450-82f7467b-c9ab-44da-aefc-df5b6ce08f98.mov

mattseddon:
Closes #2991 by addressing:

> I think we need a better way to detect failures eventually (check pid from time to time?)

### Demo

https://user-images.githubusercontent.com/37993418/210467380-f6f3638d-46e6-4eec-a98a-c22c502faa85.mov


extension/src/experiments/index.ts
```diff
@@ -602,4 +609,19 @@ export class Experiments extends BaseRepository<TableData> {
...
+
+  private async checkSignalFile() {
+    const dvcLiveOnly = await checkSignalFile(this.dvcLiveOnlySignalFile)
+
+    if (dvcLiveOnly && !this.dvcLiveOnlyCleanupInitialized) {
+      this.dvcLiveOnlyCleanupInitialized = true
+      pollSignalFileForProcess(this.dvcLiveOnlySignalFile, () => {
+        this.dvcLiveOnlyCleanupInitialized = false
+        if (this.hasRunningExperiment()) {
+          this.cliData.update()
```
mattseddon, line 621: [F] The signal file has now been removed so we request the data again to flush out the fact that an experiment is no longer running.

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3048, opened 2023-01-04
Enable the user to specify a number of concurrent jobs when starting the experiments queue

mattseddon:
# 1/2 `main` <- this <- #3049 

Relates to #3042

This PR puts an input box in front of starting the queue. The user now has to enter the number of workers that they want to run the queue with.

### Demo

https://user-images.githubusercontent.com/37993418/210483080-91e62d71-3d67-4d6e-8fd5-1b1b65b26afe.mov

PR #3049, opened 2023-01-04
Consolidate use of input box validation

mattseddon:
# 2/2 `main` <- #3048 <- this

This PR consolidates our use of validation functions with respect to the VS Code `InputBox`.

### Demo

https://user-images.githubusercontent.com/37993418/210489768-c0d124ff-fee1-4732-8ed2-1d76a2347e9b.mov

### DIFF
--- a/extension/src/vscode/inputBox.ts
+++ b/extension/src/vscode/inputBox.ts
@@ -1,18 +1,22 @@
 import { window } from 'vscode'
 import { Title } from './title'
 import { isValidStringInteger } from '../util/number'
+import { getIsoDate, isFreeTextDate } from '../util/date'
 
-export const getInput = (title: Title, value?: string) =>
+export const getInput = (
+  title: Title,
+  value?: string
+): Thenable<string | undefined> =>
   window.showInputBox({
     title,
     value
   })
 
-export const getValidInput = (
+const getValidInput = (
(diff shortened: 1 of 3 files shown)


extension/src/vscode/inputBox.ts
```diff
@@ -1,37 +1,58 @@
...
+  if (!isValidStringInteger(input)) {
+    return false
+  }
+
+  const number = Number(input)
+
+  if (!includeZero) {
+    return number > 0
+  }
+  return number >= 0
```
julieg18, line 40: Not sure if it looks better but you could turn these four lines into one: 
```
return includeZero ? number >= 0 : number > 0
```

extension/src/vscode/inputBox.ts
```diff
@@ -1,37 +1,58 @@
...
-) =>
+): Thenable<string | undefined> =>
   window.showInputBox({
     prompt: options?.prompt,
     title,
     validateInput,
     value: options?.value
   })
 
+const isValid = (
```
julieg18, line 27: Should this function be `isValidNumber` since the function is focused on checking numbers?

julieg18 (approved):

sroy3 (approved):

PR #3052, opened 2023-01-04
Stub getLastThreeCommitMessages in integration tests

mattseddon:
Not having this call stubbed seems to be causing some flakiness.


julieg18 (approved):
Looks good!

Issue #3036, opened 2023-01-02
Improve multiple commits feature


julieg18:
- [x] Fix bug with `dvc exp show -n 3` not always showing the last three commits (see [#3035 comment](https://github.com/iterative/vscode-dvc/pull/3035#discussion_r1060120952)) (#3053)
- [x] Make split between the "main" row and "past commit" rows more obvious (#3087) 
- [x] Add more commit information to table rows in a tooltip (full message, author, git tags, etc.)
- [x] Possibly show the commit messages in quick pick experiment selection (if we can show both columns and commits) 
- [x] Show the commit messages in the plots ribbon blocks
- [x] Incorporate the [git-commit](https://microsoft.github.io/vscode-codicons/dist/codicon.html) icon into commit messages 

Followup to #3016

julieg18:
> Make split between the "main" row and "past commit" rows more obvious

Looked into some other tables but had some difficulty finding ideas since no table that I had a clear "split" between rows minus when the rows were in collapsible groups. But here's one idea I experimented with: 

<img width="1008" alt="Screenshot 2023-01-05 at 8 17 01 AM" src="https://user-images.githubusercontent.com/43496356/210804635-58f6f00e-568e-40ba-81ff-20ad452bfaa0.png">

Though it's hard for me to tell when something looks good or I've been staring at it for too long 😅 Any other ideas that we want to test out? cc @iterative/vs-code, @shcheklein

shcheklein:
@julieg18 I like it! it's simple and explicit. We'll be able to add a dropdown of some other element to configure which commits to show as a next step, it's an obvious and good place for that.

I don't think we need ":"

julieg18:
Two design ideas so far for commit tooltips on the experiment table! 

<img width="310" alt="Screenshot 2023-01-13 at 11 49 16 AM" src="https://user-images.githubusercontent.com/43496356/212388401-56b5a00e-fe5b-497f-84fd-10fdbc150a08.png">
<img width="261" alt="Screenshot 2023-01-13 at 11 58 00 AM" src="https://user-images.githubusercontent.com/43496356/212388405-c8ce0f96-ac52-45ed-93de-9aaa0a5cc6b0.png">

Any preference or other ideas/inspiration? cc @iterative/vs-code, @shcheklein

shcheklein:
[...]

PR #3053, opened 2023-01-04
Fix missing commit messages in experiment branches

julieg18:
Ensures that our `git log` call returns the commit messages needed for all experiment branches since [`dvc exp show` doesn't always return the last three commits](https://github.com/iterative/vscode-dvc/pull/3035#discussion_r1060120952). 

Part of #3036


extension/src/experiments/model/collect.test.ts
```diff
@@ -7,15 +7,15 @@ import {
 } from '../../cli/dvc/contract'
 
 describe('collectExperiments', () => {
-  it('should return an empty array if no branches are present', async () => {
```
julieg18, line 10: Looks like these unneeded `async` and `await`s are leftover from #3035.

extension/src/experiments/index.ts
```diff
@@ -600,6 +600,21 @@ export class Experiments extends BaseRepository<TableData> {
     return experiment?.id
   }
 
+  private getOldestCommitHashFromData(data: ExperimentsOutput) {
```
julieg18, line 603: Is there a better way to find the oldest commit hash?

extension/src/experiments/model/collect.test.ts
mattseddon: can we turn this on: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/await-thenable.md?

extension/src/experiments/index.ts
mattseddon: I think it would be safe to assume that the last one is the oldest. From the docs: 
```
Experiments in the table are first grouped (by parent commit). 
They are then sorted inside each group, chronologically by default. 
The --sort-by and --sort-order options can change this ordering, based on any single, visible metric or param.
```

```
--sort-order {asc,desc} - sort order to use with --sort-by. Defaults to ascending (asc).
```

```
-n <num>, --num <num> - show experiments from the last num commits (first parents) starting from the --rev baseline. 
Give a negative value to include all first-parent commits (similar to git log -n).
```

extension/src/experiments/model/collect.ts
```diff
@@ -316,22 +316,32 @@ const formatCommitMessage = (commit: string) => {
   return `${lines[0]}${lines.length > 1 ? ' ...' : ''}`
 }
 
+const getCommitMessages = (commitsOutput: string) => {
+  const commits = commitsOutput.split('\u0000').map(commit => {
```
mattseddon, line 320: [Q] Could this be a constant imported from `extension/src/cli/git/constants.ts`?

extension/src/cli/git/reader.ts
```diff
@@ -7,7 +7,7 @@ import { trimAndSplit } from '../../util/stdout'
 import { isDirectory } from '../../fileSystem'
 
 export const autoRegisteredCommands = {
-  GIT_GET_LAST_THREE_COMMIT_MESSAGES: 'getLastThreeCommitMessages',
+  GIT_GET_COMMIT_MESSAGES_UP_TO_HASH: 'getCommitMessagesUpToHash',
```
mattseddon, line 10: [C] Can probably drop `upToHash` from the name.

extension/src/experiments/model/collect.ts
mattseddon: This being `'\u0000'`

mattseddon (approved):

extension/src/experiments/index.ts
julieg18: > I think it would be safe to assume that the last one is the oldest. From the docs:

The experiment data is an object though and I thought the order isn't guaranteed in JS objects 🤔 I'll go ahead and update the code for now though!

sroy3 (approved):

PR #3054, opened 2023-01-05
Add new stop queue action

mattseddon:
# 1/2 `main` <- this <- #3059

### Demo

https://user-images.githubusercontent.com/37993418/210924793-0bb59b85-67ac-430f-a7ed-cdd107c25e32.mov


sroy3 (approved):

extension/package.json
```diff
@@ -775,7 +780,11 @@
         },
         {
           "command": "dvc.startExperimentsQueue",
-          "when": "dvc.commands.available && dvc.project.available && !dvc.experiment.running"
+          "when": "dvc.commands.available && dvc.project.available"
+        },
+        {
+          "command": "dvc.stopExperimentsQueue",
+          "when": "dvc.commands.available && dvc.project.available"
```
julieg18, line 787: The stop queue command is always available even when the queue isn't running. Would we want the command to only be available when the queue is running?

julieg18 (approved):
Great work!

extension/package.json
mattseddon: We would need to interrogate `dvc queue status` to know whether or not there is anything running as the information is not readily available from `exp show` (queue can be started outside of VS Code, e.g in a terminal). It is essentially a no-op when the queue is not running. For now this is good enough.

Issue #3044, opened 2023-01-03
Improve quick pick experiment selection


julieg18:
- [x] The commands "Modify Experiment Param(s) and Resume", "Modify Experiment Param(s) and Queue" , and "Modify Experiment Param(s) and Run" are only showing the branches instead of all experiments. 
- [x] The command "Remove Experiment" (possible some others as well) doesn't let you filter by description and detail
- [x] All quick picks for experiment selection should be showing the first three columns
- [ ] Change "Remove Experiment" to "Remove Experiments"

mattseddon:
I think `Remove Experiment` should be `Remove Experiments` and moved to `canPickMany`

PR #3056, opened 2023-01-05
Fix filtering and add missing experiments on quick pick experiment selection

julieg18:
* Allow filtering by description/detail on quick pick experiment selection
* Add missing experiments on "Modify Experiment Param(s) and ..." commands

### Filtering

https://user-images.githubusercontent.com/43496356/210830190-95339af3-1356-4317-ba00-bebb5a322beb.mov

### Missing Experiments

https://user-images.githubusercontent.com/43496356/210850007-86f7f2ab-8cdf-4cb3-9684-219b7209807f.mov

Part of #3044

### DIFF
--- a/extension/src/experiments/quickPick.ts
+++ b/extension/src/experiments/quickPick.ts
@@ -21,7 +21,7 @@ export const pickExperiment = (
         label,
         value: { id, name: name || label }
       })),
-      { title }
+      { matchOnDescription: true, matchOnDetail: true, title }
     )
   }
 }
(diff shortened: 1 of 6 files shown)


extension/src/experiments/quickPick.ts
```diff
@@ -21,7 +21,7 @@ export const pickExperiment = (
         label,
         value: { id, name: name || label }
       })),
-      { title }
+      { matchOnDescription: true, matchOnDetail: true, title }
```
julieg18, line 24: I set `matchOnDescription` and `matchOnDetail` to be true for experiment selection, but we could also set it by default or on other commands like applying filters.

mattseddon (approved):

PR #3059, opened 2023-01-06
Create queue workers in background process

mattseddon:
# 2/2 `main` <- #3054 <- this

I noticed when testing #3054 that the stdout of `dvc queue start` was not being returned to a Toast message until after the worker(s) had exited. This was due to some strange behaviour with the child process. This made me realise that we need to create the process that starts the queue in a completely different way so that it can keep running when/if the main VS Code process is closed or exits.

### Demo

https://user-images.githubusercontent.com/37993418/210912631-475b2acc-6a55-48d0-bf8f-1d6d0058cb5b.mov


extension/src/test/suite/cli/background.ts
```diff
@@ -0,0 +1,7 @@
+import { Logger } from '../../../common/logger'
+import { delay } from '../../../util/time'
+
+delay(1000).then(() => {
```
mattseddon, line 4: [F] This is a script that will stay alive for a lot longer than it needs to. It is run as a child process in a couple of places.

extension/src/test/suite/cli/child.ts
```diff
@@ -0,0 +1,25 @@
+import { getOptions } from './util'
```
mattseddon, line 1: [F] The purpose of this script is to create a background process and give the pid back to the test.

extension/src/cli/index.ts
```diff
@@ -101,6 +96,60 @@ export class Cli extends Disposable implements ICli {
...
+      return new Promise<string>(resolve => {
+        const readable = backgroundProcess.all
+        readable?.on('data', chunk => {
+          resolve(
+            chunk
+              .toString()
+              .split(/(\r?\n)/g)
+              .join('\r')
+          )
+          if (backgroundProcess.connected) {
```
mattseddon, line 116: See [this answer](https://stackoverflow.com/questions/59646654/detach-a-spawn-child-process-after-the-start).

extension/src/cli/index.ts
```diff
@@ -101,6 +96,60 @@ export class Cli extends Disposable implements ICli {
...
 
+  protected createBackgroundProcess(options: ProcessOptions) {
+    const { baseEvent, stopWatch } = this.getProcessDetails(options)
+    try {
+      const backgroundProcess = this.createProcess(baseEvent, {
+        detached: true,
+        ...options
+      })
+
+      return new Promise<string>(resolve => {
```
mattseddon, line 107: [F] Here we are attaching a listener to the readable stream so that we can get the initial stdout then cut the process loose.

extension/src/test/suite/cli/child.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Logger } from '../../../common/logger'
+import { delay } from '../../../util/time'
+
+require('../../../vscode/mockModule')
```
mattseddon, line 5: `vscode` is not available in the child process but we need it in order to test.

extension/src/test/suite/cli/background.ts
```diff
@@ -0,0 +1,7 @@
+import { Logger } from '../../../common/logger'
+import { delay } from '../../../util/time'
+
+delay(2000).then(() => {
```
mattseddon, line 4: [F] This is a script that will stay alive for a lot longer than it needs to. It is run as a child process in a couple of places.

extension/src/test/suite/experiments/model/tree.test.ts
```diff
@@ -17,7 +17,7 @@ import { UNSELECTED } from '../../../../experiments/model/status'
 import {
   experimentsUpdatedEvent,
   getFirstArgOfLastCall,
-  getSafeWatcherDisposer,
+  getTimeSafeDisposer,
```
mattseddon, line 20: [F] Needed to rename because I am using it to dispose and flush a process

extension/src/processExecution.ts
```diff
@@ -26,22 +27,46 @@ export interface ProcessOptions {
...
+
+  if (!detached) {
+    return options
+  }
+
+  // a detached/non-shell/windowsHide process will error on Windows:
+  // https://github.com/nodejs/node/issues/21825#issuecomment-503766781
+  // use this workaround: https://github.com/sindresorhus/execa/issues/433
+  const detachedOption =
+    getProcessPlatform() === 'win32' ? { shell: true } : { detached }
```
mattseddon, line 55: tbh I am not massively confident that this will work as expected on Windows. At least the process doesn't crash.

sroy3: I can test this later today on my Windows machine if that helps.

sroy3 (approved):

julieg18 (commented):
Probably just my computer being slow or a local issue, but when I started the queue the command ran fine but the DVC loading icon nevery stopped. On further inspection, I noticed that the `queue start` command only initialized and never completed in the DVC output. 

https://user-images.githubusercontent.com/43496356/211066354-806bdbdc-9618-408d-b5be-90cefb1ce33b.mov

I cut the first half of the video so the video starts when the queue just finished running.

extension/src/processExecution.ts
mattseddon: Thanks

mattseddon:
> Probably just my computer being slow or a local issue, but when I started the queue the command ran fine but the DVC loading icon nevery stopped. On further inspection, I noticed that the `queue start` command only initialized and never completed in the DVC output.
> 
>  trimmed.mov 
> I cut the first half of the video so the video starts when the queue just finished running.

@julieg18 I've made an update. Can you please see if you can still recreate the issue?

julieg18 (approved):
> @julieg18 I've made an update. Can you please see if you can still recreate the issue?

The `queue start` command now completes and everything is running correctly!

PR #3067, opened 2023-01-08
Add first three columns to all experiment quick pick selection

julieg18:
## Demo

https://user-images.githubusercontent.com/43496356/211176722-453bd18a-e6ee-45e6-91b0-5823a7222d7c.mov

Part of #3044


extension/src/experiments/quickPick.test.ts
```diff
@@ -94,18 +94,18 @@ describe('pickExperiment', () => {
       name: mockedExp.name
     }
     mockedQuickPickValue.mockResolvedValueOnce(expectedDetails)
-    const experiment = await pickExperiment(mockedExpList)
+    const experiment = await pickExperiment(mockedExpList, [])
```
mattseddon, line 97: [I] Can we get a regression test here that uses the first three columns.

mattseddon (approved):

extension/src/test/suite/experiments/workspace.test.ts
```diff
@@ -534,11 +545,17 @@ suite('Workspace Experiments Test Suite', () => {
...
+              '2020-12-29T15:26:36'
+            )}, loss:-, accuracy:-`,
             label: 'f0f9186',
             value: { id: 'exp-f13bca', name: 'exp-f13bca' }
           },
           {
             description: undefined,
+            detail: `Created:${formatDate(
+              '2020-12-29T15:25:27'
+            )}, loss:-, accuracy:-`,
```
mattseddon, line 558: [Q] why don't there have a loss or accuracy value?

sroy3 (approved):

extension/src/test/suite/experiments/workspace.test.ts
julieg18: This experiment (`55d492c9c633912685351b32df91bfe1f9ecefb9`) is a failed experiment and [has no `metrics`](https://github.com/iterative/vscode-dvc/blob/121f651589f2325cd1cef20ab7c93671e55d8322/extension/src/test/fixtures/expShow/base/output.ts#L1752).

PR #3070, opened 2023-01-08
Fix flaky setup tests

mattseddon:
This PR should fix a lot of the flakiness in the integration tests surrounding our `Setup` class. We will no longer be relying on calls to `DvcReader.prototype.version` to test the desired behaviour.


extension/src/test/suite/extension.test.ts
```diff
@@ -354,77 +278,6 @@ suite('Extension Test Suite', () => {
     })
   })
 
-  describe('dvc.checkCLICompatible', () => {
-    it('should call setup', async () => {
```
mattseddon, line 358: [F] I actually dropped this test. Not worth the effort to mock.

julieg18 (approved):
Great work!

sroy3 (approved):

PR #3079, opened 2023-01-09
Use Chromatic's Turbosnap functionality

mattseddon:
See [this thread](https://iterativeai.slack.com/archives/C01R00PPQ1L/p1672922660694809).

This change whilst not perfect (I do not believe changes to the image test fixture will trigger a full rebuild) will save a lot of snapshots being produced.

Things I have tested:

- Change in the extension alone triggers 0 snapshots to be generated
- Change in a single component generates a single snapshot
- Change to an `scss` file creates a full rebuild as it is referenced as [externals](https://www.chromatic.com/docs/turbosnap#specify-external-files-to-trigger-a-full-re-test-when-they-change) (not listing as externals means the changes would be missed).

### DIFF
--- a/webview/package.json
+++ b/webview/package.json
@@ -12,7 +12,6 @@
     "test": "jest --collect-coverage",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook --webpack-stats-json",
-    "chromatic": "npx chromatic --project-token fexq6zfy6",
     "svgr": "svgr --out-dir src/shared/components/icons --ignore-existing ../node_modules/@vscode/codicons/src/icons && svgr -d src/shared/components/icons icons/"
   },
   "main": "./index.js",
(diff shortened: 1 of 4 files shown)


webview/package.json
```diff
@@ -11,7 +11,7 @@
     "build": "webpack --mode production",
     "test": "jest --collect-coverage",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
+    "build-storybook": "build-storybook --webpack-stats-json",
     "chromatic": "npx chromatic --project-token fexq6zfy6",
```
yathomasi, line 15: Do we need this line? 
We already pass it as GitHub secrets. 
https://github.com/iterative/vscode-dvc/blob/19d1c560937c2690242a2e34c8eef8fc48890763/.github/workflows/continuous-integration.yml#L56

- https://www.sabbaticaldev.co.uk/post/chromatic-storybook

mattseddon: I was thinking the same thing. I don't think we need this script or even chromatic as a dependency in the sub-repo. I'll remove.

sroy3 (approved):

julieg18 (approved):
Looks good to me!

PR #3084, opened 2023-01-10
Add kill experiments queue tasks command

### LINKED ISSUE #3042: Integrate new queue functionality
Quick wins

 - [ ] Add number of jobs functionality quick pick in front of starting queue
 - [ ] Add stop functionality when the queue is running
 - [ ] Add kill functionality when the queue is running

Consider #1995

mattseddon:
# 2/2 `main` <- #3083 <- this

Closes #3042

### Demo

https://user-images.githubusercontent.com/37993418/211462570-2964caa4-1610-43a4-92a5-4efbb462f2d6.mov

https://user-images.githubusercontent.com/37993418/211462727-699409c3-738b-42dc-9349-4a664f27799f.mov


extension/src/experiments/model/quickPick.ts
```diff
@@ -129,7 +133,63 @@
...
+    void Toast.showError(noExperimentsToSelect)
+  } else {
+    return quickPickValue<{ id: string; name: string }>(
+      getExperimentItems(experiments, firstThreeColumnOrder),
+      { matchOnDescription: true, matchOnDetail: true, title }
+    )
+  }
+}
+
+export const pickExperiments = (
```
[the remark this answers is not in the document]
mattseddon: @julieg18 you should be able to use this for `Change "Remove Experiment" to "Remove Experiments"` in #3044

sroy3 (approved):

extension/package.json
```diff
@@ -216,6 +216,11 @@
         "command": "dvc.init",
```
julieg18, line 216: Do we want a more clear message? Like "Killed experiments EXP_NAME and EXP_NAME" for example 

Screenshot 2023-01-10 at 8 47 37 AM

julieg18 (approved):
Great work!

extension/package.json
mattseddon: We just pass on whatever the CLI provides. Operation successful is what we show when there is no output.

mattseddon: For more context: If the DVC team decided to add output for the command and we have added our own custom output then we will have probably diverged forever. IMO it is better to not "paper over the cracks" for things like this.

shcheklein: Let's create a ticket / PR to update this on the DVC side with this screenshot and motivation please. Good catch @julieg18 ! Good point @mattseddon on diverging.

PR #3087, opened 2023-01-10
Add "Previous Commits" row to experiments table

julieg18:
## Demo

https://user-images.githubusercontent.com/43496356/211624306-f1b47b3e-fdde-4515-8122-1e693b2d39e4.mov

Part of #3036


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -566,6 +566,19 @@ $bullet-size: calc(var(--design-unit) * 4px);
       outline-offset: -4px;
     }
   }
+
+  .otherCommitsRow {
```
julieg18, line 570: Currently the row has no hover animation but we could change the background color on hover like we do with other rows if preferred.

shcheklein: no need for this I think

extension/src/test/e2e/extension.test.ts
```diff
@@ -37,7 +37,7 @@ describe('Experiments Table Webview', function () {
   const webview = new ExperimentsWebview('experiments')
 
   const epochs = 15
-  const headerRows = 3
+  const headerRows = 4
```
mattseddon, line 40: [N] I would say it's an extra commit row, not a header (not at the top of the table)

mattseddon: or specifically, call out "pastCommitsRow"

mattseddon (approved):

Issue #3091, opened 2023-01-11
Queue: make sure tasks are hidden and all actions make sense in the table


shcheklein:
As we discussed during the planning, I don't think exposing tasks in the table, quick pick makes sense. E.g. remove action - we should be (at least first) be able to remove any experiment including queued, failed in the queue, etc, etc. 

From the user perspective- they don't care about tasks, celery, etc ... how it's implemented etc. All that stuff should be hidden by the experiments table and actions on experiments.

- [x] Make sure that we show all experiments in the table
- [x] Make sure that same actions can be applied seamlessly to all of those experiments. E.g. I can pick a failed and successful experiment + a queued experiments and delete them (do we need DVC support for this)
- [x] DVC should be providing the same experience I think (cc @dberenbaum )
- [x] Review quick picks, etc - don't expose extra actions on the queue for now (e.g. we don't want to have two removes, etc).
- [x] Discuss why would we actually want to expose this information as an advanced scenario, etc and do a followup after that

dberenbaum:
LGTM!

> * Make sure that same actions can be applied seamlessly to all of those experiments. E.g. I can pick a failed and successful experiment + a queued experiments and delete them (do we need DVC support for this)

VS Code should only need to pass the experiment names to `dvc exp remove` and it should work:

```
$ dvc queue status
Task     Name        Created       Status
be372c8  taboo-trey  08:47 AM      Queued
f285967  sworn-cuss  08:47 AM      Failed
63ea1ab  regal-weft  Jan 06, 2023  Success

Worker status: 0 active, 0 idle

$ dvc exp remove taboo-trey sworn-cuss regal-weft
Removed experiments: taboo-trey,sworn-cuss,regal-weft
```

mattseddon:
[Q] Do we want to hide `dvc queue kill <exp>`? Seems like we would want to provide two options with `dvc queue stop`:

1. Stop the current queue workers and leave the experiments currently being processed running.
2. Kill anything that is running and stop the queue workers.

shcheklein:
@mattseddon are those two different questions? or one? I'm trying to better understand it, sorry .. :)

mattseddon:
Things that are currently implemented

1. queue start (start queue workers)
2. queue stop (stop queue workers -> leaves any currently running queue tasks to complete)
3. queue kill (stop currently processing queue tasks, i.e experiments running in the queue - user is given all executing queue tasks as options)
[...]

PR #3094, opened 2023-01-11
Update queue commands

mattseddon:
- remove dvc.removeFromExperimentsQueue command
- rename dvc.killExperimentsQueueTasks command

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -341,11 +341,6 @@
         "icon": "$(debug-continue-small)"
       },
       {
-        "title": "%command.removeFromExperimentsQueue%",
-        "command": "dvc.removeFromExperimentsQueue",
-        "category": "DVC"
-      },
-      {
         "title": "%command.startExperimentsQueue%",
         "command": "dvc.startExperimentsQueue",
         "category": "DVC",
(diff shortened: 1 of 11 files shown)


shcheklein:
> Kill Experiments Running in the Queue

- Does it kill only running experiments or all queued?
- Does kill mean stop or delete?
- Can we generalize and make it "Stop all running experiments"?
- How does it correlate with the stop button, or does it work only for the workspace now? (which is probably fine, just to make sure)
-

mattseddon:
> > Kill Experiments Running in the Queue
> 
> * Does it kill only running experiments or all queued?

Kill will "Gracefully interrupt running experiment queue tasks (equivalent to Ctrl-C)." you have to pass the IDs of the tasks you want to kill. https://dvc.org/doc/command-reference/queue/kill

> * Does kill mean stop or delete?

Kill means stop running queue tasks. 

> * Can we generalize and make it "Stop all running experiments"?

That is what I was asking [here](https://github.com/iterative/vscode-dvc/issues/3091#issuecomment-1379379205). Stopping queue workers and stopping currently running queue tasks (experiments running in the queue) are separate things. If we want to provide an option to stop both at the same time we would need to add [dvc queue stop --kill](https://dvc.org/doc/command-reference/queue/stop#--kill). Right now the user has the option to kill any of the tasks that are running. 

> * How does it correlate with the stop button, or does it work only for the workspace now? (which is probably fine, just to make sure)

The stop button only relates to the workspace.

shcheklein:
> Kill means stop running queue tasks.

what happens to them from the user perspective? do I still see them in the table?

mattseddon:
> > Kill means stop running queue tasks.
> 
> what happens to them from the user perspective? do I still see them in the table?

They become failed experiments. From the docs:

_Note that killed experiments will be considered failed runs and will not be re-added to the queue for future execution_.

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3095, opened 2023-01-11
Upgrade react-table

### LINKED ISSUE #2130: Upgrade react-table to the newest version
React table was completely rebuilt and version 8 should have lots of breaking changes.

### LINKED ISSUE #2307: Table column resizers for last item in group should span all groups
I thought the resizer would take the whole border when it's the last item of the group since it's the whole group that is resizing. I believe that it was the original behaviour and that it was changed along the way (could be wrong on that one).

https://user-images.githubusercontent.com/3683420/187740873-71586548-402e-4723-a80c-f65b4f77a65b.mov

_Originally posted by @sroy3 in https://github.com/iterative/vscode-dvc/pull/2305#pullrequestreview-1092253554_

sroy3:
Closes #2130 
Will also fix #2307

Hopefully no use cases were missed. 

In the end, if you look at the diff size, the upgrade should be very beneficial code-wise.


shcheklein:
Not sure if this is related, but I'm stuck on this:


https://user-images.githubusercontent.com/3659196/213599584-baefe6d5-b314-4ba5-a17c-dc57b91fe4cb.mov

shcheklein:
A few issues:

- shadow in the first column appears after you open the table first time
- borders are not stable when you scroll

https://user-images.githubusercontent.com/3659196/213599972-fa4ef511-c779-4571-8d9c-e34f0fe286f1.mov

shcheklein:
- Fonts look different to me (e.g. header - bold vs non bold)
- Created column layout is different (dates take too much space atm)
- Cell borders looks different on row hover in selected and non-selected state

shcheklein:
@sroy3 @julieg18 Q: I thought we had counts on collapsed chevrons for experiments - I don't see them now (in this version and in the production one) - am I missing something or is it a regression?

sroy3:
> @sroy3 @julieg18 Q: I thought we had counts on collapsed chevrons for experiments - I don't see them now (in this version and in the production one) - am I missing something or is it a regression?

Regression, I did not notice them. I'll add them.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -661,7 +663,8 @@ $bullet-size: calc(var(--design-unit) * 4px);
 }
 
 .copyButton {
```
julieg18, line 707: The copy button isn't appearing when I'm hovering on cells: 

image

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -785,7 +788,7 @@ $bullet-size: calc(var(--design-unit) * 4px);
   text-align: left;
 }
 
-.thead {
```
julieg18, line 788: I can no longer decrease the column width to be shorter than the text width itself: 

image

webview/src/experiments/components/table/CellRowActions.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { MouseEventHandler } from 'react'
```
julieg18, line 1: Not sure why, but when I try to hover of the `workspace` radio button, the tooltip appears right on top of it: 

image

julieg18 (commented):
Great work so far, the code looks a lot cleaner in multiple places!

webview/src/experiments/components/table/Cell.tsx
```diff
@@ -34,22 +40,26 @@ const RowExpansionButton: React.FC<RowProp> = ({ row }) =>
...
   } = row
   const { toggleExperiment } = rowActionsProps
 
   return (
-    <div
-      {...cell.getCellProps({
-        className: cx(styles.td, styles.experimentCell)
-      })}
-    >
+    <td className={cx(styles.experimentCell)} style={{ width: getSize() }}>
```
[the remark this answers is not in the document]
sroy3: Good catch

webview/src/experiments/components/table/styles.module.scss
sroy3: It's one of the first thing I fixed, but I guess it got lost with the merge. Fixed back now.

webview/src/experiments/components/table/styles.module.scss
sroy3: Fixed!

webview/src/experiments/components/App.test.tsx
```diff
@@ -69,40 +69,6 @@ afterEach(() => {
```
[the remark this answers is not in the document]
sroy3: Should be pretty much fixed now

sroy3:
Only thing that should be left is the gap between first cell of highlighted rows. Table with sticky components and borders don't go well together and that first row is already full of `::before` and `::after` elements. Should be able to get around this soon.

Screenshot 2023-01-20 at 4 02 57 PM

shcheklein:
New finds:

- [ ] Steange header title behavior on resize:

https://user-images.githubusercontent.com/3659196/213843725-282286a9-5aa1-4424-b38e-e97e6a9dad62.mov

- [ ] First cell border is missing, cell border are not aligned with header:

Screen Shot 2023-01-20 at 7 22 28 PM

- [ ] Something weird is happening with the shadow and borders

Screen Shot 2023-01-20 at 7 25 09 PM

- [ ] Check defaults for column sizes, they look very large by default

Screen Shot 2023-01-20 at 7 28 46 PM

- [ ] Resizer doesn't span the whole column height:

Screen Shot 2023-01-20 at 8 45 32 PM

- [ ] No cell borders on selected row on hover:

Screen Shot 2023-01-20 at 8 46 51 PM

sroy3:
> * [ ]  First cell border is missing, cell border are not aligned with header:
> 
> Screen Shot 2023-01-20 at 7 22 28 PM
> 
> * [ ]  Something weird is happening with the shadow and borders
> 
> Screen Shot 2023-01-20 at 7 25 09 PM

These are expected from what is missing regarding my last comment (https://github.com/iterative/vscode-dvc/pull/3095#issuecomment-1398934127). Looking to fix this now.

> * [ ]  Resizer doesn't span the whole column height:
> 
> Screen Shot 2023-01-20 at 8 45 32 PM

Same as the current behaviour. I do remember fixing this a long time ago, so I guess there was a regression at some point. We could probably look to fix this as a follow-up.

shcheklein:
>  Resizer doesn't span the whole column height:

Yep, it also depends on my end - sometimes it does in the current version, sometimes not. Some examples:


Screen Shot 2023-01-23 at 9 24 08 AM
Screen Shot 2023-01-23 at 9 24 14 AM
Screen Shot 2023-01-23 at 9 24 17 AM

sroy3:
I think everything was addressed. For the resizer, I'd do this as a follow-up as it isn't a regression from the upgrade (and the new behaviour is actually more predictive).

sroy3:
> New findings:
> 
> * [ ]  Steange header title behavior on resize:
> 
>  Screen.Recording.2023-01-20.at.7.26.46.PM.mov 

Couldn't reproduce. I know I was getting this behaviour as well in the  beginning. Is it fixed on your end?

shcheklein:
@sroy3 thanks! good progress, we are almost there it seems 😅 A few final things to double check:

> Couldn't reproduce. I know I was getting this behaviour as well in the beginning. Is it fixed on your end?

It still there (can be reproduced only with metric columns?)


https://user-images.githubusercontent.com/3659196/214142682-9d29bf51-8c2a-4551-813f-8ab6f5e41198.mov

>  For the resizer, I'd do this as a follow-up as it isn't a regression from the upgrade (and the new behaviour is actually more predictive).

I think it's partially a regression (see the video), but it's fine to do as a follow up.

Other things that I still can reproduce:

- Border misalignment:

Screen Shot 2023-01-23 at 12 27 20 PM

- I still don't see counts when sections are collapsed (again, not sure if this is related to this ticket or not)
- some weird stuff is happening with borders on a high contrast theme (not sure again if this is a regression or not):

Screen Shot 2023-01-23 at 12 39 15 PM

sroy3:
> @sroy3 thanks! good progress, we are almost there it seems 😅 A few final things to double check:
> 
> > Couldn't reproduce. I know I was getting this behaviour as well in the beginning. Is it fixed on your end?
> 
> It still there (can be reproduced only with metric columns?)
> 
>  Screen.Recording.2023-01-23.at.12.25.14.PM.mov 

I see now! Will look into it.

> > For the resizer, I'd do this as a follow-up as it isn't a regression from the upgrade (and the new behaviour is actually more predictive).
> 
> I think it's partially a regression (see the video), but it's fine to do as a follow up.
> 
I don't know why, but when I did that part I compared to the previous version and got the same thing. I also got longer than usual lines. But it seems like this is happening, when columns have been shuffled a lot. So you are right, it is a regression (sorry it took so long to see).

> Other things that I still can reproduce:
> 
> * Border misalignment:
> 
> Screen Shot 2023-01-23 at 12 27 20 PM
> 
> * I still don't see counts when sections are collapsed (again, not sure if this is related to this ticket or not)
We only show count for selected, starred, and plotted. We might have removed them before. Is this useful information (to know how many subrows there are when collapsed)? IMO it feels like something hidden needs my attention. 

> * some weird stuff is happening with borders on a high contrast theme (not sure again if this is a regression or not):
> 
> Screen Shot 2023-01-23 at 12 39 15 PM
I'll get all these borders right eventually.

sroy3:
🤞🏼 Everything is fixed. Will have another fresh look tomorrow, because I think I might be losing the big picture right now at the end of the day. Feel free to recheck or to wait until I have taken a second look tomorrow.

shcheklein:
Thanks, @sroy3 ! I think everything reported above is fixed.

One more thing that I found (the first column width jumps):

https://user-images.githubusercontent.com/3659196/214201649-f7ba8464-17e5-47e4-91fc-de4149013d7e.mov

Btw, on the resizer. I think it makes sense from the product perspective to show them separate in some cases:


https://user-images.githubusercontent.com/3659196/214201736-2d26471e-5c47-4325-8b23-5a7c875cb1b0.mov

They have a bit different semantics. The group one can change width evenly underneath it. While the cell one at the bottom can change only that cell width. It's not exactly consistent at the moment, but it might be easier to fix it to work that way. WDYT?

sroy3:
> Btw, on the resizer. I think it makes sense from the product perspective to show them separate in some cases:
> 
>  Screen.Recording.2023-01-23.at.6.34.50.PM.mov 
> They have a bit different semantics. The group one can change width evenly underneath it. While the cell one at the bottom can change only that cell width. It's not exactly consistent at the moment, but it might be easier to fix it to work that way. WDYT?

How about resizer lines span all the way to the top (to ancestor columns), but they do not span down to descendant and leaf columns)? I think I can do that and it makes the most sense. If you're resizing a leaf column, the parent columns also resize accordingly. If we resize a parent column, the new size applies to all child columns.

sroy3:
I went over pretty much everything this morning and everything is pretty much as it was without any border or misalignment problems. We can see from storybook that the biggest difference is in the placeholders (ancestor column vs leaf column). This is a difference inside of react-table itself and there isn't much we can do about it. Other than that, things move a little bit, but this is expected when changing the layout so much.

shcheklein:
@sroy3 

> This is what this looks like:

Idea makes total sense to me. Let's do it that way. Not a blocker for merging this if it's approved from the code perspective by @julieg18 and/or @mattseddon .

From the video though, I see that it's a bit inconsistent still, right? E.g. when you move the `step` cell the resizer should snap only that header cell, right? Not the whole height.

shcheklein (approved):

sroy3:
> @sroy3
> 
> > This is what this looks like:
> 
> Idea makes total sense to me. Let's do it that way. Not a blocker for merging this if it's approved from the code perspective by @julieg18 and/or @mattseddon .
> 
> From the video though, I see that it's a bit inconsistent still, right? E.g. when you move the `step` cell the resizer should snap only that header cell, right? Not the whole height.

It also resizes the `training/metrics.json` cell though (we cannot really change that behaviour as it is the last column for `training/metrics.json`. Rules are whatever the level, it spans to the top of the table, but it ends with the column (it won't go further down).

shcheklein:
> It also resizes the training/metrics.json cell though (we cannot really change that behaviour as it is the last column for training/metrics.json. Rules are whatever the level, it spans to the top of the table, but it ends with the column (it won't go further down).

Okay, makes sense. Let's proceed with this!

PR #3105, opened 2023-01-16
Add experiment row commit tooltips

julieg18:
* adds commit tooltips to branch row secondary cell names

## Demo

https://user-images.githubusercontent.com/43496356/212734465-2b859761-66ef-4d23-bb1f-99189be3673f.mov

Part of #3036


extension/src/cli/git/constants.ts
```diff
@@ -33,7 +33,7 @@ export enum Flag {
   NAME_ONLY = '--name-only',
   NO_EMPTY_DIRECTORY = '--no-empty-directory',
   NUMBER = '-n',
-  PRETTY_FORMAT_COMMIT_MESSAGE = '--pretty=format:%H %B',
+  PRETTY_FORMAT_COMMIT_MESSAGE = '--pretty=format:%H%n%an%n%ar%nrefNames:%D%nmessage:%B',
```
julieg18, line 36: I added a "key name" to get tags (`%D`) and the message (`%B`) since `%D` can possibly output an empty string and the message can by multiple lines.

extension/src/experiments/model/collect.ts
```diff
@@ -312,19 +317,51 @@ const collectFromBranchesObject = (
...
+    tags: []
+  }
+  const [hash, author, date, refNamesWithKey] = output
+    .split('\n')
+    .filter(Boolean)
+  data.hash = hash
+  data.author = author
+  data.date = date
+
+  const message = output.match(/\nmessage:(.+)/s) || []
```
julieg18, line 337: Collecting the message and the tag names ended up being a little more complicated than the other items (see previous comment). Any ideas on simplifying things are welcome!

webview/src/experiments/components/table/CellSecondaryName.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </p>
+      <p className={styles.message}>{message}</p>
+    </div>
+  )
+
+  return (
+    <Tooltip
+      placement="bottom-start"
+      appendTo={document.body}
+      content={tooltipContent}
```
julieg18, line 47: Originally I was going to make the tooltip interactive but ran into a bug when you try to click on the text. Should be fixable but figured it could be done in a follow-up: 

https://user-images.githubusercontent.com/43496356/212735822-dae8a6c2-a85c-4436-897b-d334a121b968.mov

shcheklein (approved):

webview/src/experiments/components/table/CellSecondaryName.tsx
sroy3: Was this with the `interactive` option set to `true` on the tooltip?

extension/src/experiments/model/collect.ts
```diff
@@ -312,19 +317,51 @@ const collectFromBranchesObject = (
...
+  data.message = message[1] || ''
```
sroy3, line 338: I believe this is going to break if `message` is empty. It'll get an out of range index errror.

sroy3 (approved):

webview/src/experiments/components/table/CellSecondaryName.tsx
julieg18: > Was this with the interactive option set to true on the tooltip?

Yes!

extension/src/experiments/model/collect.ts
julieg18: I'll adjust the code accordingly! Though, I'm a little confused why the error would be thrown. Wouldn't `message[1]` just return undefined?

sroy3: You're right. I don't know where my previous assumption came from.

julieg18: > You're right. I don't know where my previous assumption came from.

No worries, I'll just undo my last commit 😄 I think other languages would have thrown that kind of error, like Python.

PR #3110, opened 2023-01-16
Add `workflow_dispatch` event to `publish` workflow

julieg18:
Allows us to manually start the `publish` workflow in case of an [issue with the workflow automatically running](https://iterativeai.slack.com/archives/C01RB7BTG4C/p1673880179770639).


mattseddon (approved):
Break glass in case of emergency

sroy3 (approved):

PR #3154, opened 2023-01-25
Update user facing queue terminology

mattseddon:
# 1/3 `main` <- this <- #3157 <- #3158

Related to #3091.

This change updates the user-facing terminology of some of our experiments queue commands.

PR #3155, opened 2023-01-25
Add stop experiments queue processing immediately command

mattseddon:
# 2/? `main` <- #3154 <- this

This PR implements the command alluded to [here](https://github.com/iterative/vscode-dvc/pull/3154#discussion_r1086148224). This command not only stops all experiments queue workers but kills currently executing queued experiments. I've added this command so that it can be wired into the stop button.

Docs for the stop command are [here](https://dvc.org/doc/command-reference/queue/stop).

### Demo

https://user-images.githubusercontent.com/37993418/214470456-5f966ea2-dfc6-400e-8729-237c5ef442a9.mov

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -347,6 +347,11 @@
         "category": "DVC"
       },
       {
+        "title": "%command.stopExperimentsQueueAndRunning%",
+        "command": "dvc.stopExperimentsQueueAndRunning",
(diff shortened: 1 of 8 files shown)


mattseddon:
Dumping this in favour of adding the exact functionality to `dvc.stopRunningExperiments`.

PR #3156, opened 2023-01-25
Fix Experiments table multi-select context menu

mattseddon:
Experiments table multi-select context menu actions were broken by the react table upgrade. The row's new ID was being used instead of the row's original ID. Understandable mistake as the behaviour was not tested.

### Before

https://user-images.githubusercontent.com/37993418/214477748-51c5e58e-8add-4919-ab03-83e5df0bf7fc.mov

### After

https://user-images.githubusercontent.com/37993418/214477820-4a2be603-94ca-496d-bd41-a064e19c9551.mov

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -48,6 +48,7 @@ import {
   setTableData
 } from '../../test/experimentsTable'
 import { clearSelection, createWindowTextSelection } from '../../test/selection'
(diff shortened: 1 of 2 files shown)


mattseddon:
@sroy3 anywhere else this could have happened in the upgrade?

sroy3 (approved):
Thanks! This one must have slipped passed me. This was the number one cause of the bugs I had after the upgrade (other than borders). I thought I had added `original` everywhere, but I must have forgotten than one.

PR #3157, opened 2023-01-25
Add ability to stop experiments running in the queue into stop experiments command

mattseddon:
- Update user facing queue terminology
- add ability to stop experiments running in the queue into stop button


shcheklein:
Q: does it stop (kill) all the experiments. E.g. If I have one running in the queue and one in the workspace- will it kill both of them now?

mattseddon:
> Q: does it stop (kill) all the experiments. E.g. If I have one running in the queue and one in the workspace- will it kill both of them now?

Yes, it will. I thought this was the best way to unify the functionality.

sroy3 (approved):

PR #3158, opened 2023-01-25
Refactor and reduce code in context class

mattseddon:
# 3/3 `main` <- #3154 <- #3157 <- this

A little bit of housekeeping on the `Context` class.

### DIFF
--- a/extension/src/context.ts
+++ b/extension/src/context.ts
@@ -39,53 +39,37 @@ export class Context extends Disposable {
 
         void this.setIsExperimentRunning(repositories)
 
-        this.setContextFromRepositories(
+        void setContextValue(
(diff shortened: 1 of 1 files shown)


extension/src/context.ts
```diff
@@ -61,31 +59,16 @@ export class Context extends Disposable {
...
     )
 
     void setContextValue(
       ContextKey.EXPERIMENT_STOPPABLE,
-      hasRunningQueuedExperiment ||
-        (await this.experiments.hasDvcLiveOnlyExperimentRunning())
-    )
-  }
-
-  private setContextFromRepositories(
```
mattseddon, line 81: [F] Realised in the previous PR that once we use `.some` we don't really need this function.

sroy3 (approved):

PR #3162, opened 2023-01-25
docs: fixing demo link in CONTRIBUTING.md

Casyfill:
It seems the demo link is broken, is the one I passed the correct one?


shcheklein (approved):
Yes, great catch, thanks @Casyfill !

PR #3166, opened 2023-01-27
Combine remove experiments and remove queued experiment quick picks

mattseddon:
# 1/2 `main` <- this <- #3167 

Related to https://github.com/iterative/vscode-dvc/issues/3091

This PR drops the `dvc.removeQueuedExperiment` quick pick command and enables users to select queued experiments through `dvc.removeExperiments`. In other words, we are combining the commands. For now I have left the `DVC: Remove All Queued Experiments` (`dvc exp remove --queue`) command in the UI as I think it has it's own place but we can remove that as well if we want to.

### Demo

https://user-images.githubusercontent.com/37993418/215000619-73843c86-4506-4081-9d4e-994a0e323737.mov


julieg18 (approved):
Looks good!

PR #3167, opened 2023-01-27
Replace branch terminology with commit for Experiments internals

mattseddon:
# 2/2 `main` <- #3166 <- this

This PR drops the branch terminology that we were incorrectly using all over the place to describe commit records within our experiments data. I have also attempted to improve the function naming in `ExperimentsModel` so that the names better represent what the functions are doing.

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -385,9 +385,9 @@ export class Experiments extends BaseRepository<TableData> {
     }
   }
 
-  public pickCurrentExperiment() {
+  public pickExperiment() {
     return pickExperiment(
-      this.experiments.getExperimentsWithoutQueued(),
+      this.experiments.getExperiments(),
       this.getFirstThreeColumnOrder()
     )
   }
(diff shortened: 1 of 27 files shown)


extension/src/experiments/model/index.ts
```diff
@@ -63,16 +63,16 @@ export type ExperimentWithCheckpoints = Experiment & {
 
 export enum ExperimentType {
   WORKSPACE = 'workspace',
-  BRANCH = 'branch',
+  COMMIT = 'commit',
```
mattseddon, line 66: [F] This type change is why we need to change the `package.json`

extension/src/experiments/index.ts
```diff
@@ -423,20 +423,20 @@ export class Experiments extends BaseRepository<TableData> {
...
   }
 
-  public getExperiments() {
+  public getWorkspaceAndCommits() {
     if (!this.columns.hasNonDefaultColumns()) {
       return []
     }
 
-    return this.experiments.getExperiments()
+    return this.experiments.getWorkspaceAndCommits()
```
mattseddon, line 431: [F] This change is what spurred this entire PR.

sroy3 (approved):

julieg18 (approved):
Great work! Makes a lot more sense then "branch".

PR #3174, opened 2023-01-30
Fix inital ordering of experiments table

### LINKED ISSUE #3172: Created column is rendered the last by default
I think it's broken after the https://github.com/iterative/vscode-dvc/pull/3095

To reproduce - launch any new project (I was using Codespaces for that) and open the experiments table. In my case (and I've checked also locally) `Created` column is rendered at the very end.

mattseddon:
Fixes #3172

### DIFF
--- a/webview/src/experiments/util/buildColumns.tsx
+++ b/webview/src/experiments/util/buildColumns.tsx
@@ -51,7 +51,7 @@ const getMainColumnProperties = (
         CellContext<Column, CellValue>
       >,
       header: () => <TimestampHeader />,
-      id: ColumnType.TIMESTAMP
+      id: 'Created'
     }
   }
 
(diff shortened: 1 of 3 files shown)


webview/src/experiments/util/buildColumns.tsx
```diff
@@ -51,7 +51,7 @@ const getMainColumnProperties = (
         CellContext<Column, CellValue>
       >,
       header: () => <TimestampHeader />,
-      id: ColumnType.TIMESTAMP
+      id: 'Created'
```
mattseddon, line 54: [F] This mismatched with the value provided in `columnOrder` which initially placed the column at the end of the table until any reordering added `id: 'timestamp'` to the table (and extension's) state.

mattseddon: tricky one to track down

shcheklein (approved):

Issue #3058, opened 2023-01-05
Onboarding to running experiments


dberenbaum:
Most options in the sidebar are about viewing experiments, which doesn't make much sense for a new project without any experiments. It would be good to focus on running experiments as the starting point.

@shcheklein Made the comparison to `launch.json`, and I think we can borrow a lot from VS Code's [debugging setup](https://code.visualstudio.com/docs/editor/debugging). When you first try to run an experiment, we can try to funnel users through a workflow like:

1. Pick the file to run the experiment. If it's not Python, what do we do?
2. Once the file is picked, we can show the instructions to add DVCLive, and maybe we can even validate that DVCLive has been added to the code before allowing users to continue.
3. Once DVCLive is added, users can choose whether to configure a pipeline (similar to the option to create `launch.json`).
4. If they choose not to generate a pipeline, any actions to run experiments should execute the file (can it instead run the active cell in a notebook when appropriate?).
5. If they choose to generate a pipeline, we know the source code file. If we can determine the DVCLive output directory, we could pre-fill dvc.yaml with something like:
```yaml
stages:
  train:
    cmd: python source_file.py
    deps:
      - source_file.py
    outs:
      - dvclive:
          cache: false
```

------

Summary of the discussion during the product sync with @dberenbaum :

- Next step for the onboarding is to focus on "launch configurations" (helping creating a dvc.yaml e.g. when user clicks on a run button)
- We can show a + Pipeline config on the experiments side panel
- On the DVCLive snippets screen we don't show anything extra yet. Eventually we can show alternatives for other languages.

------
- [x] Add a quick pick to prompt the user for their training script to add to the `dvc.yaml` file being created
- [x] Remove beaker button to show experiments and replace with the play button to run experiment (like it is currently is when showing the table)
- [x] If there is a `dvc.yaml` file, also check for a pipeline in the file (check for stages key or run `dvc stage list`)
- [x] Add a mention in the last onboarding step ("Your project contains no data") to tell the user to run their training script after adding DVCLive. (Should be common sense to work with the extension, but not so much if expecting the setup pages to show something in between those steps - "I've added DVCLive, now what?")
- [ ] Add a button below the table if there is no `dvc.yaml` file ("Add configuration")
- [x] Either replace the "Show Experiments" button with "Run Experiment" when the table is open or replace the "Show Experiments and Plots" button with it at all times
- [x] Accept other training scripts (not just python)

shcheklein:
[...]

PR #3177, opened 2023-01-30
Ensure there is a dvc.yaml file when running experiment

sroy3:
Part of #3058 

Creates an empty (seems to be all that is needed) dvc.yaml file when none exists before running experiment.

https://user-images.githubusercontent.com/3683420/215572713-4700089d-11c1-40a3-99fc-66d38be297b1.mov

Should we add a notification saying we created the file? Or even ask the user if they would like to create the file? Seems low impact enough to me to simply create the file for them.


mattseddon:
Talked through ideas for the next step with @sroy3. 

## Ideas that came up:

### Populating dvc.yaml
Run button shows a modal which asks the user if they want to add a dvc.yaml so they run an experiment.
Quick input then asks for the relative path to their training script.
dvc.yaml is created containing: 
```
stages:
  train:
    cmd: python relative/path/to/train/script
```

### Final step of onboarding
Add another step to the end of onboarding which shows the location of the run button (editor/title) and details what clicking it will do (see above idea).

extension/src/experiments/workspace.ts
```diff
@@ -213,12 +213,19 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
...
+    commandId: CommandId,
+    ensureDvcYamlFileExists?: boolean
+  ) {
     const cwd = await this.getFocusedOrOnlyOrPickProject()
     if (!cwd) {
       return
     }
 
+    if (ensureDvcYamlFileExists) {
+      ensureOrCreateDvcYamlFile(cwd)
```
mattseddon, line 226: [Q] `findOrCreateDvcYamlFile`?

mattseddon: or just `ensureDvcYamlFile`?

mattseddon (approved):
Couple things to check

1. does running an experiment with an empty `dvc.yaml` still respect the params created by Dvclive?
2. does this create any other weirdness when running experiments?

julieg18 (approved):

sroy3:
> Couple things to check
> 
> 1. does running an experiment with an empty `dvc.yaml` still respect the params created by Dvclive?
> 2. does this create any other weirdness when running experiments?

This works great for already set up projects (like https://github.com/iterative/dvc-get-started-cv/tree/1-notebook-dvclive), but it seems like it needs more than that for newly created projects. With the `dvc.yaml` file, it'll run, but the table will still have no columns.

sroy3:
Closing this as this is too simple as a first step. Will add a prompt to the training script and re-open.

PR #3188, opened 2023-01-31
Fix experiment table showing removed experiments

### LINKED ISSUE #3103: Table is not refreshed after experiment is removed
On Windows, conda, VS Code 1.74.3, extension 0.5.36:

https://user-images.githubusercontent.com/3659196/212576087-b247156c-8914-4b69-8112-4e6177a402cf.mov

```
[version: 0.5.36, 2023-01-16T00:32:31.286Z, pid: 4972] > C:\Users\ivan\miniconda3\envs\hackathon\python.exe -m dvc exp remove picky-wads - COMPLETED (1133ms)
```

```
DVC version: 2.41.1 (pip)
---------------------------------
Platform: Python 3.10.8 on Windows-10-10.0.22621-SP0
Subprojects:
        dvc_data = 0.29.0
        dvc_objects = 0.14.1
        dvc_render = 0.0.17
        dvc_task = 0.1.9
        dvclive = 1.3.2
        scmrepo = 0.1.5
Supports:
        http (aiohttp = 3.8.3, aiohttp-retry = 2.8.3),
        https (aiohttp = 3.8.3, aiohttp-retry = 2.8.3)
Cache types: hardlink
Cache directory: NTFS on C:\
Caches: local
Remotes: https
Workspace directory: NTFS on C:\
Repo: dvc, git
```

julieg18:
## Demo
Video looks pretty glitchy due to my windows screen recording software 😓 

https://user-images.githubusercontent.com/43496356/215830573-ff340a59-6ade-412f-b941-a70a99a15063.mp4

Fixes #3103


extension/src/experiments/data/index.ts
```diff
@@ -86,9 +86,12 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
 
     return createFileSystemWatcher(
       disposable => this.dispose.track(disposable),
-      getRelativePattern(dotGitPath, '**'),
```
julieg18, line 89: I couldn't figure out why but this pattern (`.git/**`) wouldn't work correctly, only firing for files directly in the .git folder ([more details in slack thread](https://iterativeai.slack.com/archives/C01RB7BTG4C/p1675103470404509)). The only solution I could come up with was having our `.git` watcher search the entire workspace folder and return if the fired path wasn't in the `.git` folder.

mattseddon: We can't take this approach as the folder can actually be outside of the workspace and/or the dvc root (e.g monorepo).

julieg18: Oh, you're correct, I didn't think about that! Will go ahead and close this pr.

PR #3192, opened 2023-01-31
Fix minor table bugs

julieg18:
* Fix hover styles for body rows and cells
* Fix misaligned icon button on the Created column
* Make icon buttons transparent when cell is a drag target (may not actually be a bug)
* Use `::` for all `before` and `after` rules (housekeeping)

## Demo

https://user-images.githubusercontent.com/43496356/216122446-b1a40c26-7ff1-4b1b-af51-91e3e479fe93.mov


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -56,6 +56,29 @@ $bullet-size: calc(var(--design-unit) * 4px);
   }
 }
 
+%cellBorderLeft {
```
julieg18, line 59: By having all cells, both header and body, use the same `:before` method to create the left border, the borders become perfectly aligned.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -649,6 +629,10 @@ $bullet-size: calc(var(--design-unit) * 4px);
       outline-style: dashed;
       outline-color: $header-dnd-outline;
       outline-offset: -4px;
+
+      .iconMenu ul[role='menu'] {
```
julieg18, line 633: Originally, a drag target with an icon looked like this:

image


Was this intentional since hovering over this button actually won't cause the target to drop? I thought it looked a little buggy so I made the button color transparent when the cell is a drop target, but I can easily revert the change if preferred :)

shcheklein: thanks @julieg18, I think it looks better with the fix. I think it should also allow dropping there tbh (no need to fix it if it's not a single obvious line)

shcheklein:
Thanks a lot for all these fixes!

sroy3 (approved):
Great work!

mattseddon:
Will include this in the next release. Please follow up if there is anything outstanding 🙏🏻

webview/src/experiments/components/table/styles.module.scss
julieg18: > thanks @julieg18, I think it looks better with the fix. I think it should also allow dropping there tbh (no need to fix it if it's not a single obvious line)

Might be doable with `pointer-events: none;` and a little JS 🤔, will try to look into it if I have some time one of these days :)

PR #3197, opened 2023-02-01
Create dvc.yaml file with training script on run experiment

sroy3:
Part of #3058 

https://user-images.githubusercontent.com/3683420/216372454-7f4f86ea-67c4-4a88-b9bd-4a63cedb0479.mov


shcheklein:
Thanks @sroy3 !

Two things to change/improve.

- we want path to be relative to the project root in the `dvc.yaml`
- we don't need `out` `dvclive` since we are not 100% sure if that it the case or we might not know the path (it can be customized)

we can discuss as a next step what to include in the file

shcheklein:
We'll need to handle the case if path is a notebook ... (again, can be a followup) ... or a non-Python command ...

sroy3:
> Thanks @sroy3 !
> 
> Two things to change/improve.
> 
> * we want path to be relative to the project root in the `dvc.yaml`
> * we don't need `out` `dvclive` since we are not 100% sure if that it the case or we might not know the path (it can be customized)
> 
> we can discuss as a next step what to include in the file

Done.

sroy3:
Updated demo:

https://user-images.githubusercontent.com/3683420/216689309-a9da898a-375c-4239-b60b-5609b8c68b66.mov

shcheklein:
@sroy3 thanks for the update!

Q - could you please remind me what is the current logic to show the table (when there are no columns) and when we show the Settings page instead.

sroy3:
> @sroy3 thanks for the update!
> 
> Q - could you please remind me what is the current logic to show the table (when there are no columns) and when we show the Settings page instead.

That's it. I may have cheated in my demo using the beaker instead of the "Show Experiments" button, which did always show the table, but that button isn't there anymore.

sroy3:
> We'll need to handle the case if path is a notebook ... (again, can be a followup) ... or a non-Python command ...

Added a [comment](https://github.com/iterative/vscode-dvc/issues/3058#issuecomment-1419516895) for that and currently working on this.

shcheklein:
>That's it. I may have cheated in my demo using the beaker instead of the "Show Experiments" button, which did always show the table, but that button isn't there anymore.

oh, got it. I wasn't aware about this backdoor :) There none left, right?

> Updated demo:

Perfect, if everything else looks good, we can merge this, I think.

extension/src/experiments/workspace.test.ts
```diff
@@ -244,4 +254,131 @@ describe('Experiments', () => {
...
+
+      await workspaceExperiments.getCwdThenRun(mockedCommandId, true)
+
+      expect(executeCommandSpy).toHaveBeenCalledWith(
+        mockedCommandId,
+        mockedDvcRoot
+      )
+    })
+
+    it('should not run the command if the path to the taining script was not given', async () => {
```
julieg18, line 366:
```suggestion
    it('should not run the command if the path to the training script was not given', async () => {
```

extension/src/experiments/workspace.ts
```diff
@@ -429,6 +440,36 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
...
 
+  private async checkOrAddPipeline(cwd: string) {
+    const stages = await this.internalCommands.executeCommand(
+      AvailableCommands.STAGE_LIST,
+      cwd
+    )
+
+    if (!stages) {
+      const selectValue = 'select'
+      const pathOrSelect = await quickPickOneOrInput(
```
julieg18, line 451: Should we put down some kind of Toast message mentioning we're creating a dvc.yaml file for running experiments?

extension/src/experiments/workspace.ts
```diff
@@ -429,6 +440,36 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
...
+        [{ label: 'Select from file explorer', value: selectValue }],
+        {
+          defaultValue: '',
+          placeholder: 'Path to script',
```
julieg18, line 455: We should probably be more clear that only absolute paths work, not relative. Personally, I think that we should allow relative paths in the input if possible. Seems easier to type out 🤔

julieg18 (approved):
Great work!

sroy3:
> > That's it. I may have cheated in my demo using the beaker instead of the "Show Experiments" button, which did always show the table, but that button isn't there anymore.
> 
> oh, got it. I wasn't aware about this backdoor :) There none left, right?

Right!

extension/src/experiments/workspace.ts
sroy3: I'm going to open the `dvc.yaml` file automatically in a follow-up.

mattseddon:
> > > That's it. I may have cheated in my demo using the beaker instead of the "Show Experiments" button, which did always show the table, but that button isn't there anymore.
> > 
> > 
> > oh, got it. I wasn't aware about this backdoor :) There none left, right?
> 
> Right!

Can you get to the table from `DVC: Show Experiments` in the command palette?

sroy3:
> > > > That's it. I may have cheated in my demo using the beaker instead of the "Show Experiments" button, which did always show the table, but that button isn't there anymore.
> > > 
> > > 
> > > oh, got it. I wasn't aware about this backdoor :) There none left, right?
> > 
> > 
> > Right!
> 
> Can you get to the table from `DVC: Show Experiments` in the command palette?

You are right. Would it be better to hide the show commands until the project is ready or to redirect to the setup?

mattseddon:
> You are right. Would it be better to hide the show commands until the project is ready or to redirect to the setup?

Hiding is probably easier but up to you. Will need to go through the same process for plots.

shcheklein:
Let's not hide it please. Ultimately, with the onboarding we should be trying to make the product look and feel 100% the same if something is installed or not. And we should be helping as we go collecting pieces of the puzzle, ideally through the existing interface as much as possible.

In this sense, I'm considering even the existence of the Setup page as a compromise ;)

I think the jupyter extension experience is a good example.

PR #3198, opened 2023-02-01
Remove read file contents request sent to extension by language server (use is file check)

### LINKED ISSUE #2529: Test and release LSP
- [ ] Regarding the possibility of the crashes coming from parsing the files, this can be done in the usual integration test in the server. An idea to do it is to send the same files a lot of times

- [ ] General smoke tests must be added to make sure the server is loading properly with the extension.

mattseddon:
This PR removes the need for the language server to request that the extension reads the contents of files. We may need to bring this back in the future but for now we don't need it.

Comments inline.

Closes #2529 (cc @shcheklein)


languageServer/src/textDocument.ts
```diff
@@ -17,19 +17,11 @@ import {
...
-): Location => {
-  const start = Position.create(0, 0)
-  const end = textDocument.positionAt(textDocument.getText().length - 1)
-  const range = Range.create(start, end)
-
-  return Location.create(textDocument.uri, range)
-}
+export const getLocationToOpen = (uri: string): Location => {
+  const position = Position.create(0, 0)
+  const range = Range.create(position, position)
```
mattseddon, line 22: [F] Turns out that we can open the file using an empty range (previously this didn't work or I had done the wrong thing to test). This means we don't need the file contents at all (for now). We can simply check whether or not the URI relates to a file. If it does then we can create a location with the URI string and an empty range. Passing this back to the client will open the file. If there are multiple locations (files) then the user will be prompted which one they want to open. That's the reason that we can't just ask the extension to open the file without first returning information to the server.

mattseddon: We also have to check whether or not the file is a file on the extension side because the LSP might not be running in a context that can access the file system (e.g Codespaces)

sroy3 (approved):

PR #3212, opened 2023-02-03
Fix missing button

sroy3:
Re-added the play button in the top right in the experiments table webview. I removed it while cleaning the actions (without realizing I was editing the editor title actio
Screenshot 2023-02-03 at 10 53 58 AM
ns... my bad).


sroy3:
Thought this would fix e2e tests, but problem seems to be something else. This still need to be fixed though.

mattseddon (approved):

PR #3228, opened 2023-02-06
fix(3226): remove legend object from Vega spec

### LINKED ISSUE #3226: Plots: Update legend suppression (smooth template)
Now that the default template is a smooth plot our (flaky) method of suppressing the legend and then reinstating it in a zoomed plot no longer works as expected/produces a nice result:

image

image

We need to decide whether or not we are going to continue attempting to display the legend/data in the old way or if we are going to take a different approach. If we are going to change then we need suggestions for how to integrate the sidebar/rest of the app.

After deciding on the approach we will need to spend time updating the implementation.

Note: I have spent some time looking into a fix/patch for the change and it appears that the change will be non-trivial. I think we are going to have to start walking through the template hierarchy to update the different color/strokedash/shape entries.

shcheklein:
Fixes #3226

https://user-images.githubusercontent.com/3659196/216887839-723ec8b8-39ad-4f9a-8d25-32b3f7cd723b.mov

TODO (can do as a followup if we are comfortable to do a release:

- [ ] Add tests
- [ ] Reduce complexity (generalize spec traversal)

### DIFF
--- a/extension/src/plots/model/collect.ts
+++ b/extension/src/plots/model/collect.ts
@@ -37,7 +37,7 @@ import {
 } from '../../experiments/webview/contract'
 import { addToMapArray } from '../../util/map'
 import { TemplateOrder } from '../paths/collect'
-import { extendVegaSpec, isMultiViewPlot } from '../vega/util'
+import { transformVegaSpec, isMultiViewPlot } from '../vega/util'
 import {
   definedAndNonEmpty,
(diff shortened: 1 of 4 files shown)


shcheklein:
Closing for now in favor of https://github.com/iterative/dvc-render/issues/115 (we can reopen if @dberenbaum decides other way).

Issue #3077, opened 2023-01-09
Sharing experiments


dberenbaum:
Related to https://github.com/iterative/vscode-dvc/issues/2855, the extension can make it easier to share experiments. Let's discuss what's needed here?

My initial thoughts on what's needed:

1. Show a comparison of all or a subset of experiments like what you see in the table and plots views, except that it's not stuck on your local machine.
2. Merge or otherwise move forward with an experiment that you think is a keeper. 

For 1, I think it makes sense to use Studio since it already has all this functionality. The extension can upload the params, metrics, and plots to Studio like dvclive is doing for live metrics (except for the "live" part). After selecting any number of experiments, there could be an option to post to Studio. The only user friction should be having a Studio token.

For 2, I think there are a lot of ways to do it in DVC already, so it's probably not as critical, but maybe VS Code can make it smoother. With one click, the extension could create a branch with the same name as the experiment, push that to GitHub, and show the URL to create a PR (like the git cli message `Create a pull request on GitHub by visiting...`). Regardless of the decided UX, it might be better to choose one and not overwhelm the user with options/choices here.

daavoo:
With the existing endpoint for live metrics, an existing experiment could be shared with 3 REST API calls:

- `start`:

```
json={
    "type": "start",
    "repo_url": "STUDIO_REPO_URL",
    "baseline_sha": "BASELINE_SHA",
    "name": "EXP_NAME",
    "client": "vscode",  # I think `client` is just ignored by studio
},
headers={
    "Authorization": "token STUDIO_TOKEN",
    "Content-type": "application/json",
}
```

- `data`

Include here metrics, params, and plots (only linear plots are accepted by the API). 

The API was designed for sending incremental updates of the plots on each step, but it would still work if the full data is sent and `step` is set to the latest:

```
        json={
            "type": "data",
            "repo_url": "STUDIO_REPO_URL",
            "baseline_sha": "BASELINE_SHA",
            "name": "EXP_NAME",
            "step": 2,  
            "metrics": {"metrics.json": {"data": {"step": 2, "foo": 3}}},
            "params": {"params.yaml": {"fooparam": 1}},
            "plots": {"plots/foo.tsv": {"data": [{"step": 0, "foo": 1.0}, {"step": 1, "foo": 2.0}, {"step": 2, "foo": 3.0}]}},
            "client": "vscode",
        },
        headers={
            "Authorization": "token STUDIO_TOKEN",
            "Content-type": "application/json",
        },
[...]

PR #3235, opened 2023-02-07
Enable saving of Studio access token

mattseddon:
# 1/3 `main` <- this <- #3289 <- #3294 
# 1/2 `main` <- this <- #3277 

This PR adds a new connect button to our sidebar panel and a screen that walks the user through generating/saving a Studio access token in the extension. It also adds add/remove commands for the access token which can be used to bypass the webview. Once the user has a token they will be able to share experiments with Studio from the experiments table.

### Demo

https://user-images.githubusercontent.com/37993418/218623493-05e9b6ba-a80b-4962-8c10-238d08a9ba52.mov

See comments inline.

All prototype work is shown below the line.

------

(Very rough) Prototype for connecting the extension to Studio. This is something that we can implement quickly before investing the time to roll out the OAuth flow.

In order to progress + get down some ideas I have started prototyping for #3077. As stated [here](https://github.com/iterative/vscode-dvc/issues/3077#issuecomment-1419801785). My current plan is to get the user to save a token into their local VS Code config under the `dvc.studioToken` key. We will then use that token to share experiments to Studio from various places in the UI.

## Visibility:

In order to let users know that they can/should connect to Studio we have to let them know that this is possible. I think we should add a welcome view to the sidebar whenever the `dvc.studioToken` is `null`:

image

## User assistance:

This webview can either get expanded or dropped in the future depending on the outcome of the discussions in #3077. For now, it should provide a way for the user to jump through the required hoops. Button actions are described under the screenshot

image

- `Sign In`: Opens `studio.iterative.ai` in the browser.
- `Get token`: Prompts user for their `username` and redirects them to `https://studio.iterative.ai/user/{username}/profile`
- `Save`: Opens quick input where the user can paste the token.

Alternative version (that I have dismissed):

Screenshot 2023-02-07 at 12 09 39 pm


mattseddon:
I think we'll want to have the webview regardless so that we have something with plenty of space/scope to direct users when they click "Share to Studio" and they aren't authed/have a token.

mattseddon:
Posted a demo back in #3077

extension/package.json
```diff
@@ -590,6 +595,15 @@
           "description": "%config.pythonPath.description%",
           "type": "string",
           "default": null
+        },
+        "dvc.studioAccessToken": {
```
mattseddon, line 599: [I] Need to change to use the `SecretStorage` API and not save the token in the open...

extension/src/connect/index.ts
```diff
@@ -0,0 +1,113 @@
...
+  private async openStudioProfile() {
+    const username = await getInput(Title.ENTER_STUDIO_USERNAME)
+    if (!username) {
+      return
+    }
+    return openUrl(`${STUDIO_URL}/user/${username}/profile`)
+  }
+
+  private async saveStudioToken() {
+    const token = await getValidInput(
```
mattseddon, line 92: [I] Need telemetry events

extension/src/webview/repository.ts
```diff
@@ -48,7 +48,7 @@ export abstract class BaseRepository<
       this.dvcRoot,
       this.webviewIcon,
       viewColumn,
-      this.viewKey === ViewKey.SETUP
+      [ViewKey.CONNECT, ViewKey.SETUP].includes(this.viewKey)
```
mattseddon, line 51: [I] Try and remove the bypass dvcRoot logic (maybe split the class)

extension/package.json
```diff
@@ -1319,6 +1337,11 @@
           "name": "Actions",
           "when": "true"
         },
+        {
+          "id": "dvc.views.studio",
+          "name": "Studio",
+          "when": "!dvc.studio.connected"
```
mattseddon, line 1343: [F] We only show the tree view when there is no Studio token available in the secrets store. It can still be hidden by the user. We also give them the option to remove the token and save another one.

extension/src/connect/token.ts
```diff
@@ -0,0 +1,8 @@
+export const STUDIO_ACCESS_TOKEN_KEY = 'dvc.studioAccessToken'
+
+export const isStudioAccessToken = (text?: string): boolean => {
+  if (!text) {
+    return false
+  }
+  return text.startsWith('isat_') && text.length === 54
```
mattseddon, line 7: [F] This was the basic format that I gathered from each token. With the rebrand, this may change but lets see.

extension/src/webview/repository.ts
mattseddon: needs a follow up, I've taken a note.

webview/src/shared/hooks/useVsCodeMessaging.ts
```diff
@@ -10,13 +10,13 @@ const signalInitialized = () =>
...
+  handler?: (event: { data: MessageToWebview<T> }) => void
 ) {
   useEffect(() => {
     signalInitialized()
   }, [])
   useEffect(() => {
-    window.addEventListener('message', handler)
-    return () => window.removeEventListener('message', handler)
+    handler && window.addEventListener('message', handler)
+    return () => handler && window.removeEventListener('message', handler)
```
mattseddon, line 20: [Q] Any objections to this? I don't need to send this webview any data right now, it is either on or off. This will change if/when we expand the webview to connect to other products.

extension/src/connect/index.ts
```diff
@@ -0,0 +1,118 @@
...
+import { getInput, getValidInput } from '../vscode/inputBox'
+import { Title } from '../vscode/title'
+import { openUrl } from '../vscode/external'
+import { ContextKey, setContextValue } from '../vscode/context'
+import { RegisteredCommands } from '../commands/external'
+
+export class Connect extends BaseRepository<undefined> {
+  public readonly viewKey = ViewKey.CONNECT
+
+  private readonly secrets: SecretStorage
```
mattseddon, line 19: [F] This seems like a better API to use instead of saving the token directly into the User's config.

mattseddon:
Need to add at least "share to Studio" as a command before we can ship this. Added `do not merge` to remind me.

webview/src/connect/components/Studio.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const Studio: React.FC = () => {
+  const openStudio = () =>
+    sendMessage({ type: MessageFromWebviewType.OPEN_STUDIO })
+
+  const openStudioProfile = () =>
+    sendMessage({ type: MessageFromWebviewType.OPEN_STUDIO_PROFILE })
+
+  const saveStudioToken = () =>
+    sendMessage({ type: MessageFromWebviewType.SAVE_STUDIO_TOKEN })
```
sroy3, line 16: Since none of these functions rely on props on state, I'd move them outside of the component as these get recreated on each re-render. This should fix the 2 codeclimate issues as well.

sroy3 (approved):

webview/src/connect/components/Studio.tsx
mattseddon: 🙏🏻

extension/package.json
```diff
@@ -1355,6 +1387,11 @@
         "view": "dvc.views.actions",
         "contents": "[$(beaker) Show Experiments](command:dvc.showExperiments)\n[$(graph-scatter) Show Plots](command:dvc.showPlots)\n[$(play) Run Experiment](command:dvc.runExperiment)"
       },
+      {
+        "view": "dvc.views.studio",
+        "contents": "[$(plug) Connect](command:dvc.showConnect)",
+        "when": "!dvc.studio.connected"
+      },
```
mattseddon, line 1394: [I] Double check on initial view placement.

webview/src/connect/components/Studio.tsx
```diff
@@ -0,0 +1,60 @@
...
+        />
+        <Button
+          appearance="secondary"
+          isNested={true}
+          text={'Save'}
+          onClick={saveStudioToken}
+        />
+        <p>
+          {"Don't Have an account?\n"}
+          <a href={STUDIO_URL}>Sign-Up</a>
```
julieg18, line 55: Minor! Should that hyphen be there? 
```suggestion
          <a href={STUDIO_URL}>Sign Up</a>
```

extension/src/connect/token.ts
julieg18: My token was actually `53` characters long 😅 

https://user-images.githubusercontent.com/43496356/219052994-8eb69fc6-fd64-48b0-af67-25b3292b1b4b.mov

extension/src/connect/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    const token = await getValidInput(
+      Title.ENTER_STUDIO_TOKEN,
+      validateTokenInput,
+      { password: true }
+    )
+    if (!token) {
+      return
+    }
+
+    return this.storeSecret(STUDIO_ACCESS_TOKEN_KEY, token)
```
julieg18, line 60: Should we add some kind of message after the token is saved? Feels a little weird to me that I saved my token then everything related to Studio just disappears from my screen 🤔

julieg18 (approved):
Great work!

extension/src/connect/token.ts
mattseddon: I'll relax the requirement

webview/src/connect/components/Studio.tsx
mattseddon: Don't think so

extension/src/connect/index.ts
mattseddon: Yes I will add a toast/modal

mattseddon: This is what the flow looks like now: 

https://user-images.githubusercontent.com/37993418/219277549-fc5af759-06e5-4682-ada1-80eda3c77207.mov

mattseddon: Let me know what you think. We can update in a follow up.

julieg18: Looks good! Makes things more clear on the connection and what you can do now that you're connected!

Issue #3057, opened 2023-01-05
Onboarding feedback


dberenbaum:
Hey, I took another shot at onboarding and it's getting smoother each time. I found a couple minor issues:

1. If starting from a fresh Git repo, experiments will break if there's not a commit already. Is there some way we can auto commit when initializing dvc or suggest to make a commit if there is none in the repo? IMO this should be part of the "Show Setup" workflow, and we should block any other workflows until there is a commit in the repo.
2. The first thing I see in the sidebar after initializing my repo but before running any experiments is "Columns" showing the column "Created." Can we move it down in the sidebar or hide it?

--------

Actions:

- [x] Check if this is about `dvc exp run` and `dvc exp show` in the first place. If they return a reasonable error we can and should show them.
- [ ] If there are not commits we can introduce a screen with a message and a button to focus users on the SCM panel.
- [x] Disable columns section (and other sections) if we can't show the table at this moment yet to make it consistent.

shcheklein:
1. Check if this is about `dvc exp run` and `dvc exp show` in the first place. If they return a reasonable error we can and should show them.
2. If there are not commits we can introduce a screen with a message and a button to focus users on the SCM panel.
3. Disable columns section (and other sections) if we can't show the table at this moment yet to make it consistent.

dberenbaum:
> * Check if this is about `dvc exp run` and `dvc exp show` in the first place. If they return a reasonable error we can and should show them.

I'll add issues to improve the errors in the CLI, but it doesn't really impact anything here since this is about the experience before even getting to `dvc exp run` or `dvc exp show`. 
* The table won't show at this point regardless of whether there is a commit, so that's handled well already. 
* You can try to run experiments, but it's a separate issue where we should disable this action until a pipeline exists (related to https://github.com/iterative/vscode-dvc/issues/3058). 

Regardless, encouraging a Git commit after `dvc init` or when starting the extension from an empty Git repo would help (similar to the `dvc init` message "You can now commit the changes to git").

shcheklein:
> I'll add issues to improve the errors in the CLI, but it doesn't really impact anything here since this is about the experience before even getting to dvc exp run or dvc exp show.

@dberenbaum do we have an issue for this? Also, do we have something about commits for Studio (unless you do a commit you can't see metrics). Trying to consolidate my thoughts about this and also answer [this question](https://stackoverflow.com/questions/75306950/dvc-error-scmrepo-exceptions-scmerror-empty-git-repo)

[...]

PR #3239, opened 2023-02-07
Add reminder to commit after `dvc init`

julieg18:
* adds a `Toast` message after the "Initialize a DVC Project" command is run

## Demo

https://user-images.githubusercontent.com/43496356/217296643-8a8f22bd-c328-4c2c-a7e2-6b942e5a209c.mov

Part of #3057

### DIFF
--- a/extension/src/setup/index.ts
+++ b/extension/src/setup/index.ts
@@ -43,6 +43,7 @@ import { EventName } from '../telemetry/constants'
 import { WorkspaceScale } from '../telemetry/collect'
 import { gitPath } from '../cli/git/constants'
 import { DOT_DVC } from '../cli/dvc/constants'
+import { Toast } from '../vscode/toast'
 
 export type SetupWebviewWebview = BaseWebview<TSetupData>
 
@@ -310,6 +311,9 @@ export class Setup
     const root = getFirstWorkspaceFolder()
     if (root) {
       await this.dvcExecutor.init(root)
+      void Toast.infoWithOptions(
+        'Initialized DVC repository. You can now [commit the changes to git](command:workbench.view.scm).'
+      )
     }
   }
 


extension/src/setup/index.ts
```diff
@@ -310,6 +311,9 @@ export class Setup
     const root = getFirstWorkspaceFolder()
     if (root) {
       await this.dvcExecutor.init(root)
+      void Toast.infoWithOptions(
```
julieg18, line 314: Another option that came to mind was adding an action in the notification that runs `git commit` for the user.

extension/src/setup/index.ts
```diff
@@ -43,6 +43,7 @@ import { EventName } from '../telemetry/constants'
 import { WorkspaceScale } from '../telemetry/collect'
 import { gitPath } from '../cli/git/constants'
 import { DOT_DVC } from '../cli/dvc/constants'
+import { Toast } from '../vscode/toast'
```
julieg18, line 46: There have been mentions about the vscode notifications being easily ignored. Is there a better way to notify the user?

sroy3: Maybe simply add another step in the setup that stays there until the user has commited.

extension/src/setup/index.ts
sroy3: Would it be bad to simply commit for the user while running `dvc init`? Like simply adding the necessary files and commit with a message saying `dvc init`. What does everyone think @iterative/vs-code ?

mattseddon: I think we are trying to mitigate `exp show` & `exp run` from breaking if there are **no commits** in the repo. 

The DVC team is actively working on the other side of this (https://github.com/iterative/dvc/pull/8840)
I also don't think a Toast will be enough by itself. Generally, I'd agree that we could just commit automatically for the user. However, there must be some rationale behind `dvc init` not auto-committing everything (perhaps if there is a dirty workspace or the user might want to commit different files together). An extra step in Setup is probably warranted but only if there are no commits in the repo.

julieg18: Agreed! I'll go ahead and merge this for a first iteration, adding a setup step in the followup, since it will take me a bit of time to add a Setup step (Looks like I'll need to add a run a git command and check the error message to check for 0 git commits 🤔).

extension/src/setup/index.ts
```diff
@@ -310,6 +311,9 @@ export class Setup
     const root = getFirstWorkspaceFolder()
     if (root) {
       await this.dvcExecutor.init(root)
+      void Toast.infoWithOptions(
+        'Initialized DVC repository. You can now [commit the changes to git](command:workbench.view.scm).'
```
mattseddon, line 315: [F] It is also a more expected pattern to make this a button by pasting `items` to `Toast.infoWithOptions`. See https://github.com/iterative/vscode-dvc/blob/09e32508d93c62914deac0084a2578a7710612e5/extension/src/cli/dvc/discovery.ts#L47-L48 for an example

julieg18: > [F] It is also a more expected pattern to make this a button by pasting items to Toast.infoWithOptions. 

Makes sense! Would say I'll update it in the followup, but now that I think about it, we will probably want to take off the Toast once we have the Setup screen 🤔

PR #3240, opened 2023-02-07
Add jupyter notebook script to run in dvc.yaml

sroy3:
Part of #3058 

## Demo of command

https://user-images.githubusercontent.com/3683420/217304840-7cb97b74-bd7b-48c1-a114-02b4cca5bcb6.mov

## Demo of `jupyter nbconvert` command running completely

https://user-images.githubusercontent.com/3683420/217340654-e7e85f66-a975-4067-8917-cb992639ef44.mov

## When `jupyter` is not explicitly installed

https://user-images.githubusercontent.com/3683420/217340819-98f94eb6-407f-475a-ab99-82d3ca5edb2b.mov

Should we provide a way for the user to install `jupyter`? Or just let them figure it out?


shcheklein:
Looks great! Q: what happens if `dvc.yaml` already has a stage / stages?

sroy3:
> Looks great! Q: what happens if `dvc.yaml` already has a stage / stages?

It skips the creation of the train stage and simply run it the pipeline normally (we also do not prompt the user for their script).

extension/src/fileSystem/index.test.ts
```diff
@@ -1,17 +1,28 @@
...
 import { dvcDemoPath } from '../test/util'
 import { DOT_DVC } from '../cli/dvc/constants'
 
 jest.mock('../cli/dvc/reader')
+jest.mock('fs-extra', () => {
+  const actualModule = jest.requireActual('fs-extra')
+  return {
+    __esModule: true,
+    ...actualModule,
+    appendFileSync: jest.fn(),
```
mattseddon, line 22: [N] Could set this up with 
```
mockedAppendFileSync = jest.mocked(appendFileSync)
...
appendFileSync: mockedAppendFileSync
```

so that is is obvious that it is a mock if you only read the test.

mattseddon (approved):

extension/src/fileSystem/index.ts
```diff
@@ -131,16 +131,25 @@ export const isAnyDvcYaml = (path?: string): boolean =>
       basename(path) === 'dvc.yaml')
   )
 
+export const scriptCommand = {
```
julieg18, line 134: Should we make it more clear that the user needs to submit a path that either a python or jupyter notebook? Not sure how commit it is to run a script that's outside of those two types of files 🤔

julieg18 (approved):
Great work!

extension/src/fileSystem/index.ts
sroy3: It's part of the following steps.

PR #3241, opened 2023-02-07
Don't add "Created" column when there is no data

julieg18:
## Demo

https://user-images.githubusercontent.com/43496356/217324216-a8d01d67-98ac-46a0-b6a2-fbf8668319a8.mov

Part of #3057

### DIFF
--- a/extension/src/experiments/columns/collect/index.ts
+++ b/extension/src/experiments/columns/collect/index.ts
@@ -1,4 +1,5 @@
 import { join } from 'path'
+import isEqual from 'lodash.isequal'
 import { ColumnAccumulator } from './util'
 import { collectDepChanges, collectDeps } from './deps'
 import {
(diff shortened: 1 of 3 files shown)


sroy3 (approved):

extension/src/experiments/columns/collect/index.ts
```diff
@@ -47,7 +47,9 @@ export const collectColumns = (data: ExperimentsOutput): Column[] => {
   for (const commit of Object.values(rest)) {
     collectFromCommit(acc, commit)
   }
-  return Object.values(acc)
+  const columns = Object.values(acc)
+
+  return columns.length === 1 ? [] : columns
```
mattseddon, line 52: [Q] Would it be a good idea to explicitly check that `columns` is equal to only the timestamp column (`isEqual(columns, [timestampColumn])` ) and add an explanatory variable as to why we're removing it?

mattseddon (approved):

extension/src/experiments/columns/collect/index.ts
julieg18: Great idea!

PR #3242, opened 2023-02-07
Open dvc.yaml automatically after adding a pipeline

sroy3:
Part of #3058 

https://user-images.githubusercontent.com/3683420/217665512-03a7a5ca-129e-435c-a775-8c2e109d1720.mov


sroy3:
Still need to figure out a way to test this. I keep hitting walls with whatever strategy I try. `jest.requireActual('vscode')` gives the following error `Cannot find module 'vscode' from 'src/fileSystem/index.test.ts'`

extension/src/fileSystem/index.ts
```diff
@@ -136,6 +136,11 @@ export const scriptCommand = {
   PYTHON: 'python'
 }
 
+export const openDvcYamlFile = async (dvcYamlPath: string) => {
```
mattseddon, line 139: [C] This name can be generalised (I think we open a file the same way in at least one test). Can replace the other usage

extension/src/fileSystem/index.test.ts
```diff
@@ -231,4 +251,18 @@ describe('findOrCreateDvcYamlFile', () => {
       expect.stringContaining(scriptCommand.PYTHON)
     )
   })
+
+  it('should open the dvc.yaml file in the editor', () => {
+    mockedOpenTextDocument.mockResolvedValue({} as TextDocument)
+
+    void findOrCreateDvcYamlFile('/', '/train.py')
```
mattseddon, line 258: [Q] Don't want to `await` the result of these?

mattseddon (approved):
Looks good

extension/src/fileSystem/index.test.ts
sroy3: This isn't really crucial to starting running the experiments. I don't mind changing it though

julieg18 (approved):

PR #3247, opened 2023-02-08
Add ability to stop experiments running in the workspace (outside of the extension)

mattseddon:
This PR uses the value provided in `.dvc/exps/rwlock.lock` to stop experiments which are running in the workspace. This means we can now stop experiments that are started outside of the extension context.

### Demo

https://user-images.githubusercontent.com/37993418/217414543-40b7d27d-a1b6-489e-925c-767e1ccc98dd.mov

I'll follow up to add an option to stop experiments running in the workspace from the table's context menu.

Related to https://github.com/iterative/dvc/issues/8963.

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -820,7 +820,7 @@
         },
         {
           "command": "dvc.stopAllRunningExperiments",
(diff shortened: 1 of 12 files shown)


julieg18 (approved):
Great work!

PR #3253, opened 2023-02-08
Show a setup screen when project has no commits

### LINKED ISSUE #3057: Onboarding feedback

julieg18:
* adds an extra step to our setup that shows if a user has initialized git and dvc but still has no commits
* takes off notification that appears after dvc initalizes

## Demo

https://user-images.githubusercontent.com/43496356/217660720-0d7b5ff3-7034-4124-a943-4703fc8b48ed.mov

Followup to #3239, Fixes #3057


webview/src/setup/components/NeedsGitCommit.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Button } from '../../shared/components/button/Button'
+
+type NeedsGitCommitProps = { showScmPanel: () => void }
+
+export const NeedsGitCommit: React.FC<NeedsGitCommitProps> = ({
+  showScmPanel
+}) => (
+  <EmptyState>
+    <div>
+      <h1>Your project has no commits in Git</h1>
```
julieg18, line 12: Could any of these statements be improved?

mattseddon, line 12: [C] No Git commits detected

webview/src/setup/components/NeedsGitCommit.tsx
```diff
@@ -0,0 +1,20 @@
...
+      <p>
+        DVC requires your project to have atleast one commit to run experiments
+        correctly.{' '}
```
mattseddon, line 15: [C] At least one commit is required to enable experiments.

mattseddon (approved):

webview/src/setup/components/NeedsGitCommit.tsx
sroy3, line 15: That last space is probably uneccessary.

sroy3 (approved):

webview/src/setup/components/NeedsGitCommit.tsx
julieg18: Nice catch!

webview/src/setup/components/NeedsGitCommit.tsx
julieg18: Thanks!

PR #3255, opened 2023-02-09
Ask the user for the correct command to their script if it is neither a python file or a Jupyter notebook

sroy3:
Part of #3058 

https://user-images.githubusercontent.com/3683420/217886002-271eb284-2bb5-498f-b9ae-780645a34b83.mov


extension/src/experiments/workspace.ts
```diff
@@ -473,9 +473,20 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
...
 
-    return pathOrSelect === selectValue
-      ? await pickFile(Title.SELECT_TRAINING_SCRIPT)
-      : pathOrSelect
+    const trainingScript =
+      pathOrSelect === selectValue
+        ? await pickFile(Title.SELECT_TRAINING_SCRIPT)
+        : pathOrSelect
+
+    if (trainingScript) {
```
mattseddon, line 481: [N] Could also be an early return

extension/src/fileSystem/index.test.ts
```diff
@@ -177,15 +177,25 @@ describe('getModifiedTime', () => {
...
+    )
 
     expect(mockedEnsureFileSync).toHaveBeenCalledWith(`${cwd}/dvc.yaml`)
   })
 
   it('should add the stage name to the dvc.yaml file', () => {
     const cwd = '/cwd'
     const uniqueStageName = 'aWesome_STAGE_name48'
-    findOrCreateDvcYamlFile(cwd, '/script.py', uniqueStageName)
+    findOrCreateDvcYamlFile(
```
mattseddon, line 193: [Q] Why doesn't this one need `void` but the others do?

extension/src/fileSystem/index.ts
```diff
@@ -164,6 +157,23 @@ stages:
   return appendFileSync(dvcYamlPath, pipeline)
 }
 
+export const scriptCommand = {
+  JUPYTER: 'jupyter nbconvert --to notebook --inplace --execute',
+  PYTHON: 'python'
```
mattseddon, line 162: [Q] Could this be an enum?

[Q] Is there a better file to move `scriptCommand` & `getScriptCommand` to? Doesn't feel appropriate for them to be under `fileSystem`.

mattseddon (approved):

extension/src/fileSystem/index.test.ts
sroy3: Missed it during the merge

extension/src/experiments/workspace.test.ts
```diff
@@ -32,9 +32,17 @@ jest.mock('vscode')
...
-jest.mock('../fileSystem', () => ({
-  findOrCreateDvcYamlFile: jest.fn()
-}))
+
+jest.mock('../fileSystem', () => {
+  const actualModule = jest.requireActual('../fileSystem')
+  return {
+    __esModule: true,
+    ...actualModule,
+    findOrCreateDvcYamlFile: jest.fn()
```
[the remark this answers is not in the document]
sroy3: That's only because I need another function to run correctly. I might just move it in another file actually.

shcheklein:
Thanks @sroy3 ! One thing to double check - let's make sure that the second step is optional (I think DVC can run a shell script let's say w/o specifying sh). And update the copy if needed.

sroy3:
> Thanks @sroy3 ! One thing to double check - let's make sure that the second step is optional (I think DVC can run a shell script let's say w/o specifying sh). And update the copy if needed.

It is you don't have to enter anything for that and it won't block adding the script.

PR #3256, opened 2023-02-09
Show the setup instead of original webview if needed (command palette)

sroy3:
## Demo when there is no data (or simply when setup should be shown)

https://user-images.githubusercontent.com/3683420/217916697-9a535659-ef6b-447f-9e85-3f8aa795935b.mov

## Demo with the project is all set up

https://user-images.githubusercontent.com/3683420/217916763-99f2e427-2bd6-498a-b672-a4da6280c823.mov

I could have gotten away with overriding the command in the `package.json` and add a `when` clause, but since there is more than one command to override with `showSetup` (`showExperiments`, `showPlots`, and `showExperimentsAndPlots`) it was only showing the last one (`showExperimentsAndPlots`) in the command palette. This way, the show webview(s) commands will always be overridden by `showSetup` if need be, wherever the command came from.


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1022,6 +1023,8 @@ suite('Experiments Test Suite', () => {
     }).timeout(WEBVIEW_TEST_TIMEOUT)
 
     it('should be able to handle a message to compare experiments plots', async () => {
+      const mockShouldBeShown = stub(Setup.prototype, 'shouldBeShown')
+      mockShouldBeShown.callsFake(() => false)
```
mattseddon, line 1027: [I] can just be `.returns(false)`

extension/src/setup/index.ts
```diff
@@ -220,21 +220,20 @@ export class Setup
     return this.sendDataToWebview()
   }
 
-  public async sendDataToWebview() {
-    const projectInitialized = this.hasRoots()
-    const hasData = this.getHasData()
+  public shouldBeShown() {
+    return !this.cliCompatible || !this.hasRoots() || !this.getHasData()
```
mattseddon, line 224: [Q] Does this mean some complexity can be removed from the extension's `package.json`? 
i.e around this code:

```
    "viewsWelcome": [
      {
        "view": "dvc.views.webviews",
        "contents": "[Show Experiments](command:dvc.showExperiments)\n[Show Plots](command:dvc.showPlots)\n[Show Experiments and Plots](command:dvc.showExperimentsAndPlots)",
        "when": "dvc.commands.available && dvc.project.available && dvc.project.hasData"
      },
      {
        "view": "dvc.views.webviews",
        "contents": "[Show Experiments](command:dvc.showSetup)\n[Show Plots](command:dvc.showSetup)\n[Show Experiments and Plots](command:dvc.showSetup)",
        "when": "!dvc.commands.available || !dvc.project.available || !dvc.project.hasData"
      },
```

mattseddon (approved):
Needs a test

extension/src/setup/index.ts
sroy3: You're right! I'll remove that.

extension/src/experiments/commands/register.ts
```diff
@@ -318,19 +320,22 @@ const registerExperimentRunCommands = (
   internalCommands.registerExternalCommand(
     RegisteredCommands.EXPERIMENT_SHOW,
     (context: Context) =>
-      experiments.showWebview(getDvcRootFromContext(context))
+      setup.shouldBeShown()
+        ? setup.showWebview()
```
mattseddon, line 324: Had a thought, maybe we should use `commands.executeCommand` here so that we get a telemetry event for the redirect.

julieg18 (approved):
Great work!

extension/src/test/suite/experiments/commands/register.test.ts
```diff
@@ -0,0 +1,86 @@
+import { expect } from 'chai'
```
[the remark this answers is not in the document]
mattseddon: I think I can help. Will put something together after the weekend.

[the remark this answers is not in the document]
sroy3: Finally got something that works. I'll reopen the PR soon, it'll need another review though as I might have made a mess.

extension/src/extension.ts
```diff
@@ -185,11 +185,19 @@ export class Extension extends Disposable {
...
+      this.setup
+    )
+    registerPlotsCommands(this.plots, this.internalCommands, this.setup)
     this.internalCommands.registerExternalCommand(
       RegisteredCommands.EXPERIMENT_AND_PLOTS_SHOW,
       async (context: VsCodeContext) => {
+        if (this.setup.shouldBeShown()) {
+          await commands.executeCommand(RegisteredCommands.SETUP_SHOW)
+          return
+        }
```
sroy3, line 200: Didn't add a test for that one as it's super straight forward to do and it'll most likely be removed when we remove the button in the sidebar (should we actually keep the command in the palette?)

extension/src/test/suite/experiments/commands/register.test.ts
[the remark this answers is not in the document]
mattseddon: Generally, I have been testing commands through their public interface in `extension/src/test/suite/experiments/workspace.test.ts`. A lot of the setup work is done in that file already so perhaps you can move the tests there.

Also if you made a wrapper function for the different versions of showing a webview. You would only have to test one variation.

The wrapper could look something like this:

```
<T>(setup: Setup, callback: (context: Context) => Promise<T | undefined>) =>
  (context: Context) =>
    setup.shouldBeShown()
      ? commands.executeCommand(RegisteredCommands.SETUP_SHOW)
      : callback(context)
```

WDYT?

PR #3191, opened 2023-01-31
Fix table/plots not updating correctly on windows

julieg18:
## Demo

https://user-images.githubusercontent.com/43496356/215863127-cddd364f-5b42-4972-ab51-bc2677eb2f9b.mp4

Fixes #3103, fixes #3102

PR #3259, opened 2023-02-09
Add regression test for `.git` file watcher

julieg18:
Followup to #3191


extension/src/experiments/data/index.ts
```diff
@@ -74,7 +79,7 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
       this.dvcRoot
     )
 
-    const dotGitPath = getGitPath(gitRoot, gitPath.DOT_GIT)
+    const dotGitPath = this.dotGitPath || getGitPath(gitRoot, gitPath.DOT_GIT)
```
mattseddon, line 82: [Q] Are you sure you need this when `AvailableCommands.GIT_GET_REPOSITORY_ROOT` is already returning the mockRoot? Could you stub `getGitPath` instead of adding code just for testing purposes?

extension/src/test/suite/experiments/data/index.test.ts
```diff
@@ -102,6 +105,58 @@ suite('Experiments Data Test Suite', () => {
...
+      await dataUpdatedEvent
+
+      expect(managedUpdateSpy).to.be.called
+
+      await Watcher.fireWatcher(mockDotGitNestedFilePath)
+      await dataUpdatedEvent
+
+      expect(managedUpdateSpy).to.be.called
+
+      removeSync(mockDotGitPath)
```
mattseddon, line 157: [F] If the test fails the folder won't be removed.

extension/src/test/suite/experiments/data/index.test.ts
```diff
@@ -102,6 +105,58 @@ suite('Experiments Data Test Suite', () => {
       expect(managedUpdateSpy).to.be.called
     })
 
+    it('should watch the .git directory for updates when directory is inside workspace', async () => {
```
mattseddon, line 108: [N] `...when the directory is...`

mattseddon (approved):

extension/src/experiments/data/index.ts
julieg18: Good idea!

extension/src/test/suite/experiments/data/index.test.ts
```diff
@@ -67,22 +76,69 @@ suite('Experiments Data Test Suite', () => {
...
+
+      stub(FileSystem, 'getGitPath').returns(MOCK_WORKSPACE_GIT_FOLDER)
 
       const data = disposable.track(
         new ExperimentsData(
           dvcDemoPath,
           {
             dispose: stub(),
-            executeCommand: mockExecuteCommand
+            executeCommand: (command: CommandId) => {
```
julieg18, line 137: Ended up moving the `executeCommand` to be inline to avoid the `no-identical-function` eslint rule. I could be wrong, but when I tried to create a single function for the two to share it looked like I was overcomplicating things 🤔

PR #3261, opened 2023-02-10
Add missing telemetry events to onboarding

mattseddon:
This came up in a discussion with @shcheklein this morning. I checked to see that we had events for all of the actions that a user can take in the onboarding process. There were a couple that was missing so I added them.


extension/src/setup/webview/messages.ts
```diff
@@ -67,9 +67,9 @@ export class WebviewMessages {
       case MessageFromWebviewType.INITIALIZE_DVC:
         return this.initializeDvc()
```
mattseddon, line 68: The event here is covered by `dvc.init`.

sroy3 (approved):

julieg18 (approved):

PR #3271, opened 2023-02-13
Fix checkboxes unchecking when user clicks outside the table

### LINKED ISSUE #2528: Clicking outside the table removing checks from the checkboxes
https://user-images.githubusercontent.com/98249521/193937449-1999a02e-6916-4534-9cfa-e752b5eddbf2.mov

julieg18:
## Demo
You can't really tell in the demo, but I am clicking repeatedly outside the table 😅 

https://user-images.githubusercontent.com/43496356/218481794-4064ee10-66d6-4159-9649-eca7d81a980e.mov

Fixes #2528


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -258,6 +258,8 @@ $bullet-size: calc(var(--design-unit) * 4px);
 
   .tableContainer {
     flex: 1;
+    display: flex;
```
julieg18, line 261: The checkboxes are unchecking anytime a user clicks outside of the `table` element. I don't think this should be happening but I couldn't figure out why it's happening in the small amount of time I chose to look at the issue 🤔

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -258,11 +258,14 @@ $bullet-size: calc(var(--design-unit) * 4px);
...
+    align-items: stretch;
   }
 
   table {
     display: inline-block;
     border-collapse: collapse;
+    flex-grow: 1;
```
julieg18, line 268: I fixed the issue by making the `table` element the width/height of the webview screen. Everything appears to be working correctly, but are there any issues that anyone else spots with this solution?

webview/src/experiments/components/table/styles.module.scss
sroy3: I think the issue is found in the following code
```

  const clickOutsideHandler = useCallback(() => {
    clearSelectedRows?.()
  }, [clearSelectedRows])

  useClickOutside(tableRef, clickOutsideHandler)
```
On line 45 of file `webview/src/experiments/components/Table.tsx`

sroy3: Could probably remove the whole `useClickOutside`.

julieg18:
> I think removing the behaviour is the best option for now, but I do think some sort of unselecting behaviour should be needed.

Agreed! Maybe we could add an extra context menu item for selected rows, like "Unselect All Rows" 🤔

webview/src/shared/hooks/useClickOutside.ts
```diff
@@ -1,20 +0,0 @@
-import React, { useEffect } from 'react'
-
-export const useClickOutside = (
```
julieg18, line 3: Since this hook is no longer being used anywhere, I went ahead and deleted it.

sroy3 (approved):

PR #3272, opened 2023-02-13
Replace Show Experiments and Plots buttonwith a Run Experiment button and added icons to sidebar buttons

sroy3:
Part of #3058 

https://user-images.githubusercontent.com/3683420/218512593-90c523d3-8a85-4957-9933-34ff4fd18097.mov

Following @mattseddon suggestion, I also added icons to the buttons.

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -1353,7 +1353,7 @@
     "viewsWelcome": [
       {
         "view": "dvc.views.webviews",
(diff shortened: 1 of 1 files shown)


shcheklein:
> Should we remove the "Show Experiments and Plots" from the command palette as well?

I think it's fine to keep it for now.

PR #3281, opened 2023-02-14
Add add configuration button in experiments table

sroy3:
Part of #3058 

https://user-images.githubusercontent.com/3683420/218823122-6935bfe5-232e-4f39-a5de-9bc06f02a4c4.mov

Still need a followup to change the text of the button and add explanation text.


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -998,3 +998,7 @@ $badge-size: 0.85rem;
   font-size: 0.65rem;
   cursor: pointer;
 }
+
+.addConfigButton {
```
julieg18, line 1002: Would it look better if the button was directly below the table instead of stuck at the bottom of the webview? 

Screenshot 2023-02-16 at 7 54 06 AM

julieg18 (approved):
Great work!

webview/src/experiments/components/table/styles.module.scss
sroy3: You're right, changed.

mattseddon (commented):
Sorry I missed this.

PR #3289, opened 2023-02-15
Share experiment to Studio from experiments table

mattseddon:
# 2/3 `main` <- #3235 <- this <- #3294

Next step in https://github.com/iterative/vscode-dvc/issues/3077.

This PR enables users who have added a Studio access token to share experiments to Studio from both the experiments tree and table context menus.

### Demo

https://user-images.githubusercontent.com/37993418/219279540-b8494c95-aed5-4832-8103-e9fa17e2a37e.mov

Note: This is very much the happy path implementation. Still need to (at least) handle/let the user know if their token doesn't work.


extension/package.json
```diff
@@ -1591,6 +1591,7 @@
     "lodash.isequal": "4.5.0",
     "lodash.merge": "4.6.2",
     "lodash.omit": "4.5.0",
+    "node-fetch": "2.6.9",
```
mattseddon, line 1594: [F] v2 is the commonJS implementation

extension/src/test/suite/connect/index.test.ts
```diff
@@ -35,36 +24,10 @@ suite('Connect Test Suite', () => {
     return closeAllEditors()
   })
 
-  const buildConnect = async (mockSecretStorage?: SecretStorage) => {
```
mattseddon, line 38: [F] Moved this because we need the access token for to post to the Studio API.

extension/src/patch.ts
```diff
@@ -0,0 +1,187 @@
...
+  return shareWithProgress(experimentDetails, repoUrl, studioAccessToken)
+}
+
+export const registerPatchCommand = (internalCommands: InternalCommands) =>
+  internalCommands.registerCommand(
+    AvailableCommands.EXP_PUSH,
+    (...args: Args) => {
+      const [studioAccessToken, dvcRoot, name] = args
+
+      return pushExperiment(internalCommands, dvcRoot, name, studioAccessToken)
```
mattseddon, line 185: [F] The way that I've split up the functionality/built this feature means that this part can be replaced by a CLI call once it catches up 🤞🏻.

mattseddon: The above is the main reason that (for now at least) I have not made an effort to reuse existing data/fit in more with the rest of the codebase.

sroy3 (approved):

PR #3291, opened 2023-02-15
Move the comparison table header down from under the ribbon when scrolling

### LINKED ISSUE #3269: Plots: sticky sections collide
The plots webview's two sticky sections can collide.

This happens when there are multiple experiments selected and the project has images. The top ribbon overwrites most of the comparison table's sticky header:

### Demo

https://user-images.githubusercontent.com/37993418/218344736-a2906f00-5273-440c-bcd2-50dd4d34c06a.mov

sroy3:
Fixes #3269 

https://user-images.githubusercontent.com/3683420/219151748-00c34c8f-33c1-42b4-b567-9bee03814d23.mov


mattseddon (approved):
[Q] Add a scrolled snapshot that proves when there is enough data in the webview that both sticky sections are shown?

mattseddon:
Are the current storybook changes expected?

julieg18 (approved):

sroy3:
> Are the current storybook changes expected?

Yes because there is a default height of 50px for the ribbon and it's not in the stories affected. I added a state change to set it to 0 for these stories.

sroy3:
> [Q] Add a scrolled snapshot that proves when there is enough data in the webview that both sticky sections are shown?

No matter the delay I add for Chromatic, there are no change in the snapshots. But there are 2 stories for that now.

PR #3295, opened 2023-02-16
Change the add configuration button text and add more information on functionality

sroy3:
Part of #3058 

Screenshot 2023-02-16 at 1 41 40 PM

I know we talked about the fact that "pipeline stage" might be confusing to users as they might not fully understand what it means, but as we also talked about adding some text before and a "Learn More" link, I feel like "Add a pipeline stage" (exactly what the button does) does not look that confusing anymore. Let me know what you think. It's really not complicated to change any of the text.


webview/src/experiments/components/AddStage.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const AddStage: React.FC = () => (
+  <div className={styles.addConfigButton}>
+    <p>Want to easily and efficiently reproduce your experiments? </p>
+    <IconButton
+      icon={Add}
+      onClick={() =>
+        sendMessage({ type: MessageFromWebviewType.ADD_CONFIGURATION })
+      }
+      text="Add a pipeline stage"
```
mattseddon, line 16: [Q] Should this be title case? I think the rest of our buttons text use that.

webview/src/experiments/components/AddStage.tsx
```diff
@@ -0,0 +1,24 @@
+import { MessageFromWebviewType } from 'dvc/src/webview/contract'
+import React from 'react'
+import styles from './table/styles.module.scss'
+import { IconButton } from '../../shared/components/button/IconButton'
+import { Add } from '../../shared/components/icons'
+import { sendMessage } from '../../shared/vscode'
+
+export const AddStage: React.FC = () => (
+  <div className={styles.addConfigButton}>
+    <p>Want to easily and efficiently reproduce your experiments? </p>
```
mattseddon, line 10: [C] I don't think we use this style anywhere else. Maybe make it a statement instead of a question. I.e `Easily and efficiently reproduce your experiments.` but up to you.

mattseddon (approved):

webview/src/experiments/components/AddStage.tsx
sroy3: I was asking myself the same. I'll check and change.

julieg18 (approved):
Looks good! The lint appears to be failing on the pr 🤔

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -997,4 +997,10 @@ $badge-size: 0.85rem;
 
 .addConfigButton {
   margin: 20px auto;
+  text-align: center;
```
julieg18, line 1000: Should we center this info relative to the table instead of the webview?

Screenshot 2023-02-17 at 9 20 05 AM

How it would look: 

Screenshot 2023-02-17 at 9 25 32 AM

julieg18 (approved):
Great work! Seems like the lint check is failing though?

webview/src/experiments/components/table/styles.module.scss
sroy3: Made the table scrollable instead of the webview. It now looks like:


https://user-images.githubusercontent.com/3683420/219703490-fe93a4d2-4a9b-4618-8188-3fdfd9780662.mov

What do you think? I think we previously made the table 100% of the height to avoid this, but if there is other content in the page, this makes more sense.(cc @shcheklein)

julieg18: Hmmm, I'm still seeing the same result (I'm assuming because the table content is pretty small):

Screenshot 2023-02-17 at 10 25 14 AM

Though I now realize why it'd be difficult to make info relative to the table if the table took up more than the width of the webview 🤔 Probably be best (and easiest since this is minor and I'm guessing we don't want to spend too much time on it 😄) to just leave it centered relative to the webview or left align the info (not sure how that would look though).

sroy3: I think that I didn't take the time to understand your first concern, I saw the opposite of what you were showing (which was still a problem). I see it now. IMO it's better to center the "Add a Pipeline Stage" section with the webview, but this isn't a strong opinion. I feel like there might be something to do with the table. Center it? Continue the lines until the end of the webview? I don't know.

sroy3: Quick test for what it'd look like centered 

https://user-images.githubusercontent.com/3683420/219748144-059ab35d-816c-4ce4-9bb3-356351ee55a4.mov

sroy3: Quick test for full width rows (there are probably other  borders not extending correctly like the previous commits one, this is a prototype). This might be the cleanest solution.


https://user-images.githubusercontent.com/3683420/219755714-6102c28a-440a-4f28-9f55-7ca5e30cd0f0.mov

sroy3: I went with that last option after all. Don't mind changing it in a follow-up if needed.

https://user-images.githubusercontent.com/3683420/219757613-3fa10ac5-e305-4a76-af71-e61538100de3.mov

mattseddon: image

I feel like this looks like a bug. We might be optimising for very artificial situations. However, there are two things that we do need to keep in mind: 

1. Users should not be in this state for very long.
2. Not everyone will have the sidebar open and some have wide screens. E.g

Screenshot 2023-02-20 at 10 39 12 am (2)

I don't think we should optimise the overall experience for when the button is shown. IMO the original solution was fine.

sroy3: I don't know, I think completing the rows look nice and feel like all the space is used, even on wide screens. I'll revert and keep the original solution, we can change it back later.

julieg18: Agreeing with Matt! This solution causes multiple styling bugs when it comes to the header cells. Plus we have to make the table scrollable when its easier to navigate when the webview is scrollable. 

I vote keep original solution! Personally, I think left aligning the text looks fine if we want to consider that in the future as well. 

Screenshot 2023-02-20 at 8 20 03 AM

sroy3: It's reverted, but I'm not sure I understand the styling and scrolling bugs. I'm seeing only one that would have been really easy to fix (rows to go pass 100% of width). All I did was make the table take the full width of the webview (minimum). IMO it makes the table feel less broken if it's small. All of our webviews utilize the full width.

Aligning left for button and text like that isn't found anywhere else in the extension, it's always center-aligned. It feels like we forgot to add the styles or the styles are broken if we align it that way.

Issue #3264, opened 2023-02-10
Live plots do not automatically update on Codespaces


shcheklein:
Looks like something weird is happening with our watchers on Codedspaces. Local dev version runs fine for me. 

## How to reproduce

I've tried both - GPU and non-GPU setup. Two repositories: [Get Started](https://github.com/iterative/example-get-started-experiments) and [Hackathon](https://github.com/shcheklein/hackathon).

I'm getting only the first dot for the first epoch and then only at the very end it updates everything:

https://user-images.githubusercontent.com/3659196/218223477-c201c726-d819-4e23-bef5-13e7486cbdb5.mov

https://user-images.githubusercontent.com/3659196/218223521-e099b00e-6e8f-4b2a-a043-61f5f35c2e20.mov

## Environment

Extension version: 0.6.4
DVC Version:

```
DVC version: 2.44.0 (deb)
-------------------------
Platform: Python 3.10.8 on Linux-5.4.0-1101-azure-x86_64-with-glibc2.31
Subprojects:

Supports:
        azure (adlfs = 2023.1.0, knack = 0.10.1, azure-identity = 1.12.0),
        gdrive (pydrive2 = 1.15.0),
        gs (gcsfs = 2023.1.0),
        hdfs (fsspec = 2023.1.0, pyarrow = 11.0.0),
        http (aiohttp = 3.8.3, aiohttp-retry = 2.8.3),
        https (aiohttp = 3.8.3, aiohttp-retry = 2.8.3),
        oss (ossfs = 2021.8.0),
        s3 (s3fs = 2023.1.0, boto3 = 1.24.59),
        ssh (sshfs = 2023.1.0),
        webdav (webdav4 = 0.9.8),
        webdavs (webdav4 = 0.9.8),
        webhdfs (fsspec = 2023.1.0)
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/loop0
Caches: local
Remotes: https
Workspace directory: ext4 on /dev/loop0
Repo: dvc, git
```

`dvclive==2.0.2`

shcheklein:
Making it a `p1` priority since it affects us demoing this to customers. @alex000kim @daavoo do you also see this behavior?

mattseddon:
I can recreate the behaviour in `example-get-started` but not `magnetic-tiles-defects`.

Looks like there is at least one issue related to images:

[...]

PR #3299, opened 2023-02-17
Ensure that file exists before trying to read mtime

mattseddon:
Possible cause of #3264

Definitely the cause of this error:

image

What I think is happening here is that when there are a certain number of files written to disk by `plots diff` that the extension can try to access them before `plots diff` is finished/they can be accessed.


mattseddon:
@shcheklein if we can approve/merge this I'll ship and test against codespaces this afternoon.

shcheklein (approved):

PR #3303, opened 2023-02-17
Update get Studio access token link

mattseddon:
This PR updates the link used to direct a user to get an access token from Studio.

### Demo

https://user-images.githubusercontent.com/37993418/219618708-f1ba6d87-7107-42d6-baeb-cccc331844ac.mov

Note: Right now the page will not always be scrolled to the correct section but it will be some of the time. This is better than what we have so I think we should go with it.

For context see [this thread](https://iterativeai.slack.com/archives/C01APS0FHDM/p1676595777004649)

Thanks @mvshmakov 

Related to #3077

### DIFF
--- a/extension/src/connect/index.ts
+++ b/extension/src/connect/index.ts
@@ -90,7 +90,7 @@ export class Connect extends BaseRepository<undefined> {
     if (!username) {
       return
     }
-    return openUrl(`${STUDIO_URL}/user/${username}/profile`)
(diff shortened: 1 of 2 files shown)


mvshmakov (approved):
Nice, thanks for a quick fix!

julieg18 (approved):

sroy3 (approved):

PR #3304, opened 2023-02-17
Check stages for all run/queue commands

### LINKED ISSUE #3058: Onboarding to running experiments

sroy3:
Should close #3058 

https://user-images.githubusercontent.com/3683420/220189286-47940c80-00b5-4127-a0b5-4e62cdb989e3.mov


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -736,6 +737,7 @@ suite('Experiments Test Suite', () => {
     }).timeout(WEBVIEW_TEST_TIMEOUT)
 
     it("should be able to handle a message to modify an experiment's params and run a new experiment", async () => {
+      stub(DvcReader.prototype, 'listStages').resolves('train')
```
mattseddon, line 740: [I] much better to get the `dvcReader` back from `buildExperiments` and stub the instance instead of the prototype. Can even move the stub back into `buildExperiments` if it is going to be required everywhere.

mattseddon (approved):
👍🏻 

We should discuss whether or not to hide/re-route commands like `apply experiment`/`create branch` under these circumstances in the next planning meeting.

extension/src/test/suite/experiments/index.test.ts
sroy3: Unfortunately, this does not work. I'm not sure if it's because it does not go through that same `dvcReader` or something else. I can investigate and change as a follow up.

PR #3310, opened 2023-02-19
Disable unnecessary chromatic snapshots

mattseddon:
See [this thread](https://iterativeai.slack.com/archives/C01R00PPQ1L/p1672922660694809) for some more context.

In an effort to save waste, I've removed a large number of snapshots that I don't think we need. 

Happy to be corrected if you would like to see either:

- Further snapshots removed.
- Some snapshots reinstated.

🙏🏻

### DIFF
--- a/webview/src/stories/CliIncompatible.stories.tsx
+++ b/webview/src/stories/CliIncompatible.stories.tsx
@@ -10,6 +10,7 @@ export default {
     data: {}
   },
   component: CliIncompatible,
+  parameters: { chromatic: { disableSnapshot: true } },
   title: 'Setup'
 } as Meta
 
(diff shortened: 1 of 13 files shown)


julieg18 (approved):
Looks good! 

> Happy to be corrected if you would like to see either:
> * Further snapshots removed.
> * Some snapshots reinstated.

Right now we have 4 stories based on having no data in each of the four plot sections: 

Screenshot 2023-02-20 at 7 39 24 AM

Do we need all four for snapshots or should we consider disables some?

sroy3 (approved):
The smooth template story does not need all of the viewport snapshots either I think: `SmoothTemplate.parameters = chromaticParameters`

mattseddon:
thanks @sroy3 @julieg18

PR #3311, opened 2023-02-19
Redirect user to add new Studio access token on 401 response

mattseddon:
This PR adds a simple unauthorized message when a 401 is received from Studio. The user can then redirect themselves to the Connect to Studio screen in order to add a new token.

### Demo

https://user-images.githubusercontent.com/37993418/219980654-3a6bc9dd-032b-4e51-b338-d273d2d6b0c8.mov

### DIFF
--- a/extension/src/patch.ts
+++ b/extension/src/patch.ts
@@ -1,5 +1,6 @@
+import { commands } from 'vscode'
 import omit from 'lodash.omit'
-import fetch, { Response } from 'node-fetch'
+import fetch, { Response as FetchResponse } from 'node-fetch'
 import {
   EXPERIMENT_WORKSPACE_ID,
   ExperimentFields,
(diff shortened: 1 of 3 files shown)


extension/src/patch.ts
```diff
@@ -121,7 +135,16 @@ const shareWithProgress = (
...
-    await sendPostRequest(studioAccessToken, { ...base, type: 'start' })
+    const response = await sendPostRequest(studioAccessToken, {
+      ...base,
+      type: 'start'
+    })
+
+    if (response.status === 401) {
+      progress.report({ increment: 100, message: 'Access unauthorized' })
+      void showUnauthorized()
+      return Toast.delayProgressClosing()
```
mattseddon, line 146: [F] Another option that I considered was to remove the token here as it has been shown to be invalid but I thought that would be too intrusive.

sroy3 (approved):

julieg18 (approved):
Great work!

PR #3319, opened 2023-02-20
Add metrics.json to plots list of watched files

mattseddon:
Patches https://github.com/iterative/vscode-dvc/issues/3264

Seems like there are few issues with `FileSystemWatcher`s on Codespaces. This patch is good enough for the time being.

### Demo

Codespace is setup with `example-get-started-experiments`.

https://user-images.githubusercontent.com/37993418/220210459-3a50329e-fdec-4de0-beec-d36686b2fcb5.mov

https://user-images.githubusercontent.com/37993418/220210582-8075a619-3901-4900-a367-6ddb25ee84c6.mov

### DIFF
--- a/extension/src/plots/data/index.ts
+++ b/extension/src/plots/data/index.ts
@@ -31,7 +31,7 @@ export class PlotsData extends BaseData<{
           process: () => this.update()
         }
       ],
-      ['dvc.yaml', 'dvc.lock']
+      ['dvc.yaml', 'dvc.lock', 'metrics.json']
     )
     this.model = model
   }


sroy3 (approved):

alex000kim:
How would this consider that the metrics file can be located anywhere and be named anything?

mattseddon:
@alex000kim it doesn't consider the fact that it can be named anything but it does cover that it can be anywhere. It is just a patch.

alex000kim:
Thanks! I figured that before seeing your reply - that's why I removed my comment :)

PR #3324, opened 2023-02-21
Optimize rendering time of plots by making sure snap points do not trigger re-rendering

sroy3:
Although `snapPoints` were not technically changing until we resize the window, the way that it was built before changed the reference to `snapPoints` and triggered some unwanted re-rendering making the plots webview really slow, especially when resizing. This optimization greatly improves the first rendering of plots and improves the resizing experience. There are some other optimizations to be done for sure, but this is a good start.

### Before

https://user-images.githubusercontent.com/3683420/220407754-0cf43a1c-519c-49a6-bb79-a326655eb417.mov

### Now

https://user-images.githubusercontent.com/3683420/220407440-25d4a890-e3fa-4a13-bec6-43b442265848.mov


sroy3:
Closing this. Will open a PR soon with something  else that'll change pretty much all of the work in here.

PR #3325, opened 2023-02-21
Optimize rendering of plots

sroy3:
I turned on _Why Did You Render?_ and saw multiple re-rendering due to the spec prop of `ZoomablePlot` it was memoized fine for checkpoint plots, but not at all for template plots. This PR addresses the memoization of the spec of template plots. It is even more beneficial while resizing (more fluid now).

https://user-images.githubusercontent.com/3683420/220441217-3bfe3b17-77f5-42c8-9b04-22639bb8a65b.mov

Resizing is still slow, but it does lag a lot less than it used to.

### DIFF
--- a/webview/src/util/wdyr.ts
+++ b/webview/src/util/wdyr.ts
@@ -5,6 +5,6 @@ import React from 'react'
 if (process.env.NODE_ENV === 'development') {
   const whyDidYouRender = require('@welldone-software/why-did-you-render')
   whyDidYouRender(React, {
-    trackAllPureComponents: true
+    trackAllPureComponents: false
   })
 }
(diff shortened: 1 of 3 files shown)


webview/src/util/wdyr.ts
```diff
@@ -5,6 +5,6 @@ import React from 'react'
 if (process.env.NODE_ENV === 'development') {
   const whyDidYouRender = require('@welldone-software/why-did-you-render')
   whyDidYouRender(React, {
-    trackAllPureComponents: true
+    trackAllPureComponents: false
```
sroy3, line 8: WDYR is still a useful tool, but having this set to true was getting quite annoying.

sroy3:
Closing this. Will open a PR soon with something  else that'll change pretty much all of the work in here.

Issue #3334, opened 2023-02-22
Optimize re-rendering of plots


sroy3:
There is a lot of useless re-rendering and drilling down of certain props (like the whole content of the Vega plot) that make the plots UI quite slow. It makes the resizing feature practically useless as it's almost impossible to use at times. We can definitely improve on the re-rendering to make the webview faster.

sroy3:
Two PRs (#3337 and #3341 ) were open to optimize re-rendering. I think that there could be even more work done to optimize, but we could probably be spending an infinite amount of time on this with fewer increases in performance.

Only other thing left to close this would be cleaning up the code that update `ZoomedInPlot` as it's doing pretty much the same thing as the new optimization is doing.

PR #3341, opened 2023-02-23
Optimize the plots sections to reduce the number of useless re-renderings

sroy3:
Part of #3334 

### Demo before

https://user-images.githubusercontent.com/3683420/221030411-bd441d74-d36f-4d4d-ae47-5aad60815935.mov

### Demo now

https://user-images.githubusercontent.com/3683420/221030697-e30fe089-db23-4776-a15f-d382147f3efc.mov

Not a big improvement, but this should add up with the number of plots and it should make the code a little easier to read.

### DIFF
--- a/webview/src/plots/components/templatePlots/TemplatePlots.tsx
+++ b/webview/src/plots/components/templatePlots/TemplatePlots.tsx
@@ -5,7 +5,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import { MessageFromWebviewType } from 'dvc/src/webview/contract'
 import { AddedSection } from './AddedSection'
 import { TemplatePlotsGrid } from './TemplatePlotsGrid'
-import { PlotGroup, updateUserSections } from './templatePlotsSlice'
+import { PlotGroup, updateSections } from './templatePlotsSlice'
 import { removeFromPreviousAndAddToNewSection } from './util'
 import { sendMessage } from '../../../shared/vscode'
 import { createIDWithIndex, getIDIndex } from '../../../util/ids'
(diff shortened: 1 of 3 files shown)


webview/src/plots/components/templatePlots/TemplatePlots.tsx
```diff
@@ -43,29 +35,30 @@ export const TemplatePlots: React.FC = () => {
...
+      })),
+      type: MessageFromWebviewType.REORDER_PLOTS_TEMPLATES
+    })
+  }, [])
+
+  const setSections = useCallback(
+    (sections: PlotGroup[]) => {
+      /* Although the following dispatch duplicates the work the reducer will do when the state returns 
+         from the extension, this is necessary to not see any flickering in the order as the returned state 
+         sometimes takes a while to come back */
```
mattseddon, line 56: 🎉

mattseddon (approved):

mattseddon:
[Q] Can we automate re-render performance testing in some way?

sroy3:
> [Q] Can we automate re-render performance testing in some way?
> 
> Very quick look around found this: [testing-library/react-hooks-testing-library#135](https://github.com/testing-library/react-hooks-testing-library/issues/135)

I'll take a little bit of time to research this today. I think this could be very useful in the future. Even with virtualization, there is a lot of data rendered at once. Making sure we do not worsen re-rendering performance should be on own mind when developing in the plots webview (experiments table as well).

sroy3:
[Profiler API](https://reactjs.org/docs/profiler.html) is probably the best tool for testing re-renders. The problem is that we don't want to test the number of re-renders for an isolated component and we do not want to profile simply the `<App />` component either. What would be best, would be being able to wrap a certain component within the whole `<App />` component. I'll see if it's possible to create a test helper method to solve that problem.

PR #3346, opened 2023-02-24
Improve naming in experiment row context menu component

mattseddon:
Comments inline

### DIFF
--- a/webview/src/experiments/components/table/RowContextMenu.tsx
+++ b/webview/src/experiments/components/table/RowContextMenu.tsx
@@ -175,7 +175,7 @@ const getSingleSelectMenuOptions = (
 
   const notExperiment = depth !== 1
 
-  const withId = (
+  const notRunningWithId = (
     label: string,
     type: MessageFromWebviewType,
     hidden?: boolean,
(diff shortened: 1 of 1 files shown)


webview/src/experiments/components/table/RowContextMenu.tsx
```diff
@@ -175,7 +175,7 @@ const getSingleSelectMenuOptions = (
 
   const notExperiment = depth !== 1
 
-  const withId = (
+  const notRunningWithId = (
```
mattseddon, line 178: I have now been caught out by this POS a few times. It silently added `|| hasRunningExperiment` to the hidden options which leads to plenty of head scratching. Name is updated to reflect what it is actually doing.

webview/src/experiments/components/table/RowContextMenu.tsx
```diff
@@ -190,35 +190,34 @@ const getSingleSelectMenuOptions = (
...
       MessageFromWebviewType.CREATE_BRANCH_FROM_EXPERIMENT,
       isNotExperimentOrCheckpoint
     ),
-    experimentMenuOption(
-      id,
+    notRunningWithId(
       'Share to Studio',
       MessageFromWebviewType.SHARE_EXPERIMENT_TO_STUDIO,
-      notExperiment || hasRunningExperiment,
+      notExperiment,
```
mattseddon, line 206: [F] Should be no behaviour change here.

sroy3 (approved):

Issue #3178, opened 2023-01-30
Queue: wrap up e2e workflow


shcheklein:
Followup after #3091 

```[tasklist]
### Tasks
- [ ] Watch and keep updating plots
- [x] When something runs in background - can we allow running an experiment in workspace?
- [x] Attach and follow standard output / error
- [ ] Failed / stoped - add resume and / or run again action
- [x] Add to a queue action is very slow - can we add a progress bar and notification? (fixed by https://github.com/iterative/vscode-dvc/pull/3815 without adding a notification)
```

See https://github.com/iterative/vscode-dvc/issues/3091#issuecomment-1409125370

shcheklein:
I've updated the ticket.

mattseddon:
Depends on https://github.com/iterative/dvc/issues/8478

mattseddon:
@shcheklein can you please reorder the checkboxes in order of priority? I will add some comments.

shcheklein:
@mattseddon done (also, trying a new beta feature - tasklists).

mattseddon:
> When something runs in background - can we allow running an experiment in workspace?

We need to update the extension's `package.json` and the experiments table to make this happen. It should be achievable to enable. However, I have manually tested this and the experience feels very unstable.

> Attach and follow standard output / error

It can be done with `dvc queue logs <exp-id> -f` but will need a sizeable amount of wiring built around it. We would need to use the `PseudoTerminal` and build a wrapper around it that looks a lot like the `DvcRunner`. Should be enabled from the right-click context menu in the table which will bring its own set of challenges (can it be run twice for the same experiment, should it be disabled, under what circumstances, etc).

> Failed / stoped - add resume and / or run again action

I don't think this is currently possible.  Maybe we could do something with a combination of apply, re-queue and start the queue but I feel like it is unlikely that it will work as expected. Feels like this needs a FR against DVC.

> Add to a queue action is very slow - can we add a progress bar and notification?

What would the progress notification show and when would it update? From the data we get back, it would go from being empty to being full and that would be it (we have no information about intermediate steps). My suggestion for this would be to raise an issue in DVC to make the command more performant.

mattseddon:
I don't think we can mark this as done until https://github.com/iterative/dvc/issues/8763 is resolved.

shcheklein:
Thanks @mattseddon, for giving more color / details to the list.

> However, I have manually tested this and the experience feels very unstable.

Could you please add more color/details to this?

[...]

PR #3347, opened 2023-02-24
Add show logs to context menu of experiments running in the queue

mattseddon:
Part of https://github.com/iterative/vscode-dvc/issues/3178

WIP.

Lots of duplication in the PR that ~can~ will be removed. Need to think more about how to set this out so that it can be used under certain circumstances for things like `dvc pull`. Thinking maybe

1. show progress notification which contains "show logs" link
2. show logs link opens a terminal and starts piping the output in (like demo below)

### Demo

https://user-images.githubusercontent.com/37993418/221096848-df1e98c6-66c1-40e0-bb15-ae679d248c1f.mov

### DIFF
--- a/extension/src/cli/dvc/constants.ts
+++ b/extension/src/cli/dvc/constants.ts
@@ -67,6 +67,7 @@ export enum ExperimentSubCommand {
 
 export enum QueueSubCommand {
   KILL = 'kill',
+  LOGS = 'logs',
   START = 'start',
   STOP = 'stop'
 }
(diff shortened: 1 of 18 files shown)


mattseddon:
image

you got me

dberenbaum:
@mattseddon Will it work for failed/successful experiments that were run by the queue also?

shcheklein:
Looks great! Q: is there a way to control the Terminal name? To include the experiment name in it:

Screenshot 2023-02-24 at 10 22 04 AM

So that I can see the list of Tasks (experiments) / terminals like in this case ^^, switch between them, etc?

mattseddon:
> @mattseddon Will it work for failed/successful experiments that were run by the queue also?

It can. Not sure if all of the required data is actually in `exp show` but I can work it out.

> Q: is there a way to control the Terminal name?

Yep, was getting there.

mattseddon:
Need to refactor before I can get this in.

PR #3360, opened 2023-02-27
Add show logs to context menu of experiments running in the queue

mattseddon:
Replaces #3347, part of https://github.com/iterative/vscode-dvc/issues/3178

This PR adds the ability to show the logs of experiments running in the queue into the experiments table context menu

### Demo

https://user-images.githubusercontent.com/37993418/221721281-91766932-d45a-4d0c-8164-8c802ca771d2.mov

Notes: 
- For the time being I have not wired up showing of logs from the experiments tree because a running experiment can change quick enough for the command to be called with `undefined`. This was confusing behaviour to track down. Once we provide the ability to show logs of experiments that have been run via the queue I will add the command for non-running experiments.
- Working out which completed experiments have readable/persisted logs will require some of the prerequisites for https://github.com/iterative/dvc/issues/8478

> The directory being used can be determined based on the hash for the experiment. Files related to the temporary execution process/dir are stored in `.dvc/tmp/runs/<hash>`.

From some investigation, it appears that the `<hash>` shown in the directory mentioned above relates to the queue task and not the final `<hash>` of the run experiment (obviously). This means that in order to get the required information I will have to read every `.run` file under the `.dvc/tmp/runs` directory (could possibly ignore folders that don't have a `.out`) and get a mapping between the `name` entry and what is in the experiments table.

cc @dberenbaum


sroy3 (approved):
I can't seem to be able to see the "Show log" option. Is there anything specific to do to see it?

julieg18 (approved):
Working on my end!

mattseddon:
> I can't seem to be able to see the "Show log" option. Is there anything specific to do to see it?

Only works for experiments running in the queue so queue an experiment and then start the queue. LMK if that still doesn't work.

sroy3:
> > I can't seem to be able to see the "Show log" option. Is there anything specific to do to see it?
> 
> Only works for experiments running in the queue so queue an experiment and then start the queue. LMK if that still doesn't work.

Yes, I got it to work. I somehow missed "running in the queue" from the name.

PR #3363, opened 2023-02-27
Watch for dvc.yaml changes for manually added stages

sroy3:
https://user-images.githubusercontent.com/3683420/221600786-437288f1-c205-43de-90df-084201b02a24.mov

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -156,6 +156,7 @@ export class Experiments extends BaseRepository<TableData> {
         if (hadCheckpoints !== this.hasCheckpoints()) {
           this.checkpointsChanged.fire()
         }
+        this.webviewMessages.sendWebviewMessage()
       })
     )
 


extension/src/experiments/index.ts
```diff
@@ -156,6 +156,7 @@ export class Experiments extends BaseRepository<TableData> {
         if (hadCheckpoints !== this.hasCheckpoints()) {
           this.checkpointsChanged.fire()
         }
+        this.webviewMessages.sendWebviewMessage()
```
mattseddon, line 159: [Q] What happens if the `dvc.yaml` is completely broken?

mattseddon (approved):

extension/src/experiments/index.ts
sroy3: It'll ask for a stage because it cannot find any. In a way, that was always a problem if the `dvc.yaml` is broken before opening the webview, but we could probably send whether the `yaml` is valid as well and show an error in that case. I can add this as a follow up or include it into this PR.

sroy3: Actually, I kind of forgot the use of this watcher while working on this feature. It doesn't have anything to to with the top button, it's fir the button in the webview. I'll close this PR and re-open with the real solution.

PR #3365, opened 2023-02-27
Watch for dvc.yaml changes for manually added stages

sroy3:
https://user-images.githubusercontent.com/3683420/221702198-88ae87b4-9df9-4dc5-a30c-0145e05091cd.mov


extension/src/experiments/webview/messages.ts
```diff
@@ -74,8 +75,11 @@ export class WebviewMessages {
     this.addStage = addStage
   }
 
-  public async changeHasConfig() {
-    this.hasConfig = !!(await this.hasStages())
+  public async changeHasConfig(update?: boolean) {
+    const stages = await this.hasStages()
+    this.hasValidDvcYaml = stages !== undefined
+    this.hasConfig = !!stages
```
mattseddon, line 81: [Q] Is there any way that `this.hasValidDvcYaml` isn't `!this.hasStages`? Is `this.hasValidDvcYaml === !this.hasConfig` always true? What values of stages mean that we need both variables? Can we refactor to make it more obvious? (my brain is tiny)

mattseddon (approved):

webview/src/experiments/components/AddStage.tsx
```diff
@@ -5,16 +5,28 @@ import { IconButton } from '../../shared/components/button/IconButton'
...
 import { sendMessage } from '../../shared/vscode'
 
-export const AddStage: React.FC = () => (
-  <div className={styles.addConfigButton}>
+interface AddStageProps {
+  hasValidDvcYaml: boolean
+}
+
+export const AddStage: React.FC<AddStageProps> = ({ hasValidDvcYaml }) => (
+  <div className={styles.addConfigButton} data-testid="aaa">
```
julieg18, line 13: Should we name this id to be a more clear name? Though it looks like it isn't being used anywhere 🤔

julieg18 (approved):
Great work!

webview/src/experiments/components/AddStage.tsx
sroy3: Nice catch It was for quick testing (had a lot of trouble testing the shadow DOM button), but I forgot to remove it after.

extension/src/experiments/webview/messages.ts
sroy3: `dvc stage list` will return '' if no pipelines were found. It will return an error (`'./dvc.yaml' validation failed.`) if the `dvc.yaml` file is broken. the `listStages` command will return the result, but it has a `catch` clause that will return undefined. 

In other words, `hasInvalidDvcYaml = hasStages() === undefined` and `doesNotHaveConfig = hasStages() === ''`. 

We cannot test accurately for `hasConfig` if `hasValidDvcYaml` is `true`, but I chose to set it to false because it shows the message at the bottom of the table.

PR #3376, opened 2023-03-01
Rename size to nbItemsPerRow

sroy3:
Was trying to add plot height into the data and realized the `size` was a little confusing. In the end `size` refers to the width, but more specifically to the number of items per row.


webview/src/plots/components/templatePlots/templatePlotsSlice.ts
```diff
@@ -57,12 +57,12 @@ export const templatePlotsSlice = createSlice({
       return {
         ...state,
         hasData: !!action.payload,
+        nbItemsPerRow: action.payload.nbItemsPerRow,
         plotsSnapshots: snapShots,
         sections:
           JSON.stringify(plotSections) === JSON.stringify(state.sections)
             ? state.sections
-            : plotSections,
-        size: Math.abs(action.payload.size)
```
sroy3, line 65: `Math.abs` was there because when first implementing the feature, I kept making the width negative. It shouldn't be a problem now.

webview/src/stories/ComparisonTable.stories.tsx
```diff
@@ -43,22 +43,13 @@ const Template: Story = ({ plots, revisions }) => {
...
           plots,
-          revisions,
-          size: PlotSizeNumber.REGULAR
+          revisions
         }}
       >
         <WebviewWrapper>
-          <div
-            style={
-              { '--size': `${PlotSizeNumber.REGULAR}px` } as DetailedHTMLProps<
```
sroy3, line 54: This wasn't used at all.

extension/src/plots/vega/util.test.ts
```diff
@@ -275,7 +278,7 @@ describe('extendVegaSpec', () => {
   it('should update the multi-source template to remove erroneous shape encoding from the vertical line displayed on hover', () => {
     const updatedSpec = extendVegaSpec(
       multiSourceTemplate,
-      PlotSizeNumber.LARGE,
+      PlotNumberOfItemsPerRow.LARGE,
```
mattseddon, line 281: [C] Should probably change the names in the enum from LARGE etc to the numbers.

mattseddon (approved):

extension/src/plots/vega/util.test.ts
sroy3: I'm doing a follow up. I first thought of adding `height` inside of `plotSizes`, but it'll be way simpler to just add another property. I'll rename these at the same time.

PR #3378, opened 2023-03-01
Move Connect to Studio button into Actions tree

mattseddon:
First very small step based on [this feedback](https://iterativeai.slack.com/archives/C01APS0FHDM/p1677657980420979)

Seems like a reasonable step as I try to work out what is next.

### Demo

https://user-images.githubusercontent.com/37993418/222293911-5513b9af-17c3-4c8a-9559-a42c73071fa3.mov

Options I am considering for the 3rd button after Studio is connected:

1. Show settings
2. Share Experiment Live or checkbox or similar
3. Edit Studio Connection
4. No change

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -1345,11 +1345,6 @@
           "when": "true"
         },
         {
-          "id": "dvc.views.studio",
-          "name": "Studio",
-          "when": "!dvc.studio.connected"
-        },
-        {
           "id": "dvc.views.experimentsColumnsTree",
           "name": "Columns",
           "when": "dvc.commands.available && dvc.project.available"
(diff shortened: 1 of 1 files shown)


extension/package.json
```diff
@@ -1383,18 +1378,18 @@
     "viewsWelcome": [
       {
         "view": "dvc.views.actions",
-        "contents": "[$(beaker) Show Experiments](command:dvc.showExperiments)\n[$(graph-scatter) Show Plots](command:dvc.showPlots)\n[$(play) Run Experiment](command:dvc.runExperiment)",
-        "when": "!dvc.experiment.checkpoints"
```
sroy3, line 1387: So we're removing the "Run Experiment" button? Isn't there enough room for four buttons?

sroy3 (approved):

extension/package.json
julieg18: Also not sure about taking off the "Run Experiment" button. Isn't that button valuable for people just getting started with dvc?

julieg18 (approved):

extension/package.json
shcheklein: How about turning the  `+Add Pipeline` button into the `Run Experiment` ? I think the whole idea behind that button was to promote creation of the DVC config. They kinda overlap, the second one gives enough visibility already. We don't need it permanently on the side panel. WDYT folks?

sroy3: IMO it was nice to be able to run an experiment without having the experiments table open (there is also a run button available at the top of that webview). It might be just me using the product as a developer, but sometimes just having the plots webview open and running an experiment without having to switch webview is welcomed.

mattseddon: > So we're removing the "Run Experiment" button? Isn't there enough room for four buttons?

The 4th button is hidden by default.

> Also not sure about taking off the "Run Experiment" button. Isn't that button valuable for people just getting started with dvc?

There is value there for anyone with the view open but there are plenty of ways to run an experiment.

> How about turning the +Add Pipeline button into the Run Experiment ? I think the whole idea behind that button was to promote creation of the DVC config. They kinda overlap, the second one gives enough visibility already. We don't need it permanently on the side panel. WDYT folks?

Good idea.

mattseddon:
After sleeping on it this feels too artificial. There is not enough value available in connecting to Studio yet. Will revert back to running a separate tree for the time being.

PR #3381, opened 2023-03-02
Plug in the height inside plots data

stephr0y:
Prepping the field to use the height and set it. Makes it easier for both you to review and I to work on resizing the plots vertically if this is done separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "fastify": "3.29.5",
     "json5": "2.2.3",
     "loader-utils": "2.0.4",
-    "terser": "5.16.4",
+    "terser": "5.16.5",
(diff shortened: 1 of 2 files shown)


julieg18 (approved):
If I'm understanding correctly, for now, the webview isn't updating the height yet? Aka, the height will stay undefined throughout a session with the plots webview? 

Also, looks like some tests need to be corrected.

sroy3:
> If I'm understanding correctly, for now, the webview isn't updating the height yet? Aka, the height will stay undefined throughout a session with the plots webview?
> 
> Also, looks like some tests need to be corrected.

Right, nothing is used at the moment. Just adding it before any logic so that things don't get too messy.

Issue #3373, opened 2023-03-01
Improve Custom Plots


julieg18:
Tasks are sorted by importance. 

```[tasklist]
### Tasks
- [x] Remove Trends section, instead having trends plots be plots you can create in "Custom". This will simplify the code logic and duplication.
- [x] Update README/Walkthrough
- [x] Remove the possibility of a user selecting options that create a plot that is already made. Currently an error popup will appear if a user does this. (See [#3342 discussion](https://github.com/iterative/vscode-dvc/pull/3342#discussion_r1120171539))
- [x] Add commit data into custom plots
- [x] Update get started screen (see https://github.com/iterative/vscode-dvc/pull/3382#discussion_r1125568446)
- [x] Add command pallete commands
- [x] Have metric vs param plots use scatter plots instead of linear plots
- [x] Make sure that custom spec does not generate errors in `webview/src/plots/components/App.test.tsx` (https://github.com/iterative/vscode-dvc/pull/3575#discussion_r1155414065)
- [x] [Delete old Persistance metrics state](https://github.com/iterative/vscode-dvc/pull/3404#discussion_r1133350817) (#3518)
- [x] [Adjust custom plots order state](https://github.com/iterative/vscode-dvc/pull/3404#discussion_r1133350999) (#3518)
- [x] Investigate loading section test (https://github.com/iterative/vscode-dvc/pull/3404#discussion_r1133370484)
- [x] Update walkthrough images after other PRs are merged (https://github.com/iterative/vscode-dvc/pull/3404/files#r1133343393)
- [x] Check if plot spec creation can be moved to extension (https://github.com/iterative/vscode-dvc/pull/3404#discussion_r1134686640)
- [x] Consolidate type checking to the smallest space/amount of functions (https://github.com/iterative/vscode-dvc/pull/3404#discussion_r1134686640) (Most of this has been done already but more to do)
- [x] Add code to add 1 trend and 1 metric vs param custom plot to e2e test (https://github.com/iterative/vscode-dvc/pull/3404/files#r1133364032)
- [x] Improve Jest testing on plots section (see [#3342 comment](https://github.com/iterative/vscode-dvc/pull/3342#discussion_r1120188485))
```

Followup to #3342

mattseddon:
Addition of a custom plot with no experiments present breaks the empty state screen:


https://user-images.githubusercontent.com/37993418/222307123-a8adab15-0109-4698-8520-3afba4c4fec4.mov

[...]

PR #2932, opened 2022-12-14
Update plots empty states to reflect dynamic nature of available plots

mattseddon:
Follow up to https://github.com/iterative/vscode-dvc/pull/2915/files#r1046849621

The empty states are now as shown below:

image

image

image

PR #3382, opened 2023-03-02
Fix plots "get started" walkthrough

julieg18:
* shows the get started screen if there are no selected experiments/plots even if there is custom plot data
* stops the custom plots showing empty plots if there are no experiments in the table

## Main

https://user-images.githubusercontent.com/43496356/222498681-5e468cb2-8d50-47c8-94fb-4fd89d1070a3.mov

## PR

https://user-images.githubusercontent.com/43496356/222498626-153097fb-c065-4ddc-b28f-6d8560df94ae.mov

Part of #3373


webview/src/plots/components/GetStarted.tsx
```diff
@@ -5,7 +5,6 @@ import { StartButton } from '../../shared/components/button/StartButton'
 
 export type AddPlotsProps = {
   hasUnselectedPlots: boolean
-  hasSelectedRevisions: boolean
```
julieg18, line 8: I noticed this `hasSelectedRevisions` isn't being used anywhere so I deleted it. Or is it supposed to be used to hide the `Add Experiments` button?

webview/src/plots/components/Plots.tsx
```diff
@@ -54,20 +49,10 @@ const PlotsContent = () => {
     return <EmptyState>Loading Plots...</EmptyState>
   }
 
-  if (
-    !hasCheckpointData &&
-    !hasComparisonData &&
-    !hasTemplateData &&
-    !hasCustomData
-  ) {
+  if (!hasCheckpointData && !hasComparisonData && !hasTemplateData) {
```
julieg18, line 52: Explanation of the new walkthrough in more detail:
* has plots render the get started screen if there are no selected plots or experiments even if there is custom plot data 
* has the custom section render "No Data to Display" if there is no data or there are no experiments
* has the custom section render "No Plots Added" if there is experiment data but no plots added

webview/src/plots/components/GetStarted.tsx
sroy3: Looks like it was a simple [leftover](https://github.com/iterative/vscode-dvc/pull/2932/files#diff-45b20cc7f55a9ad8b30d0c84017251735c1a0bfa97a4ef787da36c2d7ecb96c3). Good catch!

sroy3 (approved):

webview/src/plots/components/Plots.tsx
mattseddon: > has plots render the get started screen if there are no selected plots or experiments even if there is custom plot data

[Q] Does this mean that we hide added custom plots even though there is something to display? How about we show a non-fullscreen `<EmptyState>` (`isFullScreen=false`) version of the get started screen above the custom plots? What would that look like?

julieg18: > [Q] Does this mean that we hide added custom plots even though there is something to display? How about we show a non-fullscreen <EmptyState> (isFullScreen=false) version of the get started screen above the custom plots? What would that look like?


Yes, we do hide it since I was thinking it would look odd but now that actually it looks pretty good! I think the first one looks better but does require a little more CSS:


Screenshot 2023-03-03 at 4 58 27 PM

Screenshot 2023-03-03 at 5 04 14 PM

If we want to keep this, I'm assuming we'd need to adjust the message since there are actually some plots to display... Maybe "There are no selected plots or experiments"? What does everyone else think?

julieg18:
Going to go ahead and merge this so we can get rid of the bug. Will update the screen in a followup if needed!

webview/src/plots/components/Plots.tsx
mattseddon: I definitely think this is better and we should not be hiding anything that has been selected based on some other hidden condition. 

> Maybe "There are no selected plots or experiments"

👍🏻. I think there are some variations on this statement but that works. 

I also think we should be displaying all commit data in custom plots as well as experiments.

mattseddon: > I also think we should be displaying all commit data in custom plots as well as experiments.

Added a checkbox in the original ticket 🙏🏻.

PR #3390, opened 2023-03-03
Remove package.nls

mattseddon:
This PR removes our `package.nls.json` by inlining all of the values. We need to do this to workaround https://github.com/microsoft/vsmarketplace/issues/489

### Current problem

https://user-images.githubusercontent.com/37993418/222683866-8e77362a-1a9d-475e-bddc-a6eb4c642bae.mov

### Demo (built VSIX)

https://user-images.githubusercontent.com/37993418/222684145-4a29b4bd-d91d-40b8-8683-8b987c107938.mov

This was a painful way to end the week.


extension/package.json
```diff
@@ -554,8 +554,7 @@
       "title": "DVC",
       "properties": {
         "dvc.focusedProjects": {
-          "title": "%config.focusedProjects.title%",
```
mattseddon, line 557: In https://github.com/iterative/vscode-dvc/pull/201 a title was added to a configuration property. For 3000+ PRs/issues we have persisted with the title field. It has no effect.

mattseddon: probably even debated those titles.

sroy3 (approved):

PR #3398, opened 2023-03-06
Have queue workers respect dvc.studio.shareExperimentsLive

mattseddon:
Related to #3077

This PR connects new queue workers with Studio. If an access token is present then the queue will start processing with the correct environment variables to send everything to Studio. This is handy because the functionality is not currently available locally.

### Demo

https://user-images.githubusercontent.com/37993418/222999310-096341cf-4438-42b2-b590-0923f8e8af52.mov

### DIFF
--- a/extension/src/cli/dvc/executor.ts
+++ b/extension/src/cli/dvc/executor.ts
@@ -42,11 +42,13 @@ export class DvcExecutor extends DvcCli {
   )
 
   private readonly getStudioLiveShareToken: () => string | undefined
+  private readonly getRepoUrl: (cwd: string) => Promise<string>
   private scmCommandRunning = false
 
   constructor(
     config: Config,
     getStudioLiveShareToken: () => string | undefined,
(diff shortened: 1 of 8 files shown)


extension/src/cli/dvc/executor.ts
```diff
@@ -38,8 +41,24 @@ export class DvcExecutor extends DvcCli {
...
   )
 
+  private readonly getStudioLiveShareToken: () => string | undefined
+  private readonly getRepoUrl: (cwd: string) => Promise<string>
   private scmCommandRunning = false
 
+  constructor(
+    config: Config,
+    getStudioLiveShareToken: () => string | undefined,
+    getRepoUrl: (cwd: string) => Promise<string>,
```
mattseddon, line 51: [F] Currently needed due to https://github.com/iterative/dvclive/issues/485.

sroy3 (approved):

PR #3399, opened 2023-03-06
Expose Open Studio Settings in the command palette

mattseddon:


shcheklein:
Good, we can go with this as a next step. I think we need to figure out something though with the number of different screens and buttons- we need to consolidate them. Btw, will the button stay "Connect Studio" initially, when there is no token yet?

mattseddon:
> Good, we can go with this as a next step. I think we need to figure out something though with the number of different screens and buttons- we need to consolidate them. Btw, will the button stay "Connect Studio" initially, when there is no token yet?

It is under the Studio tree so it says "Connect":

https://user-images.githubusercontent.com/37993418/223006883-0a367d0c-ca6e-46db-9de7-ad82058f8940.mov

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3401, opened 2023-03-06
Watch deps for experiment updates

mattseddon:
Realised that this was missing when collecting experiment data requirements.


mattseddon:
(remembered) This is not required because `deps` cannot be updated without running `dvc commit`.

PR #3404, opened 2023-03-06
Move trends plots inside of "Custom" section

julieg18:
# 1/3 main <= this <= #3466 <= #3491
## Demo

https://user-images.githubusercontent.com/43496356/224517013-fbc554e8-8c81-43c6-91b2-d8b5412e960c.mov

Part of #3373

PR #3405, opened 2023-03-06
Select the number of items per row from slider

sroy3:
Part of #2585 

### Demo

https://user-images.githubusercontent.com/3683420/223521083-fa2ce8f4-0d55-4d11-bcc4-f0d59ba40897.mov

### Demo on a bigger screen (higher maximum number of plots per row)

https://user-images.githubusercontent.com/3683420/223521933-ea4dc183-ae13-4b3e-9a44-23d164b0653a.mov

### Scope

- Remove the resizer on each plot and add a slider to select the number of plots per row
- Slider is only available if there are plots in the section
- Slider is not available in the comparison table (still no resize in that section)
- Maximum number of plots per row is dynamic. It changes with the screen size. Minimum width of one plot is 300px


sroy3:
Just used the current controls we add, but we could probably change this for a slider. We can either add more options or add a custom option at the end for the user to specify what they want. In the end, it does simplify things by a lot, it'll be way easier to maintain, and will definitely improve performance.

shcheklein:
Thanks, @sroy3 !

Let's do a few changes (going forward, I know this is just a POC):

- Add a very explicit message before the icon that this is about plots layout
- Let's not limit it by 4? I'm not sure why we have this limitation tbh.
- Q: how does it affect the large plots?

sroy3:
> Thanks, @sroy3 !
> 
> Let's do a few changes (going forward, I know this is just a POC):
> 
> * Add a very explicit message before the icon that this is about plots layout

There are tooltips that pop pretty much instantly on these icon buttons. All plots controls are centralized there. Discovery shouldn't be hard. If we add a message, we'd have to rethink that whole menu. We have more important actions without obvious labels either. As a matter of fact, most of the actions in the extension are only one icon and a tooltip. We can try to find better icons as well if neccesary.


https://user-images.githubusercontent.com/3683420/223200643-4727c5d0-6403-49a8-87a2-d9844dbcc625.mov

Screenshot 2023-03-06 at 1 34 38 PM


> * Let's not limit it by 4? I'm not sure why we have this limitation tbh.

I can do that easily (more than before). I think we had limited to 4 because on a small screen it starts being quite small and calculating break points becomes more difficult the smaller the screen is (but we do not have those break points anymore, everything is done by css). We should limit this to how much then? 10? Should we just have a numerical input instead? By the way should these icon buttons simply open a quick pick / quick input like the buttons for custom plots? We currently have 2 strategies here, custom controls and using quick pick / quick inputs. I'll generalize them at the same time (use the quick picks / inputs).

> * Q: how does it affect the large plots?

It doesn't for now. But we could maybe do some calculations to make it fit. Something like, if there are 4 items per rows and the large plot has 2 revisions in it, then place 2 plots on the row. But if there are 3 items per row or 4+ revisions, stick to 1 plot per row. This will get better if we increase the possible number of items per row.

shcheklein:
> There are tooltips that pop pretty much instantly on these icon buttons.

Yep, but I think it's not enough unfortunately. With resize the issue is that I (as a user) expect it to be naturally part of the plots (e.g. grip, an icon on the plot, etc). It's less natural and no single standard way to make it part of the menu. Thus I would advocate for explicit controls (text + something) vs buttons only. We can always optimize it back, but there is no reason for us to make it harder for people to find.

> But we could maybe do some calculations to make it fit.

Yep, sounds good (there are some details still to clarify, but initial approach sounds reasonable to me).

webview/src/plots/components/checkpointPlots/CheckpointPlotsWrapper.tsx
```diff
@@ -42,6 +44,8 @@ export const CheckpointPlotsWrapper: React.FC = () => {
       menu={menu}
       nbItemsPerRow={nbItemsPerRow}
       sectionCollapsed={isCollapsed}
+      changeNbItemsPerRow={changeSize}
```
mattseddon, line 47: [C] I just realised that updating both Custom and Trend plots is going to collide with #3404

cc @julieg18

webview/src/plots/components/styles.module.scss
```diff
@@ -343,6 +288,12 @@ $resizer-border: 2px dashed $accent-color;
   }
 }
 
+.nbItemsPerRowSlider {
+  display: flex;
+  justify-content: end;
+  padding: 0 20px;
```
mattseddon, line 294: [C] When plots are large and I am at the bottom of the section then there is a lot of scrolling involved to get back up to the top if I want to adjust the size. Although this is less important if we don't think people are using it I still think is bad UX:

https://user-images.githubusercontent.com/37993418/223609270-3d6190cc-9ec4-4977-a3ca-a85cceff3b02.mov

Could make the control sticky in a follow-up? Up to you.

webview/src/plots/components/Plots.tsx
```diff
@@ -35,7 +35,9 @@ const PlotsContent = () => {
     const onResize = () => {
       wrapperRef.current &&
         dispatch(
-          setSnapPoints(wrapperRef.current.getBoundingClientRect().width - 100)
+          setMaxNbPlotsPerRow(
+            wrapperRef.current.getBoundingClientRect().width - 100
```
mattseddon, line 39: [Q] Can this pass `wrapperRef.current.getBoundingClientRect().width` only and do the full calc in the slice? Having `-100` here confused my tiny brain. All the logic would then be in a single place too.

webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -173,11 +203,25 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
...
         </summary>
+        {changeNbItemsPerRow && hasItems && (
+          <div
+            className={styles.nbItemsPerRowSlider}
+            data-testid="nb-items-per-row-slider"
+          >
+            <MinMaxSlider
+              minimum={1}
+              maximum={setMaxNbPlotsPerRow}
+              label="Number of plots per row"
```
mattseddon, line 214: [C] I do like the idea of reversing this, giving the label `Plot Width` and having the min value = max number of plots / max = 1 plot per row.

mattseddon: underlying implementation would be the same, just change the way the slider operates.

webview/src/plots/components/webviewSlice.ts
```diff
@@ -38,17 +36,12 @@ export const webviewSlice = createSlice({
...
     ) => {
-      // Action payload here is the max width of the plots webview. When changed, we re-calculate the snap points.
-      state.snapPoints = [
-        action.payload,
-        action.payload / 2,
-        action.payload / 3,
-        action.payload / 4
-      ]
+      // Action payload here is the max width of the plots webview. When changed, we re-calculate the max number of plots per row
+      state.maxNbPlotsPerRow = Math.floor(action.payload / 300)
```
mattseddon, line 44: [N] [for completeness](https://github.com/iterative/vscode-dvc/pull/3405/files#diff-66b700260ad828bfc40b1f51e751b4cf4798759be95acb9555e81a77588d59c4R39) from this could be

```
const width = action.payload
state.maxNbPlotsPerRow = Math.floor((width - 100) / 300)
```

extension/src/plots/webview/contract.ts
```diff
@@ -1,15 +1,7 @@
...
-// It is easier to keep the numerical order than the alphabetical one
-/* eslint-disable sort-keys-fix/sort-keys-fix */
-export const PlotNumberOfItemsPerRow = {
-  ONE: 1,
-  TWO: 2,
-  THREE: 3,
-  FOUR: 4
-}
-/* eslint-enable sort-keys-fix/sort-keys-fix */
+export const DefaultNumberOfItemsPerRow = 2
```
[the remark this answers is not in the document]
sroy3: > [N] Should use screaming snake case. e.g `NUM_OF_COMMITS_TO_SHOW`

I got confused and was wondering about this because of other variables not using this format. Will change that.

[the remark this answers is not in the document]
sroy3: > [Q] Should this be calculated based on the screen size?

Good question. I could try this as a follow up. Currently the default vale is set in the extension, so I'd have to reset everything to undefined to then calculate it in the webview. I'd probably select the middle value by default.

webview/src/plots/components/App.test.tsx
```diff
@@ -1603,10 +1595,10 @@ describe('App', () => {
         expect(screen.getByRole('grid')).toBeInTheDocument()
       })
 
-      it('should not wrap the template plots in a big grid (virtualize them) when there are ten or fewer large plots', async () => {
+      it('should not wrap the template plots in a big grid (virtualize them) when there are eigth or fewer large plots', async () => {
```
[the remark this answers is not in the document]
sroy3: Mashing the same buttons multiple times instead of copy and pasting will do that.

webview/src/plots/components/styles.module.scss
sroy3: Yes was thinking the same. I'll add this to the ticket, I think this is an important follow up.

webview/src/plots/components/PlotsContainer.tsx
sroy3: Yep, sounds good to me too.

webview/src/plots/components/checkpointPlots/CheckpointPlotsWrapper.tsx
julieg18: Yes, there, for sure, will be some collisions. Though I don't think they'll be anything too hard to correct :)

webview/src/shared/components/slider/MinMaxSlider.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { FormEvent } from 'react'
```
julieg18, line 1: There appears to be little space between the plots and the slider in the custom and trends sections: 

Screenshot 2023-03-08 at 8 19 11 AM

webview/src/plots/components/Plots.tsx
sroy3: I was thinking the opposite. Having `-100` so far away from the actual item we're calculating the width from is going to cause more confusion. That said, `100` probably refers to something (margin or padding maybe?), I'll label it correctly, because even I can't remember where it comes from.

webview/src/shared/components/slider/MinMaxSlider.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface MinMaxSliderProps {
+  minimum?: number
+  maximum: number
+  step?: number
+  defaultValue: number
+  label: string
+  onChange: (newValue: number) => void
+}
+
+export const MinMaxSlider: React.FC<MinMaxSliderProps> = ({
```
julieg18, line 13: Not sure if there is anything that can be done about it but the slider jumps sometimes when the plot rows increase: 

https://user-images.githubusercontent.com/43496356/223739261-93d346ee-8f54-4a8a-addd-d407a4257499.mov

webview/src/shared/components/slider/MinMaxSlider.tsx
```diff
@@ -0,0 +1,42 @@
...
+  minimum,
```
julieg18, line 14: If the screen reaches a certain small width, the slider can't move

https://user-images.githubusercontent.com/43496356/223740405-ac7e2990-8780-4f3a-99f9-8b0f117775bd.mov

Should we hide/disable the slider in this case?

julieg18 (approved):
Great work!

webview/src/shared/components/slider/MinMaxSlider.tsx
sroy3: The way to solving that would be to make sure the last section is at least `100vh` which does not always look good, especially when it's empty or have very few plots. I'm wondering though if we shouldn't have one width slider for all plots instead of resizing per section.

sroy3:
Newest demo:

https://user-images.githubusercontent.com/3683420/223764437-065db2c4-2b90-4d71-8e32-c38c84527014.mov

webview/src/shared/components/slider/styles.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  align-items: center;
+}
+
+.label {
+  font-size: 0.6rem;
+  margin-right: 10px;
+  color: $fg-color;
+}
+
+.slider {
```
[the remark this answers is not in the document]
sroy3: Not sure I get what you mean, aren't both slider the same?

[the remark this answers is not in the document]
sroy3: I see. I'll do that as a quick follow-up, no problem

PR #3407, opened 2023-03-06
Change the aspect ratio of plots through picker

sroy3:
Part of #2585 

https://user-images.githubusercontent.com/3683420/223207832-9a3abc70-2079-4221-a811-73cf3100125e.mov

Similar to #3405 

Currently using "random" aspect ratio. Could test more to get something that look nicer. Maybe we could have a resize quick pick / input with two steps (select number items per plots and then the aspect ratio), but I do like the fact that there you can see the impact of each choice from the picker directly and pick something else quickly if it's not what you like.


sroy3:
Too many changes with the slider. Will create a new clean branch.

PR #3421, opened 2023-03-07
Add command to reset state

sroy3:
https://user-images.githubusercontent.com/3683420/223568303-8fd553b2-f731-43ac-b64e-7db0b98997fe.mov


sroy3:
Not really sure how to test this to be honnest.

mattseddon (approved):
> Not really sure how to test this to be honest.

Could have an external function that is registered in the extension. Just give the function the workspace state instead of the whole context. You could then test that function with a `mockMemento` to make sure all of the correct keys are called. the main reason for adding that test would be to document the behaviour.

Maybe add `extension/src/persistence/util.ts` & `extension/src/persistence/register.ts`. Register would take the `workspaceState` object and so would the util. The test could be done on the util (wouldn't test the registered function as it would cause unexpected result in development)... remember to stub reload window 😁

julieg18 (approved):
Looks good!

sroy3:
> > Not really sure how to test this to be honest.
> 
> Could have an external function that is registered in the extension. Just give the function the workspace state instead of the whole context. You could then test that function with a `mockMemento` to make sure all of the correct keys are called. the main reason for adding that test would be to document the behaviour.
> 
> Maybe add `extension/src/persistence/util.ts` & `extension/src/persistence/register.ts`. Register would take the `workspaceState` object and so would the util. The test could be done on the util (wouldn't test the registered function as it would cause unexpected result in development)... remember to stub reload window 😁
> 
> edit: this is a suggestion.

Thanks that was really useful.

extension/src/test/util/index.ts
```diff
@@ -16,8 +16,8 @@ export const buildMockMemento = (
   ({
     get: (key: string, defaultValue: unknown) => values[key] || defaultValue,
     keys: () => Object.keys(values),
-    update: (key: string, value: unknown) =>
-      new Promise(() => {
-        values[key] = value
-      })
+    update: (key: string, value: unknown) => {
+      values[key] = value
+      void Promise.resolve()
```
mattseddon, line 21: [Q] Should this be returned?

PR #3423, opened 2023-03-08
Remove unnecessary Vega dependency

mattseddon:
Vega was added in #803 when plots were initially created. At that point, the whole package was used for a type (`Config`). After digging through the code we don't use it much and we can replace the usage with the stub types definition and `vega-util`.

Testing locally this seems to work. 
Please confirm that I have not gone crazy.


julieg18 (approved):
Looks good to me!

sroy3 (approved):
Works great on my end. Good riddance ! 😄

PR #3427, opened 2023-03-08
Change the background color for the smooth slider over plots

stephr0y:
Following https://github.com/iterative/vscode-dvc/pull/3405#discussion_r1129926118

https://user-images.githubusercontent.com/3683420/223824128-725ff041-ecd0-43bd-963f-113c370240e9.mov

### DIFF
--- a/extension/src/patch.ts
+++ b/extension/src/patch.ts
@@ -24,23 +24,14 @@ type ExperimentDetails = {
   params: ValueTreeRoot
 }
 
-type BaseRequestBody = {
+type RequestBody = {
   client: 'vscode'
   repo_url: string
   name: string
   baseline_sha: string
-}
-
-type StartRequestBody = BaseRequestBody & { type: 'start' }
-
-type DoneRequestBody = BaseRequestBody & { type: 'done' }
-
-type DataRequestBody = BaseRequestBody & {
   metrics: ValueTreeRoot
   params: ValueTreeRoot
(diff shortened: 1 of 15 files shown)


mattseddon (approved):

mattseddon:
Up to you on whether or not to merge. Was a suggestion not a requirement.

mattseddon:
Looks weird for zoomed-in plot 🤦🏻:

image

sroy3:
> Looks weird for zoomed-in plot 🤦🏻:
> 
> image

Now (first) vs before (second) the change:

https://user-images.githubusercontent.com/3683420/224347047-fca93e3a-ac48-4aac-9695-dce52e19e734.mov

It looks like it's actually better most of the time. If we want to make it look better, we'd need something different completely. Let's leave it like this for now as it's an improvement, and change it when we think of something better.

sroy3:
> I think the previous implementation was a bit better to be honest. I would try for these sliders to blend more with the plot itself. I don't see a strong reason for all sliders to have the same background, the context is different for them.
> 
> Sorry for going back and forth @sroy3 on this.

There are times when the slider is almost the same color as the background (see 0:42, 0:44, 0:48 in the video) (just making sure we're seeing the same thing). But I can easily revert and don't have an opinion on the matter.

sroy3:
Revert in #3444. Feel free to merge if needed.

PR #3428, opened 2023-03-08
Add a slider to resize plots vertically

sroy3:


extension/src/plots/webview/contract.ts
```diff
@@ -3,6 +3,17 @@ import { Color } from '../../experiments/model/status/colors'
 
 export const DEFAULT_NB_ITEMS_PER_ROW = 2
 
+export enum PlotHeight {
+  SMALLER = 0,
+  SMALL = 1,
+  REGULAR = 2,
+  SQUARE = 3,
+  VERTICAL_NORMAL = 4,
+  VERTICAL_LARGER = 5
```
mattseddon, line 12: [Q] Could these be the aspect ratios if that is what they refer to?

webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -204,19 +222,36 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
...
-              label="Plot Width"
-              onChange={handleResize}
-              defaultValue={-nbItemsPerRow}
-            />
-          </div>
+        {changeSize && hasItems && maxNbPlotsPerRow > 1 && (
+          <>
+            <div
+              className={styles.sizeSliders}
+              data-testid="nb-items-per-row-slider"
```
julieg18, line 229: Should we rename this id since it now holds two different sliders? Or maybe give each slider their own id 🤔

webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -204,19 +222,36 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
...
-            <MinMaxSlider
-              maximum={-1}
-              minimum={-maxNbPlotsPerRow}
-              label="Plot Width"
-              onChange={handleResize}
-              defaultValue={-nbItemsPerRow}
-            />
-          </div>
+        {changeSize && hasItems && maxNbPlotsPerRow > 1 && (
+          <>
```
julieg18, line 226: Is this `<></>` wrapper needed?

extension/src/plots/webview/contract.ts
mattseddon: Would we gain/lose anything by having this as `PlotAspectRatio`?

webview/src/plots/components/styles.module.scss
```diff
@@ -156,17 +155,28 @@ $gap: 20px;
...
+.ratioRegular .plot {
+  aspect-ratio: 4 / 3;
+}
+
+.ratioSquare .plot {
+  aspect-ratio: 1;
+}
+
+.ratioVerticalNormal .plot {
+  aspect-ratio: 3 /4;
```
julieg18, line 175: I'm assuming we want to keep things consistent 😄 
```suggestion
  aspect-ratio: 3 / 4;
```

webview/src/plots/components/PlotsContainer.tsx
```diff
@@ -204,19 +222,36 @@ export const PlotsContainer: React.FC<PlotsContainerProps> = ({
...
+            <div
+              className={styles.sizeSliders}
+              data-testid="nb-items-per-row-slider"
+            >
+              <div className={styles.sizeSlider}>
```
julieg18, line 231: Maybe it's just me, but it seems the two sliders on top of each other cause the header to have some empty space 🤔 

What about having them side by side? 

Screenshot 2023-03-12 at 4 43 55 PM

julieg18 (approved):
Great work!

mattseddon (approved):
Change LGTM.

One issue that I managed to create was y-axis label overflow (cc @julieg18):

image


https://user-images.githubusercontent.com/37993418/224575668-73f49d12-dae3-4dde-8aff-e1c5ab1fcc87.mov

webview/src/plots/components/PlotsContainer.tsx
mattseddon: Could be complicated but I think we should consolidate everything into the section header and make it sticky (as per https://github.com/iterative/vscode-dvc/pull/3443#pullrequestreview-1336087886). WDYT?

sroy3: I think it'll make the whole thing too bulky. I also like the fact that when we collapse the section, we don't see the sliders anymore. It's more of a distraction otherwise. Having them side by side isn't a bad idea. But I'll make the headers sticky as well.

sroy3:
> Change LGTM.
> 
> One issue that I managed to create was y-axis label overflow (cc @julieg18):
> 
> image

You're right, I forgot that y title truncation is done using the plot width at the moment. Need to change it to use the height instead.

extension/src/plots/webview/contract.ts
sroy3: First iteration used aspect ratios instead, but it made it more complicated to hold fractions instead of numbers. It makes it way easier to change the aspect ratio at the end of the chain (stylesheet) than doing so inside the enum here.

PR #3433, opened 2023-03-08
Add zoom plot telemetry event

mattseddon:
Realised this event was missing when pulling data for #2585


webview/src/plots/components/messages.ts
```diff
@@ -0,0 +1,5 @@
+import { MessageFromWebviewType } from 'dvc/src/webview/contract'
+import { sendMessage } from '../../shared/vscode'
+
+export const zoomPlot = () =>
+  sendMessage({ type: MessageFromWebviewType.ZOOM_PLOT })
```
mattseddon, line 5: [F] trying to follow the pattern in similar parts of the codebase.

Issue #3434, opened 2023-03-09
Consolidate onboarding, connect to Studio, settings & walkthrough


mattseddon:
There are a lot of small disparate screens related to onboarding (Setp webview) and connecting to Studio (Connect webview). There are also buttons associated with these screens in the sidebar. We also have a separate walkthrough (uses VS Code API) which users will see on first-install. The purpose of this ticket is to discuss and then implement a consolidation of this information into a single place.

My idea for an initial implementation involves reusing the layout from the plots webview. Namely this (without the ribbon):

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/223903438-85987a7a-cacf-4d8d-b8d7-34f269ffd841.png">

To begin with, we can start by moving the existing sections within this layout. We can also add a new section that links to the walkthrough.

After we get everything into the same place then we can discuss the next steps. 

Please LMK your thoughts on 

- Initial section titles.
- Name for the view (Setup?).
- Whether or not we will need to expose all steps at the same time (i.e do we continue to show things like initialize Git once the action has been completed?)
- Final layout after initial implementation
- Alternatives

Note: We need to drop the code that automatically opens the walkthrough on first-install.

Related to #3331

mattseddon:
Some next steps:

- Add a reference to the walkthrough to a new section on the page.
- Split CLI setup out from the Experiments section (add into new DVC section)
- Add a reminder of how to use DVCLive in the Setup Complete step of the Experiments section.
- Add some indicator as to whether sections are optional/completed to section titles.

julieg18:
> Add a reference to the walkthrough to a new section on the page.

Do we have an idea on what we want this new section to consist of? Did we want something simple like:

<img width="846" alt="image" src="https://user-images.githubusercontent.com/43496356/230152526-b1d1bfbc-280d-44ae-98c7-149570917aae.png">

Or would there be more to it? 

> Add a reminder of how to use DVCLive in the Setup Complete step of the Experiments section.

When we say "reminder", are we just think of adding an extra sentence with a doc link or did you want to add more content like our dvclive framework tabs?

mattseddon:
Note from retro: Setup should have a plots section that will help a user to get started with plots, show some snippets they can add to their `dvc.yaml` and also a button to create custom plots. Don't need to start on this until everything else is done. Could be a quick win before starting on remote setup.

julieg18:
[...]

PR #3435, opened 2023-03-09
Clean up setup internals

mattseddon:
Related to https://github.com/iterative/vscode-dvc/issues/3434

This PR starts to clean-up the internals of `Setup`. Comments are inline but the main change is removing instances of `DvcReader`, `DvcExecutor`, `GitReader` & `GitExecutor` from `Setup`. All of the required commands should be available through `InternalCommands` which makes things in the class a bit cleaner. 

I also moved command registration out of the class (usual pattern) as then we don't have to worry about commands being "re-registered"/throwing errors when we build `Setup` for testing purposes.


extension/src/extension.ts
```diff
@@ -131,6 +133,8 @@ export class Extension extends Disposable {
       new InternalCommands(outputChannel, ...clis)
     )
 
+    const status = this.dispose.track(new Status(config, ...clis))
```
mattseddon, line 136: [F] There was a bug in the way that we were creating `Status` it did not contain the `dvcViewer`.

extension/src/test/suite/setup/util.ts
```diff
@@ -1,124 +1,60 @@
...
-    mockRoot,
-    mockRunSetup: stub(Runner, 'run').resolves(undefined),
-    mockRunSetupWithRecheck: stub(Runner, 'runWithRecheck').resolves(undefined),
-    mockVersion
-  }
-}
-
 export const buildSetup = (
   disposer: Disposer,
-  dependencies?: {
```
mattseddon, line 91: [F] This was not used anywhere

extension/src/setup/index.ts
```diff
@@ -83,38 +75,23 @@ export class Setup
   private dotFolderWatcher?: Disposer
 
   constructor(
-    stopWatch: StopWatch,
     config: Config,
-    dvcExecutor: DvcExecutor,
-    dvcReader: DvcReader,
-    dvcRunner: DvcRunner,
-    gitExecutor: GitExecutor,
-    gitReader: GitReader,
```
mattseddon, line 92: [F] All of these `ICli`s sit behind the `InternalCommands` facade. No point to include the functionality twice.

extension/src/setup/index.ts
```diff
@@ -257,10 +241,50 @@ export class Setup
     })
   }
 
+  public async selectFocusedProjects() {
```
mattseddon, line 244: [F] exposed so it can be registered as a command

extension/src/setup/index.ts
```diff
@@ -310,13 +337,6 @@ export class Setup
     }
   }
 
-  private async initializeDvc() {
```
mattseddon, line 313: [F] Registered externally and used in that way so we get a "free" telemetry event

extension/src/test/cli/util.ts
```diff
@@ -36,7 +36,7 @@ export const initializeEmptyRepo = async (): Promise<string> => {
     return ''
   }
 
-  await gitExecutor.init(TEMP_DIR)
+  await gitExecutor.gitInit(TEMP_DIR)
```
mattseddon, line 39: [F] (unfortunate but) you can't register the same command name twice in `InternalCommands`

extension/src/test/suite/setup/util.ts
```diff
@@ -1,124 +1,60 @@
...
+import { Status } from '../../../status'
 
 export const TEMP_DIR = join(dvcDemoPath, 'temp-empty-watcher-dir')
 
-const buildSetupDependencies = (
-  disposer: Disposer,
-  mockDvcRoot: string | undefined,
-  mockGitRoot: string | undefined,
-  noGitCommits = true
-) => {
```
mattseddon, line 33: [F] Now that we are using `InternalCommands` we have to build the dependencies in the normal way.

extension/src/extension.ts
mattseddon: I'll add a type in a quick follow-up

extension/src/setup/index.ts
```diff
@@ -161,7 +136,16 @@ export class Setup
...
+      if (tryGlobalCli) {
+        return await this.internalCommands.executeCommand(
+          AvailableCommands.GLOBAL_VERSION,
+          cwd
+        )
+      }
+      return await this.internalCommands.executeCommand(
+        AvailableCommands.VERSION,
+        cwd
+      )
```
mattseddon, line 148: [F] Move the logic out of the "dumb" reader and into here. Wrote some tests to cover it.

julieg18 (approved):
Great work!

sroy3 (approved):
Works great and good cleanup job!

PR #3440, opened 2023-03-10
Create (shared) section container component

mattseddon:
# 1/2 `main` <- this <- #3441 

Related to #3434.

This PR creates a `SectionContainer` component from large sections of the `PlotsContainer`. I'll be using this for the new `Setup` page. This is a refactor so there should be no change in the UI.


webview/src/shared/components/sectionContainer/SectionContainer.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { Icon } from '../Icon'
+import { ChevronDown, ChevronRight, Info } from '../icons'
+import { isSelecting } from '../../../util/strings'
+import { isTooltip } from '../../../util/helpers'
+import { sendMessage } from '../../vscode'
+import Tooltip from '../tooltip/Tooltip'
+import { IconMenu } from '../iconMenu/IconMenu'
+import { IconMenuItemProps } from '../iconMenu/IconMenuItem'
+
+export const SectionDescription = {
```
mattseddon, line 14: [F] This needs to be here to avoid circular imports and to make the types available to `SectionDecription[section]`. I think that having all of the info sections in the same place will be beneficial for style as well. LMK if you disagree.

webview/src/shared/components/sectionContainer/styles.module.scss
```diff
@@ -0,0 +1,68 @@
+@import '../../variables.scss';
+@import '../../style.scss';
+
+$gap: 20px;
```
sroy3, line 4: It's used only once and it's not really a gap, I think it'll be better to just use the value or move the `$gap` variable to the `variables.scss` file.

sroy3 (approved):

PR #3443, opened 2023-03-10
Make the plot sizing sliders sticky

sroy3:
Part of #2585 

https://user-images.githubusercontent.com/3683420/224411102-9f9a5802-0296-4721-a834-de9b737e8250.mov


shcheklein:
I think sliders should be part of the header with the section name (is it sticky now?). We should be trying to optimize the space here. Same as we do with buttons that control the section.

mattseddon (approved):
This is an improvement. I agree with what Ivan said. The section header should be condensed and sticky.

Side note: Stories/snapshots in Chromatic look really strange with the slider moved to the LHS.

sroy3:
> I think sliders should be part of the header with the section name (is it sticky now?). We should be trying to optimize the space here. Same as we do with buttons that control the section.

They have to be separate as the sliders cannot be in the header when the section is collapsed. But they can behave the same way. I'll make the section header sticky as well in a follow-up.

sroy3:
> This is an improvement. I agree with what Ivan said. The section header should be condensed and sticky.
> 
> Side note: Stories/snapshots in Chromatic look really strange with the slider moved to the LHS.

Can't seem to figure out why. It looks ok in Storybook and the slider was on the left on main screenshots as well. I'll try to fix this when merging inside the vertical sizer as this requires the CI to test.

shcheklein:
> They have to be separate as the sliders cannot be in the header when the section is collapsed. But they can behave the same way. I'll make the section header sticky as well in a follow-up.

We can hide the controls when it's collapsed. I still think it's better to take less space. It's hard to justify 3 sticky rows (ribbon + header + controls).

sroy3:
> > They have to be separate as the sliders cannot be in the header when the section is collapsed. But they can behave the same way. I'll make the section header sticky as well in a follow-up.
> 
> We can hide the controls when it's collapsed. I still think it's better to take less space. It's hard to justify 3 sticky rows (ribbon + header + controls).

I'll try to move them inside the header while making the headers sticky.

PR #3441, opened 2023-03-10
Give Setup webview a single section

mattseddon:
# 2/2 `main` <- #3440 <- this

Related to #3434.

This PR uses the shared component created in #3440 to give the Setup webview a single section.

### Demo

https://user-images.githubusercontent.com/37993418/224225566-19dbce88-97f0-4286-b564-b044848e6a42.mov

Note: This looks odd now but there will be a section above for the walkthrough and one below for Studio. I will also condense the `EmptyState`s to `EmptySection`s (`isFullScreen=false`)

PR #3448, opened 2023-03-12
Give Setup webview a single section

mattseddon:
# 2/2 `main` <- #3440 <- this

This is exactly the same PR as #3441 which was merged in error. I have manually reverted that changed and replaced it with this PR.

This PR uses the shared component created in #3440 to give the Setup webview a single section.

### Demo

https://user-images.githubusercontent.com/37993418/224225566-19dbce88-97f0-4286-b564-b044848e6a42.mov

Note: This looks odd now but there will be a section above for the walkthrough and one below for Studio. I will also condense the `EmptyState`s to `EmptySection`s (`isFullScreen=false`)


julieg18 (approved):

extension/src/setup/webview/contract.ts
```diff
@@ -8,3 +8,13 @@ export type SetupData = {
   projectInitialized: boolean
   pythonBinPath: string | undefined
 }
+
+export enum Section {
```
sroy3, line 12: Should we make this enum name unique (as well as the one in plot)? Importing after typing only section might be confusing if the wrong `Section` is imported.

sroy3 (approved):

extension/src/setup/webview/contract.ts
mattseddon: I actually got hit by this already. I will update after #3452.

PR #3460, opened 2023-03-14
Clean up column path logic in custom state

julieg18:
* cleans up the metric/param logic used for our custom plots 

Resolves https://github.com/iterative/vscode-dvc/pull/3404#discussion_r1133905101

### DIFF
--- a/extension/src/plots/model/custom.ts
+++ b/extension/src/plots/model/custom.ts
@@ -1,4 +1,3 @@
-import { FILE_SEPARATOR } from '../../experiments/columns/paths'
 import { ColumnType } from '../../experiments/webview/contract'
 import { CheckpointPlot, CustomPlot, CustomPlotType } from '../webview/contract'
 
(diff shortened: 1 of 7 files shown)


extension/src/plots/model/custom.ts
```diff
@@ -23,3 +25,20 @@ export const doesCustomPlotAlreadyExist = (
   order.some(value => {
     return value.param === param && value.metric === metric
   })
+
+const removeColumnTypeFromPath = (columnPath: string, type: string) =>
+  columnPath.startsWith(type + FILE_SEPARATOR)
```
mattseddon, line 30: @julieg18 if you import anything from `../../experiments/columns/paths` here you expose the webview to 'path' so it won't build.

mattseddon: because you import `isCheckpointPlot` into the webview

PR #3466, opened 2023-03-14
Consolidate `collectCustomPlots`

julieg18:
# 2/3 `main` <= #3404 <= this <= #3491

* consolidates our code for collecting custom plots, simplifying the logic. 

Resolves https://github.com/iterative/vscode-dvc/pull/3404#discussion_r1135538551


extension/src/plots/model/collect.ts
```diff
@@ -67,256 +50,109 @@ import {
...
-  acc: CheckpointPlotAccumulator,
-  sha: string,
-  checkpointParent: string
-): number => {
-  const iteration = getLastIteration(acc, checkpointParent) + 1
-  acc.iterations[sha] = iteration
-  return iteration
-}
+  const group = exp.name || exp.label
+  const maxEpoch = exp.checkpoints.length + 1
```
julieg18, line 69: I just realized that using this way to get iterations breaks modified experiments. I need to understand a bit more on how iterations are originally gathered and see if it's even it's a good idea to get iterations through experiments... 

If not, I either need to see if we can gather both plots through output data or rethink how to consolidate when plot plots require two different kinds of data.

julieg18: After a discussion with Matt, we decided to keep the current implementation for now, since with a future change in the exp show output, it's likely that the current way of collecting iterations will no longer work anyway.

extension/src/test/fixtures/expShow/base/customPlots.ts
```diff
@@ -34,88 +34,158 @@ export const customPlotsOrderFixture: CustomPlotsOrderValue[] = [
...
-      { group: 'exp-83425', iteration: 4, y: 0.557449996471405 },
-      { group: 'exp-83425', iteration: 5, y: 0.5926499962806702 },
-      { group: 'exp-83425', iteration: 6, y: 0.5926499962806702 },
-      { group: 'test-branch', iteration: 1, y: 0.4083833396434784 },
-      { group: 'test-branch', iteration: 2, y: 0.4668000042438507 },
-      { group: 'test-branch', iteration: 3, y: 0.4668000042438507 },
-      { group: 'exp-e7a67', iteration: 1, y: 0.3723166584968567 },
-      { group: 'exp-e7a67', iteration: 2, y: 0.3724166750907898 },
-      { group: 'exp-e7a67', iteration: 3, y: 0.3724166750907898 }
+export const experimentsWithCheckpoints: ExperimentWithCheckpoints[] = [
```
julieg18, line 37: Do we have data similar to this in a different spot in the codebase? I couldn't find any and created one based off our current `CustomPlotFixture`.

extension/src/test/fixtures/expShow/base/customPlots.ts
```diff
@@ -178,18 +248,18 @@ const data: CustomPlotsData = {
...
-        { group: 'exp-83425', iteration: 2, y: 1.9329891204833984 },
-        { group: 'exp-83425', iteration: 3, y: 1.8798457384109497 },
-        { group: 'exp-83425', iteration: 4, y: 1.8261293172836304 },
-        { group: 'exp-83425', iteration: 5, y: 1.775016188621521 },
-        { group: 'exp-83425', iteration: 6, y: 1.775016188621521 },
-        { group: 'test-branch', iteration: 1, y: 1.9882521629333496 },
-        { group: 'test-branch', iteration: 2, y: 1.9293040037155151 },
-        { group: 'test-branch', iteration: 3, y: 1.9293040037155151 },
-        { group: 'exp-e7a67', iteration: 1, y: 2.020392894744873 },
+        { group: 'exp-e7a67', iteration: 3, y: 2.0205044746398926 },
```
julieg18, line 251: Due to the way `experimentWithCheckpoints` is formatted compared to `ExperimentOutput`, our checkpoint values are kind of "backwards" :) From my understanding, this change won't affect the rendered plots, but please correct me if I'm wrong!

extension/src/plots/model/collect.ts
```diff
@@ -67,256 +50,115 @@ import {
...
-      param: CHECKPOINTS_PARAM,
-      type: CustomPlotType.CHECKPOINT,
-      values: value
+const getCheckpointValues = (
+  experiments: ExperimentWithCheckpoints[],
+  metricPath: string
+): CheckpointPlotValues => {
+  const fullValues: CheckpointPlotValues = []
+  for (const experiment of experiments) {
+    if (experiment.checkpoints) {
```
mattseddon, line 95: [I] This should be a type guard

mattseddon: then no need to cast `as ExperimentWithDefinedCheckpoints`

extension/src/plots/model/index.ts
```diff
@@ -105,8 +102,8 @@ export class PlotsModel extends ModelWithPersistence {
     this.customPlotsOrder = this.revive(PersistenceKey.PLOTS_CUSTOM_ORDER, [])
   }
 
-  public transformAndSetExperiments(data: ExperimentsOutput) {
-    this.recreateCustomPlots(data)
+  public transformAndSetExperiments() {
```
mattseddon, line 105: [I] As discussed on the call this morning.

I think this can now be dropped and `recreateCustomPlots` can be renamed/called instead of `getCustomPlots`. It will get called from `sendWebviewMessage` and it can be passed the selected revisions so that we only need to process the required experiments. Instead of processing them all and then filtering them afterwards.

extension/src/plots/model/index.ts
```diff
@@ -171,20 +168,20 @@ export class PlotsModel extends ModelWithPersistence {
...
-      this.customCheckpointPlots = collectCustomCheckpointPlots(data)
-    }
-
-    const experiments = this.experiments.getExperiments()
+  public recreateCustomPlots() {
+    const experiments = this.experiments.hasCheckpoints()
+      ? this.experiments
+          .getExperimentsWithCheckpoints()
+          .filter(({ checkpoints }) => !!checkpoints)
+      : this.experiments.getExperiments()
```
mattseddon, line 176: [B] `collectCustomPlots` should take all experiments regardless of whether or not they have checkpoints. I think the best thing to do here would be call `this.experiments.getExperimentsWithCheckpoints()` all the time as the checkpoints part of `collectCustomPlots` will have a type guard in it anyway?

extension/src/plots/webview/contract.ts
```diff
@@ -89,21 +89,21 @@ export type CheckpointPlotValues = {
...
+type CustomPlotBase = {
   id: string
-  values: CheckpointPlotValues
   metric: string
   param: string
-  type: CustomPlotType.CHECKPOINT
 }
 
+export type CheckpointPlot = {
+  values: CheckpointPlotValues
```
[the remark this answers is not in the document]
mattseddon: I think this is fine. We've limited the different logic to as fewer places as possible.

extension/src/test/fixtures/expShow/base/customPlots.ts
mattseddon: nothing significant enough

extension/src/plots/model/collect.ts
mattseddon: Asked here: https://github.com/iterative/dvc/pull/9170#issuecomment-1473001613

extension/src/plots/model/collect.test.ts
```diff
@@ -215,50 +172,6 @@ describe('collectData', () => {
...
 })
 
-describe('collectCustomCheckpointPlotsData', () => {
-  it('should return the expected data from the test fixture', () => {
-    const data = collectCustomCheckpointPlots(expShowFixture)
-
-    expect(data).toStrictEqual(checkpointPlotsFixture)
-  })
-
-  it('should provide a continuous series for a modified experiment', () => {
```
mattseddon, line 225: [F] This logic has been removed on purpose. Change is being discussed in `[WIP] exp: refactor show behavior` PR. Link is further down.

extension/src/plots/model/collect.ts
```diff
@@ -67,256 +50,115 @@ import {
...
-    param,
-    type: CustomPlotType.METRIC_VS_PARAM,
-    values: []
-  }
+const getMetricVsParamValues = (
+  experiments: ExperimentWithCheckpoints[],
+  metricPath: string,
+  paramPath: string
+): MetricVsParamPlotValues => {
+  const fullValues: MetricVsParamPlotValues = []
```
mattseddon, line 111: [N] Why not `values`?

mattseddon (approved):

extension/src/plots/model/collect.ts
mattseddon: Answer here: https://github.com/iterative/dvc/pull/9170#issuecomment-1473029067

extension/src/plots/model/index.ts
julieg18: Now that I think about it, there is definitely an easier way to filter out experiment commits. Will update how we filter for now!

sroy3 (approved):

extension/src/plots/model/index.ts
julieg18: > Now that I think about it, there is definitely an easier way to filter out experiment commits. Will update how we filter for now!

Originally was thinking I could filter by `type` or `commit` but looks like either of those could be unavailable. Ended up just renaming the variable to make things more clear.

extension/src/plots/model/index.ts
julieg18: Since this pr already has approval and all other comments have been addressed, I completed this change in followup pr, #3491

PR #3467, opened 2023-03-14
Open comparison table images when clicking on them

sroy3:
Part of #2585 

https://user-images.githubusercontent.com/3683420/225128467-f0eb8193-47f3-4395-a558-7f98901e2f71.mov


extension/src/fileSystem/index.ts
```diff
@@ -141,6 +141,9 @@ export const openFileInEditor = async (filePath: string) => {
   return document
 }
 
+export const openImageFileInEditor = async (imagePath: string) =>
+  await commands.executeCommand('vscode.open', Uri.file(imagePath))
```
mattseddon, line 145: [Q] Should this be open to the side? (`explorer.openToSide`). Definitely not a statement. Not sure of the best UX.

extension/src/plots/webview/messages.ts
```diff
@@ -411,6 +414,7 @@ export class WebviewMessages {
     if (webview) {
       return {
         ...plot,
+        resourceUrl: plot.url,
```
mattseddon, line 417: [Q] Can we encode/decode the url instead of duplicating? 

Note: might not be worth the effort.

extension/src/plots/webview/messages.ts
```diff
@@ -112,6 +112,9 @@ export class WebviewMessages {
       case MessageFromWebviewType.TOGGLE_EXPERIMENT:
         return this.setExperimentStatus(message.payload)
       case MessageFromWebviewType.ZOOM_PLOT:
+        if (message.payload) {
+          void openImageFileInEditor(message.payload)
+        }
```
mattseddon, line 117: [Q] Can we add `isImage` or similar to the telemetry event? This would mean we could quantify the split between the usage of each feature.

extension/src/plots/webview/messages.ts
mattseddon: the only place we would have to decode would be directly before `openImageFileInEditor`.

mattseddon (approved):

mattseddon:
Note: Demo doesn't show it but the image can be zoomed further once opened in the editor.

extension/src/plots/webview/messages.ts
sroy3: I'lll try something, but I know I hard a really hard time trying to decode it, that's why I decided it might be simpler to just pass a new value, but with the amount of changes in the tests, it might be worth trying again.

extension/src/fileSystem/index.ts
sroy3: I thought it might be a little small for images on a 13" laptop, but it does look good. I like the fact that you don't lose context when opening to the side, you can also open multiple images without going back and forth. I'll make the change.

extension/src/test/suite/plots/index.test.ts
```diff
@@ -529,11 +530,53 @@ suite('Plots Test Suite', () => {
...
+        { isImage: true },
         undefined
       )
     }).timeout(WEBVIEW_TEST_TIMEOUT)
 
+    it('should open an image when receiving a plot zoomed message from the webview with a payload', async () => {
+      const { plots } = await buildPlots(disposable, plotsDiffFixture)
+
+      const webview = await plots.showWebview()
+      const imagePath = 'some/path/image.jpg'
```
mattseddon, line 563: [I] Mangle this with `const url = webview.getWebviewUri(imagePath)` to test the transformation.

mattseddon (approved):

julieg18 (approved):
Great work!

PR #3470, opened 2023-03-15
Rename section enums to differentiate types

mattseddon:
Follow up from https://github.com/iterative/vscode-dvc/pull/3448#discussion_r1133880297

Note: This is an easy change to make but could be difficult to integrate with concurrent work. Happy to delay until there is less moving around in plots.


julieg18 (approved):
Looks good to me!

julieg18:
> For my part, it's really easy to find and fix during conflicts. For @julieg18's sake though, delaying might not be a bad idea.

I don't think this change will be that hard to adjust to since it's just renaming a variable, so no preference on whether we choose to delay or not :)

PR #3477, opened 2023-03-16
Bump min DVC version to 2.52.0 (plots errors)

mattseddon:
# 1/4 `main` <- this <- #3520 <- #3522 <- #3532

This PR accommodates the breaking part of the change in https://github.com/iterative/dvc/pull/9146.

TODOs: 

- [x] Confirm `errors` API changes give everything needed.
- [x] Coordinate DVC release
- [x] Fix `extension/src/test/cli/plotsDiff.test.ts` after change is merged to DVC `main`. (#3558)
- [x] Bump`MIN_CLI_VERSION` to the appropriate version once the related DVC PR is merged/released.
- [x] Reset `vscode-dvc-demo` to the appropriate commit and delete branch


extension/src/test/suite/experiments/model/tree.test.ts
```diff
@@ -114,21 +109,15 @@ suite('Experiments Tree Test Suite', () => {
 
       expect(
         messageSpy,
-        'when there are no experiments selected we dont send checkpoint type plots'
+        "when there are no experiments selected we don't send checkpoint trend plots"
```
mattseddon, line 112: [F] By default we now give the plots diff fixture instead of nothing. The test only needs to check the match assumption

skshetry:
> Confirm errors API changes give everything needed.

Hi, @mattseddon. Are you able to confirm if `errors` give you everything?

mattseddon:
> > Confirm errors API changes give everything needed.
> 
> Hi, @mattseddon. Are you able to confirm if `errors` give you everything?

It looks very likely. I am working through the last couple of parts now.

The one thing that was missing (which I am not sure if you will be able to provide), is the plot type for each error ('image' or 'vega'). Can you add that in?

skshetry:
> The one thing that was missing (which I am not sure if you will be able to provide), is the plot type for each error ('image' or 'vega'). Can you add that in?

Is `plot_id` not enough to figure that out? I am trying to think of when this can happen. The only one that I can think of is when the plot images are missing.

We choose between vega and image based on file extensions. So:
1. either the images' directory is missing, and we don't know what contents are (which makes it fallback to vega plots). 
2. or, a image file is missing, which will still be considered as image plot but won't have any data.

In both of these cases, there will be no `data` and hence no `type` in the json output.
We can say type is "image" for 2nd kind of error, but 1st one is still misleading.

skshetry:
But this is of course in case where there is no plots' data at all. In those cases, do you need to differentiate if it's a vega or an image plot?

mattseddon:
> But this is of course in case where there is no plots' data at all. In those cases, do you need to differentiate if it's a vega or an image plot?

When all we have is errors for a plot then we don't know the type. If we don't know the type then we don't know what section it should be displayed in. If it should be an image then we can't provide a row in the image comparison table. It's an edge case and we can live without the data. It's ok not to worry about it.

🙏🏻

skshetry:
@mattseddon, what do you think of supporting older dvc versions? Do you think that's possible?

mattseddon:
> @mattseddon, what do you think of supporting older dvc versions? Do you think that's possible?

If we had a lot more users and there were legitimate reasons that they could not update the CLI then I would consider attempting to make things backwards compatible but at the moment it is not worth the effort IMO.

mattseddon:
Give me a heads-up when you're going to make the release and I'll make the corresponding extension release 🙏🏻.

skshetry:
Okay, we’ll merge and release early next week, either Mon or Tue.

sroy3 (approved):

julieg18 (approved):

PR #3480, opened 2023-03-16
Resize comparison table with slider

sroy3:
Part of #2585 

https://user-images.githubusercontent.com/3683420/225697074-cd1feebc-e5df-4245-a9ec-9888245196a0.mov


sroy3:
Sorry for the renaming, it just made more sense that way. The main part of the change is found in `ComparisonTableWrapper.tsx`, `ComparisonTable.tsx`, `PlotsContainer.tsx` and `plots/components/styles.modules.scss`.

sroy3:
Won't be necessary after all. I'll revert the change in the submodule.

mattseddon:
Sticky slider collides with comparison header:

image

Can the slider/ribbon fill the space on the RHS?

image

Will both of the above be fixed by the section having its own scrollbar?

extension/src/plots/model/index.ts
```diff
@@ -56,7 +56,7 @@ export type CustomPlotsOrderValue = { metric: string; param: string }
 export class PlotsModel extends ModelWithPersistence {
   private readonly experiments: Experiments
 
-  private nbItemsPerRow: Record<PlotsSection, number>
+  private nbItemsPerRowOrWidth: Record<PlotsSection, number>
```
mattseddon, line 59: [Q] Do these need to be stored together?

mattseddon (approved):
Change looks fine outside of previously mentioned issue

extension/src/plots/model/index.ts
sroy3: Was debating this, but they function the exact same way it's just that one cannot be called `nbItemsPerRow`.  Separating them would probably mean checking whether it has one property or another multiple times.

julieg18 (approved):

sroy3:
> Sticky slider collides with comparison header:
> 
> image
> 
> Can the slider/ribbon fill the space on the RHS?
> 
> image
> 
> Will both of the above be fixed by the section having its own scrollbar?
> 
> Edit: With pinned column
> 
> image

The sticky header colliding is easy to fix now. Same for the pinned column. The header taking the full width will be easier with the local scrollbar (next item on my list), else we'll end up with sliders past smaller sections.

PR #3481, opened 2023-03-16
Move e2e tests to insiders

sroy3:
Not the source of the bug, but might the version seems to work now, so better go back to it.


mattseddon (approved):
Up to you on whether to merge or close

PR #3482, opened 2023-03-16
Revert dvc-render in the demo project

sroy3:
Fix for e2e tests and plots inside the demo project


sroy3:
Won't be necessary after all. I'll revert the submodule.

PR #3492, opened 2023-03-17
Calculate the default nbOrItemsPerRow from the screen size

sroy3:
Part of #2585 

https://user-images.githubusercontent.com/3683420/225992923-30eecad9-17bb-4ecc-ab8f-41c91955fe06.mov

https://user-images.githubusercontent.com/3683420/225992986-eb20622f-de76-4f35-90fb-6fbfebea107b.mov

Both recorder after reseting the state

### DIFF
--- a/extension/src/plots/vega/util.ts
+++ b/extension/src/plots/vega/util.ts
@@ -26,7 +26,7 @@ import {
 } from 'vega-lite/build/src/spec/repeat'
 import { TopLevelUnitSpec } from 'vega-lite/build/src/spec/unit'
 import isEqual from 'lodash.isequal'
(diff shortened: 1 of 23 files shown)


extension/src/plots/vega/util.ts
```diff
@@ -255,16 +255,16 @@ const truncateTitle = (
 
 export const truncateVerticalTitle = (
   title: Text | Title,
-  width: number,
-  height: number
+  width = 2,
+  height = 2
```
sroy3, line 259: It was great having this on the extension side, but since we rely so much on screen size now. I think it'll be best to move this back on the webview side.

PR #3499, opened 2023-03-18
Add experiment_rev into payload for share experiment to Studio

### LINKED ISSUE #3486: Share to Studio: Include `experiment_rev` field.
The post request doesn't include the [`experiment_rev` ](https://github.com/iterative/dvc-studio-client/blob/485b2d8baf056882f6c7cb59438208390a045998/src/dvc_studio_client/post_live_metrics.py#L88-L90) field:

https://github.com/iterative/vscode-dvc/blob/3c421346f0b699808a759e48bdeedfb8ad208eab/extension/src/patch.ts#L125-L132

Which is used by Studio for rendering (and, in the future, for matching against parsed experiment revisions):

The 2 experiments only displaying the name are from VSCode `Share to Studio` action:

Captura de pantalla 2023-03-17 a las 13 23 46

The `experiment_rev` should be the **full** SHA

mattseddon:
Fixes #3486

### Demo

https://user-images.githubusercontent.com/37993418/226081257-09aab265-cbbd-4c49-ae4c-a0ad31728d8a.mov

### DIFF
--- a/extension/src/patch.ts
+++ b/extension/src/patch.ts
@@ -19,6 +19,7 @@ export const STUDIO_ENDPOINT = 'https://studio.iterative.ai/api/live'
 
 type ExperimentDetails = {
   baselineSha: string
+  sha: string
   metrics: ValueTreeRoot | undefined
   name: string
   params: ValueTreeRoot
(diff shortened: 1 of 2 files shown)


shcheklein (approved):
Thanks @mattseddon !

Issue #3430, opened 2023-03-08
Unify checkbox, star, and radio buttons in the table


dberenbaum:
There's a background discussion in https://iterativeai.slack.com/archives/C01APS0FHDM/p1676329398900699.

I get confused by when to use each of these:

<img width="227" alt="Screenshot 2023-02-13 at 6 01 18 PM" src="https://user-images.githubusercontent.com/2308172/223851112-3ab5f049-6bf9-4319-9344-271ceda7a51f.png">

I especially find it confusing because the checkbox actions for multiple experiments only impact the other two columns:

![image](https://user-images.githubusercontent.com/2308172/223851590-be8dd73c-bdfc-4693-8720-14a9a4bfb425.png)

For example, if I go to star some experiments to compare, I can't then plot them automatically:

https://user-images.githubusercontent.com/2308172/223852141-233e54cf-c216-49e6-88bf-b55aa0e14f52.mov

shcheklein:
> For example, if I go to star some experiments to compare, I can't then plot them automatically:

Q: what would you expect for this? A separate action, like plot all starred as part of the Add experiment drop down?

------

There is some unification happening in Studio that is related to this. Checkboxes will behave like they do in all the other products (including MLFlow, for example) - pick items to do some mass action (vs showing plots by default). That alone should reduce the confusion.

Also, there will an eye icon (that is srandard for W&B, MLFlow, comet, neptune, etc) to toggle the plots visibility. 

I wish we could have adopted an eye icon as well here, but it's a limitation on the VS Code side, @mattseddon can add more color to this.

Stars - a bit different story. I'm trying to understand the confusion better - do you feel for example that stars are responsible for plots? If we change a circle to an icon - would that help or not? If we add quick actions as we discussed - would that help to clarify things or not?

mattseddon:
> I wish we could have adopted an eye icon as well here, but it's a limitation on the VS Code side, @mattseddon can add more color to this.

Tl;dr is that we can only have 1 icon per row in the experiments tree. If we want to blend a spinner + selected/unselected for plotting then it needs to be a circle. Unless there is some smarter option that we have missed or we want to add a custom webview to the sidebar. 

For more background, you can start here: https://github.com/iterative/vscode-dvc/issues/712

Note: Adding a webview opens up a different set of problems (matching styling/custom context menu etc) but it can be done.

mattseddon:
[...]

PR #3500, opened 2023-03-19
Add template for epic/story issues

omesser:
This might help reduce some friction on opening story/epic issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/epic-or-story.md
+++ b/.github/ISSUE_TEMPLATE/epic-or-story.md
@@ -1,0 +1,60 @@
+---
+name: Epic/Story Issue Template
+about: Template for "top" level issues - Epics (>2 weeks) / Stories (<2 weeks)
+title: 'Epic: New Feature'
+labels: epic
+assignees:
+
+---
+
+## Summary / Background
+What do you want to achieve, why? business context
+...
+
+## Scope
+
+What will be impacted and what won't be?
+What needs implementation and what is invariant?
+e.g.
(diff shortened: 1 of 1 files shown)


mattseddon:
We can discuss this in either team planning or the retro this week.

julieg18 (commented):
> Only thing I ask is that you keep it true to the original purpose - it should help share information, make stories/epics "familiar" and remind people about key pieces of information we usually care about and talk about

When you say people, are you referring to the VSCode team or the company as a whole? Personally, I'm not sure if the VSCode team needs such a template.

mattseddon:
My $0.02.

The template doesn't really fit our workflow. How we generally work is:

1. Ivan, Dave, David, etc will raise a new issue (e.g #3539, #3430, #1966).
2. We discuss that issue in the ticket.
3. After minimal discussion we assign a priority.
4. We decide on the next steps either in the ticket or in the next planning meeting. We generally discuss these steps in the planning meeting anyway to make sure we are on heading in the correct direction.
5. These steps are generally kept small. Once the step is completed we iterate and decide on what is next.

Sometimes we assign a priority to an issue before discussing it, sometimes we will continue to cut scope/refine the issue after we've started the work, and the order of these steps is not set in stone.

The idea in working this way is to avoid having to create stories/epics and we work on the next smallest chunk of work possible. We can also drop a larger piece of work when we finish a chunk as priorities change all the time. Even though they change often we are always trying to work towards the Qs goals and the most impactful change. 

----

It also feels like we are trying to solve a communication problem that we don't currently have.

PR #3503, opened 2023-03-20
Always call plots diff with workspace as the first revision (get correct templates)

mattseddon:
Following up on the ongoing discussion in https://github.com/iterative/dvc/issues/9025. 

There will be an edge case where we end up with the wrong templates for all of the displayed plots because the templates in a revision differ from the ones in the workspace. In order to combat this (for now) I've ensured that the `workspace` revision is always provided when we call `plots diff`.

Our approach may change further depending on the decisions made in https://github.com/iterative/dvc/issues/9025.

I will separate out the need for a "full update" when any `dvc.yaml` is updated into a different PR 👍🏻.


extension/src/util/array.ts
```diff
@@ -19,7 +19,7 @@ export const joinTruthyItems = (array: (string | undefined)[], sep = ' ') =>
 export const sameContents = (
   array: (null | string | number | undefined)[],
   otherArray: (null | string | number | undefined)[]
-) => isEqual(array.sort(), otherArray.sort())
+) => isEqual([...array].sort(), [...otherArray].sort())
```
mattseddon, line 22: [F] This was an interesting bug to nail down... by interesting I mean confusing and frustrating.

mattseddon: I'll add `etc/no-assign-mutated-array` once this is merged.

sroy3 (approved):

PR #3504, opened 2023-03-20
Make plots tree items decoratable

mattseddon:
Related to https://github.com/iterative/vscode-dvc/issues/1649.

This is a step that needs to be taken in order to display errors in the plots tree. It can be done now while the discussion continues.


extension/src/plots/paths/tree.ts
```diff
@@ -32,7 +34,19 @@ export class PlotsPathsTree extends BasePathSelectionTree<WorkspacePlots> {
...
+  protected getBaseTreeItem({
+    dvcRoot,
+    path,
+    collapsibleState
+  }: PathSelectionItem) {
+    const resourceUri = getDecoratableUri(
+      join(dvcRoot, path),
+      DecoratableTreeItemScheme.PLOTS
+    )
+    return new TreeItem(resourceUri, collapsibleState)
```
mattseddon, line 46: [F] By providing a "decoratable" Uri here with the `dvc.plots` scheme we won't run into the SCM decorations and we'll be able to provide a color when there is an error for that plots. See `extension/src/experiments/model/decorationProvider.ts` for a working example. 

The same approach does not work for the experiments column tree as the path can be `:` separated. Hence they have been left separate.

julieg18 (approved):
Looks good!

PR #3507, opened 2023-03-20
Fix the comparison table images width

sroy3:
Fixes the pinned column transparent background bug

https://user-images.githubusercontent.com/3683420/226371408-94713a21-7c6c-4290-b675-65591df5d4a0.mov


mattseddon (approved):

julieg18 (approved):
Looks good!

PR #3513, opened 2023-03-20
Increase coverage

sroy3:
A drop in the sea, but some quick wins to increase the coverage. Will try to increase more in upcoming PRs.


webview/setup-tests.js
sroy3, line 1: Took me way too long to realize why `getThemeValue` wasn't called unless I changed its name.

webview/jest.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   collectCoverageFrom: [
     'src/**/*.{ts,tsx}',
     '!**/*.test.*',
-    '!src/stories/*',
+    '!src/stories/**/*.{ts,tsx}',
```
sroy3, line 7: Correctly removes the `IconWrapper` and `IconWrappers` components from the coverage.

mattseddon:
We can drop the overall coverage requirement by a few percentage points but we need to get back on top of this.

mattseddon:
I guess if we are aiming for 95% the pass/fail for coverage should be 95% on a PR. Need to discuss what's meaningful this week in the retro.

sroy3:
> I guess if we are aiming for 95% the pass/fail for coverage should be 95% on a PR. Need to discuss what's meaningful this week in the retro.

That make sense.

mattseddon (approved):

PR #3514, opened 2023-03-20
Remove error for invalid dvc.yaml if no dvc.yaml file

### LINKED ISSUE #3511: "Add a pipeline stage" doesn't work with no `dvc.yaml`
For a project without any `dvc.yaml`, the "Add a pipeline stage" button is disabled and warns that the `dvc.yaml` file must contain valid yaml:

Screenshot 2023-03-20 at 1 41 08 PM

Is this expected? I thought this was the basic scenario where this button would be most useful.

sroy3:
Fixes #3511 

https://user-images.githubusercontent.com/3683420/226462424-f66d546a-24b4-42b6-ae05-936423236a0c.mov


mattseddon (approved):

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -187,6 +187,96 @@ suite('Experiments Test Suite', () => {
...
+        hasColumns: true,
+        hasConfig: false,
+        hasRunningExperiment: true,
+        hasValidDvcYaml: true,
+        rows: rowsFixture,
+        sorts: []
+      }
+
+      expect(messageSpy).to.be.calledWithExactly(expectedTableData)
+    }).timeout(WEBVIEW_TEST_TIMEOUT)
```
mattseddon, line 248: [F] This one seems like a genuine cross-platform failure

julieg18 (approved):
I think it's unrelated to the pr, but shouldn't we be able to create a dvc.yaml if a user doesn't have one? 

https://user-images.githubusercontent.com/43496356/226609237-07761271-e611-4dd7-9100-4ae83d111a57.mov

sroy3:
> I think it's unrelated to the pr, but shouldn't we be able to create a dvc.yaml if a user doesn't have one?

Yes thanks, there was definitely a regression there.

extension/src/test/suite/experiments/index.test.ts
sroy3: Nope, just me using `resolves` for the stub instead or `returns`. I had an `await` for it that I removed after the lint stage before pushing my changes, so local tests passed, but they were not doing what they should have and were changed before committing.

sroy3:
> > I think it's unrelated to the pr, but shouldn't we be able to create a dvc.yaml if a user doesn't have one?
> 
> Yes thanks, there was definitely a regression there.

Fixed

sroy3:
`ERROR: Coverage for lines (76.85%) does not meet global threshold (90%)` Not sure how it got to that point.

PR #3515, opened 2023-03-20
Drop overall test coverage requirement (unblock)

mattseddon:
Doing this to unblock us for now.

3 things we need to discuss in the retro, paying attention to:

1. Test coverage
2. Codeclimate
3. Chromatic

For reference: https://github.com/iterative/vscode-dvc/pull/890
The 95% coverage check was added to prove that the integration test suite had run. This may be a relic but shows that we have started a skid. At one point coverage was ~98%.


demo
sroy3, line 1: Is that change wanted?

sroy3 (approved):

demo
mattseddon: wtf

julieg18 (approved):

PR #3518, opened 2023-03-20
Cleanup plots workspace state

julieg18:
* deletes no longer used trends state
* updates old version of custom state

https://user-images.githubusercontent.com/43496356/226488452-3dd64a51-933b-4008-a837-fa391c64f954.mov

Part of #3373

### DIFF
--- a/extension/src/plots/model/index.test.ts
+++ b/extension/src/plots/model/index.test.ts
@@ -1,5 +1,6 @@
 import { PlotsModel } from '.'
 import {
+  CustomPlotType,
   DEFAULT_NB_ITEMS_PER_ROW,
   DEFAULT_SECTION_COLLAPSED,
   DEFAULT_SECTION_NB_ITEMS_PER_ROW_OR_WIDTH,
@@ -22,7 +23,7 @@ const mockedRevisions = [
 describe('plotsModel', () => {
   let model: PlotsModel
   const exampleDvcRoot = 'test'
-  const memento = buildMockMemento({
+  let memento = buildMockMemento({
     [PersistenceKey.PLOTS_CUSTOM_ORDER + exampleDvcRoot]:
       customPlotsOrderFixture,
     [PersistenceKey.PLOT_NB_ITEMS_PER_ROW_OR_WIDTH + exampleDvcRoot]:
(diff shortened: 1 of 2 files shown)


extension/src/plots/model/index.ts
```diff
@@ -91,6 +91,9 @@ export class PlotsModel extends ModelWithPersistence {
     )
     this.comparisonOrder = this.revive(PersistenceKey.PLOT_COMPARISON_ORDER, [])
     this.customPlotsOrder = this.revive(PersistenceKey.PLOTS_CUSTOM_ORDER, [])
+
+    this.cleanupOutdatedCustomPlotsState()
```
julieg18, line 95: How long should we wait before we remove these functions and delete the metrics persistence keys?

extension/src/plots/model/index.test.ts
```diff
@@ -45,6 +46,33 @@ describe('plotsModel', () => {
...
+    model = new PlotsModel(
+      exampleDvcRoot,
+      {
+        getFirstThreeColumnOrder: mockedGetFirstThreeColumnOrder,
+        getSelectedRevisions: mockedGetSelectedRevisions,
+        isReady: () => Promise.resolve(undefined)
+      } as unknown as Experiments,
+      memento
+    )
+    expect(model.getCustomPlotsOrder()).toStrictEqual([
```
julieg18, line 67: Couldn't really think of a simple way to test if trends state was deleted 🤔

sroy3: Instead of coupling this with the plots model data. Maybe we should add a function in `extension/persistence/util.ts` called `removedOutdatedState` with outdated keys in there. This could be called on extension load. This would be easier to test (similar to the `resetPersistedState` tests already there).

extension/src/plots/model/index.ts
sroy3: @mattseddon was the one to make the comment, so I'll leave the verdict to him. But do we even need to migrate the data? I'd just delete it and start fresh.

sroy3 (approved):

extension/src/plots/model/index.test.ts
sroy3: I commented before really reading what the whole PR does, sorry. Might be a little harder to do with what we are trying to achieve at the moment.

sroy3: You can use `memento.get('key')` to see what's actually in the state.

extension/src/plots/model/index.ts
julieg18: > @mattseddon was the one to make the comment, so I'll leave the verdict to him. But do we even need to migrate the data? I'd just delete it and start fresh.

Personally, I vote we migrate. It's not that difficult to do and it ensures users who have already created custom plots doesn't lose anything saved when they get the latest release. Though, on the other hand, I doubt that many users have created custom plots so no strong preference 🤔

mattseddon: Up to you. Please set a reminder for one month from now to turn this off.

mattseddon (approved):

Issue #2277, opened 2022-08-28
Image plots show proper message and disable button if image is not available


shcheklein:
<img width="773" alt="Screen Shot 2022-08-27 at 9 00 02 PM" src="https://user-images.githubusercontent.com/3659196/187056842-6956b726-ed90-449b-8a2a-4b702f7ffd5f.png">

This image doesn't exist at in the experiment, refresh doesn't make much sense here.

Might depend on plots error handing cc @mattseddon @pared

mattseddon:
Depends on https://github.com/iterative/dvc/issues/7692 & https://github.com/iterative/vscode-dvc/issues/1649.

shcheklein:
@shcheklein revisit this, try to reproduce.

We still need error messages, or a signal that plot doesn't exist in the commit.

mattseddon:
Can we treat the directory more like a dataset and each image more like a datapoint?

skshetry:
`dvc plots diff` does not know if there's an experiment running, and I don't think it should care. Displaying errors (or not) is contextual, so I think it's up to the extension to interpret.

dberenbaum:
@skshetry I don't think it's related to whether an experiment is running. One revision contains the file `mispredicted/6599-...` and the other does not.

@mattseddon Is there an error returned here? I don't see one when testing this. I think we just need to show a different message here than 'No plot to display: refresh` (something like N/A or no image at this revision). I see an output like this:

```json
{
  "dir/img1.png": [
    {
      "type": "image",
      "revisions": [
        "workspace"
      ],
      "url": "/Users/dave/repo/dvc_plots/workspace_dir_img1.png"
    }
  ],
  "dir/img2.png": [
    {
      "type": "image",
      "revisions": [
        "workspace"
      ],
      "url": "/Users/dave/repo/dvc_plots/workspace_dir_img2.png"
    },
    {
      "type": "image",
      "revisions": [
        "main"
      ],
      "url": "/Users/dave/repo/dvc_plots/img1.png"
    }
  ]
}
```

@shcheklein What do you want to show in this scenario?

Some thoughts on what I'd expect:
* Grouping by directory paths. Seeing all cats, dogs, muffins, etc. together would help organize it.
* Showing the number of images per revision per directory. At least then I'd easily know whether there are more misclassified dogs or cats in each revision.

mattseddon:
[...]

PR #3520, opened 2023-03-21
Use plot errors to display correct messages for missing plots

mattseddon:
# 2/4 `main` <- #3477 <- this <- #3522 <- #3532

Part of https://github.com/iterative/vscode-dvc/issues/2277

### Screenshot

image

with tooltip shown

image

I tried to make a `!` work to fit in with the errors shown in the experiments table but it did not look very good.

Note: If an image is completely broken/missing for all revisions then we do not know that it is an image so we won't be able to display it like this. We will still be able to add an error indicator into the tree for this.

Previous idea:

Screenshot 2023-03-21 at 1 08 52 pm

### DIFF
--- a/extension/src/plots/errors/collect.ts
+++ b/extension/src/plots/errors/collect.ts
@@ -1,0 +1,53 @@
+import { PlotError, PlotsOutput } from '../../cli/dvc/contract'
+
+export const collectErrors = (
+  data: PlotsOutput,
+  revs: string[],
+  errors: PlotError[],
+  cliIdToLabel: { [id: string]: string }
+) => {
+  const existingErrors = errors.filter(
+    ({ rev }) => !revs.includes(cliIdToLabel[rev] || rev)
+  )
+  const newErrors = data?.errors || []
+
+  return [
+    ...existingErrors,
+    ...newErrors.map(error => {
+      const { rev } = error
+      return {
+        ...error,
+        rev: cliIdToLabel[rev] || rev
+      }
+    })
+  ]
+}
+
+const getMessage = (error: PlotError): string => {
+  const { msg, type, source } = error
+
+  if (type === 'FileNotFoundError' && source && !msg) {
+    return `${type}: ${source} not found.`
+  }
+  return [type, msg].filter(Boolean).join(': ')
+}
+
+export const collectImageErrors = (
+  path: string,
+  revision: string,
+  errors: PlotError[]
+): string | undefined => {
+  const msgs: string[] = []
+  for (const error of errors) {
+    if (error.rev === revision && error.name === path) {
+      const msg = getMessage(error)
+      msgs.push(msg)
+    }
+  }
+
+  if (msgs.length === 0) {
+    return undefined
+  }
+
+  return msgs.join('\n')
+}
--- a/webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
+++ b/webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
@@ -5,6 +5,8 @@ import styles from './styles.module.scss'
 import { RefreshButton } from '../../../shared/components/button/RefreshButton'
 import { sendMessage } from '../../../shared/vscode'
 import { zoomPlot } from '../messages'
+import { Error } from '../../../shared/components/icons'
+import { ErrorTooltip } from '../../../shared/components/errorTooltip/ErrorTooltip'
 
 type ComparisonTableCellProps = {
   path: string
@@ -15,6 +17,7 @@ export const ComparisonTableCell: React.FC<ComparisonTableCellProps> = ({
   path,
   plot
 }) => {
+  const error = plot?.error
   const missing = plot?.fetched && !plot?.url
 
   if (!plot?.fetched) {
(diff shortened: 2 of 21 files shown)


webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
```diff
@@ -28,15 +31,25 @@ export const ComparisonTableCell: React.FC<ComparisonTableCellProps> = ({
...
-              payload: plot.revision,
-              type: MessageFromWebviewType.REFRESH_REVISION
-            })
-          }
-        />
+        {error ? (
+          <>
+            <ErrorTooltip error={error}>
+              <div>
+                <Error height={48} width={48} className={styles.errorIcon} />
```
mattseddon, line 38: [F] Tried an `$error-color` `!` as well, couldn't make it work.

webview/src/shared/components/errorTooltip/ErrorTooltip.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { ReactElement } from 'react'
 import styles from './styles.module.scss'
-import { Error } from '../../../shared/components/icons'
-import Tooltip from '../../../shared/components/tooltip/Tooltip'
+import { Error } from '../icons'
+import Tooltip from '../tooltip/Tooltip'
```
mattseddon, line 4: [Q] Should this component be grouped under `tooltip`?

webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
```diff
@@ -28,15 +31,25 @@ export const ComparisonTableCell: React.FC<ComparisonTableCellProps> = ({
...
+              onClick={() =>
+                sendMessage({
+                  payload: plot.revision,
+                  type: MessageFromWebviewType.REFRESH_REVISION
+                })
+              }
+            />
+          </>
+        ) : (
+          <p className={styles.emptyIcon}>-</p>
```
sroy3, line 51: Why don't we keep `<p>No Plot to Display.</p>`?

webview/src/shared/components/errorTooltip/ErrorTooltip.tsx
sroy3: I think that would be a good idea. For discoverability mainly.

sroy3 (approved):

shcheklein:
I like the look and feel of this iteration way better, thank @mattseddon !

webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
mattseddon: sorry, there is no `!` icon but we show `!` for errors in the experiment table. What I meant was I had `!` instead of the error icon for an error but it just looked weird (unfortunately).

extension/src/plots/errors/collect.ts
```diff
@@ -0,0 +1,53 @@
...
+      const msg = getMessage(error)
+      msgs.push(msg)
+    }
+  }
+
+  if (msgs.length === 0) {
+    return undefined
+  }
+
+  return msgs.join('\n')
```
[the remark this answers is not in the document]
mattseddon: Yes, you are right. I will move that logic out to the client.

webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
mattseddon: I stared at/reworded the statement for ~30 minutes then I decided that this is a table of data and that we could just display the same missing value that is displayed in the CLI experiments table. Could be the wrong decision but we can iterate with feedback 🙏🏻.

webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
```diff
@@ -28,15 +31,25 @@ export const ComparisonTableCell: React.FC<ComparisonTableCellProps> = ({
...
-          onClick={() =>
-            sendMessage({
-              payload: plot.revision,
-              type: MessageFromWebviewType.REFRESH_REVISION
-            })
-          }
-        />
+        {error ? (
+          <>
+            <ErrorTooltip error={error}>
```
julieg18, line 36: Are users going to want to be able to copy the error? I know that when I run into an issue that I can't figure out I need to copy the error message so I can post it somewhere 🤔

julieg18 (approved):
Great work!

webview/src/shared/components/errorTooltip/ErrorTooltip.tsx
mattseddon: I'll move in a follow up

webview/src/plots/components/comparisonTable/ComparisonTableCell.tsx
mattseddon: I will make this copyable in the follow-up where I move the component.

extension/src/plots/errors/collect.ts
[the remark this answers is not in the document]
mattseddon: I will follow-up

PR #3522, opened 2023-03-21
Show plots errors in plots tree

mattseddon:
# 3/4 `main` <- #3477 <- #3520 <- this <- #3532

Relates to #2277 

### Demo

https://user-images.githubusercontent.com/37993418/226553363-a49865fa-262c-48fb-8282-ff7192dba3bc.mov

Note: The above shows that I need to add the "full update" logic for when the dvc.yaml is changed (will be done in #3532).

### DIFF
--- a/extension/src/path/selection/tree.ts
+++ b/extension/src/path/selection/tree.ts
@@ -1,5 +1,6 @@
 import {
   Event,
+  MarkdownString,
   TreeDataProvider,
   TreeItem,
   TreeItemCollapsibleState,
@@ -24,6 +25,7 @@ export type PathSelectionItem = {
   label: string | undefined
   path: string
   iconPath: Resource | Uri
+  tooltip: MarkdownString | undefined
 }
 
 export abstract class BasePathSelectionTree<
@@ -76,7 +78,7 @@ export abstract class BasePathSelectionTree<
       return new TreeItem(resourceUri, TreeItemCollapsibleState.Collapsed)
     }
 
-    const { dvcRoot, path, description, iconPath } = element
+    const { dvcRoot, path, description, iconPath, tooltip } = element
 
     const treeItem = this.getBaseTreeItem(element)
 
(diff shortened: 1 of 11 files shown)


extension/src/plots/errors/model.ts
```diff
@@ -0,0 +1,44 @@
+import { join } from 'path'
+import { collectErrors, collectImageErrors, collectPathErrors } from './collect'
+import { Disposable } from '../../class/dispose'
+import { PlotError, PlotsOutputOrError } from '../../cli/dvc/contract'
+import { isDvcError } from '../../cli/dvc/reader'
+
+export class ErrorsModel extends Disposable {
```
mattseddon, line 7: [I] Push the creation of this back a PR.

extension/src/plots/errors/decorationProvider.ts
```diff
@@ -0,0 +1,28 @@
+import { EventEmitter, FileDecoration, Uri } from 'vscode'
+import { DecoratableTreeItemScheme, getDecoratableUri } from '../../tree'
+import { ErrorDecorationProvider } from '../../tree/errorDecorationProvider'
+import { uniqueValues } from '../../util/array'
+
+export class DecorationProvider extends ErrorDecorationProvider {
```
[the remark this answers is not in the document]
mattseddon: This is identical to `extension/src/repository/model/decorationProvider.ts` and close to `extension/src/experiments/model/decorationProvider.ts` I can follow up.

mattseddon:
Note: there is a bug where SCM file decorations can run into the plots tree. I should be able to stop this from happening but it will require a workaround. E.g

image

extension/src/path/selection/tree.ts
```diff
@@ -90,6 +92,9 @@ export abstract class BasePathSelectionTree<
     if (description) {
       treeItem.description = description
     }
+    if (tooltip) {
+      treeItem.tooltip = tooltip
```
sroy3, line 96: What's the point of checking for the element before assigning it?  If `tooltip` is `undefined`, `treeItem.tooltip` is `undefined` as well anyway, no? I'm asking because I'm probably missing something.

sroy3 (approved):

julieg18 (approved):

extension/src/path/selection/tree.ts
mattseddon: you are right, there is no point. I'll update

PR #3523, opened 2023-03-21
Show "Custom" section in "Get Started" screen

julieg18:
## Demo

~https://user-images.githubusercontent.com/43496356/228010939-427af9d3-cd2a-41f5-ac87-cb77c107a006.mov~

~https://user-images.githubusercontent.com/43496356/228312421-5275a067-87cb-48e4-bdf8-5480c8bc2ad7.mov~

https://user-images.githubusercontent.com/43496356/229543195-c379e4e2-5345-4d3a-b88b-1f7d032bdcc0.mov

Part of #3373


sroy3 (approved):

extension/src/plots/model/collect.test.ts
```diff
@@ -48,9 +48,10 @@ describe('collectCustomPlots', () => {
     expect(data).toStrictEqual(expectedOutput)
   })
 
-  it('should return only custom plots if there no selected revisions', () => {
-    const expectedOutput: CustomPlotData[] = customPlotsFixture.plots.filter(
-      plot => plot.type !== CustomPlotType.CHECKPOINT
+  it('should return checkpoint plots with empty values if there no selected revisions', () => {
+    const expectedOutput: CustomPlotData[] = customPlotsFixture.plots.map(
+      plot =>
+        plot.type === CustomPlotType.CHECKPOINT ? { ...plot, values: [] } : plot
```
mattseddon, line 54: This feels odd: 

image

I don't know

1. Which plot is which
2. What type of plot should be inside (trends or metrics vs params).

If there are no experiments at all then we end up with this:

image

Seems inconsistent.

At the very least I think the text inside of the empty box needs to mention the plot title/select a checkpoint experiment to display but I think they should be hidden with some message to say "select a checkpoint experiment to display trend plots"

Note: Chances of users seeing these screens are minimal due to the use of checkpoint experiments but I think we should attempt to get it right.

mattseddon: Also, need to consider when there are no custom plots added but we do have data available:

image


An update will definitely be needed after https://github.com/iterative/vscode-dvc/pull/3567. Probably want to add a button.

julieg18: > At the very least I think the text inside of the empty box needs to mention the plot title/select a checkpoint experiment to display but I think they should be hidden with some message to say "select a checkpoint experiment to display trend plots"

A message makes sense! Not sure where it could go though 🤔. Will start experimenting with some ideas :) 

> Also, need to consider when there are no custom plots added but we do have data available:

What's the issue with the current setup? Text says "No Plots Added" and you can add a plot with the plus button 🤔

webview/src/plots/components/styles.module.scss
```diff
@@ -12,6 +23,18 @@ $gap: 20px;
   width: 100%;
 }
 
+.plotsSectionMessage {
```
julieg18, line 26: Not sure if the message looks better or worse without borders:

Screenshot 2023-03-28 at 9 50 38 AM

webview/src/plots/components/customPlots/CustomPlotsWrapper.tsx
```diff
@@ -39,6 +48,14 @@ export const CustomPlotsWrapper: React.FC = () => {
       hasItems={hasItems}
       height={height}
     >
+      {hasMissingCheckpointData && (
+        <div className={styles.plotsSectionMessage}>
```
julieg18, line 52: Is there a better place to show a message about needing to select checkpoint experiments.

julieg18 (commented):
Updating the pr to show a message in the custom section instead of empty placeholders when we have trends plots but no selected experiments! Going to rerequest reviews since this was a major change to the pr :)

shcheklein:
@julieg18 let's put this message after the existing plots and make it similar to the regular plots section. We can say something "Some plots require experiments to be selected. Add an experiment (can be a button)". We can also show a second button "+ Add" (to add a custom plot).

julieg18:
As discussed in the planning meeting, we're going to drop checkpoints, removing trends checkpoint plot types eventually. I'll  go ahead and remove the extra logic/ui surrounding the message when there are no selected experiments in this pr!

extension/src/plots/model/collect.test.ts
mattseddon: > What's the issue with the current setup? Text says "No Plots Added" and you can add a plot with the plus button

The screen shouldn't be split. The custom section should be hidden and there should be two buttons, one to add an experiment and one to add a custom plot.

mattseddon: There are 3 things that can currently cause an "empty" plots screen:

1. No plots in the project
2. No plots selected
3. No experiments selected

I'm about to introduce a global error too which gives 4.

I think we need to distinguish between the user having added custom plots or not.

From those we end up with the following combinations:

| project has plots | plots selected | experiments selected | project has custom plots | project has global error | welcome screen content | buttons to add/create | show or hide custom plots section | 
| -- | -- | -- | -- | -- | -- | -- | -- | 
| N/A | N/A | N/A | N | Y |  Error message | custom plots | hide | 
| N | N/A | N/A | N | N | Info on adding plots to the project | custom plot | hide |
| Y | N | N | N | N | There are no selected plots or experiments | plots, experiments, custom plots | hide |
| Y | N | Y | N | N | There are no selected plots or custom plots | plots, custom plots | hide |
| Y | Y | N | N | N | The are no selected experiments or custom plots | experiments, custom plots | hide | 
| N/A | N/A | N/A | Y | Y | Error message | custom plots  | show |
| N | N/A | N/A | Y | N | Info on adding plots to the project | none | show |
| Y | N | N | Y | N | There are no selected plots or experiments | plots, experiments | show |
| Y | N | Y | Y | N | There are no selected plots | plots  | show |
| Y | Y | N | Y | N | There are no experiments selected | experiments  | show |
| Y | Y | Y | Y | N | Normal webview | N/A | show |

I added suggested welcome screen content, buttons and whether I think we should show or hide the custom plots section. IMO the section should be hidden if there is no other data sent the webview and a button should be added to create new custom plots.

I even think we should do this when there is a global CLI error.

The alternative is that we throw away the welcome screen concept in favour of improving the empty state of the existing sections. We then only have to deal with the project having no data at all.

mattseddon: Another alternative is to simplify the welcome screen into as few states as possible.

| project has plots | plots selected | experiments selected | project has custom plots | project has global error | welcome screen content | buttons to select/create | show or hide custom plots section | 
| -- | -- | -- | -- | -- | -- | -- | -- | 
| N | N/A | N/A | N | N | Info on adding plots to the project | custom plot | hide |
| Y | N/A | N/A | N | N | There are no plots to show | experiments, plots, custom plots | hide | 
| N | N/A | N/A | Y | N | Info on adding plots to the project | none | show |
| Y | N/A | N/A | Y | N | There are no plots to show | experiments, plots, experiments | show |

We can deal with global errors separately but that would give us 6 states instead of 11. Seems pretty reasonable to me.

webview/src/plots/components/App.test.tsx
```diff
@@ -283,17 +283,68 @@ describe('App', () => {
     expect(loading).toHaveLength(3)
   })
 
-  it('should render the Add Plots and Add Experiments get started button when there are experiments which have plots that are all unselected', async () => {
```
julieg18, line 286: Tried my best to cover all the possibilities for the get started screen and to make the rather lengthy text clear 😅

julieg18 (commented):
Resolved https://github.com/iterative/vscode-dvc/pull/3523#discussion_r1149984545 and ready for another round of reviews.

extension/src/plots/model/collect.test.ts
julieg18: Thanks for the clarification! Updated the logic.

webview/src/plots/components/GetStarted.tsx
```diff
@@ -5,13 +5,19 @@ import { StartButton } from '../../shared/components/button/StartButton'
...
-  hasUnselectedPlots
+  hasUnselectedPlots,
+  hasNoCustomPlots
 }: AddPlotsProps) => (
   <div>
-    <p>No Plots to Display.</p>
+    <p>
+      {hasNoCustomPlots
+        ? 'No Plots to Display'
+        : 'No Selected Plots or Experiments'}
```
mattseddon, line 19: [I] I think maybe don't switch the text, use "No Plots to Display" for everything. Here is an example where there is an experiment selected and no plots:

image

mattseddon (approved):

webview/src/plots/components/GetStarted.tsx
julieg18: Updated the text!

PR #3526, opened 2023-03-21
Filter custom plot creation options

julieg18:
* filters quick pick options, stopping a user from selecting metrics/params that create a plot that has already been made

## Demo 

~https://user-images.githubusercontent.com/43496356/226739228-6eded2e8-1ba9-4f02-8b4f-69e90608161d.mov~

https://user-images.githubusercontent.com/43496356/226982974-ccb03ba9-b6b9-4c97-aef8-6e4b9a74810c.mov

Part of #3373


mattseddon:
From demo only:

When there are no combinations left to pick for a plot type then some kind of warning message needs to be shown. Probably in a toast.

extension/src/plots/model/quickPick.ts
```diff
@@ -99,9 +100,54 @@ export const pickCustomPlotType = (): Thenable<CustomPlotType | undefined> => {
...
+
+        return !plotIds.has(getCustomPlotId(metric, param))
+      })
+    : typeColumnLikes
+}
+
+export const pickMetricAndParam = async (
+  columns: Column[],
+  customPlotOrderVals: CustomPlotsOrderValue[]
+) => {
```
mattseddon, line 150: [I] In order to provide an accurate list of options I think you need to:

1. Generate the full params x metrics list but do not include entries that already exist.
2. Present the unique list of metrics from the list created by 1 in `pickFromColumnLikes` to get the metric. If there are no metrics left then you can show the initial warning.
3. Use the metric to filter the list created by 1 and pass the resulting params into `pickFromColumnLikes` to get the param.

This would mean you could never hit `Toast.showError('There are no params to select from.')`.

The initial list can be generated by something like
```
    const acc: { metric: string; param: string }[] = []
    for (const metric of metrics) {
      for (const param of params) {
        if (
          !existingOrder.includes({
            metric,
            param,
            type: CustomPlotType.METRIC_VS_PARAM
          })
        ) {
          acc.push({ metric, param })
        }
      }
    }

```

extension/src/plots/model/quickPick.ts
```diff
@@ -99,9 +100,54 @@ export const pickCustomPlotType = (): Thenable<CustomPlotType | undefined> => {
...
+  for (const { type, metric, param } of customPlotVals) {
+    if (type === customPlotType) {
+      plotIds.add(getCustomPlotId(metric, param))
+    }
+  }
+
+  return plotIds
+}
+
+const filterColumnsBasedOffCustomPlotValues = (
```
mattseddon, line 120: [I] I would avoid combining the logic here. Have two different branches for pickMetric vs pickMetricAndParam.

All pickMetric needs to do is filter the full list against what is already selected (easy, understandable).

pickMetricAndParam is more involved.

My smol brain cannot cope with this

sroy3 (approved):
Like mentioned, I would also remove the metric when there are no more params to chose from. Then disabling the `Metric Vs Param` or  `Checkpoint Trend` option if there are no more choices. Finally disable the "+" button when there is nothing else to pick.

julieg18:
> Like mentioned, I would also remove the metric when there are no more params to chose from. Then disabling the Metric Vs Param or Checkpoint Trend option if there are no more choices. 

Unfortunately, as far as I know, quick pick does not offer a way to disable quick pick items 😞 We could hide it entirely but I feel like that might cause confusion to users 🤔 

> Finally disable the "+" button when there is nothing else to pick.

Good idea! Should be simple enough to do and I will do so in a followup since it would involve front-end props and tests.

sroy3:
> > Like mentioned, I would also remove the metric when there are no more params to chose from. Then disabling the Metric Vs Param or Checkpoint Trend option if there are no more choices.
> 
> Unfortunately, as far as I know, quick pick does not offer a way to disable quick pick items 😞 We could hide it entirely but I feel like that might cause confusion to users 🤔
> 
 It does have the `enabled` property (https://code.visualstudio.com/api/references/vscode-api#QuickPick)

julieg18:
> It does have the enabled property (https://code.visualstudio.com/api/references/vscode-api#QuickPick)

I believe that's a property that enables user input which you can add to the QuickPick itself. 

image

Quick Pick items don't seem to have any enabled/disabled properties:

image

sroy3 (approved):

PR #3532, opened 2023-03-22
Refresh cached plots data on every update

### LINKED ISSUE #2277: Image plots show proper message and disable button if image is not available

mattseddon:
# 4/4 `main` <- #3477 <- #3520 <- #3522 <- this 

Closes #2277. Patches [this problem](https://github.com/iterative/dvc/issues/9025#issuecomment-1478703597).

Tl;dr of the problem is that our cached plots data is actually mutable depending on the combination of revisions (and even order of those revisions) that we call the `plots diff` with.

This PR updates the way that we treat cached plots data in the extension. Now every time there is an update to the selected revisions we call for an update from the CLI. Please watch the demo plus read below to see the approach that I've taken.

LMK what you think.

### Demo

https://user-images.githubusercontent.com/37993418/227178145-ba1c9821-be98-4a2d-b927-74bfdfa9a0cd.mov

The way that it works. 

1. We get a call to update the data from one of the paths available (e.g change in selected revisions or change in data in the workspace).
2. We send all the data that we currently have cached to the UI.
3. At the same time we call the CLI for an update for all of the currently selected revisions.
4. We send the updated data to the UI.

This approach gives us a few of things:

1. It gives the user something to look whilst the CLI is working. 
2. It gives them the impression that something is happening.
3. In most cases the initial data send to the UI will be correct anyway.
4. The resultant data should always be correct.


extension/src/experiments/index.ts
```diff
@@ -450,12 +450,6 @@ export class Experiments extends BaseRepository<TableData> {
     return experiment?.name || experiment?.label
   }
 
-  public getMutableRevisions() {
```
mattseddon, line 453: [F] No longer required as we now refresh all data all the time.

extension/src/plots/data/index.ts
```diff
@@ -37,13 +43,12 @@ export class PlotsData extends BaseData<{
...
     this.model = model
+    this.onDidTrigger = this.triggered.event
   }
 
   public async update(): Promise<void> {
-    const revs = collectRevs(
-      this.model.getMissingRevisions(),
-      this.model.getMutableRevisions()
-    )
+    this.notifyTriggered()
```
mattseddon, line 50: [F] This is how we send the cached data before following up with a CLI update.

extension/src/plots/index.ts
```diff
@@ -223,6 +219,19 @@ export class Plots extends BaseRepository<TPlotsData> {
...
   }
 
+  private triggerDataUpdate() {
+    void this.data.managedUpdate()
+  }
+
+  private onDidTriggerDataUpdate() {
+    const sendCachedDataToWebview = () => {
+      this.plots.resetFetched()
+      this.plots.setComparisonOrder()
```
mattseddon, line 229: [F] Resetting the fetchedRevs means that we get spinners in the ribbon for all of the revisions. Setting the comparison order means we get a loading state for the comparison table:

https://user-images.githubusercontent.com/37993418/227394044-b661259a-4440-4d5a-b952-bbe74e1c9d71.mov

extension/src/plots/model/index.ts
```diff
@@ -355,6 +336,12 @@ export class PlotsModel extends ModelWithPersistence {
     return this.experiments.getSelectedRevisions().map(({ label }) => label)
   }
 
+  public getSelectedOrderedCliIds() {
+    return collectOrderedRevisions(this.experiments.getSelectedRevisions()).map(
```
mattseddon, line 340: [F] This logic should give stable semantics no matter what order the user selects the experiments. We will take the most recent version of the template and try to apply that to the rest of the data.

mattseddon, line 340: [F] This logic should give stable semantics no matter what order the user selects the experiments. We will take the most recent version of the template and try to apply that to the rest of the data.

sroy3 (approved):

julieg18 (approved):
Great work!

PR #3535, opened 2023-03-22
Add a button to show more commits

sroy3:
For #1966 

With more commits available: 

https://user-images.githubusercontent.com/3683420/227008687-269cdfd1-4197-4aed-96e8-59e3ea1392a0.mov

Without more commits to show:
Screenshot 2023-03-22 at 3 46 30 PM

Will be following up with a show less commits button and adding persistence.


shcheklein:
Thanks @sroy3 . Can wr make it please similar to Studio and similar to the Previous Commits row? Aligned to the left, more like "blended" with the table ... trxt size should be apprx the same, etc

sroy3:
> Thanks @sroy3 . Can wr make it please similar to Studio and similar to the Previous Commits row? Aligned to the left, more like "blended" with the table ... trxt size should be apprx the same, etc

Screenshot 2023-03-24 at 2 52 19 PM

extension/src/experiments/index.ts
```diff
@@ -177,6 +183,15 @@ export class Experiments extends BaseRepository<TableData> {
     return this.cliData.managedUpdate()
   }
 
+  public getMoreCommits() {
+    this.numberOfCommitsToShow = this.numberOfCommitsToShow + 2
+    void this.webviewMessages.changeHasMoreCommits(this.numberOfCommitsToShow)
```
mattseddon, line 188: [C] Feels strange that `webviewMessages` is indirectly calling itself when this becomes an anonymous function.

mattseddon (approved):

extension/src/experiments/index.ts
sroy3: I agree. I might actually be able to move `numberOfCommitsToShow` into the webview messages. I'll try.

sroy3: Done. You're right, it feels better that way.

PR #3550, opened 2023-03-24
Improve Custom Plot Creation

julieg18:
* hide "Add Plots" button if there are no more plots to create
* hide pick custom plot type options if there are no types available

## Demo

https://user-images.githubusercontent.com/43496356/227646627-91878df2-98a8-4198-89bb-85eec7acf95a.mov

Followup to #3526


extension/src/plots/model/custom.ts
```diff
@@ -16,28 +18,93 @@ export const isCheckpointValue = (
...
+): boolean => {
+  const { metrics, params } = getCustomPlotPathsFromColumns(columns)
+  const plotIds = getCustomPlotIds(
+    customPlotOrder,
+    CustomPlotType.METRIC_VS_PARAM
+  )
+
+  for (const metric of metrics) {
+    for (const param of params) {
+      if (!plotIds.has(getCustomPlotId(metric, param))) {
```
julieg18, line 83: I tried a couple things but, as far as I can tell, as long as there as an if statement inside these two for loops, I get the complexity warning. Only solutions that come to mind that don't require the if statement require going through the entire loop which I was trying to avoid 🤔 

I ended up adding a cognitive-complexity comment for now but is there a better way to do this?

extension/src/plots/model/custom.ts
```diff
@@ -16,28 +18,93 @@ export const isCheckpointValue = (
...
+  for (const { type, metric, param } of customPlotVals) {
+    if (type === customPlotType) {
+      plotIds.add(getCustomPlotId(metric, param))
+    }
+  }
+
+  return plotIds
+}
+
+export const getCustomPlotPathsFromColumns = (
```
julieg18, line 54: This function and the one above are the same as before, just moved from `quickPick` to `custom`.

webview/src/plots/components/customPlots/CustomPlot.tsx
```diff
@@ -26,7 +26,7 @@ const createCustomPlotSpec = (
     return {}
   }
 
-  if (isCheckpointPlot(plot)) {
+  if (plot.type === CustomPlotType.CHECKPOINT) {
```
julieg18, line 29: Switched the function out for an if statement so I could use FILE_SEPARATOR in `custom.ts`, reducing extra arguments being passed around.

extension/src/plots/model/custom.ts
mattseddon: In order to pass the check you need to remove the inner loop. Something like 

```
const hasCombination = (
  plotIds: Set<string>,
  metric: string,
  params: string[]
): boolean => params.some(param => !plotIds.has(getCustomPlotId(metric, param)))

export const checkForMetricVsParamPlotOptions = (
  columns: Column[],
  customPlotOrder: CustomPlotsOrderValue[]
): boolean => {
  const { metrics, params } = getCustomPlotPathsFromColumns(columns)
  const plotIds = getCustomPlotIds(
    customPlotOrder,
    CustomPlotType.METRIC_VS_PARAM
  )

  for (const metric of metrics) {
    if (hasCombination(plotIds, metric, params)) {
      return true
    }
  }

  return false
}
```

However, whether or not this is more readable is debatable.

Maybe we should move complexity to 6 in both sonarjs and code climate. 

WDYT?

cc @sroy3

extension/src/plots/model/quickPick.test.ts
```diff
@@ -123,6 +139,118 @@ describe('pickCustomPlotType', () => {
       }
     )
   })
+  it('should show checkpoint type if available', async () => {
```
mattseddon, line 142: [N] Missing `the`?

mattseddon: same in the `if` statement below

extension/src/plots/model/quickPick.ts
```diff
@@ -75,61 +68,41 @@ export const pickCustomPlots = (
...
+  if (isCheckpointAvailable) {
+    items.push({
+      description:
+        'A linear plot that shows how a chosen metric changes over selected experiments.',
+      label: 'Checkpoint Trend',
+      value: CustomPlotType.CHECKPOINT
+    })
   }
-  return { metrics, params }
+
```
mattseddon, line 102: [Q] Do we need to return early here if there are no items? Even if the path currently cannot be hit it would be as a safeguard.

mattseddon: [Q] Are add/remove custom plot available from the command palette? Should they be? Will they be?

mattseddon: Same [Q] for experiments table.

mattseddon (approved):

extension/src/plots/model/custom.ts
mattseddon: Raised https://github.com/iterative/vscode-dvc/pull/3556 to discuss.

shcheklein: One option could be is to create a function that is an iterator and is producing one by one a Cartesian product of metrics and params. It will be used in two places instead of `const { metrics, params } = getCustomPlotPathsFromColumns(columns)` and two `for loops`:

```
for plotId in allMetricVsParamsPlotIds():
   if ...
```

extension/src/plots/model/quickPick.ts
mattseddon: > but I know we want to avoid adding every possible command in the palette

I don't agree with this. I think we should add as much useful functionality as possible there. Look at what these (very popular) Git extensions do:


https://user-images.githubusercontent.com/37993418/228078301-26e65f7f-92c3-44a1-afee-52addde3870f.mov

PR #3555, opened 2023-03-26
Remove code that refreshes a single plot revision (no longer possible)

mattseddon:
Note: starting a new chain as this is number 10 away from `main`. Does not seem like a good use of time to maintain the already approved PR descriptions.

# 1/2 `main` <- this <- #3557

This PR removes the code relating to refreshing a single revision. It is no longer possible to refresh a single revision after #3532.


julieg18 (approved):
Looks good!

PR #3567, opened 2023-03-27
Use commit rows in metric vs param plots

julieg18:
## Demo
wip

Part of #3373


extension/src/test/fixtures/expShow/base/customPlots.ts
```diff
@@ -35,6 +35,18 @@ export const customPlotsOrderFixture: CustomPlotsOrderValue[] = [
 ]
 
 export const experimentsWithCheckpoints: ExperimentWithCheckpoints[] = [
```
mattseddon, line 37: [Q] Is this still the best name?

julieg18: I think so since the actual function that we get the experiments from is called `getExperimentsWithCheckpoints` 🤔  Unless, I'm missing something?

sroy3 (approved):

Issue #3222, opened 2023-02-04
Show notification on a global plots failure


shcheklein:
If `dvc.yaml` is broken, or something else is happening with `plots diff` command, we should notify users. Not suggest to add an experiment even though an experiments is already selected, + it goes into a loop (expected?) of trying to get it:

https://user-images.githubusercontent.com/3659196/216793769-aff89a74-2b00-4591-b87f-ae4dacbc4afc.mov

shcheklein:
@dberenbaum even after fixing the `dvc.yaml` command is breaking since `HEAD` is broken. We should fix `dvc plots diff` to be resilient to a failure in revision, wdyt?

dberenbaum:
> @dberenbaum even after fixing the `dvc.yaml` command is breaking since `HEAD` is broken.

@shcheklein What do you meant here? What is breaking after you fix `dvc.yaml`?

Here's what I see:

https://user-images.githubusercontent.com/2308172/216996292-60b68bbf-0888-4c8e-8cca-9c33a9703643.mov

shcheklein:
Okay, I can't reproduce this anymore. We can't get back to this if I hit it again.

shcheklein:
I see it again:


https://user-images.githubusercontent.com/3659196/218173991-0e5bb6b9-f983-49a5-a35f-12ee62d0eb4f.mov

This time it's stuck on "Loading plots ..."

dberenbaum:
@shcheklein I guess those failed revisions were before the dvclive 2.0 changes and have a different field name? So I guess we (VS Code) should show that error?

Also, do you know if the command returns the diff for the non-failed revisions? We should be able to show a plot for some revisions.

shcheklein:
> Also, do you know if the command returns the diff for the non-failed revisions? We should be able to show a plot for some revisions.

No, it breaks in the middle. Generates a partial result, no HTML.

> I guess those failed revisions were before the dvclive 2.0 changes and have a different field name? So I guess we (VS Code) should show that error?

Yes, most likely. Yes, we should throw an error at least. At best we could show non-broken revision (but that depends on DVC).

dberenbaum:
Opened https://github.com/iterative/dvc/issues/9025.

Possible steps to remediate:
1. Show the error in VS Code.
2. Parse the error and take action in VS Code -- for example, drop the broken revision and run `plots diff` again without it. Does it depend on https://github.com/iterative/dvc/issues/7692?
[...]

PR #3569, opened 2023-03-28
Display error thrown by Plots Diff in Plots tree

mattseddon:
# 1/2 `main` <- this <- #3570 

Part of #3222

This PR handles `dvc plots` diff throwing an error. The error is displayed in the plots tree. In the normal way that we display "global errors". I will delay the 2nd part of this PR (displaying global errors in the webview) until #3523 is merged.

### Demo

https://user-images.githubusercontent.com/37993418/228108929-1e72a36d-f8e0-4574-9ade-55c9ee64b4a8.mov

Note: `dvc plots diff` will actually no longer throw an error under the above circumstances.


julieg18 (approved):
Looks good!

PR #3570, opened 2023-03-28
Show errors in plots ribbon

mattseddon:
# 2/2 `main` <- #3569 <- this

This PR shows plot revision errors in the ribbon inside of the webview.

### Demo

https://user-images.githubusercontent.com/37993418/228134664-fcc43a82-8db0-456b-a336-80dca2f1fa07.mov

### DIFF
--- a/extension/src/plots/errors/collect.ts
+++ b/extension/src/plots/errors/collect.ts
@@ -30,13 +30,17 @@ export const collectErrors = (
   ]
 }
 
-const getMessage = (error: PlotError): string => {
+export const getMessage = (error: PlotError): string => {
   const { msg, type, source } = error
 
-  if (type === 'FileNotFoundError' && source && !msg) {
-    return `${type}: ${source} not found.`
+  if (msg) {
+    return msg
   }
-  return [type, msg].filter(Boolean).join(': ')
+
+  if (type === 'FileNotFoundError' && source && !msg) {
+    return `${source} not found.`
+  }
+  return type
 }
 
 export const collectImageErrors = (
--- a/extension/src/plots/errors/model.ts
+++ b/extension/src/plots/errors/model.ts
@@ -2,7 +2,8 @@ import { join } from 'path'
 import {
   collectErrors,
   collectImageErrors,
-  collectPathErrorsTable
+  collectPathErrorsTable,
+  getMessage
 } from './collect'
 import { Disposable } from '../../class/dispose'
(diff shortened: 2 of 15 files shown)


extension/src/plots/errors/model.ts
```diff
@@ -55,6 +55,16 @@ export class ErrorsModel extends Disposable {
...
   }
 
+  public getRevisionErrors(rev: string) {
+    const errors: string[] = []
+    for (const error of this.errors) {
+      if (error.rev === rev) {
+        errors.push(error.msg)
+      }
+    }
+    return errors.length > 0 ? errors.join('\n') : undefined
```
mattseddon, line 65: [I] Push array out to the client

extension/src/plots/errors/collect.ts
```diff
@@ -30,13 +30,17 @@ export const collectErrors = (
   ]
 }
 
-const getMessage = (error: PlotError): string => {
+export const getMessage = (error: PlotError): string => {
```
mattseddon, line 33: [F] Decided that I had to change the format of this now. When I push the transformation logic out to the clients (tree, etc) I'll update to transform the errors as they come in.

mattseddon: It will simplify all of the errors logic but requires updating all of the tests so will break the change out.

webview/src/plots/components/ribbon/RibbonBlockTooltip.tsx
```diff
@@ -8,20 +8,26 @@ import { Icon } from '../../../shared/components/Icon'
...
+          <ErrorTooltipContent error={errors.join('\n')} />
+        </div>
+      )}
       <table className={styles.columnsTable}>
         <tbody>
           {firstThreeColumns.map(({ path, value, type }) => (
             <tr key={path}>
-              <td className={cn(styles[`${type}Key`])}>
+              <td className={cx(styles[`${type}Key`])}>
                 {truncate(path, 45, 'left')}
```
sroy3, line 31: Out of scope and I must have been the one who did this. But certainly, truncating in Javascript is a last resort. That should be done in CSS instead. If it is absolutely necessary there should have been a comment explaining the reasoning of using `vega-util` `truncate` function here. I would even add that if truncating in Javascript is the way to go, using a library for plotting is probably not the way to go either. I'll take a not and investigate.

sroy3 (approved):

julieg18 (approved):

PR #3576, opened 2023-03-28
Ensure that data for broken revisions is dropped

mattseddon:
# 1/2 `main` <- this <- #3578 

#3532 introduced a bug where completely broken revisions would be incorrectly retained in the partial cache. This PR ensures that the data for broken revisions is dropped (highlighted in the live demo this morning cc @dberenbaum).

### Demo

https://user-images.githubusercontent.com/37993418/228389856-31dbab04-3b17-4d30-898b-00acef2e5e96.mov

This again highlights that we need further error messages for broken revisions.


sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3583, opened 2023-03-29
Add persistence to the number of commits to show

sroy3:


mattseddon (approved):
I think we'll need to add options to "Show all"/"Reset to default" but let's delay until someone asks.

extension/src/cli/dvc/constants.ts
```diff
@@ -13,7 +13,8 @@ export const DVCLIVE_ONLY_RUNNING_SIGNAL_FILE = join(
 )
 export const EXP_RWLOCK_FILE = join(TEMP_EXP_DIR, 'rwlock.lock')
 
-export const NUM_OF_COMMITS_TO_SHOW = '3'
+export const NUM_OF_COMMITS_TO_SHOW = 3
```
mattseddon, line 16: [Q] Should this reference that it is the default?

PR #3585, opened 2023-03-30
Remove experiment checkpoints from extension UI

### LINKED ISSUE #3577: Remove checkpoint experiment support from extension UI
Based on the discussion in https://github.com/iterative/dvc/issues/9221 and the need to update `exp show --json` (https://github.com/iterative/dvc/pull/9170) for #3178 & #3436. **We have decided to drop checkpoints from the UI in the extension.** 

As we are doing this at the same time as updating the parsing logic to deal with a new format of experiments data **we have decided to avoid integrating the checkpoints data for now**. However, (also for now) we do need still need to know whether or not there are checkpoints in the workspace's projects so that we can offer the option to resume an experiment.

## UI

This change will primarily affect the experiments tree and webview. I.e these rows will be removed:

image

There will be flow-on effects in the rest of the extension. Below is the list of things that will need to be updated in the UI

1. Badges that group "hidden" values for a collapsed checkpoint experiment should be moved from the experiment level up to commits. I.e these:
 
image

2. Conditional `hasCheckpoints` logic needs to be dropped for filtering

image

3. Checkpoints need to be dropped from the "select experiments to plot" quick pick. 

image

4. Custom checkpoint trend plots should be dropped altogether (including collection, spec, etc)

image

5. Logic to add the count to the experiments tree needs to be updated

image

6. All references to checkpoints should be removed from the walkthrough/docs/pictures
 
image

image

### package.json

As we have always treated checkpoints differently to experiments/commits all of the conditional logic in our package.json will need to be updated. E.g the `view/item/context` entry for `dvc.views.experiments.applyExperiment`. Again, we should (for now) keep the ability to resume a checkpoint experiment if there are checkpoint experiments detected in any of the projects in the workspace.

### experiments table context menu

Conditional logic for checkpoints context menu can be dropped and the whole component can be greatly simplified. 

https://user-images.githubusercontent.com/37993418/228397347-74e367e7-1aa5-4c6c-bfac-cd92dc747aca.mov

### patches/redundancy 

The biggest one is `collectOverrideRevisionDetails` which is used to override plots data when there is a checkpoint experiment running in the workspace. Checkpoints should also not be passed to the code that assigns each experiment a status/color, etc.

### test fixtures/stories

The test fixtures will all need to be updated to match the new data coming out of https://github.com/iterative/dvc/pull/9170. We will no doubt need to update multiple stories and potentially drop some.

# Outside of 1, the task will consist mostly of deleting code/conditional logic branches. 

Note: Sorting logic seems fine but should be revisited separately (only experiments are sorted under commits).

mattseddon:
This is the feature branch that closes #3577

Each commit in this PR relates to removal of checkpoints for another part of the UI.

Outstanding items once this is merged:

1. Update remaining screenshots from walkthrough
2. Remove checkpoint logic from experiment collection code (`extension/src/experiments/model/collect.ts`)
3. Remove override revision logic from plots (replace short sha with experiment name for calling CLI).
4. Remove checkpoint type from `ExperimentType` enum
5. Final sweep for "patches/redundancy"


julieg18 (approved):
Great work so far!

PR #3589, opened 2023-03-30
Remove experiment checkpoints from select experiments to plot

mattseddon:
# 4/5 `main` <- #3585 <- #3587 <- #3588 <- this <- #3590

This PR removes checkpoints from the select to plot quick pick.

### Demo

https://user-images.githubusercontent.com/37993418/228728196-81e7d14c-4eca-474e-a079-6de7ec40a206.mov

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -348,11 +348,13 @@ export class Experiments extends BaseRepository<TableData> {
   }
 
   public async selectExperiments() {
-    const experiments = this.experiments.getExperimentsWithCheckpoints()
+    const experiments = [
+      ...this.experiments.getWorkspaceAndCommits(),
+      ...this.experiments.getExperiments()
+    ]
 
     const selected = await pickExperimentsToPlot(
       experiments,
-      this.hasCheckpoints(),
       this.columns.getFirstThreeColumnOrder()
     )
     if (!selected) {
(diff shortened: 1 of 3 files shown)


extension/src/experiments/index.ts
```diff
@@ -348,11 +348,13 @@ export class Experiments extends BaseRepository<TableData> {
   }
 
   public async selectExperiments() {
-    const experiments = this.experiments.getExperimentsWithCheckpoints()
+    const experiments = [
+      ...this.experiments.getWorkspaceAndCommits(),
+      ...this.experiments.getExperiments()
+    ]
```
mattseddon, line 354: [F] Until I know exactly what I need in the model I am going to try and avoid adding extra functions. It seems like there is an awful lot of redundancy in there.

extension/src/experiments/model/quickPick.ts
```diff
@@ -146,10 +92,6 @@ export const pickExperimentsToPlot = (
   )
 }
 
-export type ExperimentWithName = Experiment & {
-  name?: string
```
mattseddon, line 150: Experiment already has `name?: string` 😕

mattseddon: (it was me that added this code)

sroy3 (approved):

PR #3595, opened 2023-03-31
Remove references to experiment checkpoints from walkthrough

mattseddon:
# 2/3 #3585 <- #3591 <- this <- #3596

This PR removes references/screenshots of checkpoints from the walkthrough.

https://user-images.githubusercontent.com/37993418/229012211-ceea9c21-317e-4dc5-90e3-b7d08e9e3e02.mov

Note: I have not removed every screenshot that contains references to checkpoints. E.g -

image

The reason for this is that the cost of recreation (for me anyway) is too high and there is some utility in the picture.

PR #3591, opened 2023-03-30
Add experiment state indicators to commit records in table

mattseddon:
# 1/3 #3585 <- this <- #3595 <- #3596

Look a "new" feature.

This PR moves the experiment indicators (count) from experiment rows up to commit records. Previously we only did this for checkpoints. That no longer makes sense. 

I also fixed a bug with the top-level plotted indicator. Hidden rows were being incorrectly removed from the count. 

### Demo

https://user-images.githubusercontent.com/37993418/228984127-3d3e4b88-fda0-4eb3-b2c7-644bd1d7b2bb.mov

PR #3596, opened 2023-03-31
Remove experiment checkpoints from table context menu

mattseddon:
# 3/3 #3585 <- #3591 <- #3595 <- this 

This PR does some housekeeping in the experiment table context menu. We no longer need to cater to the condition of checkpoints existing so code can be simplified.

### Demo

https://user-images.githubusercontent.com/37993418/229024379-1936e7ed-c35f-4055-89a5-178e8197c0a3.mov

Note: I also caught and fixed a bug where queued experiments could be selected for plotting through the context menu 😞

### DIFF
--- a/webview/src/experiments/components/table/RowContextMenu.tsx
+++ b/webview/src/experiments/components/table/RowContextMenu.tsx
@@ -125,38 +125,34 @@ const getMultiSelectMenuOptions = (
 }
 
 const getRunResumeOptions = (
-  withId: (
+  hideIfRunning: (
     label: string,
     type: MessageFromWebviewType,
     hidden?: boolean,
     divider?: boolean
   ) => MessagesMenuOptionProps,
   projectHasCheckpoints: boolean,
-  hideVaryAndRun: boolean,
-  depth: number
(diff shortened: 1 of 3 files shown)


webview/src/experiments/components/table/RowContextMenu.tsx
```diff
@@ -190,44 +183,44 @@ const getSingleSelectMenuOptions = (
...
       'Create new Branch',
-      MessageFromWebviewType.CREATE_BRANCH_FROM_EXPERIMENT,
-      isNotExperimentOrCheckpoint
+      MessageFromWebviewType.CREATE_BRANCH_FROM_EXPERIMENT
     ),
-    notRunningWithId(
+    hideIfRunningOrNotExperiment(
       'Share to Studio',
       MessageFromWebviewType.SHARE_EXPERIMENT_TO_STUDIO,
-      notExperiment,
```
mattseddon, line 213: [F] I think this meant you could send queued experiments to Studio.... 👍🏻

sroy3 (approved):

PR #3601, opened 2023-03-31
Send back change in commits to show

sroy3:
Found this while creating the product update. It wouldn't update the "Show Less Commits" button correctly when there weren't any previous commits shown simply because it wouldn't return is it's showing more commits or not to the webview. You can see it fixed in this product update https://www.loom.com/share/2726708141b1467cac693fa02686ae8b


extension/src/experiments/webview/messages.ts
```diff
@@ -226,12 +226,13 @@ export class WebviewMessages {
...
   }
 
-  private async changeHasMoreOrLessCommits() {
+  private async changeHasMoreOrLessCommits(update?: boolean) {
     const availableNbCommits = await this.getNumCommits()
     const nbOfCommitsToShow = this.experiments.getNbOfCommitsToShow()
     this.hasMoreCommits = availableNbCommits > nbOfCommitsToShow
     this.isShowingMoreCommits =
       Math.min(nbOfCommitsToShow, availableNbCommits) > 1
+    update && this.sendWebviewMessage()
```
julieg18, line 235: Bug appears to be fixed (If I'm understanding the issue correctly), but out of curiosity, how does this change fix the bug? I see this function called twice but neither call adds any arguments to the function call 🤔

julieg18 (approved):

extension/src/experiments/webview/messages.ts
sroy3: Thanks, it is due to a bad merge. Will fix ASAP

PR #3602, opened 2023-03-31
Update demo project and latest tested CLI version (2.52.0)

sroy3:


sroy3:
Looks like e2e test are not stable. They sometimes pass and some other times they don't. Let's wait a little before merging this.

mattseddon:
This should be combined with #3477. I'll get to it on Monday.

PR #3610, opened 2023-04-03
Remove checkpoints from custom plots

mattseddon:
#  #3585 <- this

This PR removes checkpoints from custom plots. Once this is merged into #3585 we should be able to merge that PR. I will update that PR with a final description and make note of any outstanding items.

### Demo

https://user-images.githubusercontent.com/37993418/229422864-2ec9160b-ae48-4b2c-8db8-8de47220e976.mov

1 of the 2 users who had trend plots (migrating between versions):

https://user-images.githubusercontent.com/37993418/229423991-ae0df7a5-0353-4f9f-9838-9122e5433141.mov

### DIFF
--- a/extension/src/plots/model/custom.ts
+++ b/extension/src/plots/model/custom.ts
@@ -1,28 +1,12 @@
 import { VisualizationSpec } from 'react-vega'
 import { getCustomPlotId } from './collect'
 import { Column, ColumnType } from '../../experiments/webview/contract'
-import {
-  CheckpointPlot,
-  ColorScale,
-  CustomPlot,
-  CustomPlotType
-} from '../webview/contract'
 import { FILE_SEPARATOR } from '../../experiments/columns/paths'
 
-export const CHECKPOINTS_PARAM = 'epoch'
-
 export type CustomPlotsOrderValue = {
-  type: CustomPlotType.METRIC_VS_PARAM | CustomPlotType.CHECKPOINT
   metric: string
   param: string
 }
-
-export const isCheckpointValue = (
-  type: CustomPlotType.CHECKPOINT | CustomPlotType.METRIC_VS_PARAM
-) => type === CustomPlotType.CHECKPOINT
-
-export const isCheckpointPlot = (plot: CustomPlot): plot is CheckpointPlot =>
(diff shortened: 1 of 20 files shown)


webview/src/plots/components/customPlots/customPlotsSlice.ts
```diff
@@ -58,7 +55,7 @@ export const customPlotsSlice = createSlice({
       if (!action.payload) {
         return customPlotsInitialState
       }
-      const { plots, colors, ...statePayload } = action.payload
```
mattseddon, line 61: [Q] @julieg18 why were we sending colors? I can't find any references to it.

sroy3 (approved):

webview/src/plots/components/customPlots/customPlotsSlice.ts
julieg18: Leftover from #3575! They can be deleted :)

extension/src/plots/model/custom.ts
```diff
@@ -35,23 +19,17 @@ export const getFullValuePath = (type: string, columnPath: string) =>
 export const cleanupOldOrderValue = (
   value: { metric: string; param: string } | CustomPlotsOrderValue
 ): CustomPlotsOrderValue => ({
-  // previous column paths have the "TYPE:" prefix
   metric: removeColumnTypeFromPath(value.metric, ColumnType.METRICS),
-  param: removeColumnTypeFromPath(value.param, ColumnType.PARAMS),
-  // previous values didn't have a type
-  type: (value as CustomPlotsOrderValue).type || CustomPlotType.METRIC_VS_PARAM
```
julieg18, line 42: Are we sure about wanting to delete the `type` in workspace state values? If we want to add a different type of custom plot in the future, we're going to have to adjust our state again, adding the type back. Might be easier to just keep it even if it's not being used in the code currently 🤔

julieg18 (approved):
Great work!

extension/src/plots/model/custom.ts
mattseddon: YAGNI

PR #3614, opened 2023-04-03
Add a branches view to the experiments table

sroy3:
Part of #1966 

https://user-images.githubusercontent.com/3683420/229557591-1f39d3ad-dab4-4cf3-8458-a99e7d69efe3.mov

Just a quick win to be able to view all branches. Maybe this could be the intermediary step between with an action "Add to commits view", we'll see.


extension/src/experiments/index.ts
```diff
@@ -568,7 +571,7 @@ export class Experiments extends BaseRepository<TableData> {
           AvailableCommands.GIT_GET_NUM_COMMITS,
           this.dvcRoot
         ),
-      () => this.cliData.update()
+      (...args: (ExperimentFlag | string)[]) => this.cliData.update(...args)
```
mattseddon, line 574: [I] There is a bug here where any update created by the watchers will use the `-n <number-of-commits>` call. Leaving the webview out of sync with the data being displayed:


https://user-images.githubusercontent.com/37993418/229643446-20953424-1c76-40e0-b52c-79f7565ad699.mov

There are a couple of options but I think the best is to add the `experimentsModel` into `experimentsData` and use both `model.getNbOfCommitsToShow()` and `model.getIsBranchesView()` in the update function

mattseddon: Need to fix this one before we 🚢. 

🙏🏻.

extension/src/experiments/webview/messages.ts
```diff
@@ -68,7 +73,7 @@ export class WebviewMessages {
     hasStages: () => Promise<string>,
     addStage: () => Promise<boolean>,
     getNumCommits: () => Promise<number>,
-    changeNbOfCommits: () => Promise<void>
+    changeNbOfCommits: (...args: (ExperimentFlag | string)[]) => Promise<void>
```
mattseddon, line 76: [I] I think this function name should be updated now as it is not only changing the number of commits.

If `ExperimentsData` knows how to call the CLI (either branches or number of commits) then this can simply be the update function?

sroy3: I had a flash about this last night. Last week I had left this out for later and obviously I forgot about it (definitely not the best way to work).

julieg18 (commented):
Great work!

extension/src/experiments/index.ts
sroy3: Totally make sense. Should be fixed now

mattseddon (approved):

Issue #3619, opened 2023-04-03
Remove checkpoints from extension


mattseddon:
Follow up to #3577

Remaining items

- [ ] Update remaining screenshots from walkthrough
- [x] Remove checkpoint logic from experiment collection code (extension/src/experiments/model/collect.ts)
- [x] Remove override revision logic from plots (replace short sha with experiment name for calling CLI).
- [x] Remove checkpoint type from ExperimentType enum
- [x] Final sweep for "patches/redundancy"

mattseddon:
I will start on the prerequisites for integrating https://github.com/iterative/dvc/pull/9170

mattseddon:
One thing that I have noticed from working on #3665 is that we inconsistently display an experiment's rev/name and commit/branch's (aka baseline's) name/sha/commit message. Here are the places where we show this information: 

### Experiments tree

![image](https://user-images.githubusercontent.com/37993418/232346372-32a60edb-c257-4015-935d-90148d305916.png)

### Experiments table

![image](https://user-images.githubusercontent.com/37993418/232346383-245e8e45-19a1-44d6-8fc3-cbf255e6b4f0.png)

### Plots Ribbon

![image](https://user-images.githubusercontent.com/37993418/232346365-750cac27-0cc1-44b2-9a3d-c55b07ac1d7c.png)

### Plots image comparison table header

![image](https://user-images.githubusercontent.com/37993418/232346391-0e213985-7f2c-4106-88f7-baf7590b8f47.png)

I think we should make an update. My proposal is that in the examples above `scaly-puds` would always be first with `adde938` de-emphasized as secondary text (i.e match the ribbon's style). I think it makes sense to do this because we have removed checkpoints and the ability to see multiple "revs" per experiment. 

@dberenbaum @shcheklein WDYT?

mattseddon:
Realised that I missed at least one more place from the above comment. 

### Quick Picks

<img width="600" alt="image" src="https://user-images.githubusercontent.com/37993418/232348212-9d104c12-c6a2-456f-8ee1-533be74116e5.png">

<img width="607" alt="image" src="https://user-images.githubusercontent.com/37993418/232348229-5ab1af84-f590-4e01-b9c9-cfd938b2936c.png">

## Looking at the bigger picture:

### CLI

<img width="406" alt="image" src="https://user-images.githubusercontent.com/37993418/232348246-9482affd-b579-42f8-bd72-ddd42b48b6d7.png">

### Studio

[...]

PR #3624, opened 2023-04-04
Remove checkpoints from plots

mattseddon:
Related to #3619

This removes patches from the plots code from checkpoint experiments running in the workspace. In order to do this I have started calling the CLI with experiment names instead of short shas. Ideally, the name would replace the short sha as the `revision: string` in the `Revision` type but that is a much bigger job for another day (started working on that and had to reverse out).


sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3632, opened 2023-04-04
Update webdriverio-monorepo and fix e2e tests

sroy3:
Should fix #3631


extension/src/test/e2e/extension.test.ts
```diff
@@ -151,7 +151,6 @@ describe('Plots Webview', function () {
     const plots = await webview.vegaVisualization$$
 
     for (const plot of plots) {
-      await plot.scrollIntoView()
```
sroy3, line 154: Source of the error, didn't investigate much more, but it was actually useless in the test as it was already scrolled to the section, so I opted for the easy solution of removing it.

mattseddon (approved):

PR #3634, opened 2023-04-04
Update custom plot spec

julieg18:
* have metric vs param plot use scatter plot instead of linear plot

## Demo

~https://user-images.githubusercontent.com/43496356/229911951-9abe154b-3fd2-4145-a9e6-0baaa748e980.mov~

https://user-images.githubusercontent.com/43496356/230112955-630a0208-f4dc-4a13-8965-9c4dd6202c3c.mov

## To Do

- [ ] Fill marks and look into making marks for visible

Part of #3373


shcheklein:
@dberenbaum do you remember why we use non filled shape (circle) for the scatter plot template - it might be more visible / readable to fill them, wdyt?

webview/src/plots/components/App.test.tsx
```diff
@@ -59,52 +59,6 @@ import { clearSelection, createWindowTextSelection } from '../../test/selection'
...
-                  type: 'point'
-                }
-              }
-            ]
-          }
-        ]
-      }
-    }
-  })
-} as CustomPlotsData
```
mattseddon, line 106: 🎉

mattseddon (approved):
Let's fill in the marks/points and then ship this.

sroy3 (approved):

extension/src/plots/model/custom.ts
```diff
@@ -130,26 +111,10 @@ export const createSpec = (title: string, metric: string, param: string) =>
...
-        ]
-      }
-    ],
-    transform: [
-      {
-        as: 'y',
-        calculate: "format(datum['y'],'.5f')"
+          filled: true,
+          size: 60,
+          type: 'point'
```
julieg18, line 116: Turns out just filling `point` or using `circle` make the marks even smaller:

Screenshot 2023-04-05 at 9 09 11 AM

I ended up increasing the size so that the filled points would have a similar width to none filled points:

Screenshot 2023-04-05 at 9 13 40 AM

Screenshot 2023-04-05 at 9 12 08 AM

PR #3639, opened 2023-04-05
Remove checkpoints from experiment collection

mattseddon:
# 1/2 `main` <- this <- #3641 

This PR removes checkpoints from experiment collection. I have not thoroughly cleaned the code as it parts of it will be rewritten as we integrate the new format.


extension/src/experiments/model/collect.test.ts
```diff
@@ -1,6 +1,4 @@
 import { collectExperiments } from './collect'
-import { Experiment } from '../webview/contract'
-import modifiedFixture from '../../test/fixtures/expShow/modified/output'
```
mattseddon, line 3: [F] No longer needed.

julieg18 (approved):
Great work!

sroy3 (approved):

PR #3641, opened 2023-04-05
Show duplicate revisions when experiment finishes running in the workspace

mattseddon:
# 2/2 `main` <- #3639 <- this

This PR replaces a small piece of functionality that was purged in #3624. Instead of overwriting the revision/data, we send duplicate revisions whenever an experiment finishes so that the experiment does not appear to vanish and reappear.

### Demo

https://user-images.githubusercontent.com/37993418/229999686-238e5f69-9145-44df-974d-5ff5cfc74232.mov

### Prior behaviour

https://user-images.githubusercontent.com/37993418/230000027-af71bd83-dc5a-4a8b-83c1-f812b6e243f8.mov

Note: this is a temporary patch as the [new data](https://github.com/iterative/dvc/pull/9170#issuecomment-1469423657) will fix the problem properly.

### DIFF
--- a/extension/src/test/suite/experiments/index.test.ts
+++ b/extension/src/test/suite/experiments/index.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable sort-keys-fix/sort-keys-fix */
 import { join, resolve } from 'path'
 import { after, afterEach, beforeEach, describe, it, suite } from 'mocha'
 import { expect } from 'chai'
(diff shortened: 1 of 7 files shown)


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable sort-keys-fix/sort-keys-fix */
```
sroy3, line 1: This rule really has some hits and misses. Sometimes we want to list things in a special order (sizes: large, larger, regular, small... number: one, two, three, four...), sometimes it just make sense to write have something come before ({ type: X, payload: Y}, to me it feels more natural to have type come before it limits what can come after). It does not have any say on enums though (because it's linked to a number underneath, I get it), but this make it feel inconsistent. I'd say we remove it / review it (can we enforce after 5 properties?) and use judgment when creating objects, but I don't mind if people want to keep it.

sroy3 (approved):

extension/src/test/suite/experiments/index.test.ts
mattseddon: It can definitely be annoying. I'll put an item in the retro to discuss, I like the rule because it helps me avoid using smol 🧠 and gives me a few fewer decisions to make in the sea of decisions.

PR #3642, opened 2023-04-05
Make sure changing the number of revisions will trigger an update for the scale of the multiview plot

sroy3:
https://user-images.githubusercontent.com/3683420/230112891-84fcd4d1-d51a-44d1-9ae4-f601185313c0.mov

The scale variable was only set on first render. Unless something else changed, the number of revisions stayed the same because it had no impact on the `useMemo`.  Had to change `useMemo` to `useEffect` otherwise it wouldn't pick up the "unrelated" dependency.

### DIFF
--- a/webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
+++ b/webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
@@ -118,17 +118,6 @@ export const TemplatePlotsGrid: React.FC<TemplatePlotsGridProps> = ({
     )
   }, [entries, addEventsOnViewReady, nbItemsPerRow, multiView, revisionsLength])
 
-  useEffect(() => {
-    const panels = document.querySelectorAll('.vega-bindings')
-    return () => {
-      for (const panel of Object.values(panels)) {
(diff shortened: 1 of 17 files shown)


webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
```diff
@@ -63,16 +63,14 @@ export const TemplatePlotsGrid: React.FC<TemplatePlotsGridProps> = ({
...
-  useEffect(() => {
-    const panels = document.querySelectorAll('.vega-bindings')
-    return () => {
-      for (const panel of Object.values(panels)) {
-        panel.removeEventListener('mouseenter', addDisabled)
-        panel.removeEventListener('mouseleave', removeDisabled)
-        panel.removeEventListener('click', disableClick)
-      }
-    }
-  }, [addDisabled, removeDisabled, disableClick])
```
sroy3, line 75: This was useless since there are no more panels at this stage. And because I moved the function, it dropped coverage as it is never called.

mattseddon (approved):

PR #3646, opened 2023-04-05
Add custom plot to e2e test

julieg18:
Part of #3373


sroy3:
Maybe the plot creation could be its own test and one for the removal of same plot. That way we could test for the count to have updated as well.

extension/src/test/e2e/extension.test.ts
```diff
@@ -143,7 +145,7 @@ describe('Plots Webview', function () {
 
     await browser.waitUntil(
       async () => {
-        return (await webview.vegaVisualization$$.length) === 5
+        return (await webview.vegaVisualization$$.length) === 6
```
mattseddon, line 148: On a 2nd local run I think this will be 7 then on the 3rd 8. Need to remove the plot even if the test fails.

mattseddon: I am probably wrong. I still would not rely on that behaviour. Image if it changed, how hard/confusing would it be to debug the test?

julieg18:
Thanks for the reviews! Added another plots test that check for adding and removing custom plot, plus ensured the created custom plot gets deleted. I believe we're ready for another round of reviews!

extension/src/test/e2e/extension.test.ts
```diff
@@ -130,32 +133,74 @@ describe('Experiments Table Webview', function () {
...
+    for (const plot of plots) {
+      const plotNotEmpty = await webview.plotNotEmpty(plot)
+      expect(plotNotEmpty).toBe(true)
+    }
+
+    await webview.unfocus()
+    await closeAllEditors()
+  })
+
+  after(async function () {
```
mattseddon, line 198: [Q] Can this go in the `after` block at the start of the test file? 

Having multiple before/after/beforeEach/afterEach blocks makes things harder to reason about.

extension/src/test/e2e/extension.test.ts
```diff
@@ -130,32 +133,74 @@ describe('Experiments Table Webview', function () {
...
     await webview.unfocus()
+    await closeAllEditors()
+
+    await deleteCustomPlot()
+    await workbench.executeCommand('DVC: Show Plots')
+
+    await waitForDvcToFinish()
+    await webview.focus()
+
+    await waitForAllPlotsToRender(webview, 5)
```
mattseddon, line 186: [C] Seeing as you have made this helper you might want to make another one for checking the plots are not empty.

mattseddon (approved):

extension/src/test/e2e/extension.test.ts
julieg18: Moved the code to a helper but ran in the [`expect-expect`](https://github.com/jest-community/eslint-plugin-jest/blob/v27.2.1/docs/rules/expect-expect.md#enforce-assertion-to-be-made-in-a-test-body-expect-expect) eslint warning. I disabled it with a comment for now but can revert the change if preferred!

mattseddon: You can use a comment to update the expect function names: 
```
/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectHeaders"] }] */
```

PR #3650, opened 2023-04-06
Add 20px to min width of experiments column

mattseddon:
Takes into account the 20px added to the column contents in https://github.com/iterative/vscode-dvc/pull/3635.

### Before 

image

### After

image

This is a take-it-or-leave-it change. Just something that I noticed. Feel free to close.


demo
julieg18, line 1: I'm assuming this change to the demo was accidental?

julieg18 (approved):

demo
mattseddon: I even reversed it out but then didn't push the change 😫 
image

Issue #3597, opened 2023-03-31
Clean up Experiments Table SCSS


mattseddon:
As discussed in the retro, when working on #3591 I noticed that the `scss` associated with the experiments table has organically grown unruly.

<p align=center>
<img src=https://user-images.githubusercontent.com/37993418/229031339-75a58389-bf62-4404-a360-769e9ff3def7.png width=500 />
</p>

We should take a few hours to clean up `webview/src/experiments/components/table/styles.module.scss`. 

We should also spend some investigating linters. E.g 
- https://stylelint.io/
- https://www.npmjs.com/package/stylelint-scss
- https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint

sroy3:
I was thinking about this and I think that the main problem is that we are using CSS modules, but our stylesheets aren't modular enough. We basically have one big module per webview. If we were to create a stylesheet (module) per component or per component group (table head, table body...). That way, we can see much more easily what styles are used / unused, overwritten, and know exactly where to find them. @julieg18 I think you already did a great job in cleaning up everything and especially grouping stuff. If we decide to move things to their own module, that'll streamline thing a lot. I guess we can talk about this in the next retro, or just start slowly moving things.

PR #3664, opened 2023-04-11
Clean up table styles

julieg18:
* move everything out of `.experiments`
* delete unused classes/selectors (`summary`, `.hidden`, `.depthSpacer`)

This is a start to cleaning up the table styles. 

Part of #3597

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -79,10 +79,6 @@ $bullet-size: calc(var(--design-unit) * 4px);
   background-color: $header-border-color;
 }
 
-.hidden {
-  visibility: hidden;
-}
-
 .iconMenu {
   position: absolute;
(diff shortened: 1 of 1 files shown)


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -237,404 +233,389 @@ $bullet-size: calc(var(--design-unit) * 4px);
   height: 100%;
   display: flex;
   flex-flow: column nowrap;
+}
```
julieg18, line 236: Looking at the styles, the main issues seem to be the fact that there is a lot of nesting and classes are selected at different parts throughout the file, making it difficult to actually see what styles are being applied to what. 

I figured the first step is moving everything out of experiments then start grouping the classes better.

julieg18:
I think the changes in storybooks are caused by past changes not being accepted in a different pr 😅

PR #3665, opened 2023-04-11
Bump min DVC version to 2.55.0 (live metrics for experiments running outside the workspace)

mattseddon:
This PR accounts for the upcoming breaking change in the `exp show` data (https://github.com/iterative/dvc/pull/9170). Please review this PR. I'll be raising follow-ups and merging in (much like https://github.com/iterative/vscode-dvc/pull/3585).

### What this PR currently does:

- Updates test fixtures
- Creates a test data generator
- Changes experiment id to be name or truncated rev (unless workspace)
- Renames displayName to description
- Has experiment quick picks return the ID only (not name + id as id replaced the old logic)
- Calls the CLI (plots diff) with experiment ID (instead of name or label, this means we can reduce a lot of complexity)
- Removes CLIIdToRevision mapping (no longer needed)
- Drops full vs partial updates (https://github.com/iterative/dvc/pull/9170#issuecomment-1467937058) from ExperimentsData (https://github.com/iterative/vscode-dvc/pull/3684)
- Deletes hasCheckpoints/checkpoints watcher by using `{data: {meta: hasCheckpoints}}` (https://github.com/iterative/vscode-dvc/pull/3684)
- ~Clicking on workspace should select the running experiment not the other way around (#3705)~ Makes it impossible to select running experiments (causes plots diff to crash https://github.com/iterative/dvc/pull/9170#issuecomment-1514147457 & #3712)
- Triggers plots update when underlying sha for a name changes (e.g. user makes a commit and revision may still have the same alias/id but data probably changed #3715).

### follow-ups

- Drop the custom implementation of the studio client. I.e remove exp push patch (#3701 - Do not merge until this https://github.com/iterative/dvc/pull/9257 released).

- Revisit fields sent to plots webview (e.g revision, id, group, should be standardised). Standardise how experiment name/rev are displayed (#3723 & #3725).

- Fix plots test fixture as experiment running in the workspace should not be selectable: 

image

<p align="center">

</p>


extension/src/experiments/model/index.ts
```diff
@@ -117,6 +118,21 @@ export class ExperimentsModel extends ModelWithPersistence {
     this.setColoredStatus(runningExperiments)
   }
 
+  public transformAndSet_(
```
mattseddon, line 121: `_` is the key here. Every time I have to change the function signature/test data I'll duplicate and add a `_`. This saves me trying to fix absolutely everything in one massive hit (head would explode). I'll be taking notes on everything that I see that needs to be updated downstream and refraining from stacking those requirements/follow-ups up until after the breaking change.

mattseddon: This is the bad news: 

image

extension/src/cli/dvc/contract.ts
```diff
@@ -96,6 +97,73 @@ export interface ExperimentsOutput {
...
+    }
+  | ExpWithError
+
+export type ExpRange = {
+  revs: ExpState[]
+  name?: string
+  executor: Executor
+}
+
+export type ExpShowOutput = (ExpState & { experiments?: ExpRange[] | null })[]
```
mattseddon, line 165: [I] Clean these types up

extension/src/experiments/columns/collect/index.test.ts
```diff
@@ -28,106 +32,91 @@ describe('collectColumns', () => {
...
-        }
-      },
-      otherExp: {
-        data: {
-          params: {
-            'params.yaml': {
-              data: { mixedParam: true }
-            }
-          }
+  const data: ExpShowOutput = [
```
mattseddon, line 79: [I] Creation of data has to be completely wrapped in these util as otherwise another "overhaul" could lead to another round of manual work

extension/src/experiments/model/accumulator.ts
```diff
@@ -14,7 +14,7 @@ export class ExperimentsAccumulator {
     this.runningExperiments = []
     if (isRunning(workspace?.status)) {
       this.runningExperiments.push({
-        executor: EXPERIMENT_WORKSPACE_ID,
+        executor: Executor.WORKSPACE,
```
mattseddon, line 17: [I] I don't think this is possible anymore

extension/src/plots/data/collect.ts
```diff
@@ -58,17 +58,3 @@ export const collectFiles = (
 
   return uniqueValues(acc)
 }
-
-export const collectMetricsFiles = (
```
mattseddon, line 62: [F] Encapsulated this back in `Experiments`

webview/src/experiments/components/App.test.tsx
```diff
@@ -1081,9 +1086,9 @@ describe('App', () => {
       renderTableWithoutRunningExperiments()
 
       clickRowCheckbox('4fb124a')
-      clickRowCheckbox('1ba7bcd', true)
```
mattseddon, line 1084: [F] This label no longer exists, the label is now "workspace" as per experiments running in the workspace.

extension/src/test/util/index.ts
```diff
@@ -21,3 +28,80 @@ export const buildMockMemento = (
...
+): ExpShowOutput => [generateTestExpState(EXPERIMENT_WORKSPACE_ID, data)]
+
+const isTestExpWithName = (
+  testExp:
+    | { rev: string; name: string; data: Partial<ExpData> }
+    | Partial<ExpData>
+): testExp is { rev: string; name: string; data: Partial<ExpData> } =>
+  !!(testExp as { name: string }).name
+
+export const generateCommitWithExperiments = (
```
mattseddon, line 64: [F] This is already hard to grok and understand but I actually need to make it worse.

extension/src/experiments/columns/collect/metricsAndParams.ts
```diff
@@ -8,22 +8,20 @@ import {
...
+  fileHasError,
+  MetricsOrParams
 } from '../../../cli/dvc/contract'
 import {
   buildMetricOrParamPath,
   FILE_SEPARATOR,
   METRIC_PARAM_SEPARATOR
 } from '../paths'
 
-export const typedValueTreeEntries = (value: NonNullable<ValueTree>) =>
```
mattseddon, line 24: [F] No longer need this now the types have been fixed... or TypeScript can handle typing object entries now.

extension/src/experiments/index.ts
```diff
@@ -64,7 +64,7 @@ export type ModifiedExperimentAndRunCommandId =
 
 export class Experiments extends BaseRepository<TableData> {
   public readonly onDidChangeIsParamsFileFocused: Event<string | undefined>
-  public readonly onDidChangeExperiments: Event<ExperimentsOutput | void>
+  public readonly onDidChangeExperiments: Event<void>
```
mattseddon, line 67: [F] No longer expose data outside of this class (that was a mistake)

extension/src/experiments/index.ts
```diff
@@ -466,13 +465,6 @@ export class Experiments extends BaseRepository<TableData> {
     return this.experiments.getExperiments()
   }
 
-  public getExperimentDisplayName(experimentId: string) {
```
mattseddon, line 469: [F] id is now the same as this so we can just pass the id around everywhere and remove this logic. This leads to a lot of code getting deleted from `WorkspaceExperiments`.

extension/src/experiments/index.ts
```diff
@@ -529,6 +521,10 @@ export class Experiments extends BaseRepository<TableData> {
     return this.columns.hasNonDefaultColumns()
   }
 
+  public getRelativeMetricsFiles() {
+    return this.columns.getRelativeMetricsFiles()
```
mattseddon, line 525: [F] required for plots so we don't have to send the data across

extension/src/experiments/model/collect.test.ts
```diff
@@ -105,32 +85,4 @@ describe('collectExperiments', () => {
...
-      tip1cp2: {
-        data: { checkpoint_tip: 'tip1' }
-      },
-      tip1cp3: {
-        data: { checkpoint_tip: 'tip1' }
-      }
-    }
-  }
-
-  it('should only list the tip as a top-level experiment', () => {
```
mattseddon, line 128: [F] Checkpoints are gone

webview/src/experiments/components/App.test.tsx
mattseddon: [I] We should wrap these somewhere so they are easier to change if we ever have to do it again.

extension/src/cli/dvc/contract.ts
```diff
@@ -26,76 +26,92 @@ export type DataStatusOutput = {
...
 
 export enum ExperimentStatus {
-  FAILED = 'Failed',
-  QUEUED = 'Queued',
-  RUNNING = 'Running',
-  SUCCESS = 'Success'
+  FAILED = 'failed',
+  QUEUED = 'queued',
+  RUNNING = 'running',
+  SUCCESS = 'success'
```
mattseddon, line 46: [F] These changed on the DVC side.

extension/src/experiments/columns/collect/index.test.ts
```diff
@@ -28,106 +27,91 @@ describe('collectColumns', () => {
...
-                data: { 2: 3 }
-              }
-            },
-            params: {
-              a: {
-                data: {
-                  b: 'c'
-                }
-              }
+    const columns = collectColumns(
```
mattseddon, line 30: [F] There should be no behaviour changes here. Just updates to the format of the data.

mattseddon: and even then the underlying params/metrics should also be the same

extension/src/experiments/columns/collect/index.test.ts
```diff
@@ -821,3 +720,11 @@ describe('collectChanges', () => {
     ])
   })
 })
+
+describe('collectRelativeMetricsFiles', () => {
```
mattseddon, line 724: [F] This was moved from `plots/data/collect` to encapsulate the data type inside of `Experiments`.

extension/src/cli/dvc/contract.ts
```diff
@@ -26,76 +26,92 @@ export type DataStatusOutput = {
...
-export interface ValueTreeOrError {
-  data?: ValueTree
-  error?: ErrorContents
-}
-
 type RelPathObject<T> = {
   [relPath: string]: T
 }
 
-export type ValueTreeRoot = RelPathObject<ValueTreeOrError>
```
mattseddon, line 38: [F] I think that having this in the recursive type was actually a bug

extension/src/experiments/columns/collect/index.test.ts
```diff
@@ -429,94 +395,31 @@ describe('collectColumns', () => {
...
+        data: { lr: 0.0005, seed: 473987, weight_decay: 0 }
       }
     }
   }
 
   it('should mark new dep files as changes', () => {
     const changes = collectChanges(uncommittedDepsFixture)
+    const [workspace] = uncommittedDepsFixture
+    if (experimentHasError(workspace)) {
+      throw new Error('Experiment should not have error')
```
mattseddon, line 418: [F] This will fail the test.

extension/src/experiments/columns/collect/index.test.ts
```diff
@@ -557,140 +459,153 @@ describe('collectChanges', () => {
...
+            a: { b: 1, d: { e: 100 } }
+          }),
+          {
+            data: getParams({
+              a: { b: 'c', d: { e: 'f' } }
+            }),
+            rev: '31c419826c6961bc0ec1d3900ac18bf904dcf82e'
+          }
+        )
+      )
```
mattseddon, line 497: I'm not a massive fan of how much "magic" is in the setup of this test but WDYT? Should we do this more/not at all?

mattseddon: If this starts failing debugging would not be fun

extension/src/experiments/commands/register.ts
```diff
@@ -164,7 +160,7 @@ const registerExperimentInputCommands = (
   internalCommands.registerExternalCliCommand(
     RegisteredCliCommands.EXPERIMENT_VIEW_BRANCH,
     ({ dvcRoot, id }: ExperimentDetails) =>
-      experiments.getExpNameAndInputThenRun(
+      experiments.getInputAndRun(
```
mattseddon, line 163: [F] See https://github.com/iterative/vscode-dvc/pull/3665/files#r1166194407

extension/src/experiments/model/collect.ts
```diff
@@ -44,48 +40,16 @@ export type ExperimentItem = {
...
-  const experiment = experimentsObject[sha]?.data
-  if (!experiment) {
-    return
-  }
-
-  const { name } = experiment
-  if (name) {
-    return `[${name}]`
-  }
+type ExperimentsAccumulator = {
```
mattseddon, line 43: [F] Replaces the old wacky class.

extension/src/experiments/model/collect.ts
```diff
@@ -287,44 +121,195 @@ const getCommitMessages = (
...
+      label
+    },
+    expState,
+    baseline
+  )
+
+  if (name) {
+    const description = `[${name}]`
+    experiment.description = description
+    experiment.logicalGroupName = description
```
mattseddon, line 257: [I] Get rid of this in a follow-up

extension/src/experiments/model/collect.ts
```diff
@@ -287,44 +121,195 @@ const getCommitMessages = (
...
+  if (
+    dvcLiveOnly ||
+    acc.runningExperiments.some(
+      ({ executor, id }) =>
+        executor === Executor.WORKSPACE && id !== EXPERIMENT_WORKSPACE_ID
+    )
+  ) {
+    acc.workspace.executor = Executor.WORKSPACE
+    acc.workspace.status = ExperimentStatus.RUNNING
+    acc.runningExperiments.unshift({
```
mattseddon, line 278: [I] Should be able to get rid of this altogether too (follow-up).

extension/src/experiments/model/modify/collect.test.ts
```diff
@@ -8,10 +8,10 @@ describe('collectFlatExperimentParams', () => {
     const params = collectFlatExperimentParams(rowsFixture[0].params)
     expect(params).toStrictEqual([
       { path: appendColumnToPath('params.yaml', 'code_names'), value: [0, 1] },
-      { path: appendColumnToPath('params.yaml', 'epochs'), value: 2 },
+      { path: appendColumnToPath('params.yaml', 'epochs'), value: 5 },
       {
         path: appendColumnToPath('params.yaml', 'learning_rate'),
-        value: 2.2e-7
+        value: 2.1e-7
```
mattseddon, line 14: [F] Changed the data in the workspace to match the experiment now running in the workspace in the test fixture

extension/src/experiments/model/quickPick.ts
```diff
@@ -146,7 +142,7 @@ export const pickExperiment = (
   experiments: Experiment[],
   firstThreeColumnOrder: string[],
   title: Title = Title.SELECT_EXPERIMENT
-): Thenable<ExperimentDetails | undefined> =>
+): Thenable<string | undefined> =>
```
mattseddon, line 145: [F] As there is now no difference between id/name (if there is one) we don't need the `ExperimentDetails` type.

extension/src/plots/data/collect.ts
mattseddon: And move the tests

extension/src/plots/data/index.ts
```diff
@@ -48,7 +47,7 @@ export class PlotsData extends BaseData<{
 
   public async update(): Promise<void> {
     this.notifyTriggered()
-    const revs = this.model.getSelectedOrderedCliIds()
+    const revs = this.model.getSelectedOrderedIds()
```
mattseddon, line 50: [F] Dropped the CLIId mapping.

extension/src/test/fixtures/plotsDiff/index.ts
```diff
@@ -467,9 +466,9 @@ export const getMultiSourceOutput = (): PlotsOutput => ({
 export const REVISIONS = [
   EXPERIMENT_WORKSPACE_ID,
   'main',
-  '4fb124a',
-  '42b8736',
-  '1ba7bcd'
+  'exp-e7a67',
+  'test-branch',
+  'exp-83425'
```
mattseddon, line 471: [F] As I mentioned in the retro I think we need to make the plots test fixture static/duplicate information instead of having all of this transformation logic present. This logic made it really hard to update the integration tests.

mattseddon: Although it will lead to more of this:

image

extension/src/test/suite/patch.test.ts
```diff
@@ -14,11 +14,11 @@ import { STUDIO_ENDPOINT, registerPatchCommand } from '../../patch'
 import { AvailableCommands } from '../../commands/internal'
 import expShowFixture from '../fixtures/expShow/base/output'
 import { dvcDemoPath } from '../util'
-import { ExperimentFields } from '../../cli/dvc/contract'
 import { Toast } from '../../vscode/toast'
 import { Modal } from '../../vscode/modal'
 import { Response } from '../../vscode/response'
 import { RegisteredCommands } from '../../commands/external'
+import { ExpData } from '../../cli/dvc/contract'
```
mattseddon, line 21: [F] These tests will be getting deleted as the patch is being replaced by `exp push`.

extension/src/plots/model/index.ts
```diff
@@ -198,12 +193,12 @@ export class PlotsModel extends ModelWithPersistence {
...
     const selectedRevisions: Revision[] = []
     for (const experiment of this.experiments.getSelectedRevisions()) {
-      const { commit, displayName, label, displayColor, logicalGroupName, id } =
+      const { commit, description, label, displayColor, logicalGroupName, id } =
         experiment
       const revision: Revision = {
         displayColor,
-        errors: this.errors.getRevisionErrors(label),
-        fetched: this.fetchedRevs.has(label),
+        errors: this.errors.getRevisionErrors(id),
```
mattseddon, line 200: [F] The reason that we have to switch from revision to id for identifying an experiment is that the revision (aka label) on an experiment can be duplicated for experiments that are running in the workspace (the label is also workspace).

extension/src/experiments/columns/collect/index.test.ts
julieg18: While it would be harder to debug, it does help reduce code repetition. I'm ok with doing it more! Might just be me, but would it be more readable if it was broken up more? Took me a sec to figure out what arguments were going where :)

sroy3 (approved):

extension/src/experiments/index.ts
```diff
@@ -558,6 +546,18 @@ export class Experiments extends BaseRepository<TableData> {
     return this.webviewMessages.sendWebviewMessage()
   }
 
+  private getCommitOutput(data: ExpShowOutput | undefined) {
+    if (!(data && data.length > 1)) {
```
julieg18, line 550: Similar to last comment, why not `if (!data || data.length <= 1)`?

extension/src/test/util/experiments.ts
```diff
@@ -0,0 +1,147 @@
...
+  data?: Partial<ExpData>
+  name?: string
+  rev: string
+  testExps: (
+    | {
+        data: Partial<ExpData>
+        executor: Partial<ExecutorState>
+        name: string
+        rev: string
+      }
```
julieg18, line 67: I'm seeing multiple ` { /* obj with various keys */ } | Partial<ExpData>` within these utils. Should these "obj with various keys" be moved into their own types? Might make things more readable 🤔

extension/src/test/util/experiments.ts
```diff
@@ -0,0 +1,147 @@
...
+  ExecutorState,
+  ExpData,
+  EXPERIMENT_WORKSPACE_ID,
+  ExperimentStatus,
+  ExpRange,
+  ExpShowOutput,
+  ExpState
+} from '../../cli/dvc/contract'
+
+export const generateTestExpData = (data: Partial<ExpData> = {}): ExpData => ({
```
julieg18, line 12: Should there be a couple tests for some of these utils? Might make debugging easier in the future 🤔

extension/src/test/util/experiments.ts
julieg18, line 1: Great work on creating these utils! I could understand what the function were doing pretty easily!

julieg18 (approved):
Great work!

extension/src/test/util/experiments.ts
mattseddon: I'll follow up on these comments

PR #3667, opened 2023-04-11
Group table styles

julieg18:
# 1/4 `main` <= this <= #3674 <= #3681 <= #3707

* moves selector rules that affect the entire webview or table into one place at the top of the concrete selectors. 

Note: I'm not planning to merge this until all WIP prs are done and approved! 

Part of #3597


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -683,10 +744,6 @@ td {
   font-size: 0.65rem;
 }
 
-.staticColumn {
```
julieg18, line 686: Class isn't being used anywhere.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -249,76 +353,12 @@ $bullet-size: calc(var(--design-unit) * 4px);
...
-      transition: box-shadow 0.25s;
-    }
-
-    &.headerCell {
-      &::after {
-        bottom: auto;
-        left: auto;
-        background-color: transparent;
-      }
+.bodyRow {
```
julieg18, line 356: The `tr {}` held styles that  only affect body rows (doesn't affect the head rows at all). I moved them outside of it and into the `bodyRow` selector. The body row styles can be cleaned up further but I'll save that for another chain pr that focused on cleaning up body styles.

mattseddon (approved):

sroy3 (approved):

PR #3670, opened 2023-04-11
Hidden: Add and remove branches to the experiments table

sroy3:
Part of #1966 

https://user-images.githubusercontent.com/3683420/231493040-2946a6e5-e460-4d86-aff2-ce6a74a5a7ef.mov

Hidden behind a feature flag for now until it actually does something in the table. I wanted to easily separate the collection of data from branch selection as this is already big enough. Might also be easier to merge with a feature flag than start a chain on PR.


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1422,6 +1434,73 @@ suite('Experiments Test Suite', () => {
...
+      messageSpy.resetHistory()
+      const mockMessageReceived = getMessageReceivedEmitter(webview)
+
+      mockMessageReceived.fire({
+        type: MessageFromWebviewType.SELECT_BRANCHES
+      })
+
+      expect(mockSelectBranches).to.be.calledOnce
+
+      // Needed for the test to pass, but shouldn't be needed.
```
sroy3, line 1461: Racked my brain a whole day for that. Still can't figure out why it's needed. I really wish I could remove it.

mattseddon: This is because waiting for 

```
    const selectedBranches = await this.selectBranches(
      this.experiments.getBranchesToShow()
    )
```

pushes the result into the next event loop, meaning that the assertion executes before the stub is called.

`await Time.delay(0)` is equivalent to `await new Promise(process.nextTick)` so by calling that you delay the assertion until the code under test has completed (FIFO).

I would rewrite the test like this to make it deterministic:

```
    it.only('should handle a message to select branches', async () => {
      const {
        experiments,
        experimentsModel,
        messageSpy,
        mockUpdateExperimentsData,
        mockSelectBranches
      } = setupExperimentsAndMockCommands()
      const mockSelectBranchesToShow = stub(
        experimentsModel,
        'selectBranchesToShow'
      )

      const waitForBranchesToBeSelected = new Promise(resolve =>
        mockSelectBranchesToShow.callsFake(() => resolve(undefined))
      )

      messageSpy.resetHistory()

      mockMessageReceived.fire({
        type: MessageFromWebviewType.SELECT_BRANCHES
      })

      expect(mockSelectBranches).to.be.calledOnce

      await waitForBranchesToBeSelected

      expect(mockSelectBranchesToShow).to.be.calledOnceWith(['main', 'other'])

      expect(mockUpdateExperimentsData).to.be.calledOnce
```

Hope that helps... I used to get stuck on this all the time.

mattseddon: You'll see example of this all over the integration tests
```
      const waitForSomeFunctionCall = new Promise(resolve =>
        mockSomeFunction.callsFake(() => resolve(undefined))
      )
      ...
      await waitForSomeFunctionCall
      expect(....
```

extension/src/experiments/workspace.test.ts
```diff
@@ -661,4 +672,121 @@ describe('Experiments', () => {
...
+
+      expect(mockedQuickPickManyValues).toHaveBeenCalledWith(
+        allBranches.map(branch =>
+          expect.objectContaining({ label: branch, value: branch })
+        ),
+        expect.anything()
+      )
+    })
+
+    it('should not display the current branche in the quick pick', async () => {
```
mattseddon, line 715: [N] branche

extension/src/experiments/index.ts
```diff
@@ -551,6 +558,15 @@ export class Experiments extends BaseRepository<TableData> {
...
     this.notifyColumnsChanged()
+    void this.updateAvailableBranchesToSelect()
+  }
+
+  private async updateAvailableBranchesToSelect() {
+    const allBranches = await this.internalCommands.executeCommand<string[]>(
+      AvailableCommands.GIT_GET_BRANCHES,
+      this.dvcRoot
+    )
+    this.experiments.setAvailableBranchesToShow(allBranches)
```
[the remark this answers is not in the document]
mattseddon: This call should probably go behind the feature flag as well.

mattseddon (approved):

extension/src/experiments/model/index.ts
```diff
@@ -403,6 +410,23 @@ export class ExperimentsModel extends ModelWithPersistence {
     return this.isBranchesView
   }
 
+  public selectBranchesToShow(branches: string[]) {
```
julieg18, line 413: Should this be `setBranchesToShow`? When I think `select`, I think it's going to open a quick pick or something similar.

julieg18 (approved):
Great work!

extension/src/experiments/index.ts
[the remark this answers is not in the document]
sroy3: When there were 2 buttons (one to add branches and one to remove them), this was used to decide whether or not to disable the "add branches" button (you've picked all the branches already). I cleaned it up a little when the control got changed, but I agree, this doesn't need to be called that often. I'll look into this.

extension/src/experiments/workspace.test.ts
sroy3: French muscle memory I suspect...

extension/src/test/suite/experiments/index.test.ts
sroy3: Thank you so much

webview/src/util/flags.ts
```diff
@@ -0,0 +1,3 @@
+export const featureFlag = {
```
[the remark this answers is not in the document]
sroy3: Only thing left to do. I'll do this as my next follow up. Just so we have a real branch and PR related to feature flags (this feature flag one is mainly a hack).

PR #3671, opened 2023-04-11
Disable setup section if unavailable

julieg18:
## Demo

https://user-images.githubusercontent.com/43496356/231277475-c4da8d67-5be8-442f-ab7b-60a2a9daef40.mov

Followup to #3655


julieg18:
As discussed, we're going to keep the section enabled

PR #3674, opened 2023-04-12
Clean up table head styles

julieg18:
# 2/4 `main` <= #3667 <= this <= #3681 <= #3707

* move all selectors to one place
* delete unused code
* remove unneeded nesting

Part of #3597

PR #3681, opened 2023-04-13
Group table body styles

julieg18:
# 3/4 `main` <= #3667 <= #3674 <= this <= #3707

* move all selectors to one place
* delete unused code
* remove unneeded nesting

Part of #3597

### DIFF
--- a/webview/src/experiments/components/AddStage.tsx
+++ b/webview/src/experiments/components/AddStage.tsx
@@ -22,7 +22,7 @@ export const AddStage: React.FC<AddStageProps> = ({ hasValidDvcYaml }) => (
       disabled={!hasValidDvcYaml}
     />
     {!hasValidDvcYaml && (
-      <p className={styles.error}>
+      <p className={styles.errorText}>
         Your dvc.yaml file should contain valid yaml before adding any pipeline
         stages.
       </p>
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -80,6 +80,11 @@ $badge-size: 0.85rem;
   background-color: $header-border-color;
 }
 
+%cellContentsBase {
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
 // Concrete selectors
 
 @keyframes spin {
@@ -115,7 +120,6 @@ table {
 
 tr {
   position: relative;
-  background-color: $row-bg-color;
   width: fit-content;
 }
 
@@ -190,8 +194,7 @@ table.withExpColumnShadow tr > *:first-child {
 }
 
 .cellContents {
-  overflow: hidden;
-  text-overflow: ellipsis;
+  @extend %cellContentsBase;
 }
 
 ul {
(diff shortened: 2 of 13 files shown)


webview/src/experiments/components/AddStage.tsx
```diff
@@ -22,7 +22,7 @@ export const AddStage: React.FC<AddStageProps> = ({ hasValidDvcYaml }) => (
       disabled={!hasValidDvcYaml}
     />
     {!hasValidDvcYaml && (
-      <p className={styles.error}>
```
julieg18, line 25: Similiar to the previous pr, renamed some classes to better explain what they're doing :)

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -774,18 +628,108 @@ td {
...
+      border-right: 1.5px solid currentColor;
+      border-top: 1.5px solid currentColor;
+      animation: spin 1s cubic-bezier(0.53, 0.21, 0.29, 0.67) infinite;
+      background-color: $checkbox-background;
+    }
+  }
 }
 
-.experimentCellContentsContainer {
-  @extend .cellContents;
```
julieg18, line 782: Personally, I don't think it's the best practice to extend regular classes because someone could update the class not realizing they're effecting other classes. Created another one of our "Extendable Silent Rules" for this.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -804,13 +748,18 @@ td {
 .experimentCellSecondaryName {
   color: $meta-cell-color;
   font-size: 0.75em;
-  display: flex;
-  align-items: center;
+
```
julieg18, line 751: Noticed that the text-overflow wasn't working. Fixed now:

image

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -590,70 +468,76 @@ th {
...
   font-size: 0.8rem;
   line-height: 2rem;
   align-items: center;
+  background-color: $row-bg-color;
 
   &:not(.experimentCell)::before {
     @extend %cellBorderLeft;
   }
 
+  .rowSelected & {
```
julieg18, line 523: I think it's a little better but the way we update backgrounds/borders for cells is still kind of complex. I think there are some better ways to go about this but I've already spent quite a bit of time on updating table styles, so will look at it at a later date!

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -590,70 +468,76 @@ th {
...
-      border-bottom: none;
-    }
+  td {
+    font-size: 0.6rem;
+    padding-left: 16px;
   }
+}
 
-  .innerCell {
-    padding: 5px 10px;
```
julieg18, line 623: I noticed that we have the workspace cells slightly misaligned with the other rows. I assumed it wasn't intentional and deleted the line.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -80,17 +80,9 @@ $badge-size: 0.85rem;
   background-color: $header-border-color;
```
julieg18, line 80: Looked at storybook, and apparently, storybook applies table element selectors to the plots stories. With the code outside of the `.experiments` wrapper, the plots stories are now breaking. 

Are we already aware of this issue in storybook? I'm not sure why it's happening, but if it's hard to fix, we could:

* move everything back into an `experiments` wrapper. While annoying to have another wrapper around everything, it would be the quickest way to fix the problem. 
* Don't select elements at all, just rely on classes. We'd need to add some extra classes like `.td` and `.th` and update the file accordingly.

mattseddon (approved):

webview/src/experiments/components/table/styles.module.scss
mattseddon: Discuss this on our call. Need to understand why these were being applied (seems like a bug in the way we've setup the stories). What is shown in the extension should be what is shown in Storybook.

sroy3: I agree with Matt, this looks like a bug in how we setup the stories. Couldn't find where we directly import the table styles, but we directly import the plots one. We should probably clean that up and find out where these styles are coming from. That said, I think relying on classes instead of elements is good practices. I'm the one to blame here for that. During the react-table upgrade I changed all the `div`s to table elements and removed the `.td`, `.th` classes because they weren't on the right elements. I should have moved them instead.

sroy3 (approved):

Issue #3574, opened 2023-03-28
Review the new DVC sharing experiments workflow and unify it with VS Code


shcheklein:
With https://github.com/iterative/dvc/issues/9074 and https://github.com/iterative/studio/issues/5050 DVC introduced a way to connect to Studio and a workflow that sends experiments to Studio.

We need to review and unify/make it compatible:

- [x] Token management https://github.com/iterative/dvc/issues/9265 (#3768)
- [x] Auto-sharing mode - does it contradict or not with existing DVC workflow? - no, leave as is
- [x] Drop custom client implementation (#3701)
- [x] Consolidate share actions into one - https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1505741392 (#3771 + #3781)
- [x] Improvement: Sharing multiple experiments action https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1513236062 (#3792)
- [x] When a user is not already connected to Studio prompt with a toast to connect on exp push https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1513949115 (#3781)
- [x] Add instructions to live share outside of the extension context (https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1521609421) - not needed (https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1529304891)
- [x] Sharing experiments. Can we detect that it's already being shared? Show a link?

dberenbaum:
> Can we detect that it's already being shared?  Show a link?

We could fetch from the git remote or use `dvc exp list` to see what's been pushed. Do we need anything specific to Studio?

I'm a bit behind on the Studio efforts to have more shareable links, but I think we can follow it in https://github.com/iterative/studio/issues/5419 and https://github.com/iterative/studio/issues/5150.

> * Auto-sharing mode - does it contradict or not with existing DVC workflow?

I guess it means this:

<img width="372" alt="Screenshot 2023-03-29 at 3 42 58 PM" src="https://user-images.githubusercontent.com/2308172/228650823-0fdbe88a-c67d-4e04-be5a-78af29354e29.png">

I don't think it contradicts. Does it do anything besides add the `STUDIO_TOKEN` to experiment actions @mattseddon? I think it's needed since there's no other obvious way to enable this behavior when working in the extension.

> * Do we still need to use our own client implementation or not?

I hope not.

dberenbaum:
> * Sharing experiments. Can we detect that it's already being shared? Show a link?

[...]

PR #3701, opened 2023-04-17
Use exp push to share experiments to Studio

mattseddon:
# 2/6 `main` <- #3768 <- this <- #3771 <- #3777 <- #3778 <- #3779

Related to https://github.com/iterative/vscode-dvc/issues/3574.

This is an interim step where we move "Share to Studio" from our custom implementation of the studio client onto exp push. I'll remove the other 2 sharing actions shortly and rename "Share to Studio" to "Share". I'll also make it possible to share without a Studio token, under those circumstances, I will prompt the user to add a token.

### Demo

https://user-images.githubusercontent.com/37993418/234769573-512c446a-2ef2-4873-aff9-5ea51e45933f.mov


extension/src/experiments/commands/index.ts
```diff
@@ -115,10 +115,22 @@ export const getShareExperimentToStudioCommand =
       return commands.executeCommand(RegisteredCommands.SETUP_SHOW)
     }
 
-    return internalCommands.executeCommand(
-      AvailableCommands.EXP_PUSH,
-      studioAccessToken,
-      dvcRoot,
-      id
-    )
+    return Toast.showProgress('Pushing Experiment', async progress => {
```
mattseddon, line 118: [I] Improve message (currently says Pushing experiment twice)

extension/src/experiments/commands/index.ts
```diff
@@ -115,10 +115,22 @@ export const getShareExperimentToStudioCommand =
       return commands.executeCommand(RegisteredCommands.SETUP_SHOW)
     }
 
-    return internalCommands.executeCommand(
-      AvailableCommands.EXP_PUSH,
```
mattseddon, line 119: [I] Make sure to remove this from internal commands too (it is in the types).

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3702, opened 2023-04-17
Turn off add/remove branches feature flag

mattseddon:


demo
sroy3, line 1: Careful with that

sroy3 (approved):

Issue #3436, opened 2023-03-09
Plots from temporary experiments are first updated when finished


ostromann:
If you start a temporary experiment by using `dvc exp run --temp`, the plots will show the metrics of the last completed experiment in the workspace instead of the current temporary experiment.
When the temporary experiment finishes, it disappears from the plots and needs to be re-added, showing the actual metrics then.

It would be good to see the metrics of currently running temporary experiments in realtime in the Plots section.

mattseddon:
Hi @ostromann,

Thanks for raising the issue.

There is a limitation in that DVC does not collect metrics for experiments running outside of the workspace (unless they are checkpoint based). This is being actively worked on at https://github.com/iterative/dvc/issues/8478. We have #3178 in this repo for the subsequent work needed.

By any chance are you using DVCLive to log metrics?

ostromann:
Thanks!
Yes, I do use DVCLive to log metrics.

mattseddon:
Ok, as a temporary workaround, you can get live updates by connecting to Studio, switching from `--temp` to `--queue` and starting the queue with the extension. Live updates will show up in Studio for all experiments run from the queue.

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/37993418/224280186-8a5beabe-cd0a-4c06-9a8c-a44701081d8b.png">

Docs are [here](https://dvc.org/doc/studio/user-guide/projects-and-experiments/live-metrics-and-plots). 

Hope this helps. LMK if you get stuck.

Note: I think it you start `dvc exp run --temp` with the environment variables `STUDIO_TOKEN` & `STUDIO_REPO_URL` included then you should still get live updates.

shcheklein:
> As far as I understand, that rules out DVC Studio for me, right now

There is an on-prem option for Studio (it's paid, but just in case your org can consider this). In no way saying we are not fixing this, we are working on this, it's prioritized.

mattseddon:
@dberenbaum this issue is why we want to prioritise live plot updates for experiments running in the queue/temp directories. Collecting data for the separate queue/temp revisions would be problematic for the reasons mentioned in #3676. We would have to re-implement the "merging" logic, from DVC in the extension, make separate CLI calls, cache data etc. This would be a significant effort in the extension and IMO less valuable than implementing the ability to call plots diff with these revisions in DVC. I'd be happy to implement it on the DVC side if we decide it would be the best way forwards.

LMK what you think, 🙏🏻.

[...]

PR #3705, opened 2023-04-18
Fix auto-selection of running experiments

mattseddon:
# 2/2 `main` <- #3665 <- this

This PR fixes the auto selection of running experiments. The new behaviour is as follows:

1. Experiments running in the workspace are automatically selected (not the workspace record).
2. Experiments running in the queue are not automatically selected.
3. Experiments running in the queue can no longer be selected through any other means (as calling plots diff with those revisions will cause an error).

This results in a good reduction in complexity. Once the prerequisites for #3436 land we can revisit.

### Demos

https://user-images.githubusercontent.com/37993418/232690382-b887d30c-bb7d-4b72-a59d-cf1b6bc7f16f.mov

https://user-images.githubusercontent.com/37993418/232690522-2e992e1b-8a9c-4868-ada1-3e0bd07b19bd.mov

### DIFF
--- a/extension/src/experiments/model/index.ts
+++ b/extension/src/experiments/model/index.ts
@@ -4,7 +4,6 @@ import { FilterDefinition, filterExperiment, getFilterId } from './filterBy'
 import { collectExperiments } from './collect'
 import {
   collectColoredStatus,
-  collectFinishedRunningExperiments,
   collectSelected,
   collectStartedRunningExperiments
 } from './status/collect'
@@ -24,7 +23,7 @@ import {
   RunningExperiment
 } from '../webview/contract'
 import { definedAndNonEmpty, reorderListSubset } from '../../util/array'
-import { EXPERIMENT_WORKSPACE_ID, ExpShowOutput } from '../../cli/dvc/contract'
+import { ExpShowOutput } from '../../cli/dvc/contract'
 import { flattenMapValues } from '../../util/map'
 import { ModelWithPersistence } from '../../persistence/model'
 import { PersistenceKey } from '../../persistence/constants'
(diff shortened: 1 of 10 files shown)


sroy3 (approved):

extension/src/experiments/model/index.ts
```diff
@@ -136,14 +134,15 @@ export class ExperimentsModel extends ModelWithPersistence {
...
+    )
+
     if (
-      isQueued(
-        this.getExperimentsAndQueued().find(
-          ({ id: queuedId }) => queuedId === id
-        )?.status
-      )
+      experiment &&
+      (isQueued(experiment?.status) || isRunningInQueue(experiment))
```
julieg18, line 143: While nothing happens, I can still select an experiment running in the queue when I run the `Select Experiments to Display in Plots` command 

image

extension/src/plots/model/index.ts
```diff
@@ -112,8 +112,6 @@ export class PlotsModel extends ModelWithPersistence {
```
julieg18, line 112: Probably an issue on my end, but when I run an experiment, the plots break with this error: `plots diff dolce-hugs techy-hugs 53aac68fc16182b38c509da7554c22f822fc8d63 miffy-runs -o .dvc/tmp/plots --split --json failed with ERROR: unknown Git revision 'miffy-runs'`. 

https://user-images.githubusercontent.com/43496356/232789285-4a7d06cd-793c-4325-b92c-05d88d6cedb4.mov

I can't see plots until the experiment finishes running.

mattseddon: I thought I had tested this but obviously not. Seems like a bug in DVC. I'll confirm whether or not is expected and if they can provide a workaround. If not we'll be going back to the old behaviour.

mattseddon:
Will close this as the commit may still end up being useful later. For now, I'll raise a PR that disables the selection of experiments running in the queue/temp workspaces.

PR #3712, opened 2023-04-19
Prevent plotting of running experiments

mattseddon:
# 2/2 `main` <- #3665 <- this

Replaces #3705

This PR ensures that queued/running experiments cannot be selected for plotting. In the special case that an experiment is running in the workspace and the user tries to select that experiment for plotting the workspace will be selected.

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -30,13 +30,18 @@ import { starredSort } from './model/sortBy/constants'
 import { pickSortsToRemove, pickSortToAdd } from './model/sortBy/quickPick'
 import { ColumnsModel } from './columns/model'
 import { ExperimentsData } from './data'
-import { Experiment, ColumnType, TableData } from './webview/contract'
+import {
+  Experiment,
+  ColumnType,
+  TableData,
+  isRunning
+} from './webview/contract'
 import { WebviewMessages } from './webview/messages'
 import { DecorationProvider } from './model/decorationProvider'
 import { starredFilter } from './model/filterBy/constants'
 import { ResourceLocator } from '../resourceLocator'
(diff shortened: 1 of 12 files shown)


extension/src/experiments/index.ts
```diff
@@ -217,15 +222,6 @@ export class Experiments extends BaseRepository<TableData> {
   public toggleExperimentStatus(
     id: string
   ): Color | typeof UNSELECTED | undefined {
-    if (
```
mattseddon, line 220: [F] Moved this down into the model so that all of the logic is together.

extension/src/experiments/index.ts
```diff
@@ -353,8 +349,10 @@ export class Experiments extends BaseRepository<TableData> {
     return this.experiments.getWorkspaceCommitsAndExperiments()
   }
 
-  public async selectExperiments() {
-    const experiments = this.experiments.getWorkspaceCommitsAndExperiments()
+  public async selectExperimentsToPlot() {
```
mattseddon, line 352: [F] This is the only thing this function is used for so I think the name better explains the usage

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -2072,121 +2099,4 @@ suite('Experiments Test Suite', () => {
       expect(mockUpdateExperimentsData).to.be.calledOnce
     })
   })
-
-  describe('checkForFinishedWorkspaceExperiment', () => {
```
mattseddon, line 2076: 🤦🏻

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3716, opened 2023-04-19
Clean up shared styles

julieg18:
* removes `.storybook/styles.scss`
* renames `shared/style.scss` to `shared/styles.scss`
* removed unneeded import of `shared/styles.scss`
* adds `mixins.scss` for holding reusable mixins

Followup to #3696


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -1,7 +1,7 @@
 // Variables
 
 @import '../../../shared/variables.scss';
-@import '../../../shared/style.scss';
+@import '../../../shared/styles.scss';
```
sroy3, line 4: I don't think this line is necessary as it is already imported in `webview/src/experiments/index.tsx`. There doesn't seem to be any variables or classes to extend in there.

webview/src/shared/components/sectionContainer/styles.module.scss
```diff
@@ -1,5 +1,5 @@
 @import '../../variables.scss';
-@import '../../style.scss';
+@import '../../styles.scss';
```
sroy3, line 2: Same here, I don't think we need to import this.

julieg18: Good catch!

webview/src/experiments/components/table/styles.module.scss
julieg18: Looks like we are extending an `a` element, which should be done in a mixin... I'll go ahead and create a `shared/mixins.scss` file similar to our `variables.scss` file.

PR #3715, opened 2023-04-19
Trigger plot updates whenever commit data changes

mattseddon:
# 2/2 `main` <- #3665 <- this 

This PR ensures that `plots diff` is recalled when a branch's data changes. I.e when a user makes a commit or runs `git reset`.

### Demo 

https://user-images.githubusercontent.com/37993418/233038102-28efc6c6-db1d-4b32-8735-d2f3b62fd039.mov

PR #3684, opened 2023-04-14
Remove checkpoints model and file system watcher

mattseddon:
# 2/2 `main` <- #3665 <- this

This PR accounts for a couple of the TODO items in #3665. Namely

- Drop full vs partial `exp show` updates
- Delete `hasCheckpoints`/`checkpoints` watcher and use {data: {meta: hasCheckpoints}}

PR #3723, opened 2023-04-20
Match the way Studio displays experiment name information

mattseddon:
- duplicate transform and set function with new data
- duplicate functions out to reader (tests will break)
- fix all integration tests
- update all remaining test fixtures
- fix all test data and deduplicate functions
- remove some code climate duplication
- wrap all loose test data in generator
- use breaking change version of dvc
- fix experiment status in dvc contract
- self review
- Remove checkpoints model and file system watcher (#3684)
- update demo project
- Extend timeout of run experiment test (e2e) (#3713)
- Prevent plotting of running experiments (#3712)
- Trigger plot updates whenever commit data changes (#3715)
- Match the way Studio displays experiment name information


webview/src/experiments/components/table/CellSecondaryName.tsx
```diff
@@ -1,58 +0,0 @@
-import React from 'react'
```
mattseddon, line 1: [F] Moved a simplified version of this into `ExperimentCell`

webview/src/experiments/components/table/ExperimentCell.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  return (
+    <Tooltip
+      placement="bottom-start"
+      appendTo={document.body}
+      content={tooltipContent}
+    >
+      <div className={styles.experimentCellText}>
+        <span>{label}</span>
+        <br />
```
sroy3, line 48: Can these be two `<div> tags or have `display: block;` on them instead of using a `<br />`. It's usually best to avoid `<br />` tags because it isn't semantic.

sroy3 (approved):
Is it normal that the comparison table images changed in Storybook?

webview/src/experiments/components/table/ExperimentCell.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return commit ? (
+    <CommitCell
+      commit={commit}
+      description={description}
+      label={label}
+      sha={sha}
+    />
+  ) : (
+    <div className={styles.experimentCellText}>
+      <span
```
julieg18, line 90: I was staring at the table for like ten minutes, trying to figure out why it looked so off to me when it finally hit me that the commit rows have secondary text second while the experiments have their secondary text first. 

I'm assuming most people won't notice 😅, but maybe we should keep secondary text second across all rows? Example of how this would look:

image

Technically, in design, you can have secondary text first depending on other factors but the main reason that the rows were bothering me is the fact we have a mixture of secondary being first and second. Though again, I'm probably the only person getting bothered :)

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -729,15 +730,12 @@ $badge-size: 0.85rem;
 
 .experimentCellText {
   @extend .cellContents;
-  display: flex;
-  flex-flow: row wrap;
```
julieg18, line 733: If we're going to keep things stacked on top of each other, I think we need to give the rows more height. Things look pretty crowded vertically when they're wrapped: 

image

julieg18 (approved):

webview/src/experiments/components/table/ExperimentCell.tsx
mattseddon: I know that this breaks design principles. The reason that we've arrived here is that previously experiment names were less useful (e.g `exp-43234`) and the sha was thought of as the primary information identifier. The DVC team tried to solve the problem of not being able to navigate a lot of experiments by making the name easier to read/more memorable. The name as since become the primary identifier. However, the CLI still displays the information in the same way:

image

Studio have addressed this problem in the following way:

image

For more context please read https://github.com/iterative/vscode-dvc/issues/3619#issuecomment-1510507639 & https://github.com/iterative/vscode-dvc/issues/3619#issuecomment-1510514599 but the tl;dr is - we display this information in a lot of places, we should be standardising between the products, having the name first is a better option but a much larger change across multiple products.

julieg18: Oh, ok, makes sense! I didn't realize we were interested in matching order as well as focus across the extension! Hmmm, in that case, we could make secondary text first across all the rows: 

image

Or just leave as is, up to you!

mattseddon:
This and #3725 can wait until after the release of #3665 (`0.8.0`)

webview/src/experiments/components/table/ExperimentCell.tsx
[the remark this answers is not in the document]
mattseddon: I split up the components into some different files. Please LMK what you think.

mattseddon:
Merging this as I think it is an improvement. We can iterate on it further as we get more feedback.

Issue #3286, opened 2023-02-15
Commit action doesn't work in SCM tree


shcheklein:
This repo / Codespace: https://github.com/iterative/iterative.ai

https://user-images.githubusercontent.com/3659196/218896789-4d4c1d7d-cce7-4ec8-804c-6256661550a1.mov

More weird examples:

(after reloading the container, bad error message + we are not trying to repeat + it doesn't allow to do it again):

https://user-images.githubusercontent.com/3659196/218897342-b7a28527-b2db-457e-82bd-6522edfee1a0.mov

Finally, when it works, the message is "ERROR", but also a warning. Something about stage - it's all very scary and confusing (cc @dberenbaum )

https://user-images.githubusercontent.com/3659196/218898967-25226cfd-297d-482c-bdd4-9554949df4f9.mov

shcheklein:
Make it p1 since it breaks the experience severely when we deal with blog.

mattseddon:
What does `dvc data status --granular --unchanged --json` output under these circumstances?

mattseddon:
@shcheklein, this should be fixed with the latest version. Please reopen if I have misunderstood or missed something 🙏🏻.

PR #3726, opened 2023-04-20
Force commit action by default

mattseddon:
This PR add `-f` to all of our calls to `dvc commit`. This change is based on https://github.com/iterative/vscode-dvc/issues/3286#issuecomment-1476379871

> `dvc commit` is one of the few commands that's interactive by default and always asks "Are you sure you want to commit?" if there are changes. TBH I'm not sure it's necessary at this point (looks like it is very old, see https://github.com/iterative/dvc/pull/1601). IMO for VS Code we should do dvc commit -f by default here.


sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3728, opened 2023-04-20
Fix warnings produced by integration tests

mattseddon:


extension/src/test/suite/vscode/walkthrough.test.ts
```diff
@@ -16,14 +16,19 @@ suite('Walkthrough Test Suite', () => {
 
   describe('dvc.getStarted', () => {
     it('should be able to show the walkthrough', async () => {
```
mattseddon, line 18: [F] I did double-check that the command actually works:

https://user-images.githubusercontent.com/37993418/233264604-5333e24b-08ea-464c-aee4-e66d2eac1c3e.mov

No idea why the error was being produced

julieg18 (approved):

sroy3 (approved):

PR #3735, opened 2023-04-21
Add multiple branches support inside the experiments table

sroy3:
Part of #1966 

When there will be multiple branches, it will look like:

https://user-images.githubusercontent.com/3683420/233658615-c8990cc3-6857-4475-a105-3390f5739a8a.mov

Might still need to figure out where to move the "Select branches to show" (not shown here, but will look as before) and the "Switch to all branches view"

If there is only one branch (the current branch and the current behaviour, it will look like that (same as before:
Screenshot 2023-04-21 at 9 19 25 AM


webview/src/experiments/components/table/body/TableContent.test.tsx
```diff
@@ -0,0 +1,858 @@
...
+            },
+            selected: false,
+            sha: '29ecaaf3adf216045e96e81fb8e3027c9122af52',
+            starred: false
+          },
+          subRows: []
+        }
+      ]
+    })
+  } as unknown as Table<Experiment>
```
mattseddon, line 809: [I] When we add the extension part of this it would be good to add more commits/at least one extra branch to the base test fixture.

webview/src/experiments/components/table/body/branchDivider/styles.module.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../../../../shared/variables.scss';
+
+.branchName {
+  padding: 40px 20px 10px;
```
mattseddon, line 4: [C] The next problem we will have to solve is: "how can a user compare two separate records on different branches". For me, this means we need a mechanism to display them directly above/below one another. Maybe we could allow pinning or we could start with something like "show only selected" as a multi-select option.

webview/src/experiments/components/table/body/commitsAndBranches/CommitsAndBranchesNavigation.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+            <span className={styles.separator} />
+
+            <button
+              className={styles.commitsAndBranchesNavButton}
+              onClick={
+                isBranchesView ? switchToCommitsView : switchToBranchesView
+              }
+            >
+              {isBranchesView
```
mattseddon, line 53: [Q] Do we need to/should we keep this around long term?

webview/src/experiments/components/table/body/TableContent.test.tsx
mattseddon: [Q] Would also mean we would get the design added to a Story, right?

webview/src/experiments/components/Experiments.tsx
```diff
@@ -160,13 +160,17 @@ export const ExperimentsTable: React.FC = () => {
     autoResetAll: false,
     columnResizeMode: 'onChange',
     columns: columns as ColumnDef<Commit, unknown>[],
-    data,
+    // Setting any branch for now just so that it isn't undefined. It does not matter the label for now as it is not shown
```
mattseddon, line 163: [Q] Can we track that this needs to be removed/updated somewhere?

mattseddon (approved):
Looks good. Left some comments. I do think we will need to provide a way for users to display selected rows from the table closer to each other. Also, need to consider how delta mode fits in with this as well.

I don't think it would hurt to add the `Git-merge` icon in with the other indicators (Plot, filter, sort) and have the action fire up "Select Branches" as well. The badge could relate to the number of branches selected. The idea is that it will be complementary to the text at the bottom of each branch, not replace it.

🙏🏻

webview/src/experiments/components/table/body/TableContent.test.tsx
sroy3: Yes, as soon as there is more than one branch, the new design appears. I think it might be good to add a story for it right away, so that it's possible to see it without faking it.

webview/src/experiments/components/table/body/commitsAndBranches/CommitsAndBranchesNavigation.tsx
sroy3: I think people in general feel like this won't be necessary anymore, I don't mind removing it once the whole feature is in. That being said, I personally feel like this could be useful before selecting branches to add as it gives the user an overview of what is in each branch. Also if you do not particularly case about previous commits and have too many branches to easily select, this is a quick way to view everything at once. 

This isn't too complicated to remove, nor it is to keep it. The most problematic part about keeping it is where to show it now that we have controls on every branch. I have no preferences about it.

webview/src/experiments/components/table/body/PreviousCommitsRow.tsx
```diff
@@ -0,0 +1,22 @@
...
+interface PreviousCommitsRowProps {
+  isBranchesView?: boolean
+  nbColumns: number
+}
+
+export const PreviousCommitsRow: React.FC<PreviousCommitsRowProps> = ({
+  isBranchesView,
+  nbColumns
+}) => (
+  <tbody>
```
julieg18, line 14: Same question for here as with the git branch row about this being in its own separate body.

julieg18: Just realized that this row is currently a `tbody` in `main` 🤔 Must have changed from a header row at some point :)

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -834,28 +834,6 @@ $badge-size: 0.85rem;
```
julieg18, line 834: The table bottom border is showing on the table:

image

Though in my opinion, this removed border is minor and we can just keep the border on all rows if it's hard to implement. If we decide this, we'll need to remove the corresponding CSS though.

webview/src/experiments/components/table/body/TableBody.tsx
julieg18, line 1: Is there a way to view how the different branch tables look on my end? Looking at the code, I'm not seeing a way to do this, but I might be missing something :) 

From your demo, it looks like the experiment column border might be missing but I can't confirm this since, as mentioned, I'm not sure how to test how things look.

julieg18 (approved):

webview/src/experiments/components/table/body/commitsAndBranches/CommitsAndBranchesNavigation.tsx
mattseddon: Trying to limit/fight complexity.

webview/src/experiments/components/table/body/TableBody.tsx
sroy3: Currently adding a story to see it. Otherwise, you have to manually add branches to the rows (we can duplicate the non workspace rows we send to `updateRows` in `App.tsx`)

webview/src/experiments/components/table/body/TableContent.tsx
```diff
@@ -52,20 +54,42 @@ export const TableContent: React.FC<TableContentProps> = ({
...
-          root={tableRef.current}
-          row={row}
-          instance={instance}
-          key={row.id}
-          hasRunningExperiment={hasRunningExperiment}
-          projectHasCheckpoints={hasCheckpoints}
-          batchRowSelection={batchRowSelection}
-        />
-      ))}
+      {branches.map((branch, branchIndex) => {
```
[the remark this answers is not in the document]
sroy3: I already moved it to be it's own component. That's what `TableContent` is.

PR #3737, opened 2023-04-21
Fix e2e tests failing 

julieg18:
* extends one of the timeouts

### DIFF
--- a/extension/src/test/e2e/extension.test.ts
+++ b/extension/src/test/e2e/extension.test.ts
@@ -129,7 +129,7 @@ describe('Experiments Table Webview', function () {
     expect(finalRows.length).toStrictEqual(initialRows + experimentRow)
     await webview.unfocus()
     await closeAllEditors()
-    await waitForDvcToFinish(90000)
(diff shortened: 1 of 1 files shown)


extension/src/test/e2e/extension.test.ts
```diff
@@ -129,7 +129,7 @@ describe('Experiments Table Webview', function () {
     expect(finalRows.length).toStrictEqual(initialRows + experimentRow)
     await webview.unfocus()
     await closeAllEditors()
-    await waitForDvcToFinish(90000)
+    await waitForDvcToFinish(120000)
```
julieg18, line 132: Increased the timeout that was failing but is there a better way to solve this?

sroy3 (approved):

extension/src/test/e2e/extension.test.ts
mattseddon: Thank you.

This amount of time is a bit concerning (90000) already was. I hope there isn't something else going on 🤞🏻.

PR #3744, opened 2023-04-24
Upgrade Storybook to v7

mattseddon:
Make the jump to v7 of Storybook. PTAL


webview/src/stories/ComparisonTable.stories.tsx
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
-import { Meta, Story } from '@storybook/react/types-6-0'
+import type { StoryFn, Meta } from '@storybook/react'
```
sroy3, line 2: TIL about `import type`. That's really neat!

sroy3 (approved):
LGTM! The only thing that seem to have changes is the font of buttons. I don't know if it's because it's inheriting something different. Maybe we want to look at its style make sure it does not get affected by outside factors, but that's somewhat out of scope and very minor.

webview/package.json
```diff
@@ -72,6 +73,7 @@
     "raw-loader": "4.0.2",
     "sass": "1.62.0",
     "sass-loader": "13.2.2",
+    "storybook": "7.0.7",
     "storybook-addon-designs": "6.3.1",
```
julieg18, line 77:
```
Attention: We've detected that you're using the following addons in versions which are known to be incompatible with Storybook 7:

- storybook-addon-designs@6.3.1

Please be aware they might not work in Storybook 7. Reach out to their maintainers for updates and check the following Github issue for more information:
https://github.com/storybookjs/storybook/issues/20529
```

Got this warning when I ran storybook. We can probably just remove the addon. We're only using it for one storybook and it's not working since the Figma URL is outdated, I believe.

mattseddon: Done

mattseddon: (I did see this, I just ignored it)

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -110,6 +110,9 @@ $badge-size: 0.85rem;
 }
 
 .experiments {
+  position: fixed;
+  top: 0;
+  left: 0; // position, top and left are set to fix indicators in storybook
```
mattseddon, line 115: @sroy3 @julieg18 any issue with adding this to "fix" the stories? There is no change in the production version of the app. Usage is here: https://github.com/iterative/vscode-dvc/blob/main/webview/src/experiments/components/Experiments.tsx#L207-L208

mattseddon: The answer is no it breaks scrolling

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -203,13 +203,14 @@ $badge-size: 0.85rem;
 // table head styles
 
 .tableIndicators {
```
mattseddon, line 205: @sroy3, @julieg18 I've moved the `<Indicators />` from the global scope into `<TableHeaderCell />` are we ok with this?

julieg18 (approved):

Issue #3757, opened 2023-04-25
Multi view plots do not resize correctly


sroy3:
Vegalite has [limitations](https://vega.github.io/vega-lite/docs/size.html) on responsive resizing (using the size of the plot's container). These limitations are on multi view plots. There are multiple workarounds to fix this. Some might fix everything, but might be to expensive performance-wise. Others might be more performant, but not make the plots inside resize like other plots completely.

- Change the size in the template. This will have to be done on each window resize as well as plot resize after the size is calculated. The size will need to be calculated in the webview, but the template is in the extension. A delay might appear between the resize of the window or plots and the resize of the actual plot. Changing the template on the webview side might be even worse as it might block the thread.
- We can edit the size of the SVG after its creation. We can go and resize everything needed, but that might be really complicated not knowing how VegaLite calculates everything. We can at least make sure the width and height of the SVG are not smaller than the dimensions of its container.
- We can look into the internals of VegaLite and either fork, patch, or open a pull request to fix the problem.

sroy3:
I'll create quick prototypes to test for the best solution.

sroy3:
There is already an [issue](https://github.com/vega/vega-lite/issues/5219) opened for this on the Vegalite repo. There is even a [PR](https://github.com/vega/vega-lite/pull/6672) that is requiring some [help](https://github.com/vega/vega-lite/pull/6672#issuecomment-898960194) getting merged. We can maybe take a look at it and help if we can.

mattseddon:
Decision needs to be made on whether or not we will move to Plotly and when.

mattseddon:
> Change the size in the template. This will have to be done on each window resize as well as plot resize after the size is calculated. The size will need to be calculated in the webview, but the template is in the extension. A delay might appear between the resize of the window or plots and the resize of the actual plot. Changing the template on the webview side might be even worse as it might block the thread.

@sroy3 do you remember if this calculation would be difficult? If we reliably knew where to insert the correct width for a single revision into the plot would this be doable? For some context, we are going to be filling the template on the webview side now. We're also adding anchors for a few different things. If we had an anchor to update the width of an individual plot would that make this achievable?

sroy3:
> > Change the size in the template. This will have to be done on each window resize as well as plot resize after the size is calculated. The size will need to be calculated in the webview, but the template is in the extension. A delay might appear between the resize of the window or plots and the resize of the actual plot. Changing the template on the webview side might be even worse as it might block the thread.
[...]

PR #3761, opened 2023-04-25
Make the multiview template size dynamic for each plot inside

sroy3:
Prototype for #3757 

https://user-images.githubusercontent.com/3683420/234396857-3d7a21a0-393f-47a1-ab3c-3d355fb8f2c7.mov

This might actually work. This is better than I thought it would, probably because we already parse the template anyway at that time and calculation is straight forward as it had almost everything it needed.

There are some adjustments left, but this is just a prototype, I'll fix that in a secondary step. I first need to check if having multiple multi view plots would make this look laggy.

The one thing I don't like is the duplication of aspect ratios in the webview and the extension. If we change it some place, we'll forget to change at the other place. I'll try to make it so it's only in one place.

### DIFF
--- a/extension/src/plots/model/collect.ts
+++ b/extension/src/plots/model/collect.ts
@@ -269,8 +269,8 @@ const transformRevisionData = (
 ): { revisions: string[]; datapoints: unknown[] } => {
   const field = multiSourceEncodingUpdate.strokeDash?.field
   const isMultiSource = !!field
-  const availableRevisions = selectedRevisions.filter(rev =>
-    Object.keys(revisionData).includes(rev)
+  const availableRevisions = selectedRevisions.filter(
+    rev => revisionData[rev]?.[path]
   )
   const transformNeeded =
     isMultiSource && (isMultiView || isConcatenatedField(field))
@@ -278,7 +278,7 @@ const transformRevisionData = (
   if (!transformNeeded) {
     return {
       datapoints: selectedRevisions
-        .flatMap(revision => revisionData?.[revision]?.[path])
+        .flatMap(revision => revisionData[revision]?.[path])
         .filter(Boolean),
       revisions: availableRevisions
     }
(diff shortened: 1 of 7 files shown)


sroy3:
https://user-images.githubusercontent.com/3683420/234596481-8da49421-ae98-4d06-9590-de63ff8eb959.mov

I'll need to add some debounce to the call to make this actually useful. Else it's going to move way too much.

sroy3:
I don't think this is going to work after all. The problem is that when changing the template, we force the re-rendering of the React component. This will result in something very laggy, the more plots we have (especially with virtualized plots). The problem really is inside VegaLite and our templates were never the problem.  To fix this we need to fix it after VegaLite has created the component or fix it inside VegaLite directly.

See [video](https://drive.google.com/file/d/1pyjTDJZqapH1dgpFKxBGlt0HzEZ0fpjc/view?usp=share_link) (too big for GitHub)

shcheklein:
> The problem is that when changing the template, we force the re-rendering of the React component. This will result in something very laggy, the more plots we have (especially with virtualized plots).

It's still better vs what we have now (not responding at all to resize). Even if it takes substantial amount of time to re-render. If that happens only when we change size- I think it's better to have something then.

> The problem really is inside VegaLite and our templates were never the proble

Let's create a ticket on their end to discuss it? How would you expect it to handle this? I think it would be great to describe / scope it.

shcheklein:
>  To me, it does not feel like a great experience at all.

Yep. The way you are describing it it's not good indeed. What I struggle to understand why it's not possible to implement it with changing the template on the resize w/o these delays, etc, etc.

> I made a separate fix last week that is also seen in this PR (https://github.com/iterative/vscode-dvc/pull/3774), which probably helps in making this look better.

tbh, not sure I understand how the fix is related. Could you explain it please?

PR #3763, opened 2023-04-26
Add CLI incompatible screen to Studio section

mattseddon:


webview/src/setup/components/studio/CliIncompatible.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
```
mattseddon, line 1: Feels like we're trying to head in the direction of splitting out components. LMK if I got this wrong.

sroy3: This is going to help greatly in keeping styles simple and easy to clean out. It'll also help with Codeclimate issues and even tests.

webview/src/setup/components/studio/Connect.tsx
```diff
@@ -0,0 +1,49 @@
...
+          onClick={openStudioProfile}
+        />
+        <Button
+          appearance="secondary"
+          isNested={true}
+          text="Save"
+          onClick={saveStudioToken}
+        />
+        <p>
+          {"Don't Have an account?\n"}
```
sroy3, line 43: Instead of adding a line break, I would add a `display: block;` on the following link. (It's about content vs. style).

sroy3 (approved):

julieg18 (approved):
Looks good!

webview/src/setup/components/studio/Connect.tsx
mattseddon: this line break didn't even work. I have removed

PR #3765, opened 2023-04-26
Make the table more compact with branches and actions

sroy3:
Part of #1966 

Moving the actions at the bottom of the branch section into existing row to make things more compact.

Screenshot 2023-04-26 at 3 50 04 PM

Default version with only one branch:

https://user-images.githubusercontent.com/3683420/234686822-97ad5c06-6951-46f1-8360-480c3f18b111.mov

Other than moving the more / less commits actions inside of the previous commits row and select branches into the branch row, and making the branch row the same size as the previous branches rows, there a few other changes:

- Branches row is now always there (needed for the select branches and commits action)
- ~~Previous commits row is also always there  (needed for show / more or less commit).~~ Previous commits row was removed


shcheklein:
It's a good improvement, still feels like a lot of space is taken by all these actions and auxiliary rows. I think we need to combine them into one (pretty much get rid of the previous commits one).

sroy3:
> It's a good improvement, still feels like a lot of space is taken by all these actions and auxiliary rows. I think we need to combine them into one (pretty much get rid of the previous commits one).

That does make sense. With actions like show more / less commits, it's less important to show those rows as "Previous commits" we added it in the first place because it wasn't clear what these rows were. Should we keep the default previous commits to 2 or 0 in that case?

shcheklein:
> Should we keep the default previous commits to 2 or 0 in that case?

I think we can keep the defaults the same for now (and revisit if needed). It's 2 additional commits, right?

webview/src/experiments/components/table/body/branchDivider/BranchDivider.tsx
```diff
@@ -18,6 +19,9 @@ export const BranchDivider: React.FC<PropsWithChildren> = ({ children }) => (
           {children}
         </div>
       </th>
+      <th colSpan={9999} className={styles.branchActions}>
```
mattseddon, line 22: [Q] why do we need `9999`?

mattseddon (approved):

webview/src/experiments/components/table/body/branchDivider/BranchDivider.tsx
sroy3: It makes the cell take all the available space (as long as there are less than 1000 cells, I guess at that point we don't care anyway). Instead of calculating the number of cells and passing that information around as props.

mattseddon:
[Q] How about adding "global" controls to a small sticky row underneath the workspace? This would give us space to add the control for "delta mode" when we get to it as well.

sroy3:
> [Q] How about adding "global" controls to a small sticky row underneath the workspace? This would give us space to add the control for "delta mode" when we get to it as well.

Not sure about this. It would take up space vertically all the time for only one action.

webview/src/experiments/components/table/body/commitsAndBranches/CommitsAndBranchesNavigation.tsx
```diff
@@ -17,46 +16,38 @@ export const CommitsAndBranchesNavigation: React.FC = () => {
...
-                className={styles.commitsAndBranchesNavButton}
-                onClick={showLessCommits}
-                disabled={isBranchesView}
-              >
-                Show Less Commits
-              </button>
-            )}
+    <div className={styles.commitsAndBranchesNav}>
+      {hasMoreCommits && (
+        <button
```
julieg18, line 21: Have we already considered/tried making these actions more button-like instead of link-like? 

Screenshot 2023-04-27 at 8 59 25 AM

With these actions now more immersed into the table verses at the bottom, I noticed that we don't really use this link design anywhere else in the table 🤔

julieg18 (commented):
Great work! I love how things are more compact but still clear!

webview/src/experiments/components/table/body/commitsAndBranches/CommitsAndBranchesNavigation.tsx
sroy3: These were buttons first. Maybe @shcheklein can tell more about the reasoning behind the link-like buttons.

shcheklein: I don't have any insights on this :) It just felt simpler at that point - full fledged side-panel like button will feel too noisy I think in that amount inside the table. We can definitely / and should think about some icons or lightweight elements / button. Any ideas folks?

PR #3768, opened 2023-04-27
Use dvc config to store and access studio.token

mattseddon:
# 1/6 `main` <- this <- #3701 <- #3771 <- #3777 <- #3778 <- #3779

One of the checkboxes from #3574

This PR switches us from using VS Code's secret storage API to the `dvc config` command for storing a user's `studio.token`.

### Demo

https://user-images.githubusercontent.com/37993418/234763816-2c837426-de1c-445a-a078-474ae588c5e8.mov

Note: We are not currently supporting users using `config.local` to store a `studio.token` value in monorepos.


extension/src/cli/dvc/executor.ts
```diff
@@ -71,6 +72,10 @@ export class DvcExecutor extends DvcCli {
     return this.blockAndExecuteProcess(cwd, Command.COMMIT, ...args, Flag.FORCE)
   }
 
+  public config(cwd: string, ...args: Args) {
```
mattseddon, line 75: [F] Config doesn't really fit with the `reader`/`executor` model as it can both read and write the config. For now, I've put it into `executor`.

extension/src/setup/index.ts
```diff
@@ -300,11 +281,54 @@ export class Setup
...
+      return this.internalCommands.executeCommand(
+        AvailableCommands.CONFIG,
+        cwd,
+        Flag.GLOBAL,
+        Flag.UNSET,
+        DvcConfigKey.STUDIO_TOKEN
+      )
+    }
+
+    const cwd = this.dvcRoots[0]
```
mattseddon, line 303: [F] If there is only 1 dvc root in the workspace then we try to clear out the local and global configs. If the user has set the token into their `--system` config this won't work but I think this approach will get us close enough with the vast majority of users using the extension to manage their token.

extension/src/setup/index.ts
```diff
@@ -579,6 +616,10 @@ export class Setup
     return this.workspaceChanged.fire()
   }
 
+  private getCliCompatible() {
```
mattseddon, line 619: [F] Had to wrap this so I could stub it.

extension/src/setup/index.ts
```diff
@@ -604,22 +646,54 @@ export class Setup
...
 
-  private getSecrets() {
-    return this.secrets
+  private async setStudioAccessToken() {
+    if (!this.getCliCompatible()) {
+      this.studioAccessToken = undefined
+      return
+    }
+
+    if (this.dvcRoots.length !== 1) {
```
mattseddon, line 679: [F] same thing here we are not supporting `--local` wrt to multiple projects being in the workspace.

extension/src/util/appdirs.test.ts
```diff
@@ -0,0 +1,59 @@
...
+describe('getDVCAppDir', () => {
+  it('should return the correct path for non-windows platforms', () => {
+    mockedJoin.mockImplementation((...paths: string[]) => paths.join('/'))
+    const mockedAppDir = '/app/dir'
+    mockedUserConfigDir.mockImplementationOnce((path: string) =>
+      join(mockedAppDir, path)
+    )
+    expect(getDVCAppDir()).toStrictEqual(mockedAppDir + '/' + 'dvc')
+  })
+  it('should return the correct path on Windows', () => {
```
julieg18, line 48: Minor, but don't we tent to add lines between `it` blocks?

extension/src/setup/index.ts
```diff
@@ -314,12 +338,19 @@ export class Setup
       return
     }
 
-    return this.storeSecret(STUDIO_ACCESS_TOKEN_KEY, token)
+    await this.internalCommands.executeCommand(
+      AvailableCommands.CONFIG,
+      cwd,
+      Flag.GLOBAL,
```
julieg18, line 344: Should we mention somewhere that the Studio token is saved globally? Or that users with `--system` can't remove their tokens?

julieg18 (approved):
Great work!

extension/src/setup/index.ts
mattseddon: The docs say this:

image

I did consider trying to accommodate "all possibilities" but really we are trying to help beginners with these actions. If the user is competent enough to save a token in their `--system` config then I do not think they'll try to manage it in this way. I feel like even trying to delete `--local` is really going above and beyond + makes the code messy enough.

mattseddon: if anyone complains we will revisit

PR #3769, opened 2023-04-27
Watch all workspace folders for changes in dot folders

mattseddon:
I noticed this whilst working on #3768. The explanation for the change is inline.


extension/src/setup/index.ts
```diff
@@ -546,21 +546,17 @@ export class Setup
...
     const disposer = Disposable.fn()
     this.dotFolderWatcher = disposer
     this.dispose.track(this.dotFolderWatcher)
 
-    return createFileSystemWatcher(
-      disposable => disposer.track(disposable),
-      getRelativePattern(cwd, '**'),
-      path => this.dotFolderListener(disposer, path)
-    )
+    for (const workspaceFolder of getWorkspaceFolders()) {
```
mattseddon, line 553: [F] There was a small chance of a bug here when the user has multiple workspace folders. We can easily watch multiple folders for changes.

sroy3 (approved):

julieg18 (approved):

Issue #3758, opened 2023-04-25
workspace duplicating colors of existing experiment


dberenbaum:
I'm seeing the same color for both the workspace and an existing experiment. See the screenshots below:

<img width="982" alt="Screenshot 2023-04-25 at 2 56 45 PM" src="https://user-images.githubusercontent.com/2308172/234375930-a5b3cad9-1462-430b-88be-2df7556eadf0.png">
<img width="983" alt="Screenshot 2023-04-25 at 2 56 55 PM" src="https://user-images.githubusercontent.com/2308172/234375942-187d97da-1919-49dc-a9f3-40b6c6accfd4.png">

dberenbaum:
@mattseddon You can see it around the 7:00 mark in https://iterativeai.slack.com/files/U9JC3123B/F054HL71DJB/ah_20230427.mov.

mattseddon:
Ok, so this is an experiment running in the "DVCLive" only context through a notebook (I did notice this when first watching the video). The good news (I think) is that the workspace looks like it was auto-"de-selected" here:

<img width="932" alt="image" src="https://user-images.githubusercontent.com/37993418/234838555-4c3a44dd-580d-4e48-8730-f974fa6c4045.png">

It should be showing up with a "new color" for a new experiment. I should be able to fix this.

dberenbaum:
Also note how the `bonny-lama` experiment above didn't get auto-selected but the other did.

mattseddon:
... way more broken than I thought it was.

Can you share the project with me?

dberenbaum:
Sure, it's in https://github.com/iterative/dvclive-exp-tracking.

mattseddon:
Currently unable to recreate this even when using the above project:

https://user-images.githubusercontent.com/37993418/235022208-49c37190-0c97-47bc-9da4-1da02dc48ffe.mov

I will put in some more defensive code that attempt to clean this up when it happens.

dberenbaum:
I have deleted experiments a bunch of times while iterating here. Do you think that could be the problem? I wouldn't mark it high priority if it's not occurring in more normal workflows.

PR #3780, opened 2023-04-28
Ensure duplicate colors are not made available during collection

mattseddon:
Fixes https://github.com/iterative/vscode-dvc/issues/3758

I think the issue was caused by a color being inadvertently made available twice/duplicated. Once this happened the color selection model completely breaks. I was unable to recreate locally but I have added tests around the issue and also made sure that we completely fix any existing issue when the user reloads VS Code.


julieg18 (approved):
Looks good!

PR #3781, opened 2023-04-28
Integrate remaining share experiment command with the extension (exp push)

mattseddon:
# 1/3 `main` <- this <- #3792 <- #3793

Relates to #3574.

This PR integrates `exp push` with the extension. As per [this](https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1528003781) I've renamed the action from "Share" to "Push".

#### Share with clickable link

https://user-images.githubusercontent.com/37993418/235383652-5fba2424-848f-4434-9d37-736465c6fdb3.mov

#### Error

https://user-images.githubusercontent.com/37993418/235383686-a08d3f9e-15c2-411d-8b41-665063404c2f.mov


extension/src/commands/external.ts
```diff
@@ -100,7 +100,6 @@ export enum RegisteredCommands {
   ADD_STUDIO_ACCESS_TOKEN = 'dvc.addStudioAccessToken',
   UPDATE_STUDIO_ACCESS_TOKEN = 'dvc.updateStudioAccessToken',
   REMOVE_STUDIO_ACCESS_TOKEN = 'dvc.removeStudioAccessToken',
-  EXPERIMENT_VIEW_SHARE_TO_STUDIO = 'dvc.views.experiments.shareExperimentToStudio',
```
mattseddon, line 103: `EXPERIMENT_VIEW_PUSH` was already set above 😕

extension/src/experiments/commands/index.ts
```diff
@@ -3,37 +3,76 @@ import { AvailableCommands, InternalCommands } from '../../commands/internal'
...
     const studioAccessToken = setup.getStudioAccessToken()
-    if (!studioAccessToken) {
-      return commands.executeCommand(RegisteredCommands.SETUP_SHOW)
+    if (
+      !(
+        getConfigValue(ConfigKey.DO_NOT_RECOMMEND_ADD_STUDIO_TOKEN) ||
+        studioAccessToken
+      )
+    ) {
+      void promptToAddStudioToken()
```
mattseddon, line 55: [F] This is non-blocking. User can push as many experiments as they like without setting the token.

skshetry:
@mattseddon, what do you think about making "View in Studio" to be a button like in the case of error?

mattseddon:
> @mattseddon, what do you think about making "View in Studio" to be a button like in the case of error?

In order to use a button for that link I'd have to switch APIs and send another toast message after the progress notification completes/closes. In the case of an error, the progress notification closes unexpectedly and we fallback to some existing behaviour.

julieg18 (approved):
Great work!

sroy3 (approved):

shcheklein:
LGTM! 

- One question - do we plan to have a separate "Share" action or do we expect users to understand that Push does this - essentially solves their high level use case - to see an experiment from Studio? cc @dberenbaum (I'm bit worried that it's now too low level)

- An experiment can be auto-pushed, right? And live-shared - does it affect this action? Probably all of this comes down to us being able to show a link / status of an experiment.

Not blockers, some followup questions to discuss.

mattseddon:
> Looks great @mattseddon! Agree with @skshetry that error handling is my one question here. When the cache fails, not showing what has been successful (pushing to GitHub, pinging Studio) feels confusing. Is there anything we can do to at least show the info about what parts of the push were successful like we do in CLI?

Is this important? In order to rectify the error the user is still going to have to fix the problem and push again regardless of which parts are successful.

> One question - do we plan to have a separate "Share" action or do we expect users to understand that Push does this - essentially solves their high level use case - to see an experiment from Studio? cc @dberenbaum (I'm bit worried that it's now too low level)

No plan right now, change is related to [this comment](https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1528003781):

> @dmpetrov Made a point about distinguishing between all the different options we have in the workflow today. For the dvc exp push action, what do you think about explicitly using "push" instead of "share" for the action name? I think it is more explicit and specific, giving more clarity that it's related to git push/dvc push and separate from "live" sharing.

skshetry:
@dberenbaum, I was asking for an actionable button.

> Is this important? In order to rectify the error the user is still going to have to fix the problem and push again regardless of which parts are successful.

@mattseddon, see https://github.com/iterative/dvc/issues/9342. The reasoning was that:
1. most of the metrics/plots/params are going to be git-tracked
2. and, even if we did not notify Studio, Studio could still parse them anyway if it got updates through other means.

This will be in the next release though, I just merged the PR.

shcheklein:
> Is your question about seeing what has been shared?

Yes. Connection between Studio and VS Code is still not clear to my mind.

> that doesn't tell users what's happening

why do you feel they need to know what is happening? Tbh I thought we should focus more on making it clear what is that they are getting as a result (vs an implementation details).

PR #3784, opened 2023-04-28
Simplify DVC CLI Location/Version Logic

julieg18:
* always opens "Setup" if extension status button is clicked on
* simplifies Toasts text when DVC CLI version is not found or incorrect ([details in comment](https://github.com/iterative/vscode-dvc/pull/3784#discussion_r1181981505))

## Demo

Note: Toast and extension button are hard to see in the demo since they're at the bottom of the screen 😅 

~https://user-images.githubusercontent.com/43496356/235555040-3615e203-d6df-45ef-b9f5-630a589b183e.mov~

~https://user-images.githubusercontent.com/43496356/235725878-10c11d86-e6ed-439d-8102-c4c38523a4e5.mov~

https://user-images.githubusercontent.com/43496356/235943880-d4ad6f97-d021-4278-8e6f-b25948ed7767.mov

Part of #3434


extension/src/cli/dvc/discovery.ts
```diff
@@ -12,22 +8,33 @@ import {
   getConfigValue,
   setUserConfigValue
 } from '../../vscode/config'
-import { getPythonBinPath } from '../../extensions/python'
 import { getFirstWorkspaceFolder } from '../../vscode/workspaceFolders'
 import { delay } from '../../util/time'
 import { SetupSection } from '../../setup/webview/contract'
 
+const warnWithSetupAction = async (
```
julieg18, line 15: Current possible options are:

**Version not found:**
`An error was thrown when trying to access the CLI.`
buttons: Setup, Don't Show Again

**Version can not be verified:**
`The extension cannot initialize as we were unable to verify the DVC CLI version.`
Button: Setup

**Version is found but it's not up to date:**
`The extension cannot initialize because you are using the wrong version of the CLI`
Buttons: Setup

**Version is found but it's too far ahead:**
`The extension cannot initialize because you are using the wrong version of the extension.`
Button: Setup

What do we think? Should the toasts be more/less simplified?

extension/src/status.ts
```diff
@@ -91,39 +91,23 @@ export class Status extends Disposable {
...
     }
   }
 
-  private getEnvDetails() {
+  private getEnvIndicator() {
     const dvcPath = this.config.getCliPath()
     const pythonBinPath = this.config.getPythonBinPath()
     if (dvcPath || !pythonBinPath) {
-      return {
-        indicator: '(Global)',
```
julieg18, line 112: For now, I kept the indicator and removed the tooltip entirely. Other options include removing the indicator as well, adding a more static tooltip, or just keeping things the way they were originally.

mattseddon: [Q] Outside of not having to maintain two different sets of text what is the benefit of removing the tooltip?

julieg18: > [Q] Outside of not having to maintain two different sets of text what is the benefit of removing the tooltip?

I thought that is what was requested since you can now just click on it, taking you to DVC CLI information in Setup. But I probably misunderstood something in a meeting or a comment. I can go ahead and add it back!

extension/src/cli/dvc/discovery.ts
```diff
@@ -12,90 +8,82 @@ import {
...
+  void warnWithSetupAction(
+    setup,
+    `The extension cannot initialize because you are using the wrong version of the ${update}.`
   )
 }
 
-export const warnAheadOfLatestTested = (): void => {
+const warnAheadOfLatestTested = (): void => {
   void Toast.warnWithOptions(
     `The located DVC CLI is at least a minor version ahead of the latest version the extension was tested with (${LATEST_TESTED_CLI_VERSION}). This could lead to unexpected behaviour. Please upgrade to the most recent version of the extension and reload this window.`
```
julieg18, line 45: Left this toast alone for now since we don't have any info on the latest tested cli version in Setup yet. Once we update our DVC Setup page (task is in #3434 list), we can simplify this one or remove entirely.

extension/src/status.ts
```diff
@@ -91,15 +91,8 @@ export class Status extends Disposable {
...
-    if (this.available) {
-      return {
-        command: RegisteredCommands.EXTENSION_SETUP_WORKSPACE,
-        title: Title.SETUP_WORKSPACE
-      }
-    }
-
     return {
-      command: RegisteredCommands.SETUP_SHOW,
+      command: RegisteredCommands.SETUP_SHOW_DVC,
```
julieg18, line 95: Is this supposed to go to the setup page in general or the DVC Section?

extension/src/status.ts
julieg18: Added tooltip back for now!

extension/src/status.ts
mattseddon: How about `SETUP_SHOW_DVC` if uninitialized otherwise `SETUP_SHOW`?

extension/src/cli/dvc/discovery.ts
```diff
@@ -12,90 +8,82 @@ import {
...
-  version: string,
+const warnVersionIncompatible = (
+  setup: IExtensionSetup,
   update: 'CLI' | 'extension'
 ): void => {
-  void Toast.warnWithOptions(
-    `The extension cannot initialize because you are using version ${version} of the DVC CLI. The expected version is ${MIN_CLI_VERSION} <= DVC < ${MAX_CLI_VERSION}. Please upgrade to the most recent version of the ${update} and reload this window.`
+  void warnWithSetupAction(
+    setup,
+    `The extension cannot initialize because you are using the wrong version of the ${update}.`
```
mattseddon, line 39: [C] Let's simplify this further and say something along the lines of the CLI is incompatible with this version of the extension. The user can open setup to get more details.

mattseddon: maybe get closer to `'The extension cannot initialize as we were unable to verify the DVC CLI version.'` so `'The extension cannot initialize because the DVC CLI version is incompatible.'`

extension/src/cli/dvc/discovery.ts
mattseddon: let's aim to remove it

mattseddon (approved):

extension/src/status.ts
julieg18: Done!

extension/src/cli/dvc/discovery.ts
julieg18: Updated Toasts:

**Version not found:**
`An error was thrown when trying to access the CLI.`
buttons: Setup, Don't Show Again

**Version can not be verified:**
`The extension cannot initialize as we were unable to verify the DVC CLI version.`
Button: Setup

**Version behind or ahead**
`The extension cannot initialize because the DVC CLI version is incompatible.`
Button: Setup

PR #3787, opened 2023-04-28
Fix dvc details version incorrectly showing "Not Found" 

julieg18:
* has `getDvcCliDetails()` rely on a version variable instead of calling `getCliVersion`

Possibly will fix the bug mentioned in https://github.com/iterative/vscode-dvc/issues/3434#issuecomment-1528072657


extension/src/setup/index.ts
```diff
@@ -378,9 +378,18 @@ export class Setup
...
+        version: undefined
+      }
+    }
+
+    const version = await this.getCliVersion(cwd)
+
     return {
       command,
-      version: cwd ? await this.getCliVersion(cwd) : undefined
+      version: version || (await this.getCliVersion(cwd, true))
```
julieg18, line 392: Looking at the [`discovery.ts` code](https://github.com/iterative/vscode-dvc/blob/main/extension/src/cli/dvc/discovery.ts#L243), we check for the version, then check for a global version. Our function that gets DVC CLI details, however, wasn't checking for a global. 

This could possibly be why Ivan was seeing `Version: Not Found` but the section was saying `DVC Incompatible`.

extension/src/setup/index.ts
```diff
@@ -378,9 +378,18 @@ export class Setup
     const commandArgs = args.length === 0 ? '' : ` ${args.join(' ')}`
     const command = executable + commandArgs
 
+    if (!cwd) {
```
julieg18, line 381: I believe that the test coverage is at 66% is due to this if statement not getting covered. Should I add an additional test or do I need to rewrite something?

extension/src/setup/index.ts
mattseddon: Why not save the version when setup gets run instead of recreating the logic here?

mattseddon: I had a quick look at the code and potentially `setCliCompatible` in `Setup` could be passed the version/undefined, it could then set both the version and `cliCompatible`. The version can then be passed to the webview without having to call `this.getCliVersion` each time.

mattseddon: I assume the current behaviour is why we are calling `dvc version ` all of the time too:


https://user-images.githubusercontent.com/37993418/235392775-7ec40861-df6c-48bb-89ea-eea625558d68.mov

julieg18: Thanks for the ideas, Matt! Updated the code to use a variable for version.

extension/src/test/suite/setup/index.test.ts
```diff
@@ -579,7 +578,6 @@ suite('Setup Test Suite', () => {
       mockRunSetup.restore()
       stub(config, 'isPythonExtensionUsed').returns(false)
       stub(config, 'getPythonBinPath').resolves(join('python'))
-      stub(setup, 'getDvcCliDetails').resolves(undefined)
```
julieg18, line 582: With `getDvcCliDetails` no longer calling `getDvcVersion`, we can delete related stubs.

julieg18 (commented):
Ready for another round of reviews! cc @mattseddon, @sroy3

mattseddon (approved):

PR #3799, opened 2023-05-02
Wait for config to be ready before accessing values

mattseddon:
I saw lots of type errors being sent to the integration test logs. After investigating I found and fixed some issues.


extension/src/test/suite/setup/index.test.ts
```diff
@@ -577,7 +577,7 @@ suite('Setup Test Suite', () => {
       mockExecuteCommand.restore()
       mockRunSetup.restore()
       stub(config, 'isPythonExtensionUsed').returns(false)
-      stub(config, 'getPythonBinPath').resolves(join('python'))
+      stub(config, 'getPythonBinPath').returns(join('python'))
```
mattseddon, line 580: [F] This was the cause of [these test warnings](https://github.com/iterative/vscode-dvc/actions/runs/4856306390/jobs/8655732780#step:7:677).

extension/src/setup/index.ts
```diff
@@ -389,18 +390,14 @@ export class Setup
     }
   }
 
-  private isDVCBeingUsedGlobally() {
```
mattseddon, line 392: We append the path to python to the `PATH` variable for the edge case that DVC is being used as a binary but with a virtual environment.

extension/src/setup/index.ts
```diff
@@ -374,7 +374,8 @@ export class Setup
     return this.sendDataToWebview()
   }
 
-  public getDvcCliDetails(): DvcCliDetails {
+  private async getDvcCliDetails(): Promise<DvcCliDetails> {
+    await this.config.isReady()
```
mattseddon, line 378: [F] Need to make sure that the config has set the pythonBinPath before we try to access it.

sroy3 (approved):

extension/src/setup/index.ts
mattseddon: because we append the virtual environment path to the PATH environment variable it is technically being used.

PR #3808, opened 2023-05-02
Hide push option when an experiment is running

### LINKED ISSUE #3806: push: only allow pushing completed experiments
The extension allows me to push experiments that are still running:

Screenshot 2023-05-02 at 2 55 45 PM

This will fail because the experiment commit has not yet been generated. Can we disable pushing for experiments unless they have successfully completed?

mattseddon:
Fixes #3806

### Demo

https://user-images.githubusercontent.com/37993418/235778786-676646e4-d8e6-4db9-ba39-d618740c1f25.mov

### DIFF
--- a/webview/src/experiments/components/App.test.tsx
+++ b/webview/src/experiments/components/App.test.tsx
@@ -1068,6 +1068,18 @@ describe('App', () => {
       })
     })
 
+    it('should not enable the user to share an experiment whilst an experiment is running', () => {
+      renderTable()
+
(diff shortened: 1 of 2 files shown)


sroy3 (approved):

shcheklein:
Thanks @mattseddon !Can we disable options, not hide them? I feel it's better for the menu to be consistent. Even better if we can show a tool tip why it's disabled at the moment.

PR #3809, opened 2023-05-02
Autoclose Setup Sections if completed

julieg18:


webview/src/setup/components/App.tsx
```diff
@@ -90,6 +94,10 @@ export const App: React.FC = () => {
     })
   }
 
+  const closeSection = (section: SetupSection) => {
+    setSectionCollapsed({ ...sectionCollapsed, [section]: true })
```
julieg18, line 98: Thoughts so far:

* Add completion status to steps first before adding any kind of auto-close logic. I feel like it could make things more clear.
* Don't just autoclose DVC, open other sections since neither can work until DVC is setup.
* Experiments needs to autoclose and open webview OR just leave as is. Not sure which is better yet.

webview/src/setup/components/dvc/Dvc.tsx
```diff
@@ -36,9 +38,21 @@ export const Dvc: React.FC<DvcProps> = ({
...
   pythonBinPath,
-  setSectionCollapsed,
-  isExperimentsAvailable
+  hasReceivedMessageFromVsCode,
+  setSectionCollapsed
 }) => {
+  const isSetup = projectInitialized && !!cliCompatible
+
+  const onCompletion = () => {
+    setSectionCollapsed({
```
julieg18, line 47: I decided just to leave experiments as is instead of redirecting the page, since in my opinion, it feels more natural for a redirection if there was nothing else to do on the Setup page which isn't the case since there's a studio section and more sections coming. 

But no strong preference, opening experiments does get the user a little quicker to DVC features.  With the reusable hook, it's simple enough to do so if we'd like!

webview/src/setup/hooks/useIsSectionComplete.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useState } from 'react'
+import { usePrevious } from './usePrevious'
+
+export const useIsSectionComplete = (
```
julieg18, line 4: Is there a simpler way to do this?

webview/src/setup/components/App.test.tsx
```diff
@@ -465,14 +470,29 @@ describe('App', () => {
...
-      expect(mockPostMessage).toHaveBeenCalledTimes(1)
-      expect(mockPostMessage).toHaveBeenCalledWith({
-        type: MessageFromWebviewType.OPEN_EXPERIMENTS_WEBVIEW
-      })
+      }
+
+      renderApp(dvcNotSetup)
+
+      const dvcDetails = within(
+        screen.getAllByTestId('section-container')[0]
```
julieg18, line 478: Oh wait, we probably shouldn't use array targeting as section order can change when we add more sections. Will update that.

julieg18:
Looks like storybook changes are based off of #3833

webview/src/setup/hooks/useIsSectionComplete.ts
mattseddon: Do we really need `useEffect` (and three of them)? Can't we move the state into `<Dvc>`?

If you set the initial state of `const [isComplete, setIsComplete] = useState<boolean | undefined | null>(null)` then you should know when it is updated because it should only ever be `null` when it is unset. Should then be able to use something like `if (previousIsComplete === null || isComplete === null) { return } `?

mattseddon: Also if we are going to introduce `hasReceivedMessageFromVsCode` we should introduce some loading state as well.

mattseddon (approved):

webview/src/setup/components/App.tsx
```diff
@@ -43,6 +43,8 @@ export const App: React.FC = () => {
   const [isStudioConnected, setIsStudioConnected] = useState<boolean>(false)
   const [shareLiveToStudio, setShareLiveToStudioValue] =
     useState<boolean>(false)
+  const [hasReceivedMessageFromVsCode, setHasReceivedMessageFromVsCode] =
```
mattseddon, line 46: We are definitely getting close to needing to manage all of this state differently.

sroy3: IMO we're passed that and I made a comment about that a couple of PRs ago. This is obviously not in the scope of this PR, but we should plan on doing this.

sroy3 (approved):

webview/src/setup/components/App.tsx
julieg18: Yep, already in #3434!

webview/src/setup/hooks/useIsSectionComplete.ts
julieg18: Here's what I've got so far: 

```
  const [isComplete, setIsComplete] = useState<boolean | null>(null)
  const previousIsComplete = usePrevious(isComplete)

  useEffect(() => {
    const isSetup = projectInitialized && !!cliCompatible

    if (hasReceivedMessageFromVsCode) {
      setIsComplete(isSetup)
    }

    if (previousIsComplete === false && isComplete) {
      setSectionCollapsed({
        [SetupSection.DVC]: true,
        [SetupSection.EXPERIMENTS]: false,
        [SetupSection.STUDIO]: false
      })
    }
  }, [
    projectInitialized,
    cliCompatible,
    isComplete,
    previousIsComplete,
    setSectionCollapsed,
    hasReceivedMessageFromVsCode
  ])
```

Is this what you had in mind?

PR #3813, opened 2023-05-03
Show progress notification when queueing experiments

mattseddon:
Raising this as a draft as not sure this is needed (the queue command is fast enough now).

### Demo

https://user-images.githubusercontent.com/37993418/235814901-4b60dee3-a245-4da3-9a56-8b5aac12af2c.mov

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -29,6 +29,7 @@ import { UNSELECTED } from './model/status'
 import { starredSort } from './model/sortBy/constants'
 import { pickSortsToRemove, pickSortToAdd } from './model/sortBy/quickPick'
 import { ColumnsModel } from './columns/model'
+import { getQueueWithProgress } from './commands'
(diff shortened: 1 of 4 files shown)


extension/src/experiments/index.ts
```diff
@@ -483,13 +483,9 @@ export class Experiments extends BaseRepository<TableData> {
...
-    await Toast.showOutput(
-      this.internalCommands.executeCommand<string>(
-        AvailableCommands.EXP_QUEUE,
-        this.dvcRoot,
-        ...paramsToModify
-      )
-    )
+    const queueWithProgress = getQueueWithProgress(this.internalCommands)
+
+    await queueWithProgress({ dvcRoot: this.dvcRoot, paramsToModify })
```
mattseddon, line 488: [I] Be careful with this it will delay it will stop the table from updating

mattseddon:
Doing something else.

PR #3815, opened 2023-05-03
Remove pausing of data updates when running/queueing experiments

mattseddon:
Related to #3178

Back in https://github.com/iterative/vscode-dvc/pull/1123 I added code to pause data updates whilst we queue experiments. This was because the commands used to update the data would take the read/write lock on the repository and cause the queueing process to fail. Now these commands are lockless so we don't need to pause updates anymore. The same thing happened intermittently when we started running experiments.

We need to do this to avoid trying to provide progress updates for batch-queueing experiments - e.g a grid search (see https://github.com/iterative/vscode-dvc/pull/3814)

### Demo

https://user-images.githubusercontent.com/37993418/235829745-938456d1-66c9-4671-867f-2770960467b7.mov

### DIFF
--- a/extension/src/data/index.ts
+++ b/extension/src/data/index.ts
@@ -33,7 +33,6 @@ export abstract class BaseData<
   constructor(
     dvcRoot: string,
     internalCommands: InternalCommands,
-    updatesPaused: EventEmitter<boolean>,
     updateProcesses: { name: string; process: () => Promise<unknown> }[],
     staticFiles: string[] = []
   ) {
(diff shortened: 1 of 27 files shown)


shcheklein:
so, queueing is immediate? we were slowing it down artificially by running some commands?

mattseddon:
> so, queueing is immediate? we were slowing it down artificially by running some commands?

When we queued a big batch of experiments the functions that update the data underlying the table (that call `exp show`) were paused. We waited for the entire command to complete before unpausing them.

julieg18 (approved):
Looks good!

PR #3823, opened 2023-05-03
Add completion status to setup sections

julieg18:
* replaces section tooltip with an icon depending on status

## Demo

~https://user-images.githubusercontent.com/43496356/236336529-94e7ab49-cb57-4218-a384-a43ded399a36.mov~

https://user-images.githubusercontent.com/43496356/236501613-41258c33-cd0c-428b-9799-9c18560d5f70.mov

image

image

Part of #3434


webview/src/shared/variables.scss
```diff
@@ -17,6 +17,7 @@ $error-color: var(--vscode-errorForeground);
 $meta-cell-color: var(--vscode-descriptionForeground);
 $accent-color: var(--button-primary-background);
 $accent-color-transparent: var(--accent-transparency-3);
+$passed-color: var(--vscode-testing-iconPassed);
```
julieg18, line 20: Is there a better variable for this color that I missed?

webview/src/setup/components/App.tsx
```diff
@@ -128,6 +134,7 @@ export const App: React.FC = () => {
         title="Studio"
         sectionCollapsed={sectionCollapsed}
         setSectionCollapsed={setSectionCollapsed}
+        hasError={!cliCompatible}
```
julieg18, line 137: Noticed that the studio section shows the "DVC Not Available" screen if DVC is not compatible but shows the setting screen if the project is not initialized. Is this expected?

julieg18:
After a discussion with Matt, we're going to remove the white checkmark icon, replacing it with the red error icon. It will reduce states and make things more clear :) Should be a quick change but reverting this pr to a draft since I won't be working on it till tomorrow morning!

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const InfoTooltip: React.FC<{
+  sectionKey: PlotsSection | SetupSection
+  icon?: TooltipIconType
+}> = ({ icon, sectionKey }) => {
+  const infoIcon = getIcon(icon)
+
+  const tooltipContent = (
+    <div className={styles.infoTooltip}>
+      {infoIcon}
+      {SectionDescription[sectionKey]}
```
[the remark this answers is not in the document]
mattseddon: The current text should be ok for now. If we want to revisit we should add very clear text for the expectation/usage of the section. E.g `Configure the extension to start working with DVC. Under the current configuration, the extension cannot initialize. Please follow the steps laid out in the section in order to get up and running.`

webview/src/setup/components/App.tsx
mattseddon: That is how I set it up. Connecting to Studio is a secondary concern. We also can't check to see if Studio is connected if we don't have access to the CLI (need to run `dvc config studio.token`).

webview/src/setup/components/App.test.tsx
```diff
@@ -629,6 +630,58 @@ describe('App', () => {
...
+      })
+
+      const iconWrapper = screen.getAllByTestId('info-tooltip-toggle')[0]
+
+      expect(
+        within(iconWrapper).getByTestId(TooltipIconType.ERROR)
+      ).toBeInTheDocument()
+    })
+
+    it('should show an complete icon if DVC is not setup', () => {
```
mattseddon, line 660: [Q] Should this be `should show a complete icon if the DVC CLI is compatible`?

webview/src/setup/components/App.test.tsx
```diff
@@ -823,6 +876,84 @@ describe('App', () => {
...
+      })
+
+      const iconWrapper = screen.getAllByTestId('info-tooltip-toggle')[1]
+
+      expect(
+        within(iconWrapper).getByTestId(TooltipIconType.ERROR)
+      ).toBeInTheDocument()
+    })
+
+    it('should show an passed icon if experiments are setup', () => {
```
mattseddon, line 932: [N] `a passed icon`

mattseddon: couple instances of this

webview/src/setup/components/App.tsx
```diff
@@ -128,6 +132,8 @@ export const App: React.FC = () => {
         title="Studio"
         sectionCollapsed={sectionCollapsed}
         setSectionCollapsed={setSectionCollapsed}
+        isSetup={!!cliCompatible}
+        isDisconnected={!isStudioConnected}
```
mattseddon, line 136: [Q] Why reverse this? Makes it harder to track IMO

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+import styles from './styles.module.scss'
+import { SectionDescription } from './SectionContainer'
+import Tooltip from '../tooltip/Tooltip'
+import { Icon } from '../Icon'
+import { Info, PassFilled, Error } from '../icons'
+
+export enum TooltipIconType {
+  PASSED = 'icon-pass-filled',
+  INFO = 'icon-info',
+  ERROR = 'icon-error'
```
mattseddon, line 14: [C] values probably don't need the `icon` prefix.

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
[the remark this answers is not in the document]
mattseddon: up to you

webview/src/shared/variables.scss
mattseddon: this is fine

mattseddon (approved):

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const getIcon = (icon?: TooltipIconType) => {
+  const props = {
+    className: styles.infoIcon,
+    'data-testid': icon || TooltipIconType.INFO,
+    height: 16,
+    icon: Info,
```
mattseddon, line 22: [I] Don't optionally overwrite this. Have clear logic branches to show how it is set.

mattseddon: e.g 

```
const getIcon = (icon?: TooltipIconType) => {
  const defaultProps = {
    height: 16,
    width: 16
  }

  if (icon === TooltipIconType.ERROR) {
    return (
      <Icon
        {...{
          className: cx(styles.infoIcon, styles.errorIcon),
          'data-testid': icon,
          icon: Error,
          ...defaultProps
        }}
      />
    )
  }

  if (icon === TooltipIconType.PASSED) {
    return (
      <Icon
        {...{
          className: cx(styles.infoIcon, styles.completedIcon),
          'data-testid': icon,
          icon: PassFilled,
          ...defaultProps
        }}
      />
    )
  }

  return (
    <Icon
      {...{
        className: styles.infoIcon,
        'data-testid': TooltipIconType.INFO,
        icon: Info,
        ...defaultProps
      }}
    />
  )
}
```

if duplication is an issue pull out a `getProps` (or similar) helper function.

webview/src/shared/components/icons/PassFilled.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
```
sroy3, line 1: Can you add this icon to the Icons story. I'll try to add something to auto generate new icons in there later, but it's still manual for the moment.

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
sroy3: Why not have the `TooltipIconType` match the icon directly? You'll be able to directly use it in `getIcon` for the `icon` prop.

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+  PASSED = 'icon-pass-filled',
+  INFO = 'icon-info',
+  ERROR = 'icon-error'
+}
+
+const getIcon = (icon?: TooltipIconType) => {
+  const props = {
+    className: styles.infoIcon,
+    'data-testid': icon || TooltipIconType.INFO,
+    height: 16,
```
sroy3, line 21: I'd rather see these properties that do not change directly on the component so that I don't have to check the code to see whether they've been changed or not.

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export enum TooltipIconType {
+  PASSED = 'icon-pass-filled',
+  INFO = 'icon-info',
+  ERROR = 'icon-error'
+}
+
+const getIcon = (icon?: TooltipIconType) => {
+  const props = {
+    className: styles.infoIcon,
```
sroy3, line 19: By using having `TooltipIconType` match the icon directly, I think you could move all props directly on the component and use `cx()` with conditionals.

webview/src/shared/components/sectionContainer/styles.module.scss
```diff
@@ -52,6 +52,18 @@ $gap: 20px;
 .infoIcon {
   fill: $accent-color;
   margin-left: 6px;
+
+  .infoTooltip & {
+    margin-left: 0;
```
sroy3, line 57: Is there an `infoIcon` anywhere else other than inside `infoTooltip`? If I missed it, better add another class to this one instead. If there isn't, we should remove that nesting.

sroy3 (approved):

webview/src/shared/components/sectionContainer/styles.module.scss
julieg18: Yes, it's used inside the `infoTooltip` and inside the header. I'll add another class.

webview/src/setup/components/App.test.tsx
```diff
@@ -629,6 +630,58 @@ describe('App', () => {
...
+        canGitInitialize: false,
+        cliCompatible: true,
+        dvcCliDetails: {
+          command: 'python -m dvc',
+          version: '1.0.0'
+        },
+        hasData: true,
+        isPythonExtensionUsed: true,
+        isStudioConnected: true,
+        needsGitCommit: false,
```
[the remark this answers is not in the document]
julieg18: I was thinking about this as well since that's how we write other test files when it comes to a lot of props. Will update renderApp and adjust the rest of the tests in a followup.

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
sroy3: For the name comparison, better yet: `icon.name === TooltipIconType.INFO.name`.

julieg18: After testing, I realized that enums can only hold strings or numbers. I'm not seeing another way to mimic what an enum does with component values without just letting `InfoTooltip` just take in any icon which I'd like to avoid. Will keep the icons strings for now!

sroy3: You'll need to use a const which is practically the same thing.

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
julieg18: Done!

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
julieg18: > You'll need to use a const which is practically the same thing.

Apologies, I can't seem to understand how we could use a const. Could you give me a code example? Here's what I tried:

```
const IconTypes = {
  ERROR: Error,
  PASSED: PassFilled
} as const

type Type = (typeof IconTypes)[keyof typeof IconTypes]

const getNeededIcon = (icon: Type) => {
  // get the icon here
}

// yes this works
getNeededIcon(IconType.PASSED)

// but this also works 
// since all icons have the exact same type: ((props: SVGProps<SVGSVGElement>) => JSX.Element)
getNeededIcon(Beaker)

```

webview/src/shared/components/icons/PassFilled.tsx
julieg18: Added!

webview/src/shared/components/sectionContainer/InfoTooltip.tsx
sroy3: Does it really matter if you can pass other icons? You'd be remove your `if` branches so it does not need to be a certain type anymore. If you really want to add type safety you can have an enum for the keys and a const for the tooltipIcons.

PR #3825, opened 2023-05-04
Keep experiments table context menu consistent by showing disabled items

mattseddon:
### Demo

https://user-images.githubusercontent.com/37993418/236118005-c4f1dae4-4d52-467e-bae2-458e026cba68.mov

TODO

- [x] Rename hidden to disabled
- [x] De-legacy-ify


webview/src/experiments/components/App.test.tsx
```diff
@@ -885,24 +893,12 @@ describe('App', () => {
       advanceTimersByTime(100)
       expect(screen.getAllByRole('menuitem')).toHaveLength(10)
 
-      fireEvent.click(window, { bubbles: true })
-      advanceTimersByTime(100)
-      expect(screen.queryAllByRole('menuitem')).toHaveLength(0)
-    })
-
-    it('should be removed with a left click on a different row', () => {
```
mattseddon, line 893: [F] This behaviour now relies `onClickOutside` which won't work in an environment with 0 height and width. The behaviour is included in the demo.

webview/src/experiments/components/table/body/TableBody.tsx
```diff
@@ -16,7 +16,6 @@ interface TableBodyProps extends RowProp, InstanceProp, BatchSelectionProp {
 export const TableBody: React.FC<TableBodyProps> = ({
   row,
   instance,
-  contextMenuDisabled,
```
mattseddon, line 19: [F] This starts here and drills all the way down but is not set on `<Table>`. I assumed that means I can remove it.

webview/src/experiments/components/table/body/Row.tsx
```diff
@@ -107,8 +104,7 @@ export const RowContent: React.FC<
             [styles.evenRow]: !isOdd,
             [styles.workspaceRow]: isWorkspace,
             [styles.normalRow]: !isWorkspace,
-            [styles.rowSelected]: isRowSelected,
-            [styles.rowFocused]: menuActive
```
mattseddon, line 111: this `.rowFocused` style doesn't exist which means we don't need `menuActive` which means we no longer need `onShow` and `onHide`

webview/src/experiments/components/table/body/ExperimentGroup.tsx
```diff
@@ -5,14 +5,12 @@ import { RowProp } from '../../../util/interfaces'
 
 export const ExperimentGroup: React.FC<RowProp & BatchSelectionProp> = ({
   row,
-  contextMenuDisabled,
```
mattseddon, line 8: See [this](https://github.com/iterative/vscode-dvc/pull/3825/files#r1184557186)

webview/src/shared/components/messagesMenu/MessagesMenuOption.tsx
```diff
@@ -1,24 +1,38 @@
...
+  label,
+  message,
+  hideOnClick,
+  disabled,
+  divider,
+  onOptionSelected,
+  keyboardShortcut
+}) => {
   const sendTheMessage = () => {
+    if (disabled) {
```
mattseddon, line 30: [F] This is why we need control of `hideOnClick`. If the behaviour is wrapped up in `positionContextMenuAndDisableEvents` then we cannot stop the context menu from closing if a disabled item is click (not the normal behaviour)

webview/src/shared/components/contextMenu/ContextMenu.tsx
```diff
@@ -1,47 +1,45 @@
...
 
-  instance.reference.removeEventListener('contextmenu', handleDefaultEvent)
-  instance.reference.addEventListener('contextmenu', handleDefaultEvent)
+    const hideOnClick = () => {
+      !instance.state.isDestroyed && instance.hide()
+    }
 
-  const hideOnClick = () => {
-    !instance.state.isDestroyed && instance.hide()
+    setHideOnClick(() => hideOnClick)
```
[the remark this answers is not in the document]
mattseddon: See [also](https://github.com/iterative/vscode-dvc/pull/3825/files#r1184809441)

webview/src/util/props.ts
```diff
@@ -15,9 +15,6 @@ export const clickAndEnterProps: <T>(
   onKeyDown: React.KeyboardEventHandler<T>
 } = (handler, textsForSelection = [], checkForTooltip = false) => ({
   onClick: e => {
-    e.preventDefault()
-    e.stopPropagation()
```
mattseddon, line 19: I could not find a reason for these being here. Does anyone recall the reason?

webview/src/shared/components/contextMenu/ContextMenu.tsx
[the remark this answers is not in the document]
sroy3: How about passing a prop `shouldHide` to the `ContextMenu` and hiding inside a `useEffect`? Bringing the function upward feels backward  for some reason. This might also cause re-rendering problems as arrow functions are re-created on renders and passing it back up to be added to a state will trigger a re-render...

webview/src/shared/components/messagesMenu/MessagesMenuOption.tsx
```diff
@@ -1,24 +1,38 @@
...
+      return
+    }
     !!message && sendMessage(message)
     onOptionSelected?.()
+    hideOnClick?.()
```
sroy3, line 35: This could be `setShouldHide` from where this component and `ContextMenu` are created instead.

webview/src/experiments/components/table/body/Row.tsx
```diff
@@ -68,31 +67,29 @@ export const RowContent: React.FC<
...
   }, [subRows, selectedRows])
 
-  const [menuActive, setMenuActive] = useState<boolean>(false)
-
   const running = isRunning(status)
   const queued = isQueued(status)
   const unselected = selected === false
   const isOdd = index % 2 !== 0 && !isRowSelected
 
+  const [hideOnClick, setHideOnClick] = useState<(() => void) | undefined>(
```
sroy3, line 75: I would use a boolean `shouldHide`, pass `setShouldHide` to `RowContextMenu` and pass `shouldHide` to the `ContextMenu`.  `ContextMenu` should handle its behaviour. (See further comments as to why this would be better).

webview/src/experiments/components/table/header/ContextMenuContent.tsx
```diff
@@ -51,24 +52,19 @@ export const sortOption = (
         }
 
   return {
-    hidden,
+    disabled: !isSortable || disabled,
```
sroy3, line 55: `!isSortable` could be moved to the `disabled` variable above.

webview/src/experiments/components/table/header/TableHeaderCell.tsx
```diff
@@ -140,18 +140,26 @@ export const TableHeaderCell: React.FC<{
     />
   )
 
-  const menuContent = <ContextMenuContent header={header} />
+  const [hideOnClick, setHideOnClick] = useState<(() => void) | undefined>(
```
sroy3, line 143: This should also be reversed to use a boolean instead.

webview/src/shared/components/contextMenu/ContextMenu.tsx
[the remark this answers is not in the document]
sroy3: Would a `clickOutiside` check also work here? Reducing the number of places that need to know about closing the context menu?

sroy3 (approved):
I think it would be a good idea to change the cursor for the disabled item. Either `not-allowed` or a simple `default`.

webview/src/util/props.ts
sroy3: I would move this inside the following `if` branch. We do want to stop any default or added behaviour if the user is selecting., but in a normal context, there is no reason for these to be there and it might cause some unexpected behaviour.

webview/src/shared/components/messagesMenu/styles.module.scss
```diff
@@ -25,6 +25,14 @@
   text-overflow: ellipsis;
 }
 
+.disabledItem {
+  opacity: $disabled-opacity;
+
```
julieg18, line 30: Should we be adding a `pointer: auto` or `pointer: default` to disabled items?

webview/src/shared/components/contextMenu/ContextMenu.tsx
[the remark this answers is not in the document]
julieg18: Is the only reason we're needing this `setHideOnClick` prop is to make sure clicking on disabled items doesn't close the context menu?  If so, why use a window event listener like before but just check the element that was clicked on? Aka similar to what we had before:

```
  const hideOnClick = e => {
    if (!e.target.closest('[aria-disabled=true]')) {
      !instance.state.isDestroyed && instance.hide()
    }
  }

  instance.popper.removeEventListener('click', hideOnClick)
  instance.popper.addEventListener('click', hideOnClick)

window.addEventListener('click', hideOnClick, { once: true })
```

Seems to working from a quick test on my end. For the record, as part of this test, I commented out our `const [hideOnClick, setHideOnClick] = useState...` declarations and added a `aria-disabled` attribute to our disabled menu options:

https://user-images.githubusercontent.com/43496356/236240085-a4d998ce-86c8-46b7-967a-8bc77afefde0.mov

julieg18 (approved):
Great work!

webview/src/shared/components/messagesMenu/styles.module.scss
mattseddon: Changed both to `cursor: default;` to match the VS Code context menu.

webview/src/experiments/components/App.test.tsx
mattseddon: comment is outdated

PR #3826, opened 2023-05-04
Add import/no-unused-modules rule to eslint config

mattseddon:
Noticed that we are carrying some dead code whilst working on #3825 so hunted around for this rule: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unused-modules.md


sroy3 (approved):
That's great!

PR #3832, opened 2023-05-05
Enable experiment operations when experiment(s) are running in the queue

mattseddon:
# 1/3 `main` <- this <- #3834 <- #3840

Part of #3178 

This PR makes it possible to run experiments operations when experiments are running in the queue. Actions will still be disabled on experiments running in the queue and on all experiments when there is an experiment running in the workspace.

### Demo

https://user-images.githubusercontent.com/37993418/236372876-e408d4ba-771c-4e17-a38e-9dba2ea63df9.mov


extension/src/experiments/model/index.ts
```diff
@@ -39,10 +40,11 @@ export type SelectedExperimentWithColor = Experiment & {
 }
 
 export enum ExperimentType {
-  WORKSPACE = 'workspace',
   COMMIT = 'commit',
   EXPERIMENT = 'experiment',
-  QUEUED = 'queued'
+  RUNNING = 'running',
```
mattseddon, line 45: [F] We need this to prevent actions being run from the `experimentsTree` context menu for running experiments.

extension/package.json
```diff
@@ -664,11 +664,11 @@
         },
         {
           "command": "dvc.applyExperiment",
-          "when": "dvc.commands.available && dvc.project.available && !dvc.experiment.running"
+          "when": "dvc.commands.available && dvc.project.available && !dvc.experiment.running.workspace"
```
mattseddon, line 667: [F] We only want to suppress actions when there is an experiment running in the workspace.

extension/package.json
```diff
@@ -1158,42 +1163,42 @@
         {
           "command": "dvc.views.experiments.applyExperiment",
           "group": "inline@1",
-          "when": "view == dvc.views.experimentsTree && dvc.commands.available && viewItem == experiment && !dvc.experiment.running"
+          "when": "view == dvc.views.experimentsTree && dvc.commands.available && viewItem == experiment && !dvc.experiment.running.workspace"
```
mattseddon, line 1166: [F] `viewItem == experiment` means none of these actions will be available for running experiments.

mattseddon: I'm going to add stop inline as a follow up

webview/src/experiments/components/table/body/ExperimentGroup.tsx
```diff
@@ -5,14 +5,5 @@ import { RowProp } from '../../../util/interfaces'
...
   batchRowSelection
-}) => (
-  <NestedRow
-    row={row}
-    projectHasCheckpoints={projectHasCheckpoints}
-    hasRunningExperiment={hasRunningExperiment}
-    batchRowSelection={batchRowSelection}
-  />
-)
+}) => <NestedRow row={row} batchRowSelection={batchRowSelection} />
```
mattseddon, line 9: [F] Using redux for this state instead of drilling

webview/src/test/experimentsTable.tsx
```diff
@@ -62,7 +63,15 @@ export const renderTableWithSortingData = () => {
...
   renderTable({
     ...tableDataFixture,
-    hasRunningExperiment: false
+    hasRunningWorkspaceExperiment: false,
+    rows: tableDataFixture.rows.map(row => ({
+      ...row,
+      status: ExperimentStatus.SUCCESS,
+      subRows: row.subRows?.map(subRow => ({
+        ...subRow,
+        status: ExperimentStatus.SUCCESS
```
mattseddon, line 72: [F] Have to set the status on each experiment for them to count as not running.

julieg18 (approved):
Great work!

extension/src/experiments/model/tree.ts
```diff
@@ -238,17 +238,15 @@ export class ExperimentsTree
...
     type,
     selected
   }: {
     displayColor?: string
     label: string
-    status?: ExperimentStatus
     type?: ExperimentType
     selected?: boolean
   }): ThemeIcon | Uri | Resource {
-    if (isRunning(status)) {
```
sroy3, line 251: Are these helpers still useful if they're checking only the type? It was removed here, but used again in `extension/src/experiments/model/index.ts`. We should stay consistent and either use them everywhere or remove them.

sroy3 (approved):

PR #3840, opened 2023-05-08
Combine/enhance commands used to stop experiments

mattseddon:
# 3/3 `main` <- #3832  <- #3834 <- this

This PR consolidates our "Stop Experiment" commands. There is now no external separation between stopping an experiment running in the workspace or queue. Experiments can now be stopped via an inline action shown in the experiments tree, the tree's view/title, the experiments table context menu, the editor/title and the command palette.

### Demo

https://user-images.githubusercontent.com/37993418/236997101-d7dc95e8-8b55-43a1-ae2a-799625bdbc4e.mov

https://user-images.githubusercontent.com/37993418/236997230-a74a6f7b-a106-49b8-98ad-53fd23072ba1.mov

https://user-images.githubusercontent.com/37993418/237008688-1c08b0b6-2f04-4c42-96a5-133e1d2f021e.mov

https://user-images.githubusercontent.com/37993418/236998675-ef50fe40-a5ce-4793-a056-24ef858a04b3.mov

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -1243,6 +1243,11 @@
           "group": "navigation@2"
         },
         {
+          "command": "dvc.stopAllRunningExperiments",
+          "when": "view == dvc.views.experimentsTree && dvc.experiments.webview.active && dvc.experiment.running",
+          "group": "navigation@0"
+        },
+        {
           "command": "dvc.runExperiment",
           "when": "view == dvc.views.experimentsTree && !dvc.experiment.running.workspace && !dvc.experiment.checkpoints",
           "group": "navigation@1"
@@ -1261,11 +1266,6 @@
           "command": "dvc.stopAllRunningExperiments",
           "when": "view == dvc.views.experimentsTree && !dvc.experiments.webview.active && dvc.experiment.running",
           "group": "1_run@1"
-        },
-        {
-          "command": "dvc.stopAllRunningExperiments",
-          "when": "view == dvc.views.experimentsTree && dvc.experiments.webview.active && dvc.experiment.running",
-          "group": "navigation@1"
         },
         {
           "command": "dvc.views.experimentsTree.selectExperiments",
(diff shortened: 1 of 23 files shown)


extension/src/experiments/index.ts
```diff
@@ -486,6 +491,36 @@ export class Experiments extends BaseRepository<TableData> {
     return this.notifyChanged()
   }
 
+  // eslint-disable-next-line sonarjs/cognitive-complexity
+  public stopExperiments(ids: string[]) {
```
mattseddon, line 495: collection can happen in the model, stopping can happen here

extension/src/experiments/processExecution/index.ts
```diff
@@ -1,23 +1,23 @@
...
-    return
-  }
-  if (!(await processExists(pid))) {
-    return
+export const stopWorkspaceExperiment = async (dvcRoot: string, id: string) => {
+  const pids = new Set<number>()
+
+  await collectDvcRootPids(pids, dvcRoot)
+
+  const [pid] = [...pids]
```
sroy3, line 9: How about `const pid = pids.slice(0, 1)`? It doesn't matter one or the other, I guess it's just a preference, but to me it looks like 2 steps vs. one.

extension/package.json
```diff
@@ -362,8 +362,8 @@
         "category": "DVC"
```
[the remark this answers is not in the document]
julieg18: Ah, wait looks like this bug is getting fixed in #3846 :)

extension/src/experiments/processExecution/index.ts
mattseddon: I would still have to convert the set into an array (`error TS2339: Property 'slice' does not exist on type 'Set<number>'`). Whatever I do it will look a bit whacky. This comment did make me realise that I needed another logic branch "to be safe". Thanks

sroy3: I get it now. I thought you were avoiding mutation, I didn't follow to see it was a `Set`, my bad. Thanks for explaining.

PR #3850, opened 2023-05-09
Update DVC Setup Version Details

julieg18:
* remove information about max version in setup details
* adds information about latest tested version setup details
* removes latest tested version toast

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/0b293db7-a2b0-49dd-a930-216262889676~

Screenshot 2023-05-09 at 12 09 07 PM

Part of #3434


webview/src/setup/components/dvc/DvcEnvDetails.tsx
```diff
@@ -14,9 +17,14 @@ export const DvcEnvDetails: React.FC<DvcEnvDetailsProps> = ({
   version,
   isPythonExtensionUsed
 }) => {
-  const versionText = `${
-    version || 'Not found'
-  } (${MIN_CLI_VERSION} <= required < ${MAX_CLI_VERSION}.0.0)`
+  const versionText = (
```
julieg18, line 20: Another possibility I thought of was including all info within parenthesis: 

Screenshot 2023-05-09 at 10 46 41 AM

julieg18: Any other ideas for the design?

julieg18: > required up to -> required <> and above?
> tested up to -> tested with

Here's how that looks:

Screenshot 2023-05-09 at 12 04 13 PM

webview/src/setup/components/dvc/DvcEnvDetails.tsx
```diff
@@ -14,9 +17,13 @@ export const DvcEnvDetails: React.FC<DvcEnvDetailsProps> = ({
...
+    <>
```
sroy3, line 21: I don't think the fragment is needed here.

sroy3 (approved):

webview/src/setup/components/dvc/DvcEnvDetails.tsx
mattseddon: Can we make it more obvious if the CLI is a minor version ahead of the latest tested version? Maybe add a warning symbol somewhere? The reason is that if there has been a breaking change and the extension no longer works with that version of the CLI then it will not be completely obvious as to why from just looking at the above screen. Maybe PassFilled should become the warning icon? Maybe we should put a warning icon against the version. WDYT?

mattseddon (approved):
Good iteration, can we follow up by replacing the old behaviour instead of completely removing it.

PR #3851, opened 2023-05-09
Add redux to setup webview

julieg18:
# 1/2 `main` <= this <= #3862 

Part of #3434


webview/src/setup/state/setupDataSlice.ts
```diff
@@ -0,0 +1,50 @@
...
+  hasData: undefined,
+  hasReceivedMessageFromVsCode: false,
+  isPythonExtensionUsed: false,
+  isStudioConnected: false,
+  needsGitCommit: false,
+  needsGitInitialized: false,
+  projectInitialized: false,
+  pythonBinPath: undefined,
+  sectionCollapsed: DEFAULT_SECTION_COLLAPSED,
+  shareLiveToStudio: false
```
julieg18, line 25: Should we break up this slice into sections? Aka, `studio`, `dvc`, `experiments`, and `sectionCollapsed`.

julieg18: Decided to do so :)

webview/src/setup/components/dvc/Dvc.tsx
```diff
@@ -17,54 +14,49 @@ import {
...
+  )
+  const {
+    canGitInitialize,
+    cliCompatible,
+    dvcCliDetails,
+    isPythonExtensionUsed,
+    needsGitInitialized,
+    projectInitialized,
+    pythonBinPath
+  } = useSelector((state: SetupState) => state.dvc)
```
julieg18, line 33: For now, I moved the state into the "main" section components but we can lower the amount of props being passed around further. For example `isPythonExtensionUsed` is only used by `DvcEnvCommandRow`. 

I'll take care of this in a second pr though since this one is already up to +400 lines added though.

webview/src/setup/components/studio/CliIncompatible.tsx
```diff
@@ -1,25 +1,31 @@
...
+    <EmptyState isFullScreen={false}>
+      <h1>DVC is currently unavailable</h1>
+      <p>
+        Locate DVC to connect to <a href={STUDIO_URL}>Studio</a>
+      </p>
+      <Button
+        text="Setup DVC"
+        onClick={() =>
+          dispatch(
+            updateSectionCollapsed({
```
mattseddon, line 21: [Q] Is this how we handle this in other parts of the app (e,g plots)? Should we converge on a single method?

mattseddon (approved):

webview/src/setup/components/SetupContainer.tsx
```diff
@@ -16,33 +19,32 @@ const getTooltipIconType = (isSetup: boolean, isConnected = true) => {
...
-      })
-    }
-  >
-    {children}
-  </SectionContainer>
-)
+}> = ({ children, sectionKey, title, isSetup, isConnected }) => {
+  const sectionCollapsed =
+    useSelector((state: SetupState) => state.webview.sectionCollapsed) ||
+    DEFAULT_SECTION_COLLAPSED
```
sroy3, line 29: Shouldn't that default be inside the slice?

webview/src/setup/components/SetupContainer.tsx
```diff
@@ -16,33 +19,32 @@ const getTooltipIconType = (isSetup: boolean, isConnected = true) => {
...
+
+  return (
+    <SectionContainer
+      sectionCollapsed={sectionCollapsed[sectionKey]}
+      sectionKey={sectionKey}
+      title={title}
+      icon={getTooltipIconType(isSetup, isConnected)}
+      onToggleSection={() =>
+        dispatch(
+          updateSectionCollapsed({
```
sroy3, line 40: There could be another action just to toggle a section. That would simplify things here and let the action handle the state.

webview/src/setup/components/experiments/Experiments.tsx
```diff
@@ -1,38 +1,39 @@
...
-              dvc: false,
-              experiments: true,
-              studio: true
-            })
+            dispatch(
+              updateSectionCollapsed({
+                dvc: false,
+                experiments: true,
+                studio: true
+              })
```
sroy3, line 35: Out of scope, but in `Dvc.tsx` we have:

```
  [SetupSection.DVC]: true,
  [SetupSection.EXPERIMENTS]: false,
  [SetupSection.STUDIO]: false
```
We probably should keep using section keys.

webview/src/setup/state/dvcSlice.ts
```diff
@@ -0,0 +1,72 @@
...
+export type DvcState = Pick<
+  SetupData,
+  | 'canGitInitialize'
+  | 'cliCompatible'
+  | 'dvcCliDetails'
+  | 'isPythonExtensionUsed'
+  | 'projectInitialized'
+  | 'needsGitInitialized'
+  | 'pythonBinPath'
+>
```
sroy3, line 13: Should this be `Omit` instead of `Pick` (with omitted keys)? There are way more properties to pick than to omit.

sroy3 (approved):

webview/src/setup/components/experiments/Experiments.tsx
julieg18: Good catch! I'll update that

webview/src/setup/components/studio/CliIncompatible.tsx
julieg18: Well, the plots' collapsed state is handled differently in the front end. Instead of having a section object, plots section state each have their own `isCollapsed` key. I suppose it makes sense that it was set up that way since, unlike setup, plots sections don't need to close/open other sections. 

We could possibly have both plots and setup webviews use a section collapsed object, making them more similar and allowing them to possibly share functions/methods 🤔

PR #3855, opened 2023-05-10
Update all webdriver packages

mattseddon:
Replaces https://github.com/iterative/vscode-dvc/pull/3847


mattseddon:
E2E might fail because the Chromedriver API is not currently available. Not for the same reason that #3847 doesn't work.

sroy3 (approved):

PR #3856, opened 2023-05-10
Update blocked ESM only packages (execa, process-exists)

mattseddon:
For a while, I've wanted to update `execa` and `process-exists` but have been unable to because they are [ESM only packages](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). Importing these modules is now possible (after discovering `"moduleResolution": "node16"` by chance). This PR upgrade the aforementioned packages and integrates the way we now need to import them with the rest of code base.

Please LMK if you have any questions 🙏🏻.


extension/src/process/execution.ts
```diff
@@ -2,11 +2,36 @@ import { ChildProcess } from 'child_process'
...
+type EsmExeca = typeof import('execa').execa
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+type EsmProcessExists = typeof import('process-exists').processExists
+
+const envCanImportEsm = process.env.NODE_ENV !== 'test'
+
+let execa: EsmExeca
+let doesProcessExist: EsmProcessExists
+const importEsmModules = async () => {
```
mattseddon, line 23: [F] The reason that we cannot simply `await import('execa')` before we call it is that whenever we create a process the function has to be sync. If a `Promise<Process>` is returned then we cannot attach anything to it. We can only `await` the `stdout`.

mattseddon: this is still relevant

sroy3 (approved):

julieg18 (approved):
Great work!

PR #3857, opened 2023-05-10
Send all stderr and stdout to the output channel when a DVC process fails

mattseddon:
As discussed in the cross-team meeting this morning (cc @dberenbaum). 

It is useful to send all of the stderr/stdout to the output channel when a DVC process fails. This PR makes the necessary updates.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/0ef998f6-0499-4c5d-a9db-daa8f8839cd1


extension/src/cli/dvc/runner.ts
```diff
@@ -164,8 +164,8 @@ export class DvcRunner extends Disposable implements ICli {
         {
           ...baseEvent,
           duration: stopWatch.getElapsedTime(),
-          exitCode,
-          stderr
+          errorOutput: stderr,
```
mattseddon, line 167: [F] The runner and viewer show processes in the foreground so there is no change for these

extension/src/cli/index.ts
```diff
@@ -180,8 +196,8 @@ export class Cli extends Disposable implements ICli {
       {
         ...baseEvent,
         duration,
-        exitCode: cliError.exitCode,
-        stderr: cliError.stderr
+        errorOutput: all || cliError.stderr,
```
mattseddon, line 199: [F] If we have captured a valid all string (length > 0) then we will use that, otherwise use the stderr from the generated error. I feel like this way there will always be something provided to the output channel

sroy3 (approved):

dberenbaum:
Thanks for taking this so quickly @mattseddon1

extension/src/cli/index.ts
```diff
@@ -96,18 +99,26 @@ export class Cli extends Disposable implements ICli {
...
     try {
       const backgroundProcess = this.createProcess(baseEvent, {
         detached: true,
         ...options
       })
+
+      backgroundProcess.all?.on(
+        'data',
+        chunk => (all += transformChunkToString(chunk))
+      )
```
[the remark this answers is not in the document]
mattseddon: I did have this in a separate helper function but the string was not captured as expected and passed to the catch block.

PR #3859, opened 2023-05-10
Update coverage path ignore patterns

mattseddon:
I wrote a test for #3857 and the coverage actually went down 😬 🙃. I think that there are some issues with the way we exclude from test coverage. I've updated the various configs to try and fix them up.


julieg18 (approved):
Looks good!

PR #3871, opened 2023-05-11
Cleanup icons

### LINKED ISSUE #3848: Review how we use SVGs (icons)
As seen in #3844, the Codicons can get updated occasionally. That means we can get new icons (not much of a problem if we aren't using these yet), or some icons get changed (more of a problem if we are using any of these). We are not using these icons directly, and we must convert them to React components and import them into our codebase. We could use them directly, but we use this method to stay consistent with our other icons (check, chevrons, clock, copy, dots, down arrow, lines, pin, up arrow).

We have tools to help us with our icons (SVGR with a script and a Storybook story with all the icons). None of these tools work automatically at the moment.

We should change a few things to make working with icons and Codicons easy.

- [ ] First, we should clean up our custom icons. There might be some we don't use in there and some we could directly import from Codicons. There are two places to do so. The first place is the `icons` directory at the root of the web view. The other one is the generated React components directory `icons` in the shared components of the web view. We could also remove one of the chevrons and one of the arrows to rotate them in CSS instead. (#3871 )

- [ ] Second, the script used by SVGR should have a clear list of the icons we want from Codicons and re-generate them (as a post-install script, maybe?) so that we can stay up to date. We could also import every Codicon into our codebase. Still, we need to ensure they do not all get included in our releases if we don't use them, as they will add unnecessary weight to our distribution files. (#3872)

- [ ] Third, the Storybook story should not be updated manually every time we add a new icon. Instead, the Storybook story should generate itself from the icons in the shared components directory. This way, we always get an up-to-date representation of the available icons. (#3873)

- [ ] Lastly, if we do not import every Codicon in our codebase, adding a second Storybook story to visualize the available icons from Codicons before importing them could help select the right new icon. (#3874)

sroy3:
# 1/4 main <= this <= #3872 <= #3873 <= #3874

First task of #3848

- Removed icons from the icons folder if they were inside of the Codicons folder.
- Renamed icons with different names in Codeicons

Closes #3848 when everything is merged back


julieg18 (approved):
Great work!

sroy3:
@mattseddon I'll go ahead and merge this as this is housekeeping stuff, but I'll be happy to do any follow-up/modification if you want to review this.

PR #3872, opened 2023-05-11
Read and transform svg icons from list

sroy3:
- Cleaned up svg icons
- Read and transform svg icons from list


webview/package.json
```diff
@@ -13,7 +13,7 @@
     "test": "jest --collect-coverage",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build --webpack-stats-json",
-    "svgr": "svgr --out-dir src/shared/components/icons --ignore-existing ../node_modules/@vscode/codicons/src/icons && svgr -d src/shared/components/icons icons/"
```
julieg18, line 16: This is my first time actually seeing this command! Looking at it, I'm assuming we used this to generate chosen icon components? I've just been creating them manually 🤣

webview/icons/generate.mjs
```diff
@@ -0,0 +1,103 @@
...
+  'star-full',
+  'trash'
+]
+
+const allIcons = [
+  ...customIcons,
+  ...codicons.map(codicon => `${codiconsPath}/${codicon}.svg`)
+]
+
+const toPascalCase = string =>
```
julieg18, line 68: If I'm understanding this code correctly, it's converting a kebab-case word to a PascalCase word? If so, I found this [simpler solution on SlackOverflow](https://stackoverflow.com/a/54651317) a while back:

```
function toPascalCase(text) {
  return text.replace(/(^\w|-\w)/g, clearAndUpper);
}

function clearAndUpper(text) {
  return text.replace(/-/, "").toUpperCase();
}
```

CONTRIBUTING.md
```diff
@@ -93,3 +93,27 @@ There are some discrepancies between the Storybook environment and the
...
+circumstance, none of the codicon meets your needs, you can add a custom SVG to
+the `webview/icons` directory.
+
+If you have added any icon (custom or from codicon), you will have to run
+`yarn svgr` or `yarn install` to ensure the icon is now available in the
+codebase.
+
+You should also add any added icon to `webview/src/stories/Icons.stories.tsx`.
+
+To use an icon, import it from `webview/src/shared/components/icons` (from the
```
julieg18, line 117: Do we need this last paragraph about importing? I feel like it would be obvious how to import and render an Icon by looking at the codebase (it definitely was for me when I needed an icon for the first time).

webview/icons/generate.mjs
```diff
@@ -0,0 +1,103 @@
...
+    if (path.extname(file) === '.svg') {
+      customIcons.push(`icons/${file}`)
+    }
+  }
+} catch (err) {
+  console.error(err.message)
+}
+
+const codiconsPath = '../node_modules/@vscode/codicons/src/icons'
+const codicons = [
```
julieg18, line 34: Since this is a constant that will be added to/deleted from to create icons, should we place it in a separate constants file?

webview/src/shared/components/icons/ChevronRight.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react'
-import { SVGProps } from 'react'
-const SvgChevronRight = (props: SVGProps<SVGSVGElement>) => (
-  <svg
```
julieg18, line 4: When I ran `svgr` for the first time, I got three changes to the branch:

* `icons/index.ts` being sorted into a different order
* `ChevronRight.tsx` got a new `<svg` line which I now see was missing. 
* `Check` and `ArrowDown` losing new lines

Are these changes normal/expected or could their be an issue with how `generage.mjs` works?

julieg18 (approved):
Great work! This will make handling icons a lot easier!

webview/package.json
sroy3: Yes, that is why I also added steps in the contributing guide. To be honest, I've been creating them manually as well for a while to because the process was broken. Running the command would generate every codicon, you would then have to remove all the ones you don't need.

webview/icons/generate.mjs
sroy3: Nice! I just grabbed the first thing I saw. This looks better.

CONTRIBUTING.md
sroy3: It depends where you import it. If there are no example in the file you're trying to import the icon from, looking at our other imports, you might go directly for the file. I know I've made the comment more than once in PRs. It's mostly that it goes against how we import in general (using an index file).

webview/icons/generate.mjs
sroy3: That is a great idea. Will do.

webview/src/shared/components/icons/ChevronRight.tsx
sroy3: It's probably because I merge "icons-cleanup" back into this branch and fixed conflicts manually. I'll clean this up.

sroy3: But yes, it is expected and wanted. It signals some manual changes were done and are being overwritten. I thought the `postinstall` script would run on the CI, but it might not since the `yarn.lock` didn't change between builds. I'll look into a way to make sure it runs.

sroy3: > `icons/index.ts` being sorted into a different order

This should be normalized though, I'll fix it.

.github/workflows/continuous-integration.yml
```diff
@@ -38,6 +38,8 @@ jobs:
 
       - run: yarn run install-frozen-lockfile
 
+      - run: yarn svgr
+
```
sroy3, line 42: This makes sure no one manually creates or edits icons.

webview/src/shared/components/icons/ChevronRight.tsx
sroy3: Added the script as a step to our CI and fixed the order of the `index.ts` file.

PR #3896, opened 2023-05-15
Fix broken setup stories

julieg18:
Noticed while working on #3895 that our setup stories were broken

Leftover from #3851

### DIFF
--- a/webview/src/stories/Setup.stories.tsx
+++ b/webview/src/stories/Setup.stories.tsx
@@ -1,9 +1,12 @@
 import type { StoryFn, Meta } from '@storybook/react'
 import React from 'react'
+import { Provider, useDispatch } from 'react-redux'
+import { configureStore } from '@reduxjs/toolkit'
 import { MessageToWebviewType } from 'dvc/src/webview/contract'
 import { SetupData, SetupSection } from 'dvc/src/setup/webview/contract'
 import { DISABLE_CHROMATIC_SNAPSHOTS } from './util'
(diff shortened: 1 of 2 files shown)


webview/src/stories/Setup.stories.tsx
```diff
@@ -42,7 +56,13 @@ export default {
 } as Meta
 
 const Template: StoryFn = ({ data }) => {
-  const app = <App />
+  const app = (
+    <Provider store={configureStore({ reducer: setupReducers })}>
```
sroy3, line 60: You can pass a preloaded state here instead of using dispatch and feeding the store like you currently do. The good part about this is that the story is ready on the get-go, it's not going to change while the store is being fed. Also, you won't need to export stuff from the webview simply for the story.

```
<Provider
      store={configureStore({
        preloadedState: { ...data },
        reducer: experimentsReducers
      })}
    >
```

sroy3 (approved):

mattseddon (approved):

webview/src/stories/Setup.stories.tsx
julieg18: On further inspection, we'd need to adjust our `DEFAULT_DATA` to be configured like our redux state object for us to pass it as our `preloadedState`. While we could do that, it would make the `DEFAULT_DATA` harder to adjust with its nested state objects. I'll keep the `MockedState` component for now to make story creation simpler.

PR #3904, opened 2023-05-16
Add upgrade dvc button to setup when incompatible

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/517c6355-2098-4143-8158-eadf2b3745a2

Part of #3434


extension/src/setup/autoInstall.ts
```diff
@@ -16,6 +16,28 @@ export const findPythonBinForInstall = async (): Promise<
   )
 }
 
+const showUpgradeProgress = (
+  root: string,
+  pythonBinPath: string
+): Thenable<unknown> =>
+  Toast.showProgress('Upgrading DVC', async progress => {
```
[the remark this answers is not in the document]
mattseddon: Take a look at this:

https://github.com/iterative/vscode-dvc/blob/e95cc0eddf878d3e72acb5cdd1987886a288f825/extension/src/experiments/commands/index.ts#L58-L77

Can start by making a small increment and then updating again afterwards. That gives the impression to the user that something is happening. Doesn't have to be super accurate until he have a way to give discrete progress updates.

extension/src/setup/autoInstall.ts
```diff
@@ -16,6 +16,28 @@ export const findPythonBinForInstall = async (): Promise<
...
+    progress.report({ increment: 0 })
+    try {
+      await Toast.runCommandAndIncrementProgress(
+        async () => {
+          await installPackages(root, pythonBinPath, 'dvc')
+          return 'Upgraded successfully'
```
mattseddon, line 29: The assumption here is that the user used pip to install DVC which isn't necessarily true. I think this is fine but we should add a quantifier to the button as an interim solution.

webview/src/setup/components/dvc/CliIncompatible.tsx
```diff
@@ -1,15 +1,38 @@
...
+export const CliIncompatible: React.FC<PropsWithChildren> = ({ children }) => {
+  const pythonBinPath = useSelector(
+    (state: SetupState) => state.dvc.pythonBinPath
+  )
+  const canUpgrade = !!pythonBinPath
+
+  const conditionalContents = canUpgrade ? (
+    <>
+      <div className={styles.sideBySideButtons}>
+        <Button onClick={upgradeDvc} text="Upgrade" />
```
mattseddon, line 18: perhaps just add `(pip)` i.e `Upgrade (pip)`

mattseddon (approved):

extension/src/setup/autoInstall.ts
julieg18: > The assumption here is that the user used pip to install DVC which isn't necessarily true. I think this is fine but we should add a quantifier to the button as an interim solution.

Good idea! I'll be opening an issue soon detailing what can be improved with how we handle the installation of DVC and will be sure to mention this. Out of scope, but we should probably add a quantifier to our "Install" button as well. Same issue with it assuming the use of `pip`.

extension/src/setup/autoInstall.ts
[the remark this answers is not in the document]
julieg18: Updated! Looks better, thanks :)

PR #3921, opened 2023-05-17
Fix row selection (duplicate experiments)

sroy3:
For #1966 

https://github.com/iterative/vscode-dvc/assets/3683420/912a6161-57ab-496e-8883-a78410ae553d


mattseddon:
This PR highlights a bigger problem. Having duplicate entries in the table is going to lead to some unexpected behaviour. For example selecting a single record for plotting adds duplicate entries into the `plots diff` call:


https://github.com/iterative/vscode-dvc/assets/37993418/f585f517-1ee0-4f56-b587-dcac70581d9a

☝🏻 I'll fix this right now. I think the behaviour for the checkboxes makes sense but doing things like starring/deleting etc should update entries for that record.

webview/src/experiments/components/table/body/RowContextMenu.tsx
```diff
@@ -324,7 +327,7 @@ const getContextMenuOptions = (
 
 export const RowContextMenu: React.FC<RowProp> = ({
   row: {
-    original: { status, starred, id, executor },
+    original: { branch, status, starred, id, executor },
```
mattseddon, line 330: [I] When sending back a list of IDs (e.g. for removal) can we ensure that the list is unique, please?

webview/src/experiments/components/table/body/Row.tsx
```diff
@@ -24,14 +25,15 @@ export const RowContent: React.FC<
...
+  const { branch, displayColor, error, starred, id, status, selected } =
+    original
   const [firstCell, ...cells] = getVisibleCells()
   const isWorkspace = id === EXPERIMENT_WORKSPACE_ID
   const changesIfWorkspace = isWorkspace ? changes : undefined
 
   const { toggleRowSelected, selectedRows } = useContext(RowSelectionContext)
 
-  const isRowSelected = !!selectedRows[id]
+  const isRowSelected = !!selectedRows[getCompositeId(id, branch)]
```
mattseddon, line 36: [Q] Should we call this `rowId` and put it on the experiment data?

mattseddon (approved):

julieg18 (approved):
Good work!

webview/src/experiments/components/table/body/Row.tsx
sroy3: It was more complicated doing that as it needs to be added everywhere and it's only useful for row selection (something happening on the webview side only).

PR #3923, opened 2023-05-17
Increase timeout of flaky test

mattseddon:


julieg18 (approved):

extension/src/test/suite/config.test.ts
```diff
@@ -159,6 +159,6 @@ suite('Config Test Suite', () => {
         config.getFocusedProjects(),
         'should exclude projects that are outside of the workspace'
       ).to.equal(undefined)
-    })
+    }).timeout(10000)
```
sroy3, line 162: This is fine as a hack, but there are 8 `expect` calls in that test; maybe breaking it up in multiple tests would help more than increasing the timeout.

sroy3 (approved):

Issue #3913, opened 2023-05-17
experiments table unstable after running experiment


mattseddon:
I am seeing this behaviour after running an experiment:

https://github.com/iterative/vscode-dvc/assets/37993418/4ba8b997-c444-465c-860a-654ca2e413fc

mattseddon:
Also seeing lots of duplicate entries after running an experiment:

<img width="1728" alt="image" src="https://github.com/iterative/vscode-dvc/assets/37993418/00e3ae89-4ff3-411e-8058-6a60547dba85">

Can recreate weird behaviour in a non-submodule version of the project as well:


https://github.com/iterative/vscode-dvc/assets/37993418/2e94fd49-7b58-440a-bed8-a359ee0344f5

mattseddon:
PTAL @sroy3

mattseddon:
Looks like the error is here:  https://github.com/iterative/vscode-dvc/blob/7af45f8004213733fe4bb46d176b57c9b1a77279/extension/src/experiments/data/index.ts#L82-L95

When an experiment gets run `HEAD` is moved into a detached state. Which means that we end up with `getCurrentBranch` returning:

- `main` (before experiment) 
- short sha (during experiment) 
- `main` (after experiment) 
 
We get what are essentially duplicate commits set into `experiments.branchesToShow`. Which in turn duplicates all of the entries from `exp show` and the table can start convulsing.

sroy3:
I'll look into this

sroy3:
These seem related to the cleanup when changing branches. `8c3902f` is still present while on the main branch. Working on this today so hopefully that stabilize things.

PR #3925, opened 2023-05-18
Ensure unique list of experiment passed to the rest of the app

mattseddon:
Related to #3913 and for #1966

If an experiment was displayed twice in the experiments table/tree it could still be selected for plotting. This would result in the same id being sent to `plots diff` twice and the command would not return the required data. That resulted in an infinite loop and the plots not loading. Behaviour is shown below:

https://github.com/iterative/vscode-dvc/assets/37993418/0508c45e-445b-4f30-934b-a813c5114b62

In order to combat this I've ensured that whenever we access the commits data (outside of the table) that the list is de-duplicated. I've also made sure that we don't collect the same experiment twice (as they are stored per commit).

### Demo (after change)
 
https://github.com/iterative/vscode-dvc/assets/37993418/1f5358dc-f2a7-4eac-bc88-bbd751a5cbb2

Please LMK if you have any questions

### DIFF
--- a/extension/src/experiments/model/collect.ts
+++ b/extension/src/experiments/model/collect.ts
@@ -251,16 +251,25 @@ const collectExpRange = (
   const expState = revs[0]
 
   const { name, rev } = expState
-  const { branch, id } = baseline
+  const { id: baselineId } = baseline
 
   const label =
     rev === EXPERIMENT_WORKSPACE_ID
       ? EXPERIMENT_WORKSPACE_ID
       : shortenForLabel(rev)
 
+  const experimentId = name || label
+
+  if (
+    acc.experimentsByCommit
+      .get(baselineId)
+      ?.find(({ id }) => id === experimentId)
+  ) {
+    return
+  }
+
   const experiment = transformExpState(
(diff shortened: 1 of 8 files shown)


extension/src/experiments/model/collect.ts
```diff
@@ -251,16 +251,25 @@ const collectExpRange = (
...
       : shortenForLabel(rev)
 
+  const experimentId = name || label
+
+  if (
+    acc.experimentsByCommit
+      .get(baselineId)
+      ?.find(({ id }) => id === experimentId)
+  ) {
+    return
```
mattseddon, line 268: [F] Previously we were pushing duplicates in here. By dropping the branch name we can avoid that. We just need to make sure that we add the branch name before sending the experiment to the table.

shcheklein:
Thanks @mattseddon !

shcheklein (approved):

mattseddon:
@sroy3, @julieg18 I'm going to ship this now but please review and I will retro-actively address any concerns 🙏🏻.

Issue #3867, opened 2023-05-11
Onboarding to DVC Remotes


mattseddon:
The main goal of this story is to get new users onboarded to using remotes. We should also provide a happy path for experienced users who are setting up new projects but that is a secondary concern.

## Steps

- [x] Add a remote section to the setup page which displays current remote information with pass/fail status depending on whether or not a remote is detected. (https://github.com/iterative/vscode-dvc/pull/3901)
- [x] Enable `add`, `default`, `remove`, `rename` & `modify` commands to be accessed via the setup page and command palette (#3956)
- [x] Provide links to (custom) auth portions of the [docs](https://dvc.org/doc/user-guide/data-management/remote-storage) for specific storage options (see Helping with credentials below) (#3996)
- [x] Add links to display the relevant extension per storage provider (see https://github.com/iterative/vscode-dvc/issues/3867#issuecomment-1544195549) (#4004).
- [x] Condense text for remote types (remove cloud versioning) (#4051)
- [ ] Display remote information in the sidebar (similar to the strategy and reasoning for the Studio "tree") - this could potentially also display whether or not the remote is up to date with data and have a refresh button to check on the status.
- [ ] Enable quickstart with some kind of provided storage option (more Quickstart below)

### Helping with credentials

Just from looking into the docs setting up some set of quick inputs/picks to assist with this feels like it would be a story by itself. For example AWS can take a `configpath`, `credentialpath`, `profile`... and more, Azure blob storage can take a `connection_string` or... you get the point. This task wouldn't be impossible but would be time consuming.

### Quickstart

tl;dr - I think we should provide a GET and POST endpoint from Studio that hides the implementation completely from the user.

There are a lot of questions surrounding this part of the story. If the idea is to drive Studio usage then we should tie Studio in. This then limits the implementation options. To me, the simplest thing to do would be along the lines of wrapping S3 with the Python AWS SDK in the Studio backend and providing a set of endpoints that we can hit with a specific token/repo URL/ref and associated data (for POST only). To begin with, we can write a standalone client for the extension but (IMO) this should be migrated into DVC when we have the capacity to do so. By this, I mean that "Studio" should be a valid option for the remote entry in the DVC config. The user would be able to put this in their project config along with the token in their local one. I wouldn't go as far as setting Studio as the default if a token is provided and a remote is not but that could also be an option.

[...]

PR #3929, opened 2023-05-18
Add command to add remote

mattseddon:
# 1/2 `main` <- this <- #3930 

Part of #3867

This PR adds a start button to the setup webview which allows the user to add a remote if they do not already have one in the workspace. I have also added the command to the palette. 

In the case that the user does not have a remote then I set the new remote as the default. If they do already have a remote then I show a quick pick so that they can confirm whether or not they want to set the new remote as the default.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/f77a4a91-0516-4ef2-a845-84f31e25f802


webview/src/setup/components/remote/Connect.tsx
```diff
@@ -18,5 +20,6 @@ export const Connect: React.FC = () => (
       </a>{' '}
       for details on how to connect to a remote
     </p>
+    <StartButton onClick={() => addRemote()} text="Add Remote" />
```
julieg18, line 23: Since you can have multiple remotes in a project, should we keep showing the "Add Remote" button when a user has already added remotes?

sroy3 (approved):

julieg18 (approved):

webview/src/setup/components/remote/Connect.tsx
julieg18: Actually, now that I think about it more, the goal of this section is to setup remotes, not manage remotes, so it does make sense for the button to disappear.

mattseddon: Discussed this today. For anyone else finding this. Adding the command is in the works but the reasons for leaving out the button from the remotes connected page in this PR are as follows:

1. Would be strange to only have a single option to add remotes when you have already set at least one up (need to add more actions to make this seem reasonable)
2. PR is big enough

PR #3936, opened 2023-05-18
Remove code slider fixed height

julieg18:
* stops the code slider from having extra bottom margin

### Main

Screenshot 2023-05-18 at 6 47 50 PM

### PR

Screenshot 2023-05-18 at 6 47 43 PM

Part of #3434


webview/src/shared/components/codeSlider/styles.module.scss
```diff
@@ -2,7 +2,6 @@
 
 .codeSlider {
   margin-top: 40px;
-  height: 300px;
```
julieg18, line 5: I'm guessing that this fixed height was originally there to stop the page contents below the code blocks from "jumping" when the user switches tabs. @sroy3 might be able to confirm this. I actually might have even been the one to ask for this to be fixed if I'm remembering correctly 🤔 

Hmmm, I wonder if there is a way to solve both problems... Ah, wait, what if we kept the fixed height but decreased it a bit and moved it to the `pre` block? That way the extra space would look less like unneeded margin and just part of the code block. I'll give it a try tomorrow morning!

julieg18: > Ah, wait, what if we kept the fixed height but decreased it a bit and moved it to the pre block? 

Tried it, it looks better but I think it still looks like too much spacing, just for codeblocks padding instead of tabs margin:

https://github.com/iterative/vscode-dvc/assets/43496356/05ff9bc3-1a36-44c6-a9f3-04bb28cbfda3

I'll just take off the fixed height, I doubt most people will notice page jumps since they won't really be clicking from tab to tab quickly :)

sroy3 (approved):

PR #3941, opened 2023-05-19
Use expShow once with multiple revisions

sroy3:
Replace the multiple parallel calls to exp show with one simple call with all the wanted revisions in it. Part of #1966 

https://github.com/iterative/vscode-dvc/assets/3683420/a65d2ad9-d171-4116-9f0b-f3fc63d8a36c


mattseddon (approved):

extension/src/experiments/data/index.ts
```diff
@@ -47,36 +47,33 @@ export class ExperimentsData extends BaseData<ExpShowOutput> {
...
-        data.push(...output)
-      })
+    const flags = []
+    const branchList: string[] = []
+
+    for (const branch of branches) {
+      const nbOfCommitsToShow = this.experiments.getNbOfCommitsToShow(branch)
+      for (let i = 0; i < nbOfCommitsToShow; i++) {
+        flags.push(ExperimentFlag.REV, `${branch}~${i}`)
+        branchList.push(branch)
```
mattseddon, line 62: [I] There is an issue when we call for the same revision twice (from different branches). Only the first instance of the revision will be returned. This leads to some unexpected results in the table. In the below demo the `z-last-branch` moves around because the `branchList` array is longer than the `output`:

https://github.com/iterative/vscode-dvc/assets/37993418/c3a45dd0-103b-4713-94ec-92903de2c3ec

mattseddon: Options as I see them:

1. Revert this and ask for a fix in DVC to return multiple records per revision (I don't think that will happen).
2. Create a mapping between branches and commits with git before/after we make the DVC call.
3. Revert this and don't worry about multiple calls to `exp show` for now.

mattseddon: Asked the question for 1 here: https://github.com/iterative/dvc/issues/9390#issuecomment-1560459523.

I think my vote is for 3 as the data is cached on the DVC side and we are already making a lot of calls to Git.

mattseddon: We have to call `git log` multiple times anyway (to get the commit messages) so why not move the calls before this to make the branch => commit + message mapping. E.g something like:

```
git log --no-merges --pretty=format:%H%n%an%n%ar%nrefNames:%D%nmessage:%B main~4..main~0 -z
```

We could then call `exp show` and map the deduplicated data back onto the branch data by rev... we might actually want to hold branch + commit and commit separately and only duplicate data when we send it out to the table.

sroy3:
Reverted here #3970

PR #3953, opened 2023-05-22
Fix setup redirection logic

julieg18:
* Has commands (show experiments, show plots, etc.) 
    * redirect to the DVC section if dvc is not setup
    * redirect to the Experiments section if dvc is setup but experiments isn't 

## main

https://github.com/iterative/vscode-dvc/assets/43496356/2345d6b2-d29c-4ece-b99e-e7e138e45fef

## pr

https://github.com/iterative/vscode-dvc/assets/43496356/3ddaa5cd-3751-48b0-977f-f935282a0a11

Resolves bug found in Slack thread: https://iterativeai.slack.com/archives/C01APS0FHDM/p1684790433607199


extension/src/commands/util.ts
```diff
@@ -5,7 +5,14 @@ import { Context } from '../vscode/context'
...
-      : callback(context)
+  (context: Context) => {
+    if (!setup.isDvcSetup()) {
+      return commands.executeCommand(RegisteredCommands.SETUP_SHOW_DVC)
+    }
+
+    if (!setup.isExperimentsSetup()) {
+      return commands.executeCommand(RegisteredCommands.SETUP_SHOW_EXPERIMENTS)
+    }
+
```
julieg18, line 16: Another option is to just open `SETUP_SHOW`, keeping all sections open no matter what is setup. I chose to write checking logic instead since I don't think it's overly complex and gets the user to where they need to go to get started with the extension a little faster :)

extension/src/test/suite/setup/index.test.ts
```diff
@@ -835,6 +835,24 @@ suite('Setup Test Suite', () => {
       )
     }).timeout(WEBVIEW_TEST_TIMEOUT)
 
+    it('should be to tell us if dvc and experiments is setup', async () => {
```
mattseddon, line 838: check on this

extension/src/setup/index.ts
```diff
@@ -244,8 +244,12 @@ export class Setup
     return !!this.webview?.isActive
   }
 
-  public shouldBeShown() {
-    return !this.getCliCompatible() || !this.hasRoots() || !this.getHasData()
```
mattseddon, line 248: [Q] Why not return an enum from this? so that callers don't need to remember to call `isDvcSetup` and then `isExperimentsSetup` in that order.

mattseddon (approved):

extension/src/setup/index.ts
julieg18: Good idea!

PR #3955, opened 2023-05-22
Standardize title entries to title case

mattseddon:
I was in this code whilst working on remote management. PR standardizes the casing to title-case.


sroy3 (approved):
Is there any lint rule we can add to this file specifically that'll do the checkup automatically?

mattseddon:
> Is there any lint rule we can add to this file specifically that'll do the checkup automatically?

I did look, I couldn't find anything 😞.

PR #3962, opened 2023-05-23
Add experiments table footer and move select branches action

sroy3:
Part of #1966 

Experimenting with a table footer for generic actions.

Demo

https://github.com/iterative/vscode-dvc/assets/3683420/b6e37404-ddeb-497e-a902-c8128ff5d1cc

IMO, the icon at the top for the action feels more appropriate. By not adding other generic actions anywhere else in the table, we educate the user that table actions can be performed by clicking one of these icons. I would even repeat the context menu when there is at least one row selected (like GMail does).

### DIFF
--- a/webview/src/experiments/components/table/TableFooter.tsx
+++ b/webview/src/experiments/components/table/TableFooter.tsx
@@ -1,0 +1,13 @@
+import React from 'react'
+import { AddAndRemoveBranches } from './commitsAndBranches/AddAndRemoveBranches'
+import styles from './styles.module.scss'
+
+export const TableFooter: React.FC = () => (
+  <tfoot className={styles.footer}>
+    <tr>
+      <td colSpan={99999}>
+        <AddAndRemoveBranches />
+      </td>
+    </tr>
+  </tfoot>
+)
--- a/extension/src/test/e2e/extension.test.ts
+++ b/extension/src/test/e2e/extension.test.ts
@@ -53,7 +53,9 @@ describe('Experiments Table Webview', function () {
   const workspaceRow = 1
   const commitRows = 3
   const branchRow = 1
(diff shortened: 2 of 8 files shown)


webview/src/experiments/components/table/TableFooter.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { AddAndRemoveBranches } from './commitsAndBranches/AddAndRemoveBranches'
+import styles from './styles.module.scss'
+
+export const TableFooter: React.FC = () => (
+  <tfoot className={styles.footer}>
```
julieg18, line 6: On a screen with a table shorter than the screen, this footer is difficult to spot:

image

We discussed the footer being at the bottom of the table instead in planning but I forgot exactly what conclusion was reached 🤔 Other possible ideas to make it stand out more:

* add a border
* add a box shadow like we do with our sticky header and column
* change the background

webview/src/experiments/components/table/TableFooter.tsx
```diff
@@ -0,0 +1,13 @@
...
+    <tr>
+      <td colSpan={99999}>
+        <AddAndRemoveBranches />
```
julieg18, line 9: Out of scope, but I'll try to find some time to look into a more button-like design that could possibly look better for actions like these in our table and setup page :)

julieg18 (approved):
Great work!

julieg18:
Minor, but I noticed that the fixed header isn't fixed on storybook: 

https://github.com/iterative/vscode-dvc/assets/43496356/dadb5b39-b32a-4f04-a4e1-398976b53300

sroy3:
Will only use icons instead.

PR #3966, opened 2023-05-24
Reduce the number of calls to git log

mattseddon:
Whilst debugging #3941 I noticed that we are making multiple calls to `git log` (each time that we call `exp show`) when we only need to make a single call.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/e5f0bf35-36d0-4c66-8b6a-fbdfa9f327bb

### DIFF
--- a/extension/src/cli/git/reader.ts
+++ b/extension/src/cli/git/reader.ts
@@ -49,15 +49,16 @@ export class GitReader extends GitCli {
     return !output
   }
 
-  public async getCommitMessages(cwd: string, sha: string): Promise<string> {
+  public async getCommitMessages(
+    cwd: string,
+    ...shas: string[]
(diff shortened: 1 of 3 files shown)


extension/src/cli/git/reader.ts
```diff
@@ -49,15 +49,16 @@ export class GitReader extends GitCli {
...
-  public async getCommitMessages(cwd: string, sha: string): Promise<string> {
+  public async getCommitMessages(
+    cwd: string,
+    ...shas: string[]
+  ): Promise<string> {
     const options = getOptions(
       cwd,
       Command.LOG,
-      sha,
+      ...shas,
```
mattseddon, line 59: This call to git log returns the entire log.

PR #3972, opened 2023-05-24
Use icons to show more or less commits and select branches

sroy3:
https://github.com/iterative/vscode-dvc/assets/3683420/5f01673b-e9ff-408c-9ef2-aa6093a97ee7

We can also label the show more / less commits like so:
Screenshot 2023-05-24 at 1 15 16 PM


webview/src/experiments/components/table/Indicators.tsx
```diff
@@ -139,6 +146,7 @@ export const Indicators = () => {
           branchesSelected,
           'Selected'
         )}
+        disabled={!hasBranchesToSelect}
```
mattseddon, line 149: [Q] Can we give the same branch icon next to the branch name the same action? We could even extend the behaviour on to the branch name. I.e if the user clicks on the branch icon or the branch name then the picker is launched. Having the + - closer may make it more obvious that all of this is clickable.

mattseddon (approved):
Code looks good. Left one question for follow-up in the review. Also, should we make all of the icons/indicators the same color?

webview/src/experiments/components/table/Indicators.tsx
sroy3: I don't know. We're moving all actions to icons here; making the branch name clickable at the same time might be counterintuitive. The goal of this PR was also to remove the generic action from repeating on all branches. Maybe it'll work still.  I'll explore this in a follow-up.

webview/src/experiments/components/table/body/commits/CommitsNavigation.tsx
```diff
@@ -0,0 +1,41 @@
...
+    {
+      action: () => showLessCommits(branch),
+      disabled: !isShowingMoreCommits[branch],
+      icon: Remove,
+      moreOrLess: 'Less'
+    }
+  ]
+
+  return (
+    <div className={styles.commitsNav}>
```
julieg18, line 35: At first glance, it felt unclear that the button increased/decreased commits, but the instant tooltips help a lot!

webview/src/experiments/components/table/body/commits/CommitsButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const CommitsButton: React.FC<CommitsButtonProps> = ({
+  icon,
+  moreOrLess,
+  action,
+  disabled
+}) => {
+  const text = `Show ${moreOrLess} Commits`
+  return (
+    <Tooltip content={<>{text}</>}>
+      <button
```
julieg18, line 21: Maybe it's just me and my bad eyesight 👓, but the icons look pretty small compared to other table elements/icons. Should we consider increasing the size  by a couple pixels?

julieg18 (approved):

Issue #3864, opened 2023-05-10
Simplify Studio onboarding screen and workflow


shcheklein:
Two things that I found recently trying it:

Related https://github.com/iterative/vscode-dvc/issues/3434

- [x] I think we should have a single button Get Token (Studio can redirect to the right page, and even if doesn't now I think we should be fine for now). @iterative/studio-frontend can Studio redirect now? What is the complexity of this? To clarify - if I sign up I want it to immediately open the page with my token. We next need to think about also creating a project automatically or semi-automatically since we know the repo URL potentially and can pass it in the studio URL.
- [ ] If an ENV variable with a token detect we need to at least say something about it. ENV var in Codespaces is enough to log metrics actually, and some options on this page can contradict with it. E.g. should we show that checkbox enabled it token ENV var is found and if ppl disable it then unset the env?
- [x] After https://github.com/iterative/dvc/pull/9433, live sharing will be enabled by default when the studio token is saved to the dvc config, although it can be disabled with studio.offline = true. To comply with this, we need to make the "Share new experiments live" checkbox correspond to the studio.offline config value and stop setting the environment variable.

<img width="1248" alt="Screenshot 2023-05-10 at 10 48 22 AM" src="https://github.com/iterative/vscode-dvc/assets/3659196/fb0b0389-f1ea-4d00-aed6-c9fbbca7a880">

mattseddon:
The first item is covered by the last item in https://github.com/iterative/vscode-dvc/issues/3574.

> Drop the sign-in button from Studio setup page (https://github.com/iterative/vscode-dvc/issues/3574#issuecomment-1513949115 - related to https://github.com/iterative/studio/issues/5715)

The Studio issue for frontend redirects is https://github.com/iterative/studio/issues/5715.

We can remove the original action item from https://github.com/iterative/vscode-dvc/issues/3574 if you like.

shcheklein:
@mattseddon thanks, yep. Dropped it in that ticket (making it closer to close that one).

dberenbaum:
> I think we should have a single button Get Token (Studio can redirect to the right page, and even if doesn't now I think we should be fine for now). 

👍 

> @iterative/studio-frontend can Studio redirect now? What is the complexity of this? To clarify - if I sign up I want it to immediately open the page with my token. We next need to think about also creating a project automatically or semi-automatically since we know the repo URL potentially and can pass it in the studio URL.

[...]

PR #3976, opened 2023-05-25
Bump min version of DVC to 2.57.0 (Live share to studio config option)

mattseddon:
# 1/2 `main` <- this <- #3987

Relates to one of the checkboxes in #3864

This PR removes our custom logic for sharing experiments live to Studio if the user has a `studio.token` config value set. We now correctly respect the new `studio.offline` DVC config value instead of the old `dvc.studio.shareExperimentsLive` extension config value. The default behaviour is that experiments will be shared live if a token is set (as per the [docs](https://dvc.org/doc/user-guide/project-structure/configuration#studio)).

#### Checkbox updating config

https://github.com/iterative/vscode-dvc/assets/37993418/76a08dd5-f42a-4d1b-8f20-fada8d21c98e

#### Workspace run - sharing enabled

https://github.com/iterative/vscode-dvc/assets/37993418/c727b074-d9bb-46cb-b65e-a27ab383ff2a

#### Queue run - sharing enabled

https://github.com/iterative/vscode-dvc/assets/37993418/6c867d8a-6997-41a3-aeb8-a2b0a2a182d5

#### Workspace run - sharing disabled

https://github.com/iterative/vscode-dvc/assets/37993418/f6296e80-8c89-4414-a850-4ed7803f8e80

#### Queue run - sharing disabled

https://github.com/iterative/vscode-dvc/assets/37993418/32ece748-06f4-430b-8b3c-da0d07e5ee1f


webview/src/setup/components/studio/Settings.tsx
```diff
@@ -30,9 +31,16 @@ export const Settings: React.FC<{
...
-            Share New Experiments Live
+            Share New Experiments Live*
           </VSCodeCheckbox>
         </p>
+        <p className={styles.smallFont}>
+          *The checkbox reflects the{' '}
+          <a href="https://dvc.org/doc/user-guide/project-structure/configuration#studio">
+            studio.offline
+          </a>{' '}
+          config option.
```
mattseddon, line 42: [F] I have added this small bit of text instead of trying to workout which config file(s) `studio.offline` is set in and then trying to update the appropriate level(s) of config. Much like updating the `studio.token` this button will only call `config` with the `--global` flag. If users want to save `studio.token`/`studio.offline` in project/local configs then they should be capable of managing this themselves. If this does however become an issue I will fix it retroactively.

mattseddon: I've started the process of discussing with the DVC team how we can make this easier to deal with.

julieg18: Should we hint that were setting things globally in the note? For example, `The checkbox reflects the global [studio.offline](https://dvc.org/doc/user-guide/project-structure/configuration#studio) config option.`

webview/src/setup/components/studio/styles.module.scss
```diff
@@ -0,0 +1,3 @@
+.smallFont {
+  font-size: 0.6rem;
```
julieg18, line 2: I don't have the best eyesight, but I had to squint a bit to read this 🤣. Should we increase it a bit? Example of `0.7rem`: 

image

julieg18 (approved):
Great work!

webview/src/setup/components/studio/Settings.tsx
mattseddon: The text in the example isn't necessarily true. If the user has set the option at the `project` or `local` level then that option will override what we set at the global level. This is from the docs and hopefully helps to explain a bit further:

image

I have only added the text so that in the (rare) case that the user has set `studio.offline=true` at the project/local level that the user will have some chance of debugging what is going on.

In order to fully explain this I think I would need to add 2 paragraphs which IMO is not necessary at this stage for onboarding. 

This issue does seem to be coming up a lot though so I think I may have to build some representation of `dvc config -l --show-origin` to add to the page.

PR #3980, opened 2023-05-25
Add a mapping from revision to branches while using only one exp show call

sroy3:
For #1966 

https://github.com/iterative/vscode-dvc/assets/3683420/3f85d354-2b1e-4334-b1dc-39a53bbea8d6

Named revision might still change (only one is being returned by dvc), but no more shift of commits between branches.


extension/src/experiments/data/index.ts
```diff
@@ -86,6 +93,58 @@ export class ExperimentsData extends BaseData<ExpShowOutput> {
     this.collectedFiles = collectFiles(data, this.collectedFiles)
   }
 
+  private async collectRevisionGitDetails(
```
sroy3, line 96: I created those to reduce complexity, but they're more like helper functions than belonging to a class. Had to keep them in the class because of `this.experiments.getNbOfCommitsToShow(branch)` and `this.internalCommands.executeCommand()`, but arguments shouldn't be moved to private properties as they renew with each  `update` call.

extension/src/experiments/data/index.ts
```diff
@@ -86,6 +93,58 @@ export class ExperimentsData extends BaseData<ExpShowOutput> {
...
+
+  private addDetailsToRevision(
+    out: ExpState & {
+      experiments?: ExpRange[] | null | undefined
+    },
+    data: ExpShowOutput,
+    hashes: Record<string, HashInfo>
+  ) {
+    const revision = hashes[out.rev]
+    if (revision) {
```
sroy3, line 135: This `if` was added to get tests to pass at this stage. Might be a good thing to remove it once fixtures are in place correctly.

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1402,7 +1402,7 @@ suite('Experiments Test Suite', () => {
   })
 
   describe('Sorting', () => {
-    it('should be able to sort', async () => {
+    it.only('should be able to sort', async () => {
```
sroy3, line 1405: Only failing test (at the moment).

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1470,6 +1470,8 @@ suite('Experiments Test Suite', () => {
         rows: [
```
sroy3, line 1470: It really fails here, but when diffing both, they're identical. Not sure what's happening and I spent way too much time on this.
Screenshot 2023-05-31 at 2 32 59 PM

extension/src/experiments/data/index.ts
```diff
@@ -86,6 +93,58 @@ export class ExperimentsData extends BaseData<ExpShowOutput> {
...
   }
 
+  private async collectRevisionGitDetails(
+    branch: string,
+    hashes: Record<string, HashInfo>,
+    flags: (ExperimentFlag | string)[]
+  ) {
+    const nbOfCommitsToShow = this.experiments.getNbOfCommitsToShow(branch)
+
+    for (let i = 0; i < nbOfCommitsToShow; i++) {
```
mattseddon, line 103: I think we can get all of these in 1 hit with something like 
```
git log --pretty=format:%H%n%an%n%ar%nrefNames:%D%nmessage:%B map-revisions -z -n 6 ```

File: extension/src/experiments/data/index.ts [hunk 6]
```diff
@@ -86,6 +93,61 @@ export class ExperimentsData extends BaseData<ExpShowOutput> {
...
+  ) {
+    const revision = hashes[out.rev]
+    if (revision) {
+      const { branches, ...commit } = revision
+      for (const branch of branches) {
+        data.push({
+          ...out,
+          branch,
+          commit,
+          description: formatCommitMessage(commit.message)
```
mattseddon, line 145: [Q] Can I move this upstream into the collect call?

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1464,59 +1465,24 @@ suite('Experiments Test Suite', () => {
...
-                displayColor: undefined,
-                description: '[exp-3]',
-                id: 'exp-3',
-                label: '333333',
-                params: { 'params.yaml': { test: 3 } },
-                selected: false,
-                starred: false
-              }
-            ]
+      const getIds = (rows: Commit[]) =>
```
mattseddon, line 1468: [F] I always run into/get annoyed by this test. It was previously testing more than it needed to. It is not testing only the functionality that it needs to.

extension/src/experiments/data/index.ts
```diff
@@ -47,43 +52,58 @@ export class ExperimentsData extends BaseData<ExpShowOutput> {
...
+      AvailableCommands.EXP_SHOW,
+      this.dvcRoot,
+      ...args
     )
 
-    this.collectFiles(data)
+    this.collectFiles({ expShow })
 
-    return this.notifyChanged(data)
+    return this.notifyChanged({ currentBranch, expShow, gitLog, order })
```
mattseddon, line 75: [F] By doing the least amount of processing possible here we can update the test fixture(s) with raw data and make sure that everything still works.

extension/src/cli/git/reader.ts
```diff
@@ -49,15 +49,19 @@ export class GitReader extends GitCli {
...
     const options = getOptions(
       cwd,
       Command.LOG,
-      sha,
+      revision,
       Flag.PRETTY_FORMAT_COMMIT_MESSAGE,
       Flag.SEPARATE_WITH_NULL,
       Flag.NUMBER,
-      '1'
+      revisions
```
mattseddon, line 64: [F] Now we're getting the logs in fewer calls to git. This does include merge commits. We might want to get rid of those later.

extension/src/data/index.ts
```diff
@@ -9,8 +9,15 @@ import { uniqueValues } from '../util/array'
 import { DeferredDisposable } from '../class/deferred'
 import { isSameOrChild } from '../fileSystem'
 
+export type ExperimentsOutput = {
+  currentBranch: string
+  expShow: ExpShowOutput
+  order: { branch: string; sha: string }[]
```
mattseddon, line 15: [F] Order is the key to duplication between branches. It is created from making multiple calls to `git log`.

extension/src/experiments/data/index.ts
```diff
@@ -47,43 +47,58 @@ export class ExperimentsData extends BaseData<ExpShowOutput> {
...
+  private async collectGitLogAndOrder(
+    gitLog: string,
+    branch: string,
+    order: { branch: string; sha: string }[],
+    args: Args
+  ) {
+    const nbOfCommitsToShow = this.experiments.getNbOfCommitsToShow(branch)
+
+    const branchGitLog = await this.internalCommands.executeCommand(
+      AvailableCommands.GIT_GET_COMMIT_MESSAGES,
```
mattseddon, line 86: [F] As a follow up I think we want to pull the timestamp from the git log so that commits are displayed in chronological order in the tree.

extension/src/experiments/index.ts
```diff
@@ -169,13 +169,19 @@ export class Experiments extends BaseRepository<TableData> {
...
+    gitLog,
+    order
+  }: ExperimentsOutput) {
+    const data = combineOutputs(currentBranch, expShow, gitLog, order)
+
     const hadCheckpoints = this.hasCheckpoints()
     const dvcLiveOnly = await this.checkSignalFile()
-    const commitsOutput = await this.getCommitOutput(data)
     await Promise.all([
       this.columns.transformAndSet(data),
```
mattseddon, line 183: [I] This does not need to be duplicated. The columns are the same.

extension/src/experiments/data/index.ts
mattseddon: already done in here

extension/src/data/index.ts
```diff
@@ -9,8 +9,15 @@ import { uniqueValues } from '../util/array'
 import { DeferredDisposable } from '../class/deferred'
 import { isSameOrChild } from '../fileSystem'
 
+export type ExperimentsOutput = {
+  currentBranch: string
+  expShow: ExpShowOutput
+  rowOrder: { branch: string; sha: string }[]
```
mattseddon, line 15: [F] `rowOrder` is the key to duplication between branches. It is created from making multiple calls to `git log`. We no longer duplicate the commit data anywhere. We only send "duplicate" records to the experiments table. I have even sorted the commits by `Created` so they are displayed chronologically in the tree.

extension/src/experiments/model/collect.test.ts
```diff
@@ -48,44 +43,19 @@ describe('collectExperiments', () => {
...
     const { experimentsByCommit } = collectExperiments(
       expShowWithTwoCommits,
-      false,
-      ''
+      false
     )
     expect(experimentsByCommit.get('branchB')).toBeUndefined()
   })
 
-  it('should add data from git to commits if git log output is provided', () => {
```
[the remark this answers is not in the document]
mattseddon: covered by the integration tests/`it('should return the expected rows when given the base fixture'`

extension/src/experiments/model/index.test.ts
```diff
@@ -344,79 +365,23 @@ describe('ExperimentsModel', () => {
...
+    model.pruneBranchesToShow(['A', 'B', 'C', 'four', 'five', 'six'])
 
     expect(memento.get(PersistenceKey.EXPERIMENTS_BRANCHES)).toStrictEqual([
-      'one',
-      'two',
-      'three'
-    ])
-  })
-
-  it('should handle duplicate commits being returned in the data', () => {
```
mattseddon, line 367: [F] This can no longer happen

extension/src/experiments/model/index.ts
```diff
@@ -449,20 +454,6 @@ export class ExperimentsModel extends ModelWithPersistence {
     return this.currentSorts.findIndex(({ path }) => path === pathToRemove)
   }
 
-  private getCommits() {
```
mattseddon, line 452: [F] This is no longer required because we're not holding duplicates

extension/src/test/fixtures/expShow/base/customPlots.ts
```diff
@@ -30,6 +30,30 @@ export const experimentsWithCommits: Experiment[] = [
     },
     params: { 'params.yaml': { dropout: 0.122, epochs: 5 } }
   },
+  {
+    branch: 'main',
+    id: 'fe2919b',
```
mattseddon, line 35: [F] There are changes to tests and fixtures because I added two more commits to the base fixture.

extension/src/experiments/index.ts
```diff
@@ -169,13 +168,12 @@ export class Experiments extends BaseRepository<TableData> {
     return this.data.managedUpdate()
   }
 
-  public async setState(data: ExpShowOutput) {
+  public async setState({ expShow, gitLog, rowOrder }: ExperimentsOutput) {
```
mattseddon, line 171: [I] `getNumCommits` should come out of `WebviewMessages` and called in `ExperimentsData` as well.

extension/src/experiments/index.ts
```diff
@@ -596,7 +575,7 @@ export class Experiments extends BaseRepository<TableData> {
     }
 
     return await pickExperiment(
-      this.experiments.getUniqueList(),
+      this.experiments.getCombinedList(),
```
mattseddon, line 578: [F] This name had to change as the old name was misleading (data is no longer duplicated)

extension/src/experiments/index.ts
mattseddon: I've got an item on the board to follow up and do this but will do it next.

extension/src/experiments/model/collect.ts
```diff
@@ -325,14 +310,14 @@ const hasCheckpoints = (data: ExpShowOutput) => {
 }
 
 export const collectExperiments = (
-  output: ExpShowOutput,
-  dvcLiveOnly: boolean,
-  commitsOutput: string | undefined
+  expShow: ExpShowOutput,
+  gitLog: string,
```
mattseddon, line 314: [F] To me it makes more sense to put the `gitLog` output before `dvcLiveOnly` because it is more important for the data.

extension/src/experiments/model/index.ts
```diff
@@ -110,18 +112,22 @@ export class ExperimentsModel extends ModelWithPersistence {
...
     const {
       workspace,
       commits,
       experimentsByCommit,
       runningExperiments,
       hasCheckpoints
-    } = collectExperiments(data, dvcLiveOnly, commitsOutput)
+    } = collectExperiments(expShow, gitLog, dvcLiveOnly)
 
+    commits.sort((a, b) => (b.Created || '').localeCompare(a.Created || ''))
```
mattseddon, line 128: [F] Doing this means we get commits sorted in chronological order in the tree.

extension/src/test/suite/experiments/util.ts
```diff
@@ -19,8 +21,10 @@ import { PersistenceKey } from '../../../persistence/constants'
 
 export const buildExperiments = (
   disposer: Disposer,
-  experimentShowData = expShowFixture,
-  dvcRoot = dvcDemoPath
+  expShow = expShowFixture,
+  dvcRoot = dvcDemoPath,
+  gitLog = gitLogFixture,
+  rowOrder = rowOrderFixture
```
mattseddon, line 27: [I] Use VS Code `Refactor...` `Convert parameters to destructured object`

extension/src/test/suite/plots/util.ts
```diff
@@ -17,15 +19,27 @@ import { BaseWorkspaceWebviews } from '../../../webview/workspace'
...
+} from '../../../cli/dvc/contract'
 import { ErrorsModel } from '../../../plots/errors/model'
 import { PersistenceKey } from '../../../persistence/constants'
 
-export const buildPlots = async (
-  disposer: Disposer,
-  plotsDiff: PlotsOutput | undefined = undefined,
-  expShow = expShowFixtureWithoutErrors
-) => {
+export const buildPlots = async ({
```
mattseddon, line 30: [F] Did the same refactoring for this as it also has too many inputs now.

webview/__mocks__/@vscode/webview-ui-toolkit/react.tsx
```diff
@@ -36,3 +36,4 @@ const MockComponentWithChildren: React.FC<PropsWithChildren> = ({
 export const VSCodePanels = MockComponentWithChildren
 export const VSCodePanelTab = MockComponentWithChildren
 export const VSCodePanelView = MockComponentWithChildren
+export const VSCodeTag = MockComponentWithChildren
```
mattseddon, line 39: [F] Adding tags to the rows fixture meant that this was triggered and broke all of the jest tests.

julieg18 (approved):
Great work!

PR #3981, opened 2023-05-25
Simplify setup section text blocks

julieg18:
* helps consolidate section steps by using one text block in "Git Uninitialized" step and "Connect to Studio" step

Part of #3434

### DIFF
--- a/webview/src/setup/components/dvc/GitUnitialized.tsx
+++ b/webview/src/setup/components/dvc/GitUnitialized.tsx
@@ -10,26 +10,25 @@ export const GitUninitialized: React.FC<PropsWithChildren> = ({ children }) => {
     (state: SetupState) => state.dvc.canGitInitialize
   )
 
+  const startingSentence =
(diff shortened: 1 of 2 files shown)


webview/src/setup/components/dvc/GitUnitialized.tsx
```diff
@@ -10,26 +10,25 @@ export const GitUninitialized: React.FC<PropsWithChildren> = ({ children }) => {
...
 
+  const startingSentence =
+    'A Git repository is a prerequisite of project initialization.'
+
   const conditionalContent = canGitInitialize ? (
-    <Button onClick={initializeGit} text="Initialize Git" />
-  ) : (
     <>
-      <p>
-        The extension is unable to initialize a Git repository in this
```
julieg18, line 18: Figured this sentence isn't needed but, if thought otherwise, we can add it back with a little tweaking to make the sentences "flow" more.

mattseddon (approved):

PR #3982, opened 2023-05-25
Fix typo in setup "Git Uninitialized" setup step

julieg18:


julieg18:
Nevermind, not actually a typo 🤣

PR #3986, opened 2023-05-26
Fix issues with messages from the plots webview

mattseddon:
This PR fixes a couple of small issues with messages coming from the plots webview. Comments inline.

### DIFF
--- a/extension/src/plots/webview/messages.ts
+++ b/extension/src/plots/webview/messages.ts
@@ -13,7 +13,6 @@ import { Logger } from '../../common/logger'
 import { Experiments } from '../../experiments'
 import { sendTelemetryEvent } from '../../telemetry'
 import { EventName } from '../../telemetry/constants'
-import { Toast } from '../../vscode/toast'
 import {
   MessageFromWebview,
   MessageFromWebviewType,
@@ -30,6 +29,7 @@ import { RegisteredCommands } from '../../commands/external'
 import { ErrorsModel } from '../errors/model'
 
 export class WebviewMessages {
+  private readonly dvcRoot: string
   private readonly paths: PathsModel
   private readonly plots: PlotsModel
   private readonly errors: ErrorsModel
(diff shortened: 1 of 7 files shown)


extension/src/plots/webview/messages.ts
```diff
@@ -79,7 +78,10 @@ export class WebviewMessages {
   public handleMessageFromWebview(message: MessageFromWebview) {
     switch (message.type) {
       case MessageFromWebviewType.ADD_CUSTOM_PLOT:
-        return commands.executeCommand(RegisteredCommands.PLOTS_CUSTOM_ADD)
+        return commands.executeCommand(
+          RegisteredCommands.PLOTS_CUSTOM_ADD,
+          this.dvcRoot
```
mattseddon, line 83: [F] If there were multiple dvc projects in the workspace then clicking on the add custom plot button in the webview would have prompted the user for which project they wanted to run the add command against.

extension/src/plots/webview/messages.ts
```diff
@@ -101,9 +103,15 @@ export class WebviewMessages {
...
       case MessageFromWebviewType.REMOVE_CUSTOM_PLOTS:
-        return commands.executeCommand(RegisteredCommands.PLOTS_CUSTOM_REMOVE)
+        return commands.executeCommand(
+          RegisteredCommands.PLOTS_CUSTOM_REMOVE,
+          this.dvcRoot
+        )
       case MessageFromWebviewType.REFRESH_REVISIONS:
-        return this.refreshData()
+        return commands.executeCommand(
+          RegisteredCommands.PLOTS_REFRESH,
```
mattseddon, line 112: [F] We had code and an extra command in here that was really duplicating logic unnecessarily.

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #3987, opened 2023-05-26
Add DvcConfig class

mattseddon:
# 2/2 `main` <- #3976 <- this

We need to access `dvc config` & `dvc remote` from `Setup`. There are a couple of issues with this 

1. both of these commands are read/write and don't really fit in with the `DvcExecutor`/`DvcReader` pattern
2. these commands can be run when there is no access to the CLI

Previously the commands were located in `DvcExecutor` and wrapped in `Setup` to avoid them breaking the rest of the app. I've now moved them to their own class (`DvcConfig`) and wrapped access to the CLI in that class (all errors will be caught).


sroy3 (approved):

julieg18 (approved):
Great idea!

PR #3993, opened 2023-05-27
Consolidate Setup and "Get Started" walkthrough

julieg18:
# 1/2 `main` <= this <= #3994

* keeps Setup focused on helping user setup DVC and the walkthrough focused on extension features 

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/a1e94c7e-054d-4574-be3f-4096c296794c

~https://github.com/iterative/vscode-dvc/assets/43496356/952dc855-5ae0-4c21-bb39-0a4dfa0662b3~

~https://github.com/iterative/vscode-dvc/assets/43496356/6ea6915f-15f7-4f26-9d2f-5b0ea84e2c8a~

image

image

Part of #3434


webview/src/setup/components/dvc/CliUnavailable.tsx
```diff
@@ -33,7 +44,13 @@ export const CliUnavailable: React.FC<PropsWithChildren> = ({ children }) => {
     <EmptyState isFullScreen={false}>
       <h1>DVC is currently unavailable</h1>
       {children}
-      {contents}
+      <p>
+        New to DVC? Check out <a href="https://dvc.org/">dvc.org</a> to learn
```
julieg18, line 48: Since these DVC steps will now be the first thing that users new to DVC will see, I added some extra sentences/links that were once in the walkthrough.

mattseddon (approved):

shcheklein:
A few comments (might not be related to this PR, sorry if I missed some changes before, but since we have good screenshots here, it's worth mentioning). Not a blocker.

- messages / screens became too verbose
- all types if installation- is it true?
- mentioning docs and demo before action items is misleading (they won't be able to try? Docs - we don't want people reading docs, only when it's needed for some advanced setup. Extension should be enough to do the setup.
- in the initialize project- do we need both first sentences?

julieg18:
> all types if installation- is it true?

We have global (`dvc`), manual (`INPUTED_PATH -m dvc`), and auto python environments. Looking at the docs, these cover all bases as far as I know. 

> in the initialize project- do we need both first sentences?

A bit out of scope, but will be easy enough to tweak these a bit! 

> Docs - we don't want people reading docs, only when it's needed for some advanced setup.

Makes sense! Will update the docs link. 

> mentioning docs and demo before action items is misleading

> messages / screens became too verbose

Will give these some more thought and save for a followup 🤔

shcheklein:
> these cover all bases as far as I know.

Ah, sorry. I read it more as "extension helps you to install it in all possible ways". Maybe we can rephrase the paragraph a bit.

extension/resources/walkthrough/command-palette.md
```diff
@@ -1,3 +1,7 @@
+> ℹ️ The extension's features cannot be accessed until DVC is installed and a
+> DVC project is available in the workspace. Please refer to the
+> [setup page](command:dvc.dvc.showDvcSetup) if you have not setup DVC yet.
+
```
sroy3, line 4: Should this be above the title?

webview/src/setup/components/dvc/CliUnavailable.tsx
```diff
@@ -33,7 +44,13 @@ export const CliUnavailable: React.FC<PropsWithChildren> = ({ children }) => {
...
+        more or our{' '}
```
sroy3, line 49: Because "to learn more" was added after the link, it feels like a verb is missing in front of "our demo".  Maybe rephrase the second part to be " or quickly try the extension with our demo"?

sroy3 (approved):

extension/resources/walkthrough/command-palette.md
julieg18: I added it at the top to better show that this admonition is for the walkthrough as a whole vs just this step. But if it looks buggy to you, I'm happy to move it. 

Comparison of the two:

image

image

webview/src/setup/components/dvc/CliUnavailable.tsx
julieg18: Good idea, I've updated the sentence.

extension/resources/walkthrough/command-palette.md
sroy3: Looks and feels better after the title.

PR #3994, opened 2023-05-27
Show setup on new install

julieg18:
# 2/2 `main` <= #3993 <= this

* prompts setup to be shown on new install instead of our get started walkthrough

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/94d3b9b3-3726-44f5-be63-f0effa398e16

Part of #3434


extension/package.json
```diff
@@ -615,8 +615,8 @@
           "type": "boolean",
           "default": null
         },
-        "dvc.doNotShowWalkthroughAfterInstall": {
```
julieg18, line 618: If any user has this setting set to true, I believe it will get undone with the name change. We could possibly check for this setting and set the new one to true for them, but probably not worth the effort with the amount of users using it.

README.md
```diff
@@ -146,17 +146,17 @@ These are the VS Code [settings] available for the Extension:
...
+| `dvc.dvcPath`                       | Path or shell command to the DVC binary. Required unless Microsoft's [Python extension] is installed and the `dvc` package found in its environment.     |
+| `dvc.pythonPath`                    | Path to the desired Python interpreter to use with DVC. Should only be utilized when using a virtual environment without Microsoft's [Python extension]. |
+| `dvc.experimentsTableHeadMaxHeight` | Maximum height of experiment table head rows.                                                                                                            |
+| `dvc.studio.shareExperimentsLive`   | Automatically share all new experiment metrics and plots logged with DVCLive to Studio. This option will only take effect once Studio is connected.      |
+| `dvc.focusedProjects`               | A subset of paths to the workspace's available DVC projects. Using this option will override project auto-discovery.                                     |
+| `dvc.doNotShowSetupAfterInstall`    | Do not prompt to show the Get Started page after installing. Useful for pre-configured development environments                                          |
```
mattseddon, line 156: [I] No longer refers to the Get Started page

extension/src/setup/util.test.ts
```diff
@@ -24,59 +24,59 @@ beforeEach(() => {
...
     expect(mockedSetConfigValue).toHaveBeenCalledTimes(1)
     expect(mockedSetConfigValue).toHaveBeenCalledWith(
-      ConfigKey.DO_NOT_SHOW_WALKTHROUGH_AFTER_INSTALL,
+      ConfigKey.DO_NOT_SHOW_SETUP_AFTER_INSTALL,
       true
     )
   })
 
-  it('should show the walkthrough if the user responds with show', async () => {
+  it('should show the setup if the user responds with show', async () => {
```
mattseddon, line 59: [N] setup page

mattseddon: all the statements need to get updated

mattseddon (approved):

extension/src/setup/util.test.ts
julieg18: Thanks for spotting that!

README.md
julieg18: Good catch!

mattseddon:
> @mattseddon, I forgot to mention in my review earlier. I couldn't figure out a way to replicate a "fresh" install and couldn't properly test if this was working correctly. I tried installing a `dvc.vsix` on codespaces, installing a `dvc.vsix` locally after deleting DVC, but neither worked to get the environment to read as a fresh install 🤔

Looks like this is a bug and we should not be using this variable as it relates only to VS Code itself. We should be using the `globalState` from the `ExtensionContext`.

```
export function showOnFirstInstall(context: ExtensionContext) {
  const installedKey = 'dvc.installed';

  const installed = context.globalState.get(installedKey, false);

  if (!installed) {
    context.globalState.update(installedKey, true);
    // do the stuff
  }
...
}
```

☝🏻 as a workaround for existing users not having this currently set we should probably check for any values that we set in the `workspaceState`. If they exist then we should set installed to true and not show the setup page.

sroy3 (approved):

PR #3996, opened 2023-05-29
Add storage types details to remote setup page

mattseddon:
Part of #3867

This PR adds a `<VSCodePanels/>` component with a tab for each of the three main storage types and a fourth tab for the remaining types.

The panels will be shown no matter what state the remotes section is in. The information in these pages is lifted with some slight adjustments from the following docs pages:

- https://dvc.org/doc/user-guide/data-management/remote-storage/amazon-s3
- https://dvc.org/doc/user-guide/data-management/remote-storage/google-cloud-storage
- https://dvc.org/doc/user-guide/data-management/remote-storage/azure-blob-storage

I have included details on how to auth, how to enable cloud versioning and a link to the docs for custom authentication.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/cefd9138-8ed0-495c-be48-0fb980cdea4c


webview/src/setup/components/App.test.tsx
```diff
@@ -18,7 +18,6 @@ import { TooltipIconType } from '../../shared/components/sectionContainer/InfoTo
 import { setupReducers } from '../store'
 
 jest.mock('../../shared/api')
-jest.mock('../../shared/components/codeSlider/CodeSlider')
```
mattseddon, line 21: [Q] @sroy3 is there a reason that this is under shared components? Shouldn't it be under the setup webview?

webview/src/setup/components/remotes/StorageSlider.tsx
```diff
@@ -0,0 +1,184 @@
...
+      </div>
+    ),
+    title: 'Other'
+  }
+]
+
+export const StorageSlider: React.FC = () => (
+  <div className={styles.storageSlider}>
+    <VSCodePanels>
+      {storageTypes.map(storageType => (
```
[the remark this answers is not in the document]
mattseddon: This is clearly a rip-off of the `CodeSlider` but I feel like it is different enough to have a separate component. WDYT?

webview/__mocks__/@vscode/webview-ui-toolkit/react.tsx
```diff
@@ -28,3 +28,11 @@ export const VSCodeCheckbox: React.FC<MockCheckboxProps> = ({ onClick }) => {
...
   return <div />
 }
+
+const MockComponentWithChildren: React.FC<PropsWithChildren> = ({
+  children
+}) => <div>{children}</div>
+
+export const VSCodePanels = MockComponentWithChildren
+export const VSCodePanelTab = MockComponentWithChildren
+export const VSCodePanelView = MockComponentWithChildren
```
mattseddon, line 38: [F] Once we mock these components we no longer need `jest.mock('../../shared/components/codeSlider/CodeSlider')` in the setup App tests.

webview/src/setup/components/remotes/AmazonS3.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
```
mattseddon, line 1: [F] I have split these components right out to a very granular level.

[Q] Have I gone too far?

webview/src/setup/components/App.test.tsx
sroy3: I can't remember, but I think it might be because there was no data at all that would make it purely a setup component, and it could be reused anywhere if necessary. Components were still pretty big back then, so I saw the `shared` components as being the dumb components would go. A little like a component library. But it's not really the direction we are moving in at the moment. It's fine to move it.

webview/src/setup/components/remotes/AmazonS3.tsx
sroy3: I would have gone further as `AmazonS3` and `AzureBlobStorage` share a wrapper `<div className={styles.storageDetails}>` and a `<CloudVersioning>` block, so I don't think you went too far. It's fine as is though.

sroy3 (approved):

PR #4006, opened 2023-05-30
Create shared panels component

mattseddon:
# 2/4 `main` <- #4005 <- this <- #4007 <- #4004

This PR creates a shared `<Panels />` component that pulls the duplication out of what was previously the `<CodeSlider />` and `<StorageSlider />` (which have been renamed).

PR #4004, opened 2023-05-30
Add link to show applicable extension for each storage provider

mattseddon:
# 4/4 `main` <- #4005 <- #4006 <- #4007 <- this 

Related to https://github.com/iterative/vscode-dvc/issues/3867

This PR adds some clickable links to tabs in the`<StorageSlider/>` added in #3996. Each link will show the appropriate extension for the storage provider in the extensions view.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/6ae8f44c-cf63-49e9-8f78-5a1d29943e02

PR #4007, opened 2023-05-30
Fix styles of supported storage panels

mattseddon:
# 3/4 `main` <- #4005 <- #4006 <- this <- #4004

This PR fixes the styles applied to the `<SupportedStorage />` component.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/89f611b0-cfcf-408f-a58e-2ce072362ce4

### DIFF
--- a/webview/src/setup/components/remotes/AmazonS3.tsx
+++ b/webview/src/setup/components/remotes/AmazonS3.tsx
@@ -24,7 +24,7 @@ export const AmazonS3 = () => (
       <a href="https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html">
         S3 versioning
       </a>{' '}
-      enabled on the bucket.
+      enabled on the bucket.{'\n'}
     </CloudVersioning>
     <CustomAuth href="https://dvc.org/doc/user-guide/data-management/remote-storage/amazon-s3#custom-authentication" />
   </div>
--- a/webview/src/setup/components/remotes/AmazonS3.tsx
+++ b/webview/src/setup/components/remotes/AmazonS3.tsx
@@ -11,20 +11,18 @@ export const AmazonS3 = () => (
       <a href="https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html">
         AWS CLI config
       </a>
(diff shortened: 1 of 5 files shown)


webview/src/setup/components/remotes/CloudVersioning.tsx
```diff
@@ -7,7 +7,7 @@ export const CloudVersioning: React.FC<PropsWithChildren> = ({ children }) => (
   <>
     <InfoText>
       <CloudVersioningLink /> requires {children}
```
sroy3, line 9: `children` should probably be in a styled `span` that fixes the formatting problem rather than inserting newlines. Or maybe `To enable run:` should be inside its own `p` element.

sroy3: I missed a few `\n` in the previous review as well. We should not rely on them for styling / formatting, but rather use  tags and css to do so.

webview/src/setup/components/remotes/AmazonS3.tsx
```diff
@@ -24,7 +24,7 @@ export const AmazonS3 = () => (
       <a href="https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html">
```
julieg18, line 24: Out of scope of this pr, but noticed a validation error on nested `p` tags:

image

webview/src/setup/components/remotes/AzureBlobStorage.tsx
```diff
@@ -15,7 +15,7 @@ export const AzureBlobStorage = () => (
       <a href="https://learn.microsoft.com/en-us/azure/storage/blobs/versioning-overview">
         Blob versioning
       </a>{' '}
-      enabled on the storage account and container.
+      enabled on the storage account{'\n'}and container.{' '}
```
julieg18, line 18: Maybe it has been already discussed in a previous pr, but why are we adding these new lines in the first place? I'm not seeing the docs use new lines unless they're starting a new paragraph entirely. 

image

Are we doing it to help emphasize certain lines? If so, maybe we should consider rewording text or using methods such as making keywords **bold**/*italic*?

julieg18 (approved):

PR #4024, opened 2023-06-01
Fix "show toast on new install" logic

julieg18:
* uses global state to check if a extension is being used for the first time

## Demo

I tested locally and on codespaces. Note that global extension state stays even after you uninstall the extension. 

https://github.com/iterative/vscode-dvc/assets/43496356/16d3ab77-eb6d-407f-9b43-e89a5c3061a0

Resolves https://github.com/iterative/vscode-dvc/pull/3994#issuecomment-1569567208

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -1,4 +1,4 @@
-import { commands, env, ExtensionContext, ViewColumn } from 'vscode'
+import { commands, ExtensionContext, ViewColumn } from 'vscode'
 import { DvcConfig } from './cli/dvc/config'
 import { DvcExecutor } from './cli/dvc/executor'
 import { DvcRunner } from './cli/dvc/runner'
@@ -254,9 +254,13 @@ class Extension extends Disposable {
       ).contributes.walkthroughs[0].id
     )
 
-    registerPersistenceCommands(context.workspaceState, this.internalCommands)
+    registerPersistenceCommands(
+      context.workspaceState,
+      context.globalState,
+      this.internalCommands
+    )
 
-    void showSetupOnFirstUse(env.isNewAppInstall)
+    void showSetupOnFirstUse(context.globalState, context.workspaceState)
     this.dispose.track(recommendRedHatExtensionOnce())
 
     this.dispose.track(new LanguageClient())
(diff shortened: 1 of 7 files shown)


extension/src/extension.ts
```diff
@@ -1,4 +1,4 @@
-import { commands, env, ExtensionContext, ViewColumn } from 'vscode'
```
julieg18, line 1: As mentioned in https://github.com/iterative/vscode-dvc/pull/3994#issuecomment-1569567208, we shouldn't be using env in the scenario.

extension/src/persistence/util.ts
```diff
@@ -1,9 +1,15 @@
 import { commands, Memento } from 'vscode'
 
-export const resetPersistedState = async (workspaceState: Memento) => {
+export const resetPersistedState = async (
```
julieg18, line 3: Updated our `resetPersistedState` to include the global state since it is a form of persisted state.

extension/src/setup/util.ts
```diff
@@ -25,3 +25,20 @@ export const showSetupOnFirstUse = async (
...
+
+  if (installed) {
+    return
+  }
+
+  void globalState.update(GlobalPersistenceKey.INSTALLED, true)
+
+  // old users won't have the installedKey even if it's not a new install
+  const workspaceKeys = workspaceState.keys()
+  return showSetupToast(workspaceKeys.length === 0)
```
julieg18, line 43: We check for workspaceKeys length since an old user will have these set. I suppose we could take off this extra check in about a month like we've done in the past.

extension/src/persistence/util.ts
mattseddon: [Q] Are we sure that all of the keys under here only relate to our extension?

extension/src/setup/util.ts
mattseddon: [Q] Same question. Are you sure everything in the workspaceState only relates to our extension?

julieg18: I believe that both workspaceState and globalState both relate only to our extension since looking at the docs for `ExtensionContext`:

```
     * An extension context is a collection of utilities private to an
     * extension.
```

Plus I was console logging both states before I touched anything. `globalState` was completely empty and `workspaceState` only had keys set by our extension.

mattseddon (approved):

PR #4036, opened 2023-06-05
Move git call into experiments data

mattseddon:
# 2/2 `main` <- #3980 <- this

This is a follow-up from #3980 ([this comment](https://github.com/iterative/vscode-dvc/pull/3980/files#r1214163024)).

The main motivation behind doing this is to make it easier to follow data being updated in a linear fashion from the watcher, through the model and into the webview.

### DIFF
--- a/extension/src/experiments/webview/messages.ts
+++ b/extension/src/experiments/webview/messages.ts
@@ -37,15 +37,12 @@ export class WebviewMessages {
 
   private hasConfig = false
   private hasValidDvcYaml = true
-  private hasMoreCommits: Record<string, boolean> = {}
-  private isShowingMoreCommits: Record<string, boolean> = {}
 
   private readonly addStage: () => Promise<boolean>
   private readonly selectBranches: (
     branchesSelected: string[]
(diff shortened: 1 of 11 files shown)


extension/src/experiments/webview/messages.ts
```diff
@@ -72,11 +68,9 @@ export class WebviewMessages {
     this.hasStages = hasStages
     this.addStage = addStage
     this.selectBranches = selectBranches
-    this.getNumCommits = getNumCommits
     this.update = update
 
     void this.changeHasConfig()
-    void this.changeHasMoreOrLessCommits()
```
mattseddon, line 79: [F] I found all of this hard to track so move the call out of `WebviewMessages` and into `ExperimentsData`. What this means is that all of the data/transformation logic should be trapped inside of `ExperimentsModel` making the overall flow easier to understand.

julieg18 (approved):

mattseddon:
Looks like #3980 destroyed the stories (and some were broken separately). I'll fix them up.

PR #4037, opened 2023-06-05
Add default value for branch name when creating from experiment

### LINKED ISSUE #4026: create branch: don't require a name
By default, `dvc exp branch` now generates a name like `couth-bean-branch` for an exp named `couth-bean`. Could we auto-fill a default name for the branch that matches that naming scheme?

mattseddon:
Closes #4026.

> By default, dvc exp branch now generates a name like couth-bean-branch for an exp named couth-bean. Could we auto-fill a default name for the branch that matches that naming scheme?

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/5ea3ca35-bddb-474d-b979-aa9d136636d1

### DIFF
--- a/extension/src/experiments/commands/register.ts
+++ b/extension/src/experiments/commands/register.ts
@@ -129,11 +129,7 @@ const registerExperimentInputCommands = (
 ): void => {
   internalCommands.registerExternalCliCommand(
     RegisteredCliCommands.EXPERIMENT_BRANCH,
-    () =>
-      experiments.getCwdExpNameAndInputThenRun(
-        getBranchExperimentCommand(experiments),
-        Title.ENTER_BRANCH_NAME
-      )
+    () => experiments.createExperimentBranch()
   )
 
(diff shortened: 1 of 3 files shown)


extension/src/experiments/commands/register.ts
```diff
@@ -129,11 +129,7 @@ const registerExperimentInputCommands = (
 ): void => {
   internalCommands.registerExternalCliCommand(
     RegisteredCliCommands.EXPERIMENT_BRANCH,
-    () =>
-      experiments.getCwdExpNameAndInputThenRun(
-        getBranchExperimentCommand(experiments),
-        Title.ENTER_BRANCH_NAME
-      )
+    () => experiments.createExperimentBranch()
```
mattseddon, line 132: [F] This function is no longer general so may as well give it a better name and remove the need to pass values.

julieg18 (approved):
Looks good!

PR #4041, opened 2023-06-06
Fix table stories and data types

mattseddon:
Follow up from #3980 / #4036 (Table stories got broken)


extension/src/cli/dvc/contract.ts
```diff
@@ -62,7 +62,6 @@ export const fileHasError = (file: FileDataOrError): file is DvcError =>
 
 export const DEFAULT_EXP_SHOW_OUTPUT = [
   {
-    branch: undefined,
```
mattseddon, line 65: [F] This is only added when sending data to the webview. No need for it to be in the output.

julieg18 (approved):
Whoops, forgot to check storybook when I was reviewing :woman_facepalming:

PR #4044, opened 2023-06-06
Exclude codicons.mjs from eslint

mattseddon:
Suppresses this error:

image


demo
julieg18, line 1: I don't think the demo project update should be here :)

julieg18 (approved):

Issue #1815, opened 2022-06-02
Exp run fails when launched via Experiments tab, but runs via command line


alex000kim:
See https://www.loom.com/share/bbefd727e9044467b56cd1891d9e4452

When I launch an experiment via the Experiments tab, it fails because of an import error. 
Launching the same experiment via the command line works fine.
It looks like there's a difference in environment variables. 
Is it possible that the extension is not using my virtual environment?
I am using `pipenv` to manage the virtual env in this project.

PR #4045, opened 2023-06-06
Add PYTHONPATH to environment variables if the Python extension makes it available

### LINKED ISSUE #1791: Improve Python environment activation
Currently for Python virtual environments we use `ms-python.python`'s API to grab the currently selected Python interpreter and then we run:

`$python_bin -c "import sys; print(sys.executable)"`

This gives us the absolute path of the interpreter. We then call `dvc` with `path/to/interpreter/python -m dvc with some args`. We also prepend the directory that the python interpreter is in to the system $PATH variable before running the command. This works for the basic case of DVC being installed within the virtual environment and is [the suggested approach](https://github.com/microsoft/vscode-python/discussions/18978#discussioncomment-2618085) from an ongoing github discussion. 

Our current approach also works for global installs of DVC as long as the required dependencies are installed within the virtual environment. E.g for the demo project a global install of DVC does not work unless `dvclive` and `ruamel.yaml` are added to the virtual environment. I don't think there is a way around this for the brew installation. I looked into what can be done for a global install via `pip` but I can recreate the same behaviour via the terminal after activating the virtual environment.

I've been looking into how other extensions do this again and have found the following issues/discussions:

- https://github.com/microsoft/vscode-python/issues/11039
- https://github.com/microsoft/vscode-python/discussions/18978
- https://github.com/guildai/guildai-vscode/issues/1
- https://github.com/brettcannon/python-launcher/discussions/168

This seems to be an area that the `vscode-python` team is actively working on. We should keep an eye on those discussion tickets and see where they go. If they could expose the activation command in their API that would be ideal. Maybe we can get the correct PATH details from [EnvironmentVariableCollection](https://github.com/microsoft/vscode/blob/17c4bd39f0fc49886004a2685b8350ab541a1e4d/src/vscode-dts/vscode.d.ts#L10769) in the extension context.

mattseddon:
Closes #1791

This PR uses the `ms-python.python` extension's new API to grab their `PYTHONPATH` variable and add it to the environment variables used to create DVC processes. In order for users to set this up they need to add a file containing `PYTHONPATH=`. This can be a `.env` file at the root of the workspace (default for the extension) or they need to set the `python.envFile` configuration option.

### Demo (fixes issue in [magnetic-tiles-defect](https://github.com/iterative/magnetic-tiles-defect))

https://github.com/iterative/vscode-dvc/assets/37993418/c4166c36-5a41-4d73-bc63-3acefc89965c

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -38,6 +38,7 @@
     "MNIST",
     "nfiles",
     "pseudoterminal",
+    "PYTHONPATH",
     "rwlock",
     "sonarjs",
     "spyable",
--- a/README.md
+++ b/README.md
@@ -160,6 +160,13 @@ These are the VS Code [settings] available for the Extension:
 > **Note** that the `Setup The Workspace` command helps you set up the basic
 > ones at the [Workspace level] (saved to `.vscode/setting.json`).
 
+### Python
+
+This extension is integrated with Microsoft's [Python extension]. When possible,
+the Python extension's selected interpreter will be used to locate DVC. The
+`PYTHONPATH` environment variable identified via the [python.envFile] config
+setting is also respected.
+
 [python extension]:
   https://marketplace.visualstudio.com/items?itemName=ms-python.python
 [studio.token]:
@@ -167,6 +174,8 @@ These are the VS Code [settings] available for the Extension:
 [Studio]: https://studio.iterative.ai
 [workspace level]:
   https://code.visualstudio.com/docs/getstarted/settings#_workspace-settings
+[python.envFile]:
+  https://code.visualstudio.com/docs/python/environments#_use-of-the-pythonpath-variable
 
 ## Debugging
 
--- a/extension/src/cli/dvc/config.test.ts
+++ b/extension/src/cli/dvc/config.test.ts
@@ -42,6 +42,7 @@ describe('DvcConfig', () => {
   const dvcConfig = new DvcConfig(
     {
       getCliPath: () => undefined,
+      getPYTHONPATH: () => undefined,
       getPythonBinPath: () => undefined
     } as unknown as Config,
     {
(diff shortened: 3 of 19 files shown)


julieg18 (approved):
> Add PYTHONPATH to environment variables if the Python extension makes it available

Is PYTHONPATH moved to stable or is it still deprecated?

shcheklein:
> In order for users to set this up they need to add a file containing PYTHONPATH=. This can be a .env file at the root of the workspace (default for the extension) or they need to set the python.envFile configuration option.

I'm still lost on why PYTHONPATH is required. #1791 seems to be broader - it's about other envs, other variables, etc. Unless I'm missing something.

mattseddon:
> > Add PYTHONPATH to environment variables if the Python extension makes it available
> 
> Is PYTHONPATH moved to stable or is it still deprecated?

It has been moved to stable:

- https://github.com/microsoft/vscode-python/blob/main/src/client/apiTypes.ts#L60
- https://github.com/microsoft/vscode-python/wiki/Python-Environment-APIs

mattseddon:
> > In order for users to set this up they need to add a file containing PYTHONPATH=. This can be a .env file at the root of the workspace (default for the extension) or they need to set the python.envFile configuration option.
> 
> I'm still lost on why PYTHONPATH is required. #1791 seems to be broader - it's about other envs, other variables, etc. Unless I'm missing something.

#1815 is the ticket that spawned #1791. Under #1791 two separate users detailed the same issue ([here](https://github.com/iterative/vscode-dvc/issues/1791#issuecomment-1312546018) and [here](https://github.com/iterative/vscode-dvc/issues/1791#issuecomment-1322634094)). They want to use relative imports for the modules within their project. By using this API we will respect whatever they've set for `PYTHONPATH` through either a `.env` file or the `terminal.integrated.env.*` settings (details here: https://code.visualstudio.com/docs/python/environments#_use-of-the-pythonpath-variable). Outside of that, no one has really complained about environment activation which leads me to think that this is a good enough solution to close the ticket.

mattseddon:
We can leave the ticket open but this is still useful IMO.

Issue #3935, opened 2023-05-18
Improve DVC Setup "DVC is Unavailable" step


julieg18:
Currently, vscode-dvc will see if the python extension has a python environment selected, then offer to install DVC and DVCLive inside said environment with `pip`. There are couple things to possibly improve with this:

![image](https://github.com/iterative/vscode-dvc/assets/43496356/b9017f6d-e858-48e7-a902-5d3a229a17c3)

* The python extension will tend to select a global environment by default if there is no virtual env. Installing DVC there is most likely unwanted and even dangerous. 
*  Some users aren't using `pip` to install their packages.
* It's unclear that we are finding the environment using the python extension

Related comments: https://github.com/iterative/vscode-dvc/issues/3434#issuecomment-1544772389, https://github.com/iterative/vscode-dvc/issues/3434#issuecomment-1544775451, https://github.com/iterative/vscode-dvc/pull/3904#discussion_r1197163784

### Task List

- [x] Change our "Install" button to "Install (pip)". (A quick way to clear things up a little more for now)
- [x] Research the possibility of recognizing that the user is not in a virtual environment and/or creating a virtual environment for them (look at Jupyter extension)
- [x] Research how we can install packages other than pip  
- [x] ~Use different install commands based on user's environment~ (no longer needed for now)
- [x] Incorporate the Python Extension's Create Env... into the setup. (We could add an extra quick pick in our Configure command that adds the Create Env option OR show an extra button in the Setup page)
- [x] Rethink step contents, and review what we can adjust to make DVC configuration more clear (Main points to review: Configure button name, information on how we got selected environment, and adding a "Select Python Interpreter" button)

julieg18:
>  Research the possibility of recognizing that the user is not in a virtual environment and/or creating a virtual environment for them (look at Jupyter extension)

Looks like we do both of these things with help from the Python Extension! We can find out if a user is using a virtual environment with the extension's API (specifically their `environments` utils) and we can run the `Python: Create Environment...` command to create a new environment.

julieg18:
>  Research how we can install packages other than pip

Looking at Jupyter, they are running commands, decided on args based off the environment information given by the Python Extension. `conda install PACKAGE_NAME --name NAME` for a named conda environment, `pip install -U PACKAGE_NAME` for a python venv, etc. We can probably do something similar to install DVC and DVCLive instead of just using pip for any environment like we currently are. 

With issue research done, I'll start testing some of these ideas and post some design possibilities next!

[...]

PR #4048, opened 2023-06-06
Have setup run different install commands depending on python environment

julieg18:
* has setup run a different install command depending on the data given to us by the Python Extension API and defaults to `pip` if environment type is unknown

## Demo

Note: Cut the installation progress to make the demo shorter

https://github.com/iterative/vscode-dvc/assets/43496356/e35e9960-9563-432d-948b-053de77faffe

Part of #3935

### DIFF
--- a/extension/src/extensions/python.ts
+++ b/extension/src/extensions/python.ts
@@ -11,12 +11,25 @@ interface Settings {
   }
 }
 
+export interface Environment {
+  id: string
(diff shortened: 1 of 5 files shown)


julieg18:
Going to close this for now since it looks like we can use `pip` to install for most cases. Will reuse some of the logic here for improving the setup though!

PR #4058, opened 2023-06-07
Add Python Extension's "Create Env" command into DVC Setup

julieg18:
* adds a "Update Python Extension" button that gives the user a choice between creating a new environment or selected from already created ones

## Demo

wip

## To Do

- [ ] Add more tests

Part of #3935

### DIFF
--- a/webview/src/setup/components/dvc/CliUnavailable.tsx
+++ b/webview/src/setup/components/dvc/CliUnavailable.tsx
@@ -4,11 +4,15 @@ import styles from './styles.module.scss'
 import { Button } from '../../../shared/components/button/Button'
 import { EmptyState } from '../../../shared/components/emptyState/EmptyState'
 import { SetupState } from '../../store'
-import { installDvc, setupWorkspace } from '../messages'
+import {
+  installDvc,
+  updatePythonEnvironment,
+  setupWorkspace
+} from '../messages'
 
 export const CliUnavailable: React.FC<PropsWithChildren> = ({ children }) => {
-  const pythonBinPath = useSelector(
-    (state: SetupState) => state.dvc.pythonBinPath
+  const { pythonBinPath, isPythonExtensionUsed } = useSelector(
+    (state: SetupState) => state.dvc
   )
   const canInstall = !!pythonBinPath
   const installationSentence = (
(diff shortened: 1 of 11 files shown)


webview/src/setup/components/dvc/CliUnavailable.tsx
```diff
@@ -29,6 +33,9 @@ export const CliUnavailable: React.FC<PropsWithChildren> = ({ children }) => {
       </p>
       <div className={styles.sideBySideButtons}>
         <Button onClick={installDvc} text="Install (pip)" />
+        {isPythonExtensionUsed && (
+          <Button onClick={updatePythonEnvironment} text="Update Env" />
```
julieg18, line 37: This pr's focus is adding the logic and quick pick for updating the python environment or creating a new one. I'm going to do separate prs for:

* adding a `--user` to the command if the environment is global
* improving the step's contents (for example, updating the text for "Configure")

We can merge this as is or I'll add a `do not merge` label until the other two prs are ready :)

extension/src/setup/index.ts
```diff
@@ -499,6 +501,16 @@ export class Setup
...
   }
 
+  private async updatePythonEnvironment() {
+    const value = await pickPythonExtensionAction()
+
+    if (!value) {
+      return
+    }
+
+    return value === 1 ? createPythonEnv() : selectPythonInterpreter()
```
mattseddon, line 511: [I] Let's make the values members of an enum so that they are connected. (i.e if 1 changes in `pickPythonExtensionAction` it will automatically be updated here. It would be good to name them appropriately as well.

mattseddon (approved):

mattseddon:
Fine to merge.

extension/src/setup/index.ts
julieg18: Good idea!

shcheklein:
QQ - does it means that we don't need Configure anymore though?

julieg18:
> QQ - does it means that we don't need Configure anymore though?

Configure (aka "Setup the Workspace" command) would be needed if the user has DVC globally installed, like with `brew`.

shcheklein:
> Setup the Workspace

what does it give on top of Update Env (why update btw?) ? Global vs env choice? Can we combine those somehow into one flow?

PR #4062, opened 2023-06-08
Show CLI errors generated by exp show

mattseddon:
# 1/3 `main` <- this <- #4075 <- #4077

Related to https://github.com/iterative/dvc/issues/9565

This issue was reported on Discord today: https://discord.com/channels/485586884165107732/842220310585147452/1116030033480392836

The user seems to be experiencing a CLI error which means that their experiments appear as if they have no data (technically correct). That means that any attempt to open the experiments webview will result in the setup page being shown. 

Now when the CLI throws an error we will let the user open the experiments webview and show an error state. The error state will be shown any time the webview is open and the CLI throws an error. IMO this is more useful than adding an experiments error state to the setup webview (I did originally add both but changed my mind).

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/3ae5bae3-7a61-427f-883b-d1487b334ec9

Note: A user will still see this if they purposefully go to the setup page when there is a CLI error.

image

This is an edge case though as I believe the most confusing thing was the redirect to setup when there was an error. If we run into this again then I'll add an error state to the Experiments section of the setup page.


julieg18 (approved):
Great work! 

> The test coverage on the diff in this pull request is 80.3% (85% is the threshold).
> This pull request will bring the total coverage in the repository to 95.1% (-0.1% change).

Is this expected?

PR #4068, opened 2023-06-08
fix(plots): correct the empty screen link to the dvc docs

### LINKED ISSUE #4067: Update link when no plots are available
When no plots are available, the extension links to https://dvc.org/doc/studio/user-guide/projects-and-experiments/visualize-and-compare, which seems a bit odd since that page is mostly focused on different views available in Studio. Not sure if that was an intentional choice or an oversight. IMO it would be better to link to https://dvc.org/doc/user-guide/experiment-management/visualizing-plots.

shcheklein:
Minor udpate. Fixes #4067

### DIFF
--- a/webview/src/plots/components/GetStarted.tsx
+++ b/webview/src/plots/components/GetStarted.tsx
@@ -77,7 +77,7 @@ export const Welcome: React.FC = () => (
     />
     <p>
       Learn how to{' '}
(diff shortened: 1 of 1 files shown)


mattseddon (approved):

mattseddon:
I fixed the failing tests in #4070

PR #4075, opened 2023-06-09
Refactor empty state code for experiments and plots

mattseddon:
# 2/3 `main` <- #4062 <- this <- #4077

This PR refactors some of the complexity/duplication out of our empty state components (get started and error state) across the plots and experiments webviews.

PR #4077, opened 2023-06-09
Standardize use of send message across setup, experiments and plots

mattseddon:
# 3/3 `main` <- #4062 <- #4075 <- this

This PR standardises the use of message util files to send messages from the webviews back to the extension.


julieg18 (approved):
Looks a lot better ✨

PR #4084, opened 2023-06-11
Prevent duplicate branches being shown when an experiment is running

mattseddon:


extension/src/experiments/model/index.ts
```diff
@@ -449,24 +450,24 @@ export class ExperimentsModel extends ModelWithPersistence {
     return this.numberOfCommitsToShow
   }
 
-  public setBranchesToShow(branches: string[]) {
-    this.branchesToShow = branches
+  public setBranches(currentBranch: string, allBranches: string[]) {
```
julieg18, line 453: I'm still seeing duplicate branches when I'm running an experiment:

https://github.com/iterative/vscode-dvc/assets/43496356/2aa590d8-b436-423c-9d34-bc3520ad05b7

mattseddon: The top branch is on a different commit, what does the table look like before the experiment and after the experiment finishes?

mattseddon: And what is the one branch that is selected?

julieg18: Oh, right I forgot that a detached HEAD is a different commit instead of a different branch. But I'm also seeing duplicates when I'm checked out to a different branch. Or maybe I'm misunderstanding the problem we're trying to solve 🤔 

I'm in a branch called `other-branch` and I have `main` selected. Before the experiment:

image

After the experiment:

image

mattseddon: The problem that we are trying to solve here is that previously `other-branch` would be shown twice. This change shouldn't affect the other branches that are shown. It should only stop the detached head branch from being duplicated. does that make sense?

mattseddon: Sorry for the description not being clearer.

julieg18: Oh, ok! Thanks for the clarification!

julieg18 (approved):

PR #4091, opened 2023-06-12
Add `--user` flag to global dvc auto installation

julieg18:
* uses the python extension api to get information on selected environment and, if env is global, adds a `--user` flag to the install command 

Part of #3935


extension/src/extensions/python.ts
```diff
@@ -11,15 +11,24 @@ interface Settings {
   }
 }
 
-type EnvironmentVariables = { readonly [key: string]: string | undefined }
+type EnvironmentVariables = { readonly [key: string]: undefined }
```
mattseddon, line 14: [Q] is this change intentional?

extension/src/setup/autoInstall.ts
```diff
@@ -64,7 +81,12 @@ const showInstallProgress = (
     try {
       await Toast.runCommandAndIncrementProgress(
         async () => {
-          await installPackages(root, pythonBinPath, 'dvc')
+          await installPackages(
```
mattseddon, line 84: I think some duplication could be removed here: 

```
    try {
      await Toast.runCommandAndIncrementProgress(
        async () => {
          await installPackages(
            root,
            pythonBinPath,
            ...getProcessGlobalArgs(isGlobalEnv),
            'dvc'
          )
          return 'DVC Installed'
        },
        progress,
        75
      )
      return Toast.delayProgressClosing()
    } catch (error: unknown) {
      return Toast.reportProgressError(error, progress)
    }
```
(especially with the upgrade command being basically the same with a different return value)

extension/src/extensions/python.ts
```diff
@@ -11,15 +11,24 @@ interface Settings {
...
+    type: string
+  }
+}
+
 export interface VscodePython {
   ready: Thenable<void>
   settings: Settings
   environments: {
+    known: Environment[]
+    getActiveEnvironmentPath: () => { id: string }
```
mattseddon, line 31: [I] If we are using this we also need to use `onDidChangeActiveEnvironmentPath` so that updates are taken into account... unless an update will be triggered in a different way.

[Q] Can we replace our other use of the API with this now?

mattseddon (approved):

extension/src/extensions/python.ts
julieg18: > [I] If we are using this we also need to use onDidChangeActiveEnvironmentPath so that updates are taken into account... unless an update will be triggered in a different way.

Updates are triggered with `onDidChangePythonExecutionDetails` since that runs when the environment path changes. 

> [Q] Can we replace our other use of the API with this now?

Apologies, I'm not sure what you mean. What other use of the API could be replaced? Looking at the API, they all seem to be doing different things to me 🤔

extension/src/setup/autoInstall.ts
julieg18: Cleaned up some of the duplication!

extension/src/extensions/python.ts
julieg18: Going to merge this, but happy to update the use of the API in a followup!

mattseddon: We use `getExecutionDetails` to get the Python executable for the active environment. Isn't that returned by `getActiveEnvironmentPath`?

```
export type EnvironmentPath = {
    /**
     * The ID of the environment.
     */
    readonly id: string;
    /**
     * Path to environment folder or path to python executable that uniquely identifies an environment. Environments
     * lacking a python executable are identified by environment folder paths, whereas other envs can be identified
     * using python executable path.
     */
    readonly path: string;
};
```

☝🏻 maybe we can't rely on this as it could return a folder and not an executable. Nevermind.

Issue #4088, opened 2023-06-12
Modify and run experiment not working as expected with dvclive params


dberenbaum:
When I try to modify and run an experiment, it looks like the extension tries to modify all parameters, including those which haven't changed and those which are not dependencies of the pipeline (`results/train/params.yaml` are outputs of dvclive). This is from a fresh clone of https://github.com/iterative/example-get-started-experiments:

https://github.com/iterative/vscode-dvc/assets/2308172/90315710-9cfb-4183-b159-57a100420927

Here's the error output:

```
Running: dvc exp run -S params.yaml:train.img_size=1025 -S params.yaml:train.batch_size=512 -S results/train/params.yaml:model=DynamicUnet -S results/train/params.yaml:batch_size=8 -S results/train/params.yaml:batch_per_epoch=8 -S results/train/params.yaml:frozen=false -S results/train/params.yaml:frozen_idx=0 -S results/train/params.yaml:transforms=[Pipeline: PILBase.create, Pipeline: partial -> PILBase.create] -S params.yaml:base.random_seed=42 -S params.yaml:data_split.test_pct=0.15 -S params.yaml:train.valid_pct=0.1 -S params.yaml:train.arch=shufflenet_v2_x2_0 -S params.yaml:train.fine_tune_args.epochs=8 -S params.yaml:train.fine_tune_args.base_lr=0.01 -S params.yaml:evaluate.n_samples_to_save=10

ERROR: unexpected error - LexerNoViableAltException: transforms=[Pipeline: PILBase.create, Pipeline: partial -> PILBase.create]
                                                                                    ^
See https://hydra.cc/docs/1.2/advanced/override_grammar/basic for details
```

Besides this error, we should not show `results/train/params.yaml` parameters in the dropdown at 0:05 since these are not modifiable dependencies of the pipeline.

mattseddon:
> it looks like the extension tries to modify all parameters,

All of the parameters are passed to `exp run` as that is how we reflect the experiment that was used as a base. This is how we get around having to apply the experiment before kicking off a run. Do you have any suggestions on how to workaround that problem? Applying the experiment first would mean that the data from the base would also be used (related to https://github.com/iterative/vscode-dvc/issues/2394). WDYT?

> Besides this error, we should not show results/train/params.yaml parameters in the dropdown at 0:05 since these are not modifiable dependencies of the pipeline.

How would we exclude these? We are currently using the `exp show` data to grab all of the parameters for an experiment.

[...]

PR #4092, opened 2023-06-13
Correctly handle string arrays in modify and commands

mattseddon:
# 1/3 `main` <- this <- #4094 <- #4095

Part of #4088 

This PR adds support for strings that contain an array being passed around inside the "modify and" commands.

### Demo (can now modify and run in `example-get-started-experiments`)

https://github.com/iterative/vscode-dvc/assets/37993418/38dcbf38-0520-4874-b07d-b7b8cdd49ca1

### DIFF
--- a/extension/src/experiments/model/modify/quickPick.test.ts
+++ b/extension/src/experiments/model/modify/quickPick.test.ts
@@ -17,7 +17,7 @@ describe('pickAndModifyParams', () => {
     mockedQuickPickManyValues.mockResolvedValueOnce(undefined)
 
     const paramsToQueue = await pickAndModifyParams([
-      { path: 'params.yaml:learning_rate', value: 2e-12 }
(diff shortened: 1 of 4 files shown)


extension/src/experiments/model/modify/quickPick.test.ts
```diff
@@ -78,8 +109,57 @@ describe('pickAndModifyParams', () => {
...
       `params.yaml:code_names=${thirdInput}`,
       '-S',
+      `params.yaml:transforms='${fourthInput}'`,
+      '-S',
       [unchanged.path, unchanged.value].join('=')
     ])
-    expect(mockedGetInput).toHaveBeenCalledTimes(3)
+  })
+
+  it('should convert any unselected params into the required format', async () => {
```
mattseddon, line 118: [F] This is actually getting dropped in the next PR as we are disabling modify and on anything that isn't the workspace (see #4088 for more details).

julieg18 (approved):

PR #4094, opened 2023-06-13
Drop commit/experiment support for modify and run/queue (only workspace supported as a base)

mattseddon:
- Correctly handle string arrays in modify and commands
- Drop commit/experiment support for modify and run/queue (only workspace supported as a base)


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -850,16 +846,14 @@ suite('Experiments Test Suite', () => {
...
       const webview = await experiments.showWebview()
       const mockMessageReceived = getMessageReceivedEmitter(webview)
-      const mockExperimentId = 'mock-experiment-id'
       const mockRunExperiment = stub(dvcRunner, 'runExperiment').resolves(
         undefined
       )
 
       const tableChangePromise = experimentsUpdatedEvent(experiments)
 
       mockMessageReceived.fire({
```
julieg18, line 855: Should we be renaming the `it` text for tests that are being updated as well? For example, change `should be able to handle a message to modify an experiment's params reset and run a new experiment` to `should be able to handle a message to modify the workspace's params reset and run a new experiment`

julieg18 (approved):
Great work!

PR #4104, opened 2023-06-14
Remove garbage collect experiments command

mattseddon:
`dvc exp gc` was dropped in `3.0.0`. This PR removes the command from the extension.


extension/src/test/e2e/util.ts
```diff
@@ -104,25 +104,22 @@ export const deleteAllExistingExperiments = async () => {
...
+  await browser
+    .action('key')
+    .down(Key.Shift)
+    .down(Key.Tab)
+    .up(Key.Tab)
+    .up(Key.Shift)
+    .perform()
 
-  await browser.waitUntil(() => deleteAllNonTagExperiments.elem.isDisplayed())
+  await browser.keys('Space')
```
mattseddon, line 121: [F] Selects all experiments if there are any

extension/src/test/e2e/util.ts
```diff
@@ -104,25 +104,22 @@ export const deleteAllExistingExperiments = async () => {
   const workbench = await browser.getWorkbench()
 
   const deleteNonWorkspaceExperiments = await workbench.executeCommand(
-    'DVC: Garbage Collect Experiments'
+    'DVC: Remove Experiment(s)'
   )
   await browser.waitUntil(() =>
     deleteNonWorkspaceExperiments.elem.isDisplayed()
```
mattseddon, line 110: [F] may timeout if there are no experiments

mattseddon: it does... fixed.

shcheklein (approved):

PR #4103, opened 2023-06-14
Update demo project and latest tested CLI version (3.0.0)

mattseddon:
### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/2fde1451-9a2d-4aeb-8555-1f5132799362

PR #4106, opened 2023-06-14
Fix e2e tests

mattseddon:
This PR reinstates the e2e test assertions dropped in #4103. I have also taken this as an opportunity to consolidate some of the util code in the e2e tests.

### DIFF
--- a/extension/src/test/e2e/extension.test.ts
+++ b/extension/src/test/e2e/extension.test.ts
@@ -29,9 +29,7 @@ before('should finish loading the extension', async function () {
 after(async function () {
   this.timeout(60000)
 
-  try {
-    await deleteCustomPlot()
-  } catch {}
+  await deleteCustomPlot()
   await dismissAllNotifications()
 
   return waitForDvcToFinish()
--- a/extension/src/test/e2e/extension.test.ts
+++ b/extension/src/test/e2e/extension.test.ts
@@ -4,7 +4,7 @@ import {
   closeAllEditors,
   createCustomPlot,
   deleteAllExistingExperiments,
-  deleteCustomPlot,
+  deleteCustomPlots,
   dismissAllNotifications,
   expectAllPlotsToBeFilled,
   findDecorationTooltip,
@@ -29,7 +29,7 @@ before('should finish loading the extension', async function () {
 after(async function () {
   this.timeout(60000)
 
-  await deleteCustomPlot()
+  await deleteCustomPlots()
   await dismissAllNotifications()
 
   return waitForDvcToFinish()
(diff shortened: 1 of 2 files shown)


extension/src/test/e2e/extension.test.ts
```diff
@@ -29,9 +29,7 @@ before('should finish loading the extension', async function () {
 after(async function () {
   this.timeout(60000)
 
-  try {
-    await deleteCustomPlot()
```
mattseddon, line 33: [F] If we check for the appropriate notification this will not fail.

extension/src/test/e2e/extension.test.ts
```diff
@@ -230,12 +228,16 @@ describe('Source Control View', function () {
...
       },
       {
         interval: 5000,
         timeout: 60000
       }
     )
+
+    expectedScmItemLabels.sort()
+    dvcTreeItemLabels.sort()
+    expect(expectedScmItemLabels).toStrictEqual(dvcTreeItemLabels)
```
mattseddon, line 241: [F] This reinstates the test that was previously removed to get a release out.

extension/src/test/e2e/util.ts
```diff
@@ -107,23 +107,38 @@ export const deleteAllExistingExperiments = async () => {
...
 
-    await browser
-      .action('key')
-      .down(Key.Shift)
-      .down(Key.Tab)
-      .up(Key.Tab)
-      .up(Key.Shift)
-      .perform()
+  await browser.waitUntil(async () => {
+    const notifications = await workbench.getNotifications()
```
mattseddon, line 113: [I] Pull out pattern

extension/src/test/e2e/util.ts
```diff
@@ -193,11 +231,19 @@ export const expectAllPlotsToBeFilled = async (webview: PlotsWebview) => {
...
   const sourceControlIcon = await activityBar.getViewControl('Source Control')
 
   await sourceControlIcon?.openView()
 
+  const visibleItems = await findCurrentTreeItems()
+
+  await visibleItems[visibleItems.length - 1].click()
+
+  for (let i = 0; i < 20; i++) {
+    await browser.keys('ArrowDown')
```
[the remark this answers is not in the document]
mattseddon: [F] Because we now have more tree items (https://github.com/iterative/dvc/issues/9602) the items towards the bottom of the tree were not shown and because they weren't shown they weren't in the document (at all!). We need to make them visible to be able to assert against them which is why we need this code.

julieg18 (approved):

PR #4116, opened 2023-06-14
Revert string quoting behaviour in modify and commands

### LINKED ISSUE #4114: quoting causing grid search to fail
Probably due to the new quoting behavior, we now don't handle grid search syntax correctly in some cases. See here how `train.arch` takes `resnet18,shufflenet_v2_x2_0` as a literal instead of splitting into 2 comma-separated values: 

https://github.com/iterative/vscode-dvc/assets/2308172/43e5204e-7f5f-4d14-8b20-a9f9c91c5342

The output channel shows `dvc exp run --queue -S params.yaml:train.arch='resnet18,shufflenet_v2_x2_0' -S params.yaml:train.fine_tune_args.base_lr=range(0.001,0.01,0.001)`

mattseddon:
Closes #4114

This PR reverts the recently added behaviour which quoted param values passed to DVC that use string parameters as a base.

The change broke grid search functionality for the modify and run/modify and queue commands. 

Reverting the change provides the user with more freedom to change parameters as they see fit. However, under certain circumstances, they will need to provide quotes for "strings that look like arrays/objects".

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/1f41ba16-ae53-40c6-99c3-2d20d7abe70e

### Edge case behaviour (requires the user to quote)

https://github.com/iterative/vscode-dvc/assets/37993418/9d73622c-117e-4456-937b-b9ec41d14e3b

### DIFF
--- a/extension/src/experiments/model/modify/collect.test.ts
+++ b/extension/src/experiments/model/modify/collect.test.ts
@@ -8,42 +8,34 @@ describe('collectFlatExperimentParams', () => {
     const params = collectFlatExperimentParams(rowsFixture[0].params)
     expect(params).toStrictEqual([
       {
-        isString: false,
         path: appendColumnToPath('params.yaml', 'code_names'),
         value: [0, 1]
       },
       {
-        isString: false,
         path: appendColumnToPath('params.yaml', 'epochs'),
         value: 5
       },
       {
-        isString: false,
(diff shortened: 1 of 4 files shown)


julieg18 (approved):
I wonder if we could mention to users somewhere that certain values need to wrapped in quotes. Not seeing a place though 🤔

PR #4124, opened 2023-06-15
Point users to Python extension in DVC Setup

julieg18:
* add link to python extension if users don't use it and disable buttons instead of removing completely

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/aec5ac6e-d39b-4799-aa52-3468a0fd8e03~

https://github.com/iterative/vscode-dvc/assets/43496356/63d0546c-35cc-408c-bdab-a55b7e4eaac7

Screenshot 2023-06-19 at 11 01 48 AM

Part of #3935

### DIFF
--- a/webview/src/setup/components/App.test.tsx
+++ b/webview/src/setup/components/App.test.tsx
@@ -133,7 +133,7 @@ describe('App', () => {
       })
     })
 
-    it('should tell the user they cannot install DVC without a Python interpreter', () => {
+    it('should tell the user they cannot install DVC without a Python env', () => {
       renderApp({
         cliCompatible: undefined,
         dvcCliDetails: {
@@ -147,7 +147,6 @@ describe('App', () => {
           /DVC & DVCLive cannot be auto-installed as Python was not located./
         )
       ).toBeInTheDocument()
-      expect(screen.queryByText('Install')).not.toBeInTheDocument()
     })
 
     it('should tell the user they can auto-install DVC with a Python interpreter', () => {
@@ -169,7 +168,7 @@ describe('App', () => {
       expect(screen.getByText('Install (pip)')).toBeInTheDocument()
     })
 
-    it('should let the user find another Python interpreter to install DVC when the Python extension is not installed', () => {
+    it('should let the user locate DVC when the Python extension is not installed', () => {
       renderApp({
         cliCompatible: undefined,
         dvcCliDetails: {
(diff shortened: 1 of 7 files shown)


webview/src/setup/components/App.test.tsx
```diff
@@ -187,6 +186,28 @@ describe('App', () => {
       })
     })
 
+    it('should show "Set Env" and "Install (pip)" button tooltips when dvc is unavailable and Python extension is not installed', () => {
```
julieg18, line 189: My original plan was to add tests for checking if buttons and tooltips were disabled depending on circumstances but I for some reason, tooltips and button weren't disabling in the tests. They are working in Storybook and in development but not in testing. Any ideas what's going on?

webview/src/setup/components/dvc/CliUnavailable.tsx
```diff
@@ -58,6 +78,35 @@ export const CliUnavailable: React.FC<PropsWithChildren> = ({ children }) => {
       <h1>DVC is currently unavailable</h1>
       {children}
       {conditionalContents}
+      <div className={styles.sideBySideButtons}>
+        <span className={styles.buttonWrapper}>
```
julieg18, line 82: Ran into Tippy warnings that lead me to adding extra span wrappers on the buttons.

webview/src/setup/components/dvc/CliUnavailable.tsx
```diff
@@ -10,6 +10,30 @@ import {
...
+import Tooltip from '../../../shared/components/tooltip/Tooltip'
+
+const PythonExtensionTooltip: React.FC<{
+  dataTestId: string
+  disabled: boolean
+  children: ReactElement
+}> = ({ dataTestId, disabled, children }) => (
+  <Tooltip
+    content={
+      <span data-testid={dataTestId}>
```
julieg18, line 23: Not sure if we need the extra tooltips since it does tell you to install the python extension in the text 🤔

webview/src/setup/components/App.test.tsx
mattseddon: the magic virtual dom probably doesn't set the properly correctly

webview/src/setup/components/dvc/CliUnavailable.tsx
mattseddon: I don't think we need the tooltip either. It looks very strange with almost the same text popping up next to the original.

webview/src/setup/components/remotes/ShowExtension.tsx
```diff
@@ -2,14 +2,13 @@ import React from 'react'
 import styles from './styles.module.scss'
 import { Icon } from '../../../shared/components/Icon'
 import { Extensions } from '../../../shared/components/icons'
+import { ExtensionLink } from '../shared/ExtensionLink'
```
mattseddon, line 5: [Q] Why not move the text out from the original paragraph, put it under the buttons and use this component? 

I think that would 

1. draw attention to the text
2. remove the need to show a tooltip.

mattseddon (approved):

webview/src/setup/components/remotes/ShowExtension.tsx
julieg18: Good idea! Here's what that looks like so far:

Screenshot 2023-06-19 at 11 01 48 AM

I adjusted the paragraph styling so that it looks more connected to the disabled buttons vs a separate paragraph entirely.

PR #4135, opened 2023-06-17
Adjust plot ribbon tooltips

### LINKED ISSUE #3942: Plots: make it easier to identify exp info
When looking at plots, I find it hard to identify information about the experiments from which it came. I can hover and see some of the metrics like this:

Screenshot 2023-05-19 at 10 32 19 AM

A couple minor suggestions to improve it:

1. Add an icon or something to make it more obvious this exists.
2. Show params rather than metrics on hover. The plots likely already give an idea about the metrics. I'm more likely interested in which parameters are associated with each experiment.

julieg18:
* removes Created column
* adds table description
* adds info icon to ribbon blocks

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/f2c24368-d01f-4064-a18b-1c6c4c41f201~

https://github.com/iterative/vscode-dvc/assets/43496356/cb6334a0-3cfe-46b2-976b-e64f9e783a20

Screenshot 2023-06-20 at 10 56 09 AM

Fixes #3942


extension/src/experiments/model/util.ts
```diff
@@ -1,24 +1,16 @@
 import get from 'lodash.get'
-import { formatDate } from '../../util/date'
 import { formatNumber } from '../../util/number'
 import { truncateFromLeft } from '../../util/string'
 import { splitColumnPath } from '../columns/paths'
 import { Experiment } from '../webview/contract'
 
 type Value = undefined | null | [] | string | number
 
-const isDate = (value: Value): boolean =>
```
julieg18, line 10: We had some utils that I believe were just used to handle dates in table tooltips which can now be deleted.

webview/src/plots/components/ribbon/RibbonBlock.tsx
```diff
@@ -47,6 +47,7 @@ export const RibbonBlock: React.FC<RibbonBlockProps> = ({
       style={{ borderColor: displayColor }}
       data-testid={`ribbon-${id}`}
     >
+      <Info className={styles.infoIcon} />
       <div className={styles.label}>
         {description ? (
           <>
```
julieg18, line 53: Noticed while testing things that copy buttons are broken in the ribbon tooltips. Will fix this!

julieg18: #4147

webview/src/plots/components/ribbon/RibbonBlockTooltip.tsx
```diff
@@ -42,6 +42,9 @@ export const RibbonBlockTooltip: React.FC<{
           ))}
         </tbody>
       </table>
+      <p className={styles.tooltipTableDescription}>
+        *Reflects the first three columns (minus Created) in the table
```
mattseddon, line 46: `*Reflects the first three columns (minus Created) in the table`

- Could swap `*` with another info icon. 
- Change `minus` to `excluding`.
- Need to mention that it is the experiments table.
- Give instructions on how to update. Something like "Drag and drop columns to the start of the table to update.

This could also go after the commit information. 

I'm almost tempted to suggest turning the ribbon into a mini version of the table that always displays the selected columns but I think that is a story in itself.

mattseddon (approved):
Good iteration. Some notes: I wouldn't assume that we won't need to deal with dates now. I'd like to see what the tooltip looks like if we include the first 3 params and the first 3 metrics.

extension/src/experiments/model/util.ts
julieg18: Good point! I'll add the date util back and make sure one of our tests had a date value!

webview/src/plots/components/ribbon/RibbonBlockTooltip.tsx
julieg18: > Need to mention that it is the experiments table.
> Change minus to excluding.

👍 

> Give instructions on how to update. Something like "Drag and drop columns to the start of the table to update.

I considered that but didn't want to make the message too verbose since it's already rather lengthy... I'll post an image and see what others think though. 

> Could swap * with another info icon.

Good idea, I'll try it and see how it looks! 

> This could also go after the commit information.

Why would we want to move the table below the commit information? I figured that the table information is more important than the commit hence why it's first 🤔

julieg18: Here's what it looks like with extra instructions:

Screenshot 2023-06-20 at 11 00 04 AM

It felt way too lengthy to me. Though I suppose if we increase the table rows, that could change 🤔

Issue #3331, opened 2023-02-22
Review onboarding experience and get started


shcheklein:
We've done a lot of things recently greatly simplifying user experience when they first come and try to use DVC through VS Code. Namely, we introduced the onboarding flow where we share DVCLive snippet, we introduced a way to migrate to pipelines (`dvc.yaml` generation). we create a way to connect Studio and VS Code.

We need to spend some time reviewing everything one more time and update messages, documentation, Gif if needed.

- [ ] "Add pipeline button" and its message.
- [ ] `dvc.yaml` that we generate. Do we need to put outputs, etc. More comments.
- [ ] Settings screen
- [ ] Get started documentation
- [ ] Animation in the docs
- [ ] Connect Studio page

dberenbaum:
From https://iterativeai.slack.com/archives/CUSNDR35K/p1686138538724399?thread_ts=1686068465.203069&cid=CUSNDR35K and related to https://github.com/iterative/studio/issues/5933#issuecomment-1580736124, it might be good to explain under experiments that the experiments are saved in Git and the user doesn't need to commit each experiment themselves.

julieg18:
Reviewed our "pipeline" process and reviewed a couple things that could possibly be improved:

### "Create Pipeline" message

<img width="404" alt="image" src="https://github.com/iterative/vscode-dvc/assets/43496356/9f9fb881-c502-4eab-8e9a-d7e2f5a6be9d">

* Is just mentioning experiment reproduction enough info to explain what the command is doing? Should we expand a bit more on what a pipeline is or is it most likely clear enough for users new to DVC? 

### **Run Experiment Button**
When you click the "Run Experiment" button without dvc.yaml, it runs the create a pipeline command:

https://github.com/iterative/vscode-dvc/assets/43496356/24e29acd-d9e9-4111-90ed-7a1dea702f4a

* Similar to the previous comment, is it obvious enough that creating a pipeline is how we run the experiment? Otherwise, it might seem odd that "Run Experiment" opens a quick pick for setting up a pipeline. 

### Generated dvc.yaml

```

# Read about DVC pipeline configuration (https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#stages)
# to customize your stages even more
stages:
  train:
    cmd: python requirements.txt
    deps:
      - requirements.txt
```

* There's an empty line on top of the file. Minor but the file will look a little cleaner without it.

PR #4148, opened 2023-06-19
Remove Setup's "Get Started" Section

julieg18:
* removes the "Get Started" section in the Setup page since its contents can just be shown in DVC's "Setup Complete" 

https://github.com/iterative/vscode-dvc/assets/43496356/3e419f0a-fc5e-4c93-8699-3f53bc3b3146

Related to #3331


mattseddon (approved):

webview/src/setup/components/dvc/Dvc.tsx
```diff
@@ -57,10 +59,16 @@ export const Dvc: React.FC = () => {
...
+
   return (
     <EmptyState isFullScreen={false}>
       <h1>Setup Complete</h1>
       {children}
+      <p>
+        New to the extension? Go through the walkthrough to familiarize yourself
+        with the different features.
+      </p>
+      <Button onClick={showWalkthrough} text="Show Walkthrough" />
```
mattseddon, line 71: [Q] Should this be shown the entire time?

mattseddon: or does it distract from the real purpose of the section?

julieg18: > [Q] Should this be shown the entire time?

Would be nice to only show it for new users but I'm not seeing a way to detect that. We could show the content if the user went through the setup (like we do with autoclosing) but could that possibly lead to unexpected behavior? For example, if a user goes through the setup but comes back after closing the webview, they will no longer see the text/button 🤔 

> or does it distract from the real purpose of the section?

In my opinion, I don't think it's distracting. We expect new users to take a look at the walkthrough after setting up DVC so it makes sense to me to have a "Show Walkthrough" button there. 

But ultimately no strong preference, I'll wait a bit before merging and we can always revert/update in a followup!

Issue #4111, opened 2023-06-14
error: index is locked


dberenbaum:
I think both me and @shcheklein have encountered this issue. The error is from git/dvc but I have only seen it in vs code. Not sure how to reliably reproduce, but maybe we can think about ways to reduce calls that require the git lock.

```
[version: 1.0.3, 2023-06-14T20:58:40.397Z, pid: 75883] > /Users/dave/miniforge3/envs/dvc/bin/dvc exp run --queue -S params.yaml:train.arch='resnet18,shufflenet_v2_x2_0' -S params.yaml:train.fine_tune_args.base_lr=range(0.001,0.01,0.001) - FAILED with code 255 (1389ms)
Queueing with overrides '{'params.yaml': ["train.arch='resnet18,shufflenet_v2_x2_0'", 'train.fine_tune_args.base_lr=0.001']}'.

Queued experiment 'after-rook' for future execution.

Queueing with overrides '{'params.yaml': ["train.arch='resnet18,shufflenet_v2_x2_0'", 'train.fine_tune_args.base_lr=0.002']}'.

Queued experiment 'cruel-fibs' for future execution.

Queueing with overrides '{'params.yaml': ["train.arch='resnet18,shufflenet_v2_x2_0'", 'train.fine_tune_args.base_lr=0.003']}'.

Queued experiment 'silky-lays' for future execution.

ERROR: unexpected error - the index is locked; this might be due to a concurrent or crashed process
```

shcheklein:
@pmrowla do we / can we retry operations on a locked Git index?

pmrowla:
We don't retry any operations on a locked git index. We could consider trying to sleep and then retry after some timeout, but it would be helpful to know where this issue is actually happening. We can't just add a generic retry for every git operation in DVC because some of them depend on state (where retrying whatever operation tried to read the index actually requires rolling back some # of git operations and then rerunning the entire set of operations)

mattseddon:
Do we want to run all of our Git processes with `GIT_OPTIONAL_LOCKS=0`?

mattseddon:
@dberenbaum @shcheklein please let me know if you run into this again.

PR #4150, opened 2023-06-20
Support more languages (git branch)

mattseddon:
Follow up from #4137

In response to [this](https://discord.com/channels/485586884165107732/1120132608962084914/1120391881940009123).

### DIFF
--- a/extension/src/cli/git/reader.test.ts
+++ b/extension/src/cli/git/reader.test.ts
@@ -54,6 +54,7 @@ describe('GitReader', () => {
       expect(mockedCreateProcess).toHaveBeenCalledWith({
         args: ['branch'],
         cwd,
+        env: { LANG: 'en_US.UTF-8' },
         executable: 'git'
       })
     })
@@ -82,6 +83,7 @@ describe('GitReader', () => {
       expect(mockedCreateProcess).toHaveBeenCalledWith({
         args: ['branch'],
         cwd,
+        env: { LANG: 'en_US.UTF-8' },
         executable: 'git'
       })
     })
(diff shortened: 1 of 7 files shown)


extension/src/cli/git/reader.test.ts
```diff
@@ -54,6 +54,7 @@ describe('GitReader', () => {
       expect(mockedCreateProcess).toHaveBeenCalledWith({
         args: ['branch'],
         cwd,
+        env: { LANG: 'en_US.UTF-8' },
```
mattseddon, line 57: [F] Rather than try to replace all possible languages' version of `HEAD detached at` it seems smarter to always ask for the English version and convert that.

extension/src/cli/git/reader.ts
```diff
@@ -151,4 +161,11 @@ export class GitReader extends GitCli {
   private getUris(repositoryRoot: string, relativePaths: string[]) {
     return relativePaths.map(path => resolve(repositoryRoot, path))
   }
+
+  private cleanUpBranchName(branch: string) {
```
mattseddon, line 165: [F] This is not a very general function so I moved it out of `stdout`.

extension/src/cli/git/reader.test.ts
julieg18: Makes sense! Though I wonder how other extensions handle things like this 🤔

julieg18 (approved):
Looks good!

extension/src/cli/git/reader.test.ts
mattseddon: Gitlens:

image

image

image

tl;dr - if the branch name has whitespace mangle the sha into `(${short sha}...)`

mattseddon: We can overhaul when we get to #4111

PR #4155, opened 2023-06-20
Use first three metrics and params in table tooltips & quick picks

julieg18:
# 1/2 `main` <= this <= #4159

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/0340b40a-25d4-4d08-afc3-a9a9dca68f58

Followup to https://github.com/iterative/vscode-dvc/pull/4135#pullrequestreview-1487659893


extension/src/test/suite/experiments/workspace.test.ts
```diff
@@ -561,57 +563,66 @@ suite('Workspace Experiments Test Suite', () => {
...
+            detail:
+              'code_names:[0,1], epochs:5, learning_rate:2.1e-7, loss:2.0488560, accuracy:0.34848332, val_loss:1.9979370',
             label: 'main',
             value: 'main'
           },
           {
             description: '[exp-e7a67]',
-            detail: 'loss:2.0205045, accuracy:0.37241668, val_loss:1.9979371',
+            detail:
+              'code_names:[0,1], epochs:2, learning_rate:2e-12, loss:2.0205045, accuracy:0.37241668, val_loss:1.9979371',
```
julieg18, line 574: Is using six column values too lengthy for the quick pick? 

image

extension/src/experiments/columns/model.test.ts
```diff
@@ -258,14 +258,16 @@ describe('ColumnsModel', () => {
       expect(model.getColumnOrder()).toStrictEqual(persistedState)
     })
 
-    it('should return the first three none hidden columns (minus created) from the persisted state', async () => {
+    it('should return the first three metric and param columns (none hidden) from the persisted state', async () => {
```
mattseddon, line 261: [N] should return the first three visible columns for both metrics and params

extension/src/experiments/columns/model.ts
```diff
@@ -45,12 +45,21 @@ export class ColumnsModel extends PathSelectionModel<Column> {
...
-      )
-      .slice(0, 3)
+    const metrics: string[] = []
+    const params: string[] = []
+
+    for (const path of this.columnOrderState) {
+      if (this.status[path] !== 2) {
+        continue
+      }
+      if (path.startsWith(ColumnType.METRICS)) {
```
mattseddon, line 55: [C] If there are already 3 we could continue (add `||`s to the initial if statement) or could also add the condition to the if statement (could be helpful with lots of params/metrics)

extension/src/experiments/columns/model.ts
```diff
@@ -45,12 +45,21 @@ export class ColumnsModel extends PathSelectionModel<Column> {
...
-      .filter(
-        path =>
-          this.status[path] && this.status[path] === 2 && path !== 'Created'
-      )
-      .slice(0, 3)
+    const metrics: string[] = []
+    const params: string[] = []
+
+    for (const path of this.columnOrderState) {
+      if (this.status[path] !== 2) {
```
mattseddon, line 52: [I] We should always use the enum instead of the raw value (`Status.SELECTED`). The reason being it took me a few seconds to work out what this is doing.

extension/src/test/suite/experiments/workspace.test.ts
mattseddon: No, I think it is fine.

webview/src/plots/components/ribbon/RibbonBlockTooltip.tsx
```diff
@@ -44,8 +44,7 @@ export const RibbonBlockTooltip: React.FC<{
       </table>
       <p className={styles.tooltipTableDescription}>
         <Info className={styles.infoIcon} width={14} height={14} />
-        Reflects the first three columns (excluding Created) in the experiments
-        table.
+        Reflects the first three params and metrics in the experiments table.
```
mattseddon, line 47: [C] Could add a mention of "both" here to make this less ambiguous (same as the it statement updates).

mattseddon (approved):

PR #4163, opened 2023-06-22
Further reduce the number of calls to Git

mattseddon:
Related to #4111 

This PR further reduce the number of calls that we make to Git. It does this by moving the git branch parsing logic out of the `GitReader` class and into the `ExperimentsModel`.

This is some prework for the issue/fix detailed [here](https://iterativeai.slack.com/archives/CNQ95CG1K/p1687046095794719).

Edit: Adding a fix for the issue may not be required.


extension/src/experiments/data/index.ts
```diff
@@ -109,18 +109,12 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
...
-      this.internalCommands.executeCommand<string>(
-        AvailableCommands.GIT_GET_CURRENT_BRANCH,
-        this.dvcRoot
-      ),
-      this.internalCommands.executeCommand<string[]>(
-        AvailableCommands.GIT_GET_BRANCHES,
-        this.dvcRoot
-      )
-    ])
+    const allBranches = await this.internalCommands.executeCommand<string[]>(
```
mattseddon, line 112: [F] I realised that this did not make sense when digging further into the underlying calls/logic.

mattseddon: I should have picked this up earlier.

julieg18 (approved):

PR #4179, opened 2023-06-26
Fix table webview stuck on error page

### LINKED ISSUE #4176: exp show error
Screenshot 2023-06-26 at 10 21 09 AM

How to reproduce:

1. Use https://github.com/iterative/example-get-started-experiments/ (clone and pull)
2. Queue experiments that will fail with `dvc exp run --queue -S "params.yaml:train.arch='resnet18,shufflenet_v2_x2_0'" -S 'params.yaml:train.fine_tune_args.base_lr=range(0.001,0.01,0.001)'`
3. Start the queue and the error should appear

julieg18:
* Keeps the table updating correctly when `cliError` is undefined

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/b31e0de4-f087-4d83-b7e1-7a5dff9cce70

Fixes #4176

### DIFF
--- a/extension/src/experiments/webview/contract.ts
+++ b/extension/src/experiments/webview/contract.ts
@@ -93,7 +93,7 @@ export interface Column extends ColumnAggregateData {
 
 export type TableData = {
   changes: string[]
-  cliError: string | undefined
+  cliError: string | null
   columnOrder: string[]
   columns: Column[]
   columnWidths: Record<string, number>
--- a/webview/src/experiments/state/tableDataSlice.ts
+++ b/webview/src/experiments/state/tableDataSlice.ts
@@ -16,7 +16,7 @@ export interface TableDataState extends TableData {
 export const tableDataInitialState: TableDataState = {
   branches: ['current'],
   changes: [],
-  cliError: undefined,
+  cliError: null,
   columnOrder: [],
   columnWidths: {},
   columns: [],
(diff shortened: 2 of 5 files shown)


extension/src/experiments/webview/contract.ts
```diff
@@ -93,7 +93,7 @@ export interface Column extends ColumnAggregateData {
 
 export type TableData = {
   changes: string[]
-  cliError: string | undefined
```
julieg18, line 96: If a key in the object sent is undefined, the key will not exist in the data received by the webview. I believe this is because VSCode expects a json serializable object. 

```
         * Post a message to the webview content.
         *
         * Messages are only delivered if the webview is live (either visible or in the
         * background with `retainContextWhenHidden`).
         *
         * @param message Body of the message. This must be a string or other json serializable object.
```

webview/src/experiments/state/tableDataSlice.ts
```diff
@@ -49,7 +49,7 @@ export const tableDataSlice = createSlice({
     updateChanges: (state, action: PayloadAction<string[]>) => {
       state.changes = action.payload
     },
-    updateCliError: (state, action: PayloadAction<string | undefined>) => {
```
julieg18, line 52: Another possible solution is how we update our data as a whole. Right now we loop through existing keys. Maybe we check for keys that didn't exist in `data.data` and update missing keys to undefined?

mattseddon: Switching undefined to null is the way I've fixed this previously.

mattseddon (approved):

PR #4181, opened 2023-06-26
Add "Export Raw Data" option to plot popup

### LINKED ISSUE #2713: Export "raw data" from plots
Small feature request. It should be possible to export the raw data from a dvc vscode plot as raw data to .csv or something similar. This would be useful if:

1. Somebody wants to share raw results with somebody who does not use dvc
2. You want to publish the plot in an article or similar and want to tinker with the "visual look" of the plot

image

julieg18:
* adds an "Export Raw Data" for zoomed in custom and template plots. 

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/8556c318-4bb2-415d-8cd6-6118d8be4693

Fixes #2713


webview/src/plots/components/ZoomedInPlot.tsx
```diff
@@ -23,8 +29,31 @@ export const ZoomedInPlot: React.FC<ZoomedInPlotProps> = ({
     }
   }, [])
 
+  const onPlotChange = () => {
```
julieg18, line 32: Tried to write a jest test for this custom action but it was failing for some reason:

```
  it('should add a custom action in zoomed in plot modal for exporting raw data', async () => {
    renderAppWithOptionalData({
      template: complexTemplatePlotsFixture
    })

    expect(screen.queryByTestId('modal')).not.toBeInTheDocument()

    const plot = within(screen.getAllByTestId(/^plot_/)[0]).getByRole('button')

    fireEvent.click(plot)

    const modal = screen.getByTestId('modal')

    expect(screen.getByTestId('modal')).toBeInTheDocument()

    await waitFor(
      () =>
        // eslint-disable-next-line testing-library/no-node-access
        expect(modal.querySelectorAll('.marks')[0]).toBeInTheDocument(),
      { timeout: 5000 }
    )

    const actionsMenu = within(modal).getByTitle('Click to view actions')

    expect(actionsMenu).toBeInTheDocument()

    fireEvent.click(actionsMenu)

    expect(screen.getByText('Save as Raw Data')).toBeInTheDocument() // fails 
  })
```

Debugging showed that the append seemed to be running but the HTML of the vega actions menu only showed two actions instead of three 🤔

julieg18: > 92.4%

Looked over the code coverage, I believe the missing percentage is due the new lines in `ZoomedInPlot`. Any ideas why the test above isn't working or an alternative way to test for the custom action?

webview/src/plots/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const useMutationObserver = (
+  targetEl: MutableRefObject<HTMLElement | null>,
+  onChange: () => void
+) => {
+  const observer = useRef<MutationObserver | null>(null)
+
+  useEffect(() => {
+    if (targetEl.current && !observer.current) {
+      observer.current = new MutationObserver(onChange)
```
julieg18, line 13: The only way I could think of to update the vega actions is with the help fo the `MutationObserver` API. I wonder if there is a "less hacky" way of doing this though 🤔

webview/src/plots/components/ZoomedInPlot.tsx
```diff
@@ -23,8 +29,31 @@ export const ZoomedInPlot: React.FC<ZoomedInPlotProps> = ({
...
+    const createdRawDataAction = actions?.querySelector(
+      `.${styles.vegaCustomAction as string}`
+    )
+    if (createdRawDataAction) {
+      return
+    }
+    const rawDataAction = document.createElement('a')
+    rawDataAction.textContent = 'Save as Raw Data'
+    rawDataAction.addEventListener('click', () => {
+      exportPlotAsRawData(id, props.data)
```
julieg18, line 43: From my understanding of how our plots work, custom plot data will always be in `props.data`, but I wasn't sure if template plots will always hold their data in`props.spec.data` (where our plots have it). I got the data for template plots from the "back-end" for now.

webview/src/plots/hooks/useMutationObserver.ts
mattseddon: IIUC the below screenshot shows that the vega-actions are already in the DOM when the modal is opened and the button has not been click on yet:

image

Doesn't that mean that you can append the new action in the original `useEffect` in `ZoomedInPlot` no need to use the `MutationObserver`?

mattseddon (approved):
Few things to work through before merging this one.

webview/src/plots/hooks/useMutationObserver.ts
julieg18: The `vega-actions` do render before the button has been clicked but they don't render at the time of mounting, they render a little later after that:

```
  useEffect(() => {
    const modalOpenClass = 'modal-open'
    document.body.classList.add(modalOpenClass)

    debugger

    return () => {
      document.body.classList.remove(modalOpenClass)
    }
  }, [])
```

image

I use the `MutationObserver` to detect when the `vega-actions` are actually rendered the DOM.

webview/src/plots/components/ZoomedInPlot.tsx
julieg18: > try adding a waitFor around expect(screen.getByText('Save as Raw Data')).toBeInTheDocument() to see if there is a timing issue there

Tried that as well with an extended timeout but I just hit a timeout error :(

julieg18: Figured out the issue! Look me way too long to realize it 🤦‍♀️ but the code wasn't actually running correctly due to us mocking the MutationObserver in `setup-tests.js`. I think we're safe to delete the mocked function since it was added in #1365 for the use of the api that no longer exists(https://github.com/iterative/vscode-dvc/blob/f026ac15f0d8d6cae7263132e66abee6b9246d0c/webview/src/shared/components/theme/Theme.tsx#L3).  

Will delete the mock for now!

extension/src/plots/webview/messages.ts
```diff
@@ -337,4 +343,25 @@ export class WebviewMessages {
   private getCustomPlots() {
     return this.plots.getCustomPlots() || null
   }
+
+  private async exportPlotAsRawData(plotId: string, data?: PlainObject) {
+    const file = await window.showSaveDialog({
```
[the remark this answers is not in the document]
julieg18: > [Q] Can we wrap this and keep window out of this file?

Done!

extension/src/plots/model/index.ts
```diff
@@ -223,6 +226,19 @@ export class PlotsModel extends ModelWithPersistence {
...
+  public saveAsPlotRawData(
+    selectedRevisions: Revision[],
+    plotId: string,
+    filePath: string,
+    data?: PlainObject
+  ) {
+    const rawData =
+      data || this.getSelectedTemplateRawData(selectedRevisions, plotId)
+
+    writeJson(filePath, rawData as Record<string, unknown>, true)
```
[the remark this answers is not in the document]
julieg18: Added an check!

julieg18:
Resolved/responded to all comments! Going to re-request reviews since I altered quite a bit of code :)

mattseddon (approved):

PR #4195, opened 2023-06-29
Fix unresolved promises in local e2e tests

julieg18:
Noticed tests were failing on my end and spotted a bug with our "remove" util


extension/src/test/e2e/util.ts
```diff
@@ -40,9 +40,13 @@ const notificationShown = async (
   message: string
 ): Promise<boolean> => {
   const notifications = await workbench.getNotifications()
-  return notifications.some(
```
julieg18, line 43: Promises don't resolve correct with array functions and an unresolved promise is an promise which is truthy, causing this function to give a false positive.

mattseddon: Any docs on this?

mattseddon (approved):

extension/src/test/e2e/util.ts
julieg18: > Async functions always return a promise. If the return value of an async function is not explicitly a promise, it will be implicitly wrapped in a promise.

[Via mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description)

If I'm understanding things correctly, `.some` returns `true` if at least one callback returns `true`. Since the `callback` function we provided was an `async` function, it was always returning a Promise which is truthy, giving the false positive.

julieg18: Apologies, my original comment wasn't very clear. Wrote it while my blood sugar was dropping 😅

PR #4197, opened 2023-06-29
Update latest tested version to `3.2.3`

julieg18:


demo
julieg18, line 1: The remote `e2e` tests did fail remotely on the first attempt with a timeout error for `Experiments Table Webview should update with new data for each DVCLive step when an experiment is running`. Probably just a one-time thing but just in case I missed something can someone else check if the e2e tests are running correctly on their end?

mattseddon (approved):

Issue #3223, opened 2023-02-05
Improvements to smooth control for plots


shcheklein:
After smooth template became default we need to find another way to control smoothness:

- We should be able to save the value in between sessions. Atm, if you open plots -> change the value -> close plots -> open them again you are loosing your changes to the value.
- On some high contrast light / dark themes it blends with the background behind the plots. Should be the color of the plot canvas instead.
- Can we align the control to the corner and make the background transparent?
- Should be a button per plot that open that control and/or control per section that can change all at once?

<img width="549" alt="Screenshot 2023-02-04 at 3 50 07 PM" src="https://user-images.githubusercontent.com/3659196/216795828-b978abd6-433f-44ad-bded-fe8225df17cc.png">

mattseddon:
Can we close this after #3250?

shcheklein:
@mattseddon I've updated the description and repurposed the ticket a bit.

mattseddon:
Related to #3837

domoritz:
All state in Vega charts is stored in the data or the signals which you can read/write using the [Vega View API](https://vega.github.io/vega/docs/api/view/#signals). You should be able to use this functionality to store the state across chart changes. You can also use the same API to coordinate changes across multiple charts to address the last point. It may even be possible to bind the param for smoothing from multiple Vega views to the same input DOM element (https://vega.github.io/vega/docs/signals/#bind) but I haven't tried that.

julieg18:
> Can we align the control to the corner and make the background transparent?

Tried this, and while it looked ok on larger plots, it looks buggy on smaller plots: 

<img width="1272" alt="image" src="https://github.com/iterative/vscode-dvc/assets/43496356/c75aeea1-d1fc-4a77-b026-6c60b583aa63">

<img width="1270" alt="image" src="https://github.com/iterative/vscode-dvc/assets/43496356/e012295e-b5b8-4d7f-a9ce-6bbf8ee6e005">

julieg18:
As discussed synchronously, going to close this as most tasks have been completed and we aren't planning to add a across-plot smooth control feature at the moment.

PR #4202, opened 2023-06-30
Update plot smooth control background color

julieg18:
Tries out comment in #3223:

> On some high contrast light / dark themes it blends with the background behind the plots. Should be the color of the plot canvas instead.

We can go ahead and close this if we prefer main :) 

## `main`

image

image

## PR

image

image


mattseddon:
Did you check the zoomed in plot?

julieg18:
> Did you check the zoomed in plot?

Yes!

image

mattseddon (approved):

PR #4218, opened 2023-07-05
Use `onNewView` instead of `MutationObserver` in zoomed in plot

julieg18:
Followup to https://github.com/iterative/vscode-dvc/pull/4181#discussion_r1248284802


webview/src/plots/components/ZoomedInPlot.tsx
```diff
@@ -66,6 +57,7 @@ export const ZoomedInPlot: React.FC<ZoomedInPlotProps> = ({
           export: true,
           source: false
         }}
+        onNewView={onNewView}
```
julieg18, line 60: Really surprised that I forgot about this prop when working on this earlier 😅. Looks to be working without issues!

mattseddon (approved):

mattseddon:
Thanks for following up.

PR #4220, opened 2023-07-05
Save smooth plot values across sessions

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/78e401d5-b467-46a0-8a55-6d06a4c169ab

Part of #3223


webview/src/plots/hooks/useSetupSmoothPlot.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    if (defaultValue) {
+      view.setState({
+        ...state,
+        signals: { ...state.signals, smooth: defaultValue }
+      })
+    }
+
+    if (isZoomedIn) {
```
julieg18, line 32: I couldn't think of a way to update the zoomable plots' values when the value changes inside of a plot model. 

We could save the `view` into a reference so we could call `setState` later but I'm not sure how we could detect when the zoomable plots need an update. I disabled running `setSmoothPlotValue` on model slider change for now but any ideas?

webview/src/plots/components/App.test.tsx
```diff
@@ -2122,5 +2122,91 @@ describe('App', () => {
       fireEvent(panel, clickEvent)
       expect(clickEvent.stopPropagation).toHaveBeenCalledTimes(1)
     })
+
+    describe('Smooth Plots', () => {
+      it('should send a message to save the value when a vega panel slider is interacted with', async () => {
```
julieg18, line 2127: This test is failing on windows testing with:

```
FAIL src/plots/components/App.test.tsx (54.55 s)
    ● App › Vega panels › Smooth Plots › should send a message to save the value when a vega panel slider is interacted with
  
      expect(jest.fn()).toHaveBeenCalledWith(...expected)
   ERROR  run failed: command  exited (1)
  
      - Expected
      + Received
  
        Object {
      -   "payload": Object {
      -     "template\\smooth.tsv": 0.4,
      -   },
      -   "type": "update-smooth-plot-values",
      +   "type": "initialized",
        },
  
      Number of calls: 1
  
        2[139](https://github.com/iterative/vscode-dvc/actions/runs/5510498105/jobs/10044789728?pr=4220#step:7:142) |         fireEvent.change(slider as HTMLInputElement, { target: { value: 0.4 } })
        2[140](https://github.com/iterative/vscode-dvc/actions/runs/5510498105/jobs/10044789728?pr=4220#step:7:143) |
      > 2[141](https://github.com/iterative/vscode-dvc/actions/runs/5510498105/jobs/10044789728?pr=4220#step:7:144) |         expect(mockPostMessage).toHaveBeenCalledWith({
             |                                 ^
        2[142](https://github.com/iterative/vscode-dvc/actions/runs/5510498105/jobs/10044789728?pr=4220#step:7:145) |           payload: { [smoothId]: 0.4 },
        2[143](https://github.com/iterative/vscode-dvc/actions/runs/5510498105/jobs/10044789728?pr=4220#step:7:146) |           type: MessageFromWebviewType.SET_SMOOTH_PLOT_VALUES
        2[144](https://github.com/iterative/vscode-dvc/actions/runs/5510498105/jobs/10044789728?pr=4220#step:7:147) |         })
  
        at Object.toHaveBeenCalledWith (src/plots/components/App.test.tsx:2141:33)
```

I checked the feature on my windows machine and it was working correctly but the tests were still failing. I've tried:

* using other `fireEvent` methods like `focus` and `keydown`
* wrapping the `expect` in a `waitFor`
* using the `userEvent` module

But nothing seems to be working. Any ideas on what's the issue?

mattseddon: I'll take a look

mattseddon: As the failure is on Windows I would suspect that there is an unexpected conversion somewhere with the `smoothId` and that the payload doesn't match. I'll run a couple of tests

mattseddon: https://github.com/iterative/vscode-dvc/pull/4253/commits/540fb7af0543a8c79d308ef83d9aeeceba0ddee4 worked. One last thing to try.

mattseddon: https://github.com/iterative/vscode-dvc/actions/runs/5514227454/jobs/10053229194?pr=4253

changing the setup to 
```
        renderAppWithOptionalData({
          template: {
            ...withVegaPanels,
            smoothPlotValues: { [smoothId]: 0.00001 }
          }
        })
```

gives the following result: 

image

☝🏻 based on the above run it looks like the `fireEvent.change` is not picked up correctly.

mattseddon: if you can confirm the behaviour on Windows I am ok with removing the test.

webview/src/plots/components/ZoomablePlot.tsx
```diff
@@ -30,6 +31,7 @@ export const ZoomablePlot: React.FC<ZoomablePlotProps> = ({
   const { data, content: spec } = useGetPlot(section, id, createdSpec)
   const dispatch = useDispatch()
   const currentPlotProps = useRef<VegaLiteProps>()
+  const onPlotViewReady = useSetupSmoothPlot(id)
```
mattseddon, line 34: [C] I think this could have a more descriptive name.

mattseddon: [Q] Also, would it be a better idea to pass this in for smooth plots only instead of calling this for all zoomable plots?  Unless I have missed something custom plots should never need to call `useSetupSmoothPlot` and having the hook floating around for all plots makes things harder to follow.

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
```diff
@@ -100,6 +100,7 @@ export const TemplatePlotsGrid: React.FC<TemplatePlotsGridProps> = ({
           <div
             key={plot}
             id={plot}
+            data-id={plot}
```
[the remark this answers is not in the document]
mattseddon: Removing this gives a reliable failure on the flaky test (which can be fixed by updating the query in the listener).

webview/src/plots/hooks/useSetupSmoothPlot.ts
```diff
@@ -0,0 +1,50 @@
...
+      return
+    }
+
+    const smoothRange = document.querySelector(
+      `[data-id="${id}"] input[name="smooth"]`
+    )
+    smoothRange?.addEventListener('change', (event: Event) => {
+      if (event.target) {
+        dispatch(
```
mattseddon, line 41: [Q] Why not send the message here as well? That would mean we wouldn't have to `useEffect` in the other component which makes things harder to track.

mattseddon: also don't have to send the entire object, could just send a partial update. I have a feeling this may fix the test as well.

webview/src/plots/hooks/useSetupSmoothPlot.ts
mattseddon: what is the difference between updating in the modal and non-modal states?

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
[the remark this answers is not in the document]
mattseddon: I wouldn't add this extra key. We don't have `data-id` anywhere else in the app and I don't really understand the reason to add it now.

extension/src/plots/webview/messages.ts
```diff
@@ -285,7 +297,8 @@ export class WebviewMessages {
       nbItemsPerRow: this.plots.getNbItemsPerRowOrWidth(
         PlotsSection.TEMPLATE_PLOTS
       ),
-      plots
+      plots,
+      smoothPlotValues: this.plots.getSmoothPlotValues()
```
mattseddon, line 301: [Q] What was the reason for keeping this separate from the plots data?

mattseddon (approved):
Few things to work through

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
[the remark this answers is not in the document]
julieg18: > I wouldn't add this extra key. We don't have data-id anywhere else in the app and I don't really understand the reason to add it now.

We're using it to target the plot's smooth range. Though If we want to avoid adding this, we should be able to use a HTML ref instead 🤔

extension/src/plots/webview/messages.ts
julieg18: Didn't think of it :woman_facepalming:. That would probably make updating the plots' vega state easier too.

webview/src/plots/hooks/useSetupSmoothPlot.ts
julieg18: > what is the difference between updating in the modal and non-modal states?

We're updating the state `onViewReady` aka when the plot first renders. Users' always see the modal on first render so it's easy to update the state there unlike the non-modals which stay on the page.

Though we might be able to detect when the state needs to be updated even after first render by tracking the `smoothPlotValue` and holding the `view` in a reference so we can check for differences and update if they're different 🤔

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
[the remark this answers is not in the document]
julieg18: Using a HTML ref now!

webview/src/plots/components/App.test.tsx
julieg18: >  I think the failure is to do with the way that we are relying on dispatch + useEffect to send the message.

Yep, fixed with removal of that combination!

extension/src/plots/webview/messages.ts
julieg18: Ended up keeping it separate for now so we could watch for changes to values by just watching the values instead of `snapshots` which would cause unnecessary checks.

webview/src/plots/hooks/useSetupSmoothPlot.ts
julieg18: > Though we might be able to detect when the state needs to be updated even after first render by tracking the smoothPlotValue and holding the view in a reference so we can check for differences and update if they're different 🤔

Modals state is now corrected!

julieg18:
Alright resolved/responded to comments and added a debounce timer for keyboard events (thanks for spotting that, @mattseddon!). Going to request reviews since the code has changed quite a bit!

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
```diff
@@ -48,8 +48,9 @@ export const TemplatePlotsGrid: React.FC<TemplatePlotsGridProps> = ({
 
   const addDisabled = useCallback(
     (e: Event) => {
-      const disabledId = (e.currentTarget as HTMLFormElement).parentElement
-        ?.parentElement?.parentElement?.id
+      const disabledId = (e.currentTarget as HTMLFormElement).closest(
```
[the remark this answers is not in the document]
mattseddon: You would need to go back and check around https://github.com/iterative/vscode-dvc/pull/2747

mattseddon (approved):

webview/src/plots/components/templatePlots/TemplatePlotsGrid.tsx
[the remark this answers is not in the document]
julieg18: Thanks for the info! Not seeing any comments on this particular line in the prs relating to this code, but my guess on why we were using `parentElement` is because it's technically faster since `closest` keeps checking up to the `root`. I don't think this code gets fired too often (only when user's mouse enters `.vega-actions`) so should be alright to use 🤔

PR #4224, opened 2023-07-06
Update DVCLive examples in Setup

mattseddon:
Related to https://github.com/iterative/studio/issues/6422

This PR updates the DVCLive example snippets for PyTorch Lightning, Keras, HuggingFace and the general Python API. It also moves the general Python API to first in the tab list. I have also added an "Other" tab which provides links to all of other currently supported frameworks.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/465c24f1-5432-4e32-ac74-70c925db8c25

Some questions for @daavoo 

1. Are PyTorch Lightning, Keras & HuggingFace still the correct frameworks to show from the complete list?
2. Are the copied snippets now correct or do they need to be updated again?
3. Is the General Python API snippet the correct one?

Thanks


daavoo (approved):
Everything looks correct.

Those frameworks are still the ones that get more views in docs

yathomasi (approved):
Nice work @mattseddon. I will propagate the same to the Studio as well.

I think we would also want the same changes in https://dvc.org/doc/start/experiments/experiment-tracking#running as well. cc: @daavoo @dberenbaum.

webview/src/setup/components/experiments/styles.module.scss
```diff
@@ -8,3 +8,15 @@
     margin: 0 auto;
   }
 }
+
```
julieg18, line 11: Minor, but with this change, a couple tabs are wider/shorter than the others causing the content to "jump" in width:

https://github.com/iterative/vscode-dvc/assets/43496356/e7ec0234-f76e-484a-957a-02d2024ea9fd

This can be fixed if we give each tab's content a width of 600px:

https://github.com/iterative/vscode-dvc/assets/43496356/e720b080-901d-42d3-ba0c-a550d9899645

julieg18 (approved):
Great work!

shcheklein (approved):

webview/src/setup/components/experiments/styles.module.scss
mattseddon: We should do the same for all of the panels. I'll follow up.

Issue #4222, opened 2023-07-06
Add targets to "run experiment"


BradyJ27:
Currently, hitting "run experiment" will prompt for a stage in ./dvc.yaml. A user should be able to specify a path and or a stage for the target of "run experiment". This will allow to run an entire dvc.yaml through the UI, along with allowing to run a dvc.yaml that exists in a subdirectory of the current working directory. 

An example of using this with the CLI would be `dvc exp run sub-dir/dvc.yaml` or, `dvc exp run sub-dir/dvc.yaml:stage`

mattseddon:
@BradyJ27 do you have an example project that you can share?

BradyJ27:
> @BradyJ27 do you have an example project that you can share?

@mattseddon I currently do not. I can see if I can throw together a quick example project that follows a "mono-repo" style.

mattseddon:
For the record the extension supports multiple projects being available in the workspace in a mono-repo (i.e each project is initialized with `--subdir`). As far as the extension is concerned a project is denoted by an individual `.dvc` folder. Under these circumstances when starting an experiment the extension will ask the user which project they want to run the experiment against. This can be bypassed by setting the `dvc.focusedProjects` option to a single project or opening the individual project at the workspace's root. 

What we don't currently support is multiple pipelines underneath a single project (i.e. a single `.dvc` directory and nested `dvc.yaml`s without a `dvc.yaml` at the project root). We will be working through this setup shortly though and we should be able to come up with a solution that gives extra flexibility without breaking the experience for users with less complicated setups.

shcheklein:
@mattseddon I've created a repo to try the scenario - https://github.com/shcheklein/monorepo-slices-example . It has two directories with the same project inside. It might be confusing a bit test bc of this, but I hope that still should be enough to test. Let me know.

mattseddon:
@BradyJ27 the new functionality will be in the next release. LMK how you go and if you have any feedback 🙏🏻.

Issue #4194, opened 2023-06-29
Improve `dvc.yaml` generation


julieg18:
### Current
**UI:**

<img width="404" alt="image" src="https://github.com/iterative/vscode-dvc/assets/43496356/9f9fb881-c502-4eab-8e9a-d7e2f5a6be9d">

**Generated `dvc.yaml` or update to already created file:**

```

# Read about DVC pipeline configuration (https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#stages)
# to customize your stages even more
stages:
  train:
    cmd: python requirements.txt
    deps:
      - requirements.txt
```

### Improvements

- [x] Try alternative phrasing in place of "pipeline" for more clarity. Referring to it as`dvc.yaml` for example. 
- [x] Expand UI text a bit more, giving a more full explanation on what pressing the button will do for the user
- [x] Add more comments to `dvc.yaml`, giving a more in-depth explanation of `dvc.yaml` with some configuration examples 

Related #3331

SoyGema:
Hey @julieg18 ! 👋
What an impactful feature you are working on! 
If you allow me, would like to give you some food for thought regarding _pipeline blocks that users face regarding design_ that might have an **impact on how this feature is designed or guided in a multiple stage scenario.** 

This might influence  two of the three points of Improvement .
1. [Random execution of pipeline stages](https://iterative.ai/blog/august-22-community-gems#im-constructing-a-pipeline-with-several-stages-inside-the-dvcyaml-file). This  Q that has arised on discord and appears in some past CG and is one of  your common questions.  In order to ensure linear order in the pipeline, _the user should concatenate all pipeline stages, taking into account that the previous stage output will be the next dependency_, from the beginning to the end of your pipeline. 
Please make sure that you specify dependencies and outputs for each stage: that will introduce the order to provide an end result
<img width="711" alt="Captura de pantalla 2023-06-29 a las 20 23 08" src="https://github.com/iterative/vscode-dvc/assets/24204714/72fd7189-96e1-4b45-b17d-844c0e28b90a">


Would you think about how this might be in UI terms or construct this having this **common roadblock** in mind? 


Thinking about this, there might be 3 cases
1. First stage. You don´t have dependency but you have an **output** ( ej : load data )
2. In-between-stages (Image above) You have an **output**(to next stages )  and a **dependency**( from previous stages) (ej: all feature engineering process , normalization, feautre selection, filtering, etc)
[...]

PR #4225, opened 2023-07-06
Update copy around adding a new stage/defining a pipeline for a project without one

mattseddon:
Part of #4194

Took the basic step of updating the copy. Further changes will be required for #4222 as well and I think people will see this messaging even less.

### Screenshot

image


shcheklein:
I think we still need to iterate on the copy. Some ideas:

1. Make button "Run experiment" (the same as all other Run button behave), and to avoid mental load with pipelines and stages
2. Change text to something more explicit. "To be able to run, queue, pass parameters, track data, etc- create a project file". Project file can be a link to `dvc.yaml`.

@dberenbaum @daavoo any other ideas on the copy here?

mattseddon:
`Run experiment` is far less explicit than `Add Stage` as (unless I am mistaken) running experiments is not possible without having a stage.

I don't think we should be changing/inventing terms to try and reduce mental load. I may have missed one but I cannot find a single reference to "project file" relating to a `dvc.yaml` in the docs. If we are worried about introducing the pipeline term here then we should de-couple/remove this step from the table.

shcheklein:
> Run experiment is far less explicit than Add Stage as (unless I am mistaken) running experiments is not possible without having a stage.

In this case the whole intention is to make experiments run, modify and run, etc to work. There is not intention at this stage to explain the benefits of the pipelines as is (cache intermediate artifacts), etc. Even the name pipeline, stage - etc -why should people bother about them at this page?

> I cannot find a single reference to "project file" relating to a dvc.yaml in the docs.

that's fine, we can make it "dvc.yaml" or Project `dvc.yaml` file, or configuration file ... something that comes from the basic user experience, not something new for them to lear

> If we are worried about introducing the pipeline term here then we should de-couple/remove this step from the table.

The whole intention of this step was not about introducing the pipelines, but about enabling all the experiments features in the extension. Please try to look at this from that angle. WDYT.

Again, curios about @dberenbaum and @daavoo 's take on this (since I see some benefits of using the pipelines term as well, just want to calibrate if I'm too biased into experiments).

dberenbaum:
I'd like to find time to work on how we talk about pipelines and running experiments in the docs and then come back to this. Right now, I don't have a strong sense now of what's better and can see the pros and cons of both approaches. I'm also not sure yet that wording it one way or the other is going to make an impactful difference.

shcheklein:
Let's try to make about experiments then please until we have more sense about the pipelines. To not confuse people with new terminology.

Issue #4246, opened 2023-07-10
Improve zoomed in plots export actions


julieg18:
<img width="335" alt="image" src="https://github.com/iterative/vscode-dvc/assets/43496356/5db06744-1829-410c-9706-7047b66df0e8">

- [x] "Save as Raw Data" exports data to a json file. Let's add csv/tsv options
- [x] Remove `dvc_data_version_info` from raw data export
- [x] It's unclear that these export actions exist. Can we make them more obvious? 
     - [x] ~Look into showing export actions across all plots (disable while an experiment is running?)~
     - [x] ~Possibly create our own actions menu~
- [x] ~Try to fix broken (currently disabled) vega actions ("View Vega", "Open In Vega Editor", etc.)~ (https://github.com/iterative/vscode-dvc/issues/4246#issuecomment-1634997776)

julieg18:
>  Try to fix broken (currently disabled) vega actions ("View Vega", "Open In Vega Editor", etc.)

Looks like these actions are not possible to run since they require certain permissions that aren't set in VSCode webviews. You will just get an `request was made in a sandboxed frame whose 'allow-popups' permission is not set` error being thrown.

julieg18:
> It's unclear that these export actions exist. Can we make them more obvious?

The only way I could think of is adding an expand icon to the plot with a tooltip, notifying the user of the export actions. I tried adding an icon or a button to the plots but both design attempts looked strange:

### **Button (using secondary button colors)**

<img width="741" alt="Screenshot 2023-07-17 at 10 51 07 AM" src="https://github.com/iterative/vscode-dvc/assets/43496356/997e0a6f-25f1-4582-b060-c1ae22d8de30">
<img width="748" alt="Screenshot 2023-07-17 at 10 51 35 AM" src="https://github.com/iterative/vscode-dvc/assets/43496356/2951e61e-2654-4a4d-8b65-e90aa47a5457">

### **Plain Icon**


<img width="747" alt="Screenshot 2023-07-17 at 10 49 10 AM" src="https://github.com/iterative/vscode-dvc/assets/43496356/70a48d01-6b6c-4fbb-89a5-e38520b7b37c">
<img width="749" alt="Screenshot 2023-07-17 at 10 49 27 AM" src="https://github.com/iterative/vscode-dvc/assets/43496356/d6714338-2e5c-4c16-98a1-6b1aa09278fb">

Or I've just been staring at them too long 🤔 Will step away and take another look later.

[...]

PR #4252, opened 2023-07-10
Add CSV export option to zoomed in plots

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/33c7088e-e956-4157-ad7d-35397cd9a042

Part of #4246


extension/src/fileSystem/index.ts
```diff
@@ -218,6 +219,15 @@ export const writeJson = <T extends Record<string, unknown>>(
   return writeFileSync(path, json)
 }
 
+export const writeCsv = async (
+  path: string,
+  arr: Array<{ [key: string]: unknown }>
+) => {
+  ensureFileSync(path)
+  const csv = await json2csv(arr)
```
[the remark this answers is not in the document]
shcheklein: Yes! let's please add the second option also. TSV is way simpler to use in a lot of cases.

(I would check if data needs to be sanitized in some way)

extension/package.json
```diff
@@ -1651,6 +1651,7 @@
     "execa": "5.1.1",
     "fs-extra": "11.1.1",
     "js-yaml": "4.1.0",
+    "json-2-csv": "^4.0.0",
```
julieg18, line 1654: Whoops, this `^` needs to be removed

extension/src/plots/model/index.ts
```diff
@@ -237,7 +242,14 @@ export class PlotsModel extends ModelWithPersistence {
       : this.getSelectedTemplatePlotData(plotId)
 
     try {
-      writeJson(filePath, rawData as unknown as Record<string, unknown>, true)
+      if (type === PlotExportType.JSON) {
+        writeJson(filePath, rawData as unknown as Record<string, unknown>, true)
+      } else {
+        await writeCsv(
+          filePath,
+          rawData as unknown as Array<{ [key: string]: unknown }>
```
mattseddon, line 250: [Q] How can the data types for the raw data be different depending on the type of export?

extension/src/plots/model/index.ts
```diff
@@ -227,7 +228,11 @@ export class PlotsModel extends ModelWithPersistence {
     return selectedRevisions
   }
 
-  public savePlotData(plotId: string, filePath: string) {
+  public async savePlotData(
```
mattseddon, line 231: [I] Rather than have a flag argument for `savePlotData` I would have different functions `savePlotDataAsJson`, etc. Even if the `write` functions are passed to a base function.

mattseddon: that would then give different messages from the webview which could have different telemetry events and tests.

mattseddon (approved):

extension/src/plots/model/index.ts
julieg18: These types were actually just a bandaid so I could test exports. Will adjust the types!

extension/src/plots/model/index.ts
julieg18: Took off the flag argument and used separate functions instead!

PR #4262, opened 2023-07-11
Remove `dvc_data_version_info` from plot values

julieg18:
## Demo 

https://github.com/iterative/vscode-dvc/assets/43496356/12fb3717-8db0-43e8-8f45-f75eaf43c1d9

Part of #4246


extension/src/plots/model/collect.ts
```diff
@@ -156,12 +157,16 @@ const collectDatapoints = (
...
     const dvc_data_version_info = getDvcDataVersionInfo(value)
-    const data: { rev: string } = {
+    const data: { rev: string; dvc_data_version_info?: unknown } = {
       ...value,
       ...dvc_data_version_info,
       rev
     }
 
+    if (hasKey(data, 'dvc_data_version_info')) {
+      delete data.dvc_data_version_info
```
julieg18, line 167: Looking at the code, `revisionData` is only used for template plots, so should be safe to delete the key here. But please let me know if I missed something or if there is a better spot.

mattseddon (approved):

PR #4283, opened 2023-07-17
Exclude sub-project files from file and editor watchers

mattseddon:
This PR fixes some sub-project-related issues with our file system and editor watchers. We were previously not correctly excluding sub-project files from their parent project's watchers.


extension/src/pipeline/data.ts
```diff
@@ -27,8 +37,10 @@ export class PipelineData extends BaseData<{
     ])
 
     const dvcYamlsDirs = new Set<string>()
+    this.collectedFiles = []
     for (const file of fileList) {
-      if (file.startsWith(this.dvcRoot)) {
+      if (isPathInProject(file, this.dvcRoot, this.subProjects)) {
+        this.collectedFiles.push(file)
```
mattseddon, line 43: [B] Not having `collectedFiles` for this class was actually a bug.

mattseddon: discovered by writing a test for this functionality

extension/src/fileSystem/index.test.ts
```diff
@@ -408,3 +409,51 @@ describe('findOrCreateDvcYamlFile', () => {
...
   })
 })
+
+describe('isPathInProject', () => {
+  it('should return true if the path is in the project', () => {
+    const path = join(dvcDemoPath, 'dvc.yaml')
+    const dvcRoot = dvcDemoPath
+    const subProjects: string[] = []
+    expect(isPathInProject(path, dvcRoot, subProjects)).toBe(true)
+  })
```
julieg18, line 419: Looks like we're missing a new line here:
```suggestion
  })
  
```

julieg18 (approved):

extension/src/fileSystem/index.test.ts
mattseddon: I wonder if there is a prettier rule for this

mattseddon: maybe this https://eslint.org/docs/latest/rules/padding-line-between-statements

Issue #4229, opened 2023-07-06
Have a way to see information about an experiment (or many?) by hovering / clicking on them


shcheklein:
Based on user feedback. If there are two many parameters, table becomes too long and it's hard to navigate it. It's a usual story for a repo like YOLO. 

We need a way for users to be able to see a preview of an experiment, or some summary etc.

It reminds me the compare panel in Studio (table pivot), we can take that an idea (multiple experiments, pivot, highlight differences, etc), but we need a better UX for that + I think we can do better tooltips, etc. 

As an idea for UX - can we use the preview panel in VS Code for this? (besides tooltips).

mattseddon:
I think that some way of moving columns/groups of columns to the left of the table might be low-hanging fruit. I think we also need a way to make this quick pick more obvious/get people to use it:

<img width="1726" alt="image" src="https://github.com/iterative/vscode-dvc/assets/37993418/ee364761-203f-4cfc-aff9-f78faf7d6ac9">

We could expand on the "leaderboard" idea/usage from #4269 and show a condensed version of the table that can be sorted by not only columns but a formula. We would have to limit the number of columns that can be shown to < 20 or users would run into the same issue. The main difference would be that records would be shown outside of the usual commit/experiment hierarchy.

#1994 would also be a potential solution for this issue (or at least help with the problem).

shcheklein:
- Followup if user was aware about DnD cc @shcheklein 
- Check other trackers - how do they show / do they show this information?

shcheklein:
DnD - yes, it was known. It doesn't solve the issue. What is needed is a side panel (like in MD preview) that could be open and update when we select experiments.

mattseddon:
As discussed in last week's planning meeting. We could add another tree view which shows "focused" experiments transposed. Here is a screenshot with a prototype view moved to the "secondary sidebar":

<img width="1728" alt="image" src="https://github.com/iterative/vscode-dvc/assets/37993418/cb6f8972-e69b-4a48-bb22-e274fd70fd22">

It should be noted that these views can be searched/filtered for/to specific values:

https://github.com/iterative/vscode-dvc/assets/37993418/3f42c0de-c865-45ba-9890-7173dd122f43

shcheklein:
@mattseddon thanks! Can it be open / close automatically or on a button click on the "right" side (similar to MD previews)?

mattseddon:
[...]

PR #4294, opened 2023-07-18
Add select first experiment table columns quick pick

mattseddon:
Related to #4229

This PR adds a "Select Columns to Display First in the Experiments Table" command to the palette and Columns tree view/title. The command can be used to quickly move selected columns to the LHS of the experiments table.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/0183a9a7-6357-4b5a-9d92-15b193986004

Note: I plan to add this in some form to the experiments table. I think adding both this and Select Columns to the column header context menu will help users.

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -572,6 +572,12 @@
         "icon": "$(list-filter)"
       },
       {
+        "title": "Select Columns to Display First in the Experiments Table",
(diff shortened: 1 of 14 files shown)


julieg18 (approved):
Great work!

PR #4296, opened 2023-07-18
Only allow moving selected columns to the left of the experiments table

mattseddon:
- Add select first experiment table columns quick pick
- Add select column options to table header cell context menu
- Only allow moving selected columns to the left of the experiments table


mattseddon:
try adding information to the quick pick

PR #4307, opened 2023-07-19
Standardise column like quick pick

mattseddon:
# 2/2 `main` <- #4306 <- this

I noticed that we had some inconsistencies in the way that we displayed some quick picks for paths.

### Before

image

### After

https://github.com/iterative/vscode-dvc/assets/37993418/7a8e55b1-9fba-4053-8190-5e10563f8f5e

Note: I initially did all of the work to convert all quick picks to match the style in the before picture but after looking at the UX I decided to go the other way. Given the way that people invariably set out their data with multiple versions of a column in different folders (i.e train vs test) I think that only showing the final part of the path in the label is less useful than showing the whole thing.


julieg18 (approved):
Makes sense!

Issue #4269, opened 2023-07-12
Table: better way to show most important information


shcheklein:
Based on the customer feedback:

- [x] Filter branches by prefix (e.g. show only branches like "dvc-exps/%")
- [x] Setting to show only the top of each branch
- [x] Remove all records where one column is not defined (only-changed?) @shcheklein to confirm if this is about experiments or columns.
- [x] Add a way to reset number of commits per branch. As a first step show a reset action per branch + we agreed to spend some time on the performance of these actions (especially `-` since we don't have to send any commands). That should mitigate the issue with being able to quickly see only HEADs.

mattseddon:
Filtering branches by a prefix is already possible:


https://github.com/iterative/vscode-dvc/assets/37993418/41d094cf-6857-47df-acd9-f2b154d38414

mattseddon:
> Remove all record where one column is not defined

This cannot currently be done via filters because of https://github.com/iterative/vscode-dvc/pull/1732. I think we should reverse that decision. I am not sure what the effort is without looking but we could potentially only apply that logic to queued experiments.

mattseddon:
I've added a "not missing" filter which will be available in the next release

https://github.com/iterative/vscode-dvc/assets/37993418/93a3b8a9-c3e5-4c2f-865b-75282d665c00

I've also limited the number of commits for selected branches to 1.

shcheklein:
Feedback:

1. Old commits are still there and the extension lags a lot (it's almost impossible to use it because I'm waiting for several seconds for every single action)
2. Non-null filtering does not help. I filtered the output by final_score is not null, but it seems that the filtering is done on branch level rather that commit level

Can it be for some branches that we hit +/- we remembered their state? Should we have a reset to defaults or something similar?

mattseddon:
> Old commits are still there and the extension lags a lot (it's almost impossible to use it because I'm waiting for several seconds for every single action)

Lag/performance is a separate story that we need to address.

> Non-null filtering does not help. I filtered the output by final_score is not null, but it seems that the filtering is done on branch level rather that commit level

Yes, my mistake. Filters only work on experiments not commits/branches. If we wanted to change this we could go back to the logic that we used for checkpoints. I.e we filter commits/branches but if a commit/branch does not meet the filter criteria but a single experiment under the parent meets the criteria then we would continue to show the parent as well as the unfiltered experiment.

[...]

PR #4308, opened 2023-07-19
Add not missing filter for experiments table

mattseddon:
Related to #4269.

Could partly solve the third checkbox depending on the answer we get back from the user.
> Remove all records where one column is not defined (only-changed?) @shcheklein to confirm if this is about experiments or columns.

The additional filter can be used to exclude experiments that are missing a value for a column.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/93a3b8a9-c3e5-4c2f-865b-75282d665c00

Note: In the demo project I queued up an experiment and metrics are no longer missing. To be discussed with @shcheklein.

### DIFF
--- a/extension/src/experiments/model/filterBy/index.test.ts
+++ b/extension/src/experiments/model/filterBy/index.test.ts
@@ -1,0 +1,271 @@
+import { FilterDefinition, filterExperiment, Operator } from '.'
+import { buildMetricOrParamPath } from '../../columns/paths'
+import { Experiment, ColumnType } from '../../webview/contract'
+
+describe('filterExperiment', () => {
+  const paramsFile = 'params.yaml'
+  const experiments = [
+    {
+      Created: '2020-12-29T12:00:01',
+      id: 1,
+      params: {
+        'params.yaml': {
+          bool: true,
+          filter: 1,
+          sort: 1,
+          text: 'abcdefghijklmnop'
+        }
+      }
+    },
+    {
+      Created: '2020-12-30T12:00:01',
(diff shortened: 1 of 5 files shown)


extension/src/experiments/model/filterBy/index.test.ts
```diff
@@ -0,0 +1,271 @@
+import { FilterDefinition, filterExperiment, Operator } from '.'
```
mattseddon, line 1: [F] This file got purged in https://github.com/iterative/vscode-dvc/pull/3585. I brought it back/modified it to bring back some coverage.

extension/src/experiments/model/filterBy/quickPick.ts
```diff
@@ -21,6 +21,12 @@ export const OPERATORS = [
     types: ['number', 'string'],
     value: Operator.NOT_EQUAL
   },
+  {
+    description: 'Not Missing',
```
julieg18, line 25: Would an antonym of missing make more sense?  Like "Defined" or "Contains"?

julieg18 (approved):

shcheklein:
No matter the answer is, I think it's a great option to have anyways. Let's go for it and show this to the user. Thanks @mattseddon !

extension/src/experiments/model/filterBy/quickPick.ts
mattseddon: "Missing"/"Missing values" is a Data Science term that most users will be familiar with.

PR #4314, opened 2023-07-20
Add shown/hidden indicator to select first column quick pick

mattseddon:
Follow up from a conversation that we had in the planning meeting about #4296.

This PR will give a user insights as to which columns are hidden/shown when they are selecting column to move to the start of the table.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/b84b8de1-ce43-44ea-86dc-7e6118342910


julieg18 (approved):
Looks good!

PR #4319, opened 2023-07-20
Add "image by step" plots to comparison section

### LINKED ISSUE #1640: plots: How to display images and plots per step
Depends on https://github.com/iterative/dvclive/issues/425

In the current demo setup, we override the saved image and plot at the end of each step. 

Unless using checkpoints, this makes it impossible to visualize images and plots logged at previous steps. And even using checkpoints the current UI doesn't easily allow to check the image and plot from previous steps (requires `Apply to workspace` and refreshing the view)

~In order to enable visualizing images from previous steps **with or without checkpoints**, in dvclive we implemented the `log_image` function in a way that images are saved under a `step` subfolder (https://dvc.org/doc/dvclive/api-reference/live/log_image#step-updates).~ (not true anymore, need to reconsider this, edited by @shcheklein)

However, using `log_image` in the VSCode UI produces a noisy output, as each `step` image is displayed in an individual row:

Captura de Pantalla 2022-05-02 a las 17 49 21

I believe that allowing to display "per-step" images and plots is a relevant scenario for VSCode.

---

In tensorboard, logging an image each step gets rendered using a slider:

confusion matrix
predictions

---

Related https://github.com/iterative/dvc-render/issues/6

julieg18:
# `main ` <=  this <= #4372

* updates comparison plot types and collection, allowing multiple images per plot
* adds a "image by step" webview fixture for testing
* breaks up `ComparisonTableCell` component

Fixes #1640


extension/src/plots/model/collect.ts
```diff
@@ -216,6 +219,94 @@ export const collectData = (output: PlotsOutput): DataAccumulator => {
...
+  path: string
+  pathLabel: string
+  getComparisonPlotImg: (id: string, path: string) => ComparisonPlotImg
+}) => {
+  const doesPlotExist = acc[pathLabel].revisions[id]
+  if (!doesPlotExist) {
+    acc[pathLabel].revisions[id] = {
+      id,
+      imgOrImgs: [],
+      isMulti
```
mattseddon, line 246: [Q] could the data format be `{ id, imgs }`? do we really need to differentiate between the two and could we always save under a directory? In order to differentiate in the webview we would just need to check the length of `imgs `.

mattseddon: I can't spot where `isMulti` is used in the webview.

julieg18: > > could we always save under a directory

> Oh, yes that makes sense. I think grouping images in comparisonData will make handling the path data easier as well.

After trying this, I realized that this can't work correctly because we need to send missing path plots to the frontend. So our image grouping can only happen when we're collecting selected plots since that's when we know the list of paths.

extension/src/plots/model/collect.ts
```diff
@@ -216,6 +218,89 @@ export const collectData = (output: PlotsOutput): DataAccumulator => {
...
+  path,
+  selectedRevisionIds,
+  getComparisonPlotImg
+}: {
+  acc: ComparisonPlotsAcc
+  path: string
+  selectedRevisionIds: string[]
+  getComparisonPlotImg: (id: string, path: string) => ComparisonPlotImg
+}) => {
+  const pathLabel = path
```
julieg18, line 260: This gets updated to be either the `/image` director or `path` in the next pr.

mattseddon:
@julieg18 I've tried updating `train.py` in the demo project to log an image per step but it isn't getting picked up by the extension, is this expected? Does there need to be updates made to DVCLive/DVC to get this to work?

julieg18:
Converting this to a draft since I'm going to apply https://github.com/iterative/vscode-dvc/pull/4372#discussion_r1278620296.

mattseddon (approved):

PR #4324, opened 2023-07-21
Show Experiment Remote Status

mattseddon:
Related to https://github.com/iterative/vscode-dvc/issues/3574

This PR extends the experiment status indicator to the right of an experiment in the table. 

The indicator will now show whether or not an experiment has been pushed to the git remote (the extension uses `git ls-remote origin 'refs/exps/*'` to determine whether or not an experiment is on the remote). 

If the experiment is not on the remote then an icon is shown that the user can click to push. If it is on the remote then an icon is shown to indicate that status. If the experiment is pushed via the extension then we show a spinner to indicate that something is happening.

Limitations: 

- Experiments pushed via the CLI will not show a spinner
- There are no "remote watchers" to trigger table updates for things like other team members deleting your experiments from the remote or even a user running `dvc exp remove -g origin <exp-name>` themselves.
- Performance could be an issue

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/600eb158-130f-44de-bdd6-5abe1b93743b

https://github.com/iterative/vscode-dvc/assets/37993418/89f7474b-bf5e-4626-8033-4e6f5072f365

cc @dberenbaum


extension/src/experiments/model/collect.ts
```diff
@@ -233,6 +234,21 @@ const collectExecutorInfo = (
...
+  experiment: Experiment,
+  remoteExpShas: Set<string>
+): void => {
+  if (
+    !experiment.sha ||
+    ![undefined, ExperimentStatus.SUCCESS].includes(experiment.status)
+  ) {
+    return
+  }
+  experiment.pushed = remoteExpShas.has(experiment.sha)
```
mattseddon, line 247: [I] Rename from pushed

extension/src/experiments/model/collect.ts
```diff
@@ -340,10 +358,20 @@ const collectCliError = (
   }
 }
 
+const collectRemoteExpShas = (remoteExpRefs: string): Set<string> => {
+  const remoteExpShas = new Set<string>()
+  for (const ref of remoteExpRefs.split('\n').filter(Boolean)) {
```
mattseddon, line 363: [I] Use `trimAndSplit`

extension/src/test/suite/experiments/util.ts
```diff
@@ -34,6 +35,7 @@ export const buildExperiments = ({
   dvcRoot = dvcDemoPath,
   expShow = expShowFixture,
   gitLog = gitLogFixture,
+  remoteExpRefs = '',
```
mattseddon, line 38: [I] Add fixture so that both statuses are shown in the storybook

webview/src/experiments/components/table/body/Cell.tsx
```diff
@@ -61,11 +65,31 @@ export const FirstCell: React.FC<
       <div className={styles.innerCell} style={{ width: getSize() }}>
         <CellRowActions status={status} {...rowActionsProps} />
         <RowExpansionButton row={row} />
-        {isRunning(status) && (
+        {(isRunning(status) || pushed === PushedStatus.PUSHING) && (
```
mattseddon, line 68: [I] Add component `<ExperimentPushedStatus />` (better name)

extension/src/experiments/commands/index.ts
```diff
@@ -56,23 +60,39 @@ export const getPushExperimentCommand =
...
+        progress.report({
+          increment: remainingProgress,
+          message: convertUrlTextToLink(stdout)
+        })
 
-      return Toast.delayProgressClosing(15000)
+        return Toast.delayProgressClosing(15000)
+      } catch (error: unknown) {
+        void updateOnCompletion()
+        throw error
```
mattseddon, line 95: [F] This will be caught by `InternalCommands` and offer to show the output channel to the user (which will show the underlying error)

extension/src/experiments/index.ts
```diff
@@ -199,6 +201,15 @@ export class Experiments extends BaseRepository<TableData> {
     return this.notifyChanged()
   }
 
+  public setPushing(ids: string[]) {
+    this.experiments.setPushing(ids)
+    return this.notifyChanged()
+  }
+
+  public unsetPushing(ids: string[]) {
+    this.experiments.unsetPushing(ids)
```
mattseddon, line 210: [F] We don't notify here because we want to fetch the data again.

mattseddon: I've moved `update` into here to make sure the function is never accidentally called without it.

extension/src/experiments/model/collect.ts
mattseddon: remember to follow up with other instances that can be moved over.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -637,17 +637,28 @@ $badge-size: 0.85rem;
...
   border-radius: $checkbox-border-radius;
   border: $checkbox-border-width solid $checkbox-border;
   cursor: pointer;
   display: flex;
   height: $checkbox-size;
   width: $checkbox-size;
+}
+
+.plotBox {
+  @extend %iconBox;
```
mattseddon, line 651: [Q] Should I be using `cx` instead of extending a base class?

webview/src/experiments/components/table/body/ExperimentStatusIndicator.tsx
```diff
@@ -0,0 +1,55 @@
...
+type ExperimentStatusIndicatorProps = {
+  status: ExperimentStatus | undefined
+  gitRemoteStatus: GitRemoteStatus | undefined
+  id: string
+}
+
+export const ExperimentStatusIndicator: React.FC<
+  ExperimentStatusIndicatorProps
+> = ({ id, status, gitRemoteStatus }) => {
+  if (isRunning(status) || gitRemoteStatus === GitRemoteStatus.PUSHING) {
```
mattseddon, line 25: [I] This `status` field should be renamed to `executorStatus` or similar in a follow up

extension/src/cli/git/reader.ts
```diff
@@ -68,6 +69,18 @@ export class GitReader extends GitCli {
     }
   }
 
+  public async getRemoteExperimentRefs(cwd: string): Promise<string> {
+    const options = getOptions({
```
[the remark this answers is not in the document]
shcheklein: I think we should start doing status non-essential heavy checks in background for experiments. If we already do - what is your concern then?

webview/src/experiments/components/table/styles.module.scss
julieg18: Extending a base class sounds like a better option to me since this class isn't conditionally added. 

Though  `%iconBox` should be moved to the "Extendable Silent Rules" section (starts on line 15).

extension/src/cli/git/reader.ts
```diff
@@ -68,6 +69,18 @@ export class GitReader extends GitCli {
...
+      args: [Command.LS_REMOTE, DEFAULT_REMOTE, 'refs/exps/*'],
```
julieg18, line 74: > There are no "remote watchers" to trigger table updates for things like other team members deleting your experiments from the remote or even a user running dvc exp remove -g origin <exp-name> themselves.

Might have already been discussed, but have we considered adding a refresh button that will run this command again? Could be a generic one that refreshes all table data entirely or adding an extra action to our "remote state" tooltips 🤔

julieg18 (approved):
Great work!

extension/src/cli/git/reader.ts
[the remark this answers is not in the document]
mattseddon: do you mean polling?

extension/src/cli/git/reader.ts
mattseddon: We have purposefully avoided a refresh button up to this point. We may have to introduce one for remote action but let's hold off until after we decide on a strategy for fetching the data.

webview/src/experiments/components/table/body/ExperimentStatusIndicator.tsx
```diff
@@ -0,0 +1,55 @@
...
+        >
+          <Icon className={styles.cloudBox} icon={CloudUpload} />
+        </div>
+      </CellHintTooltip>
+    )
+  }
+
+  if (gitRemoteStatus === GitRemoteStatus.ON_REMOTE) {
+    return (
+      <CellHintTooltip tooltipContent="Experiment on remote">
```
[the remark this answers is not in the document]
mattseddon: for some reason I thought 16 was the default, updated

extension/src/cli/git/reader.ts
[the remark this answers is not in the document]
mattseddon: We should look at setting up WebSockets in Studio.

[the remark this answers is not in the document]
shcheklein: I mean that it's important that all these statutes don't affect UI. It's fine that some icons/ statuses are missing and appear in a while. WebSockets won't be enough since we need to run it the first time? (I might be missing the context of what we do here btw).

PR #4343, opened 2023-07-25
Disable exp push when experiment is running in the workspace

mattseddon:
This PR disables pushing an experiment via the icon introduced in #4324 when there is an experiment running in the workspace (matches context menu behaviour).

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/11f84b3a-82ce-415a-8d3a-53de61f21577

### DIFF
--- a/webview/src/experiments/components/table/styles.module.scss
+++ b/webview/src/experiments/components/table/styles.module.scss
@@ -30,6 +30,18 @@ $badge-size: 0.85rem;
   margin: 12px 6px;
 }
 
+%rowArrowContainer {
+  width: 20px;
+  height: 32px;
+  padding: 0;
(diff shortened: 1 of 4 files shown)


webview/src/experiments/components/table/styles.module.scss
```diff
@@ -657,6 +669,10 @@ $badge-size: 0.85rem;
   @extend %iconBox;
 
   padding: 1px;
+
+  &[aria-disabled='true'] {
+    cursor: default;
```
julieg18, line 674: Why not adjust the opacity here so we don't have to add two `[aria-disabled='true']` blocks for the button in our styles?

julieg18 (approved):

PR #4346, opened 2023-07-25
Remove column merge logic

mattseddon:
# 1/3 `main` <- this <- #4353 <- #4354

I suspect that the column parsing logic is a performance bottleneck (walks all params/metrics/deps for every record AND previously compared values of params/metrics for all records). There could be some quick wins that improve the performance here, this PR is a start. We now no longer re-collect/update terminal nodes once they have been collected. In subsequent PRs I'll be looking to improve performance further by reworking the logic.


extension/src/experiments/columns/collect/util.ts
```diff
@@ -136,47 +120,22 @@ export const mergeAncestors = (
...
-
-  if (valueType === 'number') {
-    newColumn.maxNumber = value as number
-    newColumn.minNumber = value as number
-  }
-
-  return newColumn
 }
 
 export const mergeValueColumn = (
```
mattseddon, line 141: [I] Needs a rename

extension/src/experiments/columns/collect/util.ts
```diff
@@ -136,60 +120,34 @@ export const mergeAncestors = (
...
   parentPath: string,
   pathArray: string[],
   label: string,
   value: Value
 ) => {
-  const valueType = getValueType(value)
-
-  const newColumn: Column = {
+  return {
+    firstValueType: getValueType(value),
```
mattseddon, line 131: [F] This is now called first value type because it only takes into account the first found value for a column

PR #4264, opened 2023-07-12
Handle non-standard experiment pipeline configurations

mattseddon:
Closes #4222

This PR separates out checking of stages from experiments. Separating out the logic makes it easier for us to target different pipelines when running experiments.

The pipeline selection logic is as follows:

- Each `dvc.yaml` is counted as a potential pipeline within a project.
- In order to be counted as a pipeline the stage list must either fail (return undefined) or be non-empty
- When there are no pipelines show the `Add Stage` button in the experiments table and block running/queuing experiments.
- If there is only one pipeline then use it as the cwd for experiment commands
- If there are multiple pipelines then automatically select the root pipeline first
- If there are multiple pipelines and none at the root then offer the user the choice each time they try to run a command.

I have added comments inline to help with the size of the change.

#### Add pipeline from Run Experiment

https://github.com/iterative/vscode-dvc/assets/37993418/412115d2-97cd-4e27-b34f-b25023e5bf5a

#### Invalid vs empty pipeline

https://github.com/iterative/vscode-dvc/assets/37993418/c8e8a45b-de0f-4475-bf61-259bde84bd9b

#### Non-root dvc.yaml

https://github.com/iterative/vscode-dvc/assets/37993418/2528dd5f-00fe-4034-91aa-68fce28c6ceb

#### Choice of pipelines

https://github.com/iterative/vscode-dvc/assets/37993418/99b610ab-5f83-4823-9b96-611510b0637a

#### Demo project

https://github.com/iterative/vscode-dvc/assets/37993418/c7cc2077-b442-4f7e-83bc-5d4976ce2e65

PR #4349, opened 2023-07-25
Fix Setup Experiments being stuck on loading

julieg18:
## Main

https://github.com/iterative/vscode-dvc/assets/43496356/156ac5ea-b427-4b00-980c-417e7afb643f

## PR

https://github.com/iterative/vscode-dvc/assets/43496356/2fbd4402-3ead-4693-807b-c2c2a4c081c0

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -579,10 +579,10 @@ export class Experiments extends BaseRepository<TableData> {
   private setupInitialData() {
     const waitForInitialData = this.dispose.track(
       this.onDidChangeExperiments(async () => {
-        await this.pipeline.isReady()
         this.deferred.resolve()
         this.dispose.untrack(waitForInitialData)
         waitForInitialData.dispose()
+        await this.pipeline.isReady()
         this.dispose.track(
           this.pipeline.onDidUpdate(() =>
             this.webviewMessages.sendWebviewMessage()


extension/src/experiments/index.ts
```diff
@@ -579,8 +579,8 @@ export class Experiments extends BaseRepository<TableData> {
   private setupInitialData() {
     const waitForInitialData = this.dispose.track(
       this.onDidChangeExperiments(async () => {
-        await this.pipeline.isReady()
         this.deferred.resolve()
```
julieg18, line 582: `getHasData` (how setup checks if experiments is done loading) was returning `undefined` since this function doesn't fire `.resolve()` until the pipeline is ready. I moved `this.diferred.resolve()` to be first which does fix the problem but there's probably a better way to solve this 🤔

mattseddon: I'll take a look at this today.

mattseddon: I think we will need an event for getHasData changing

mattseddon: Adding something like this to the `Setup` constructor is the most obvious for what is needed:

```
    this.getHasData = () => experiments.getHasData()
    this.getExpShowError = () => experiments.getCliError()

    void experiments.isReady().then(() => {
      const updateProjectHasData = () =>
        Promise.all([
          this.sendDataToWebview(),
          setContextValue(ContextKey.PROJECT_HAS_DATA, this.getHasData())
        ])
      void updateProjectHasData()
      const onDidChangeHasData = experiments.columnsChanged.event
      this.dispose.track(onDidChangeHasData(() => updateProjectHasData()))
    })
```
I would pull the new code out into a new method named something like `watchExoerimentHasDataChanges` 

Hope that helps.

mattseddon: We also can't initialize experiments before the pipeline is ready because...(had to use GitLens to find out this was added in https://github.com/iterative/vscode-dvc/pull/4264) we need `Pipeline` to have calculated the value for `hasConfig` before we can send anything to the webview.

PR #4353, opened 2023-07-26
Improve performance of experiment column collection

mattseddon:
# 2/3 `main` <- #4346 <- this <- #4354

This PR uses promises to make parts of column collection asynchronous, fixes a bug (see inline comment) and improves our early return logic to avoid unnecessary processing.

### DIFF
--- a/extension/src/experiments/columns/collect/index.ts
+++ b/extension/src/experiments/columns/collect/index.ts
@@ -28,38 +28,50 @@ const collectFromExperiment = (
 
   const { data } = experiment
   if (data) {
-    collectMetricsAndParams(acc, data)
-    collectDeps(acc, data)
+    return Promise.all([
+      collectMetricsAndParams(acc, data),
+      collectDeps(acc, data)
+    ])
   }
 }
 
-const collectFromExperiments = (
+const collectFromExperiments = async (
   acc: ColumnAccumulator,
   experiments?: ExpRange[] | null
 ) => {
   if (!experiments) {
     return
   }
+
+  const promises = []
   for (const { revs } of experiments) {
-    collectFromExperiment(acc, revs[0])
+    promises.push(collectFromExperiment(acc, revs[0]))
   }
+
+  await Promise.all(promises)
 }
 
-export const collectColumns = (output: ExpShowOutput): Column[] => {
-  const acc: ColumnAccumulator = {}
+export const collectColumns = async (
+  output: ExpShowOutput
+): Promise<Column[]> => {
+  const acc: ColumnAccumulator = { collected: new Set(), columns: {} }
 
-  acc.timestamp = timestampColumn
+  acc.columns.timestamp = timestampColumn
 
+  const promises = []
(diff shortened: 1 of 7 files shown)


extension/src/test/fixtures/expShow/deeplyNested/maxHeight.ts
```diff
@@ -352,20 +352,21 @@ export const deeplyNestedColumnsWithHeightOf2: Column[] = [
 export const deeplyNestedColumnsWithHeightOf1: Column[] = [
```
mattseddon, line 352: [F] This fixture was actually broken. The paths were wrong and one column was even missing. Bug demonstrated here:

https://github.com/iterative/vscode-dvc/assets/37993418/ad766e37-9475-44a7-af41-49d375076621

and fixed:

https://github.com/iterative/vscode-dvc/assets/37993418/b1a5ac9b-4990-482a-a768-093eda5bab06

extension/src/experiments/columns/collect/util.ts
```diff
@@ -106,17 +109,15 @@ export const mergeAncestors = (
   limitedDepthAncestors: string[],
   join: (...pathArray: string[]) => string
 ) => {
-  if (!acc[path]) {
```
mattseddon, line 109: [F] I've changed the way that the code is structure to avoid coming into this function if the path has already been collected. I've taken the same approach for `collectColumn` below

extension/src/experiments/columns/collect/metricsAndParams.ts
```diff
@@ -89,17 +96,28 @@ const walkMetricsOrParamsFile = (
...
 ) => {
   const { metrics, params } = data
-  if (metrics) {
-    walkMetricsOrParamsFile(acc, ColumnType.METRICS, metrics)
-  }
-  if (params) {
-    walkMetricsOrParamsFile(acc, ColumnType.PARAMS, params)
-  }
+  return Promise.all([
+    // eslint-disable-next-line sonarjs/no-use-of-empty-return-value
```
mattseddon, line 116: [F] These are false +ves. May need to turn the rule off as we move towards more promises like this.

extension/src/experiments/columns/collect/index.ts
```diff
@@ -28,38 +28,50 @@ const collectFromExperiment = (
...
     return
   }
+
+  const promises = []
   for (const { revs } of experiments) {
-    collectFromExperiment(acc, revs[0])
+    promises.push(collectFromExperiment(acc, revs[0]))
   }
+
+  await Promise.all(promises)
```
mattseddon, line 51: [F] I am now going to turn on https://eslint.org/docs/latest/rules/no-await-in-loop in the next PR so get ready to see this pattern a lot

mattseddon: ☝🏻 actually didn't end up turning it on but did use it to identify issues

julieg18 (approved):
Looks good!

PR #4354, opened 2023-07-26
Remove await from inside of loops

mattseddon:
# 3/3 `main` <- #4346 <- #4353 <- this

I turned on eslint's [no-await-in-loop](https://eslint.org/docs/latest/rules/no-await-in-loop) and reviewed all instances case-by-case. Of the 30 or so instances of a rule violation, the two in this PR were the ones that stood out.

### DIFF
--- a/extension/src/experiments/data/index.ts
+++ b/extension/src/experiments/data/index.ts
@@ -56,20 +56,15 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
   private async updateExpShow() {
     await this.updateBranches()
     const branches = this.experiments.getBranchesToShow()
-    let gitLog = ''
-    const rowOrder: { branch: string; sha: string }[] = []
     const availableNbCommits: { [branch: string]: number } = {}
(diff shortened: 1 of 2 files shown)


extension/src/experiments/data/index.ts
```diff
@@ -56,21 +56,16 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
...
-        branch,
-        rowOrder,
-        availableNbCommits,
-        args
-      )
+      promises.push(this.collectGitLogByBranch(branch, availableNbCommits))
     }
 
+    const branchLogs = await Promise.all(promises)
+    const { args, gitLog, rowOrder } = this.collectGitLogAndOrder(branchLogs)
```
mattseddon, line 67: [F] The order that we collect the information from the promises is important so we need to wait for them all to resolve before we can proceed. This should still give us a performance boost for a large number of branches.

julieg18 (approved):

PR #4355, opened 2023-07-26
Add reset number of commits to show for a branch button to experiments table

mattseddon:
Related to https://github.com/iterative/vscode-dvc/issues/4269

> Add a way to reset number of commits per branch.

### Demo 

https://github.com/iterative/vscode-dvc/assets/37993418/b7a92575-bac3-4511-a1ae-8c8c622c8661


extension/src/experiments/model/index.ts
```diff
@@ -478,6 +478,10 @@ export class ExperimentsModel extends ModelWithPersistence {
     this.persistNbOfCommitsToShow()
   }
 
+  public resetNbfCommitsToShow(branch: string) {
+    delete this.numberOfCommitsToShow[branch]
```
julieg18, line 482: I've looked over the code multiple times but I can't figure out how this change is saved in the state across sessions.  The function only touches `this.numberOfCommitsToShow` yet if I reload the window, the change is saved so the `numberOfCommitsToShow` is being persisted somewhere, right? How is it working though 😅 What am I missing?

julieg18 (approved):
Great work!

extension/src/experiments/model/index.ts
mattseddon: The memento could just be holding a reference to the original object. I will make an update.

PR #4363, opened 2023-07-27
Apply experiment table filters to commits

mattseddon:
# 1/2 `main` <- this <- #4364 

Related to #4269 

This PR applies experiment filters to the commits (as well as experiments). The logic for commits is as follows:

- Filter experiments using the provided filters.
- Only use the provided filters on commits which do not have unfiltered experiments.
- The workspace is currently excluded. The main reason for this approach is that it mitigates the need for a new empty state when all records in the table have been filtered. More on this at the bottom of the description. 

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/dbb33357-fbb0-4641-9956-588478b42754

If all records are filtered I am thinking of adding another screen that looks like this:

image

We can iterate on this over the next couple of days. LMK if you have any suggestions.

**We could also be in need of a more obvious indicator for records being filtered.**

Note: There is a bug shown in the demo where filtered branches do not show in the selected count. This is fixed in the subsequent PR.

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -341,6 +341,10 @@ export class Experiments extends BaseRepository<TableData> {
     return this.experiments.getFilteredCount()
   }
 
+  public getRecordCount() {
+    return this.experiments.getRecordCount()
+  }
+
   public getExperimentCount() {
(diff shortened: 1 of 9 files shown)


julieg18 (approved):
Great work!

PR #4364, opened 2023-07-27
Refactor experiment table branch data

mattseddon:
# 2/2 `main` <- #4363 <- this

This PR refactors the way that branch data is passed from the extension to the experiments webview. The main reason for the change is because of a bug surfaced when demoing the previous PR in that filtered branches would not show up as selected under the old logic.

Turns out the mapping of rows back to branches in the experiments webview was a little bit brittle. I have broken it on a few occasions and struggled to get it back working. The idea is to stabilise the logic by keeping it contained in a couple of places and slightly improve efficiency.

### Demo (bug from previous PR)

https://github.com/iterative/vscode-dvc/assets/37993418/eafa25db-aa6b-4de0-b691-83dca24399b8

### DIFF
--- a/webview/src/experiments/components/table/body/TableContent.tsx
+++ b/webview/src/experiments/components/table/body/TableContent.tsx
@@ -1,10 +1,8 @@
 import React, { Fragment, RefObject, useCallback, useContext } from 'react'
-import { useSelector } from 'react-redux'
-import { EXPERIMENT_WORKSPACE_ID } from 'dvc/src/cli/dvc/contract'
 import { TableBody } from './TableBody'
+import { collectBranchWithRows } from './util'
 import { BranchDivider } from './branchDivider/BranchDivider'
 import { RowSelectionContext } from '../RowSelectionContext'
-import { ExperimentsState } from '../../../store'
 import { InstanceProp, RowProp } from '../../../util/interfaces'
 
 interface TableContentProps extends InstanceProp {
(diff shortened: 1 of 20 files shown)


webview/src/experiments/components/table/body/util.ts
```diff
@@ -0,0 +1,32 @@
+import { Row } from '@tanstack/react-table'
+import {
+  Experiment,
+  WORKSPACE_BRANCH
+} from 'dvc/src/experiments/webview/contract'
+
+export const collectBranchWithRows = (
+  rows: Row<Experiment>[]
+): [string | typeof WORKSPACE_BRANCH, Row<Experiment>[]][] => {
+  const branchesWithRows: [
```
mattseddon, line 10: [F] Usually I would have just used an object here but you can't key an object using `undefined` | `null` so I opted for this slightly wacky collection pattern.

webview/src/experiments/components/table/body/TableContent.tsx
```diff
@@ -52,11 +49,9 @@ export const TableContent: React.FC<TableContentProps> = ({
 
   return (
     <>
-      {branches.map(branch => {
-        const branchRows = rows.filter(row => row.original.branch === branch)
```
mattseddon, line 56: [F] probably just because the focus is on efficiency this week but I decided that we should get this collection done in a single loop.

julieg18 (approved):

PR #4368, opened 2023-07-27
Increase timeouts of webview tests

mattseddon:
There is a comment inline.


extension/src/test/suite/plots/index.test.ts
```diff
@@ -519,7 +519,7 @@ suite('Plots Test Suite', () => {
         undefined,
         undefined
       )
-    })
+    }).timeout(WEBVIEW_TEST_TIMEOUT)
```
mattseddon, line 522: [F] These tests all open a webview so should have the longer timeout.

julieg18 (approved):

PR #4372, opened 2023-07-27
Add "Image by Step" plots

julieg18:
# `main` <= #4319 <= this

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/e84c4fd1-a6fd-43e4-95a3-b5bbc3b5b649~

~https://github.com/iterative/vscode-dvc/assets/43496356/b49f1401-c955-4507-ba8a-c25e99be5d78~

https://github.com/iterative/vscode-dvc/assets/43496356/a3328376-0c0f-47a7-a279-716afdfa10c9


webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -76,7 +77,11 @@ export const ComparisonTableRow: React.FC<ComparisonTableRowProps> = ({
                 data-testid="row-images"
                 className={cx(styles.cell, { [styles.cellHidden]: !isShown })}
               >
-                <ComparisonTableCell plot={plot} path={path} />
+                {plot.isMulti ? (
```
mattseddon, line 80: [Q] Could this rely on the length of `imgs`? Could drop isMulti from the type. Could `isMulti` ever be false for multiple images and visa versa?

extension/src/test/fixtures/plotsDiff/index.ts
```diff
@@ -809,9 +821,7 @@ export const getComparisonWebviewMessage = (
       ? getImageDataWithMultiImgs(baseUrl, joinFunc)
       : getImageData(baseUrl, joinFunc)
   )) {
-    const multiImagePath = joinFunc('plots', 'image')
-    const isMulti = path.includes(multiImagePath)
-    const pathLabel = path
+    const pathLabel = path.includes('image') ? join('plots', 'image') : path
```
julieg18, line 824: We had to use `join` here so the tests would pass on windows (we rebuild the `image` path when collecting the selected plots).

extension/src/plots/paths/model.ts
```diff
@@ -126,14 +127,21 @@ export class PathsModel extends PathSelectionModel<PlotPath> {
...
+    const { paths } = this.sortComparisonPaths(
       this.getPathsByType(PathType.COMPARISON)
     )
+
+    return performSimpleOrderedUpdate(this.comparisonPathsOrder, paths)
   }
 
   public setComparisonPathsOrder(order: string[]) {
-    this.comparisonPathsOrder = order
+    const { multiImagePaths } = this.sortComparisonPaths(
```
julieg18, line 138: Is there a better way to handle the paths?

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
```diff
@@ -0,0 +1,84 @@
...
+    dispatch(changeDisabledDragIds([]))
+  }, [dispatch])
+
+  let imageContent = (
+    <button
+      className={cx(styles.imageWrapper, styles.multiImageWrapper)}
+      onClick={() => zoomPlot(url)}
+      data-testid="image-plot-button"
+    >
+      <img
```
julieg18, line 42: Things that I thought could be done in a followup:

* Currently we show the step based off the index of the array which wouldn't work correctly if the user deselected some of the images in the plots sideview tree. A better way to calculate the step could be relying on the image file name or getting the index when collecting the images on the backend. 
* Save the `multiValues` across sessions.

mattseddon: > Currently we show the step based off the index of the array which wouldn't work correctly if the user deselected some of the images in the plots sideview tree. A better way to calculate the step could be relying on the image file name or getting the index when collecting the images on the backend.

Should we treat the whole thing as a single plot and not allowed selecting/de-selecting of steps?

> Save the multiValues across sessions.

Are the plot steps reset when an experiment runs, or is that handled?

mattseddon:
@julieg18 I've tried updating `train.py` in the demo project to log an image per step but it isn't getting picked up by the extension, is this expected? Does there need to be updates made to DVCLive/DVC to get this to work?

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
julieg18: > Should we treat the whole thing as a single plot and not allowed selecting/de-selecting of steps?

Yes, we could do that! That would make image and image path collection easier! Apologies if that was your original idea in https://github.com/iterative/vscode-dvc/pull/4319#discussion_r1277116939, I misunderstood 😊

julieg18:
> @julieg18 I've tried updating train.py in the demo project to log an image per step but it isn't getting picked up by the extension, is this expected? Does there need to be updates made to DVCLive/DVC to get this to work?

What change did you make? 

This was the change I made to get the images:

```
 # Iterate over training epochs.
    for epoch in range(params["epochs"]):
        # ...rest of code 
        live.log_image(f"image/{epoch}.jpg", missclassified)
```

mattseddon:
> > @julieg18 I've tried updating train.py in the demo project to log an image per step but it isn't getting picked up by the extension, is this expected? Does there need to be updates made to DVCLive/DVC to get this to work?
> 
> What change did you make?
> 
> This was the change I made to get the images:
> 
> ```
>  # Iterate over training epochs.
>     for epoch in range(params["epochs"]):
>         # ...rest of code 
>         live.log_image(f"image/{epoch}.jpg", missclassified)
> ```

Yeah, that is the change I made: `live.log_image(f"misclassified/{epoch}.jpg", missclassified)`

☝🏻 is `image` being in the path hard-coded to identify multi-step plots?

julieg18:
> Yeah, that is the change I made: live.log_image(f"misclassified/{epoch}.jpg", misclassified)
> ☝🏻 is image being in the path hard-coded to identify multi-step plots?

Yes, that is what was decided in https://github.com/iterative/dvclive/issues/425. You'd need to put `image` as the directory.

mattseddon:
> > Yeah, that is the change I made: live.log_image(f"misclassified/{epoch}.jpg", misclassified)
> > ☝🏻 is image being in the path hard-coded to identify multi-step plots?
> 
> Yes, that is what was decided in [iterative/dvclive#425](https://github.com/iterative/dvclive/issues/425). You'd need to put `image` as the directory.

Can you please double-check that `image` was not just a placeholder for the actual image name? What if there are two images that you want to save for each step? Would you have to do:

`misclassified/image/{step}` & `other_image/image/{step}`

?

julieg18:
> Can you please double-check that image was not just a placeholder for the actual image name? What if there are two images that you want to save for each step?

Oh, you're right, that's probably what was implied and I took things too literally 🤦‍♀️. I'll adjust the check we use.

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
mattseddon: It definitely wasn't the original idea but let's try and get rid of `isMulti` as a flag if possible.

extension/src/plots/model/collect.ts
```diff
@@ -153,7 +157,7 @@ const collectImageData = (
     acc[id][pathLabel] = []
   }
 
-  acc[id][pathLabel].push(plot)
+  acc[id][pathLabel].push({ ...plot, path })
```
julieg18, line 160: We need access to the path for sorting.

extension/src/plots/model/collect.ts
```diff
@@ -209,6 +213,16 @@ const collectPathData = (acc: DataAccumulator, path: string, plots: Plot[]) => {
   }
 }
 
+const sortComparisonImgPaths = (acc: DataAccumulator) => {
+  for (const [label, paths] of Object.entries(acc.comparisonData)) {
+    for (const path of Object.keys(paths)) {
+      acc.comparisonData[label][path].sort((img1, img2) =>
+        img1.path.localeCompare(img2.path, undefined, { numeric: true })
```
julieg18, line 220: `dvc plots diff` doesn't give you directly images in the step order so we need to sort them on our end.

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
julieg18: Adjusted the code to turn off selection/de-selection of steps. 

> Are the plot steps reset when an experiment runs, or is that handled?

Nope, the plot steps would only reset if you close the "session". They'll keep their current value during an experiment run :)

webview/src/plots/components/comparisonTable/styles.module.scss
```diff
@@ -170,12 +151,45 @@ $gap: 4px;
...
+  padding-bottom: 25%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.multiImageWrapper {
+  .image,
+  .noImageContent {
+    height: 380px;
```
julieg18, line 178: Had to have a fixed height for multi images since we need all images in the multi plots to be the same height:

https://github.com/iterative/vscode-dvc/assets/43496356/96067397-e567-4f23-be5a-369d156b39a4

There should be some ways to at least get an average height based off the images instead of just using a fixed number but I figured it could be saved for a followup.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
julieg18: Done!

extension/src/plots/paths/collect.test.ts
```diff
@@ -75,6 +76,47 @@ describe('collectPaths', () => {
     ])
   })
 
+  it('should return the expected data from the comparison multi image test fixture', () => {
+    expect(
+      collectPaths([], multiImagePlotsDiffFixture, REVISIONS)
```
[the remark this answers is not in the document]
mattseddon: Also saves having to add and maintain `extension/src/test/fixtures/plotsDiff/comparison/multiVscode.ts` & `extension/src/test/fixtures/plotsDiff/comparison/multi.ts`

extension/src/plots/model/collect.ts
```diff
@@ -139,7 +141,9 @@ const collectImageData = (
   path: string,
   plot: ImagePlot
 ) => {
-  const pathLabel = path
+  const pathLabel = MULTI_IMAGE_PATH_REG.test(path)
+    ? (getParent(getPathArray(path), 0) as string)
```
mattseddon, line 145: [Q] Can we make this another function? `getMultiImagePath` or something like that?

mattseddon (approved):

webview/src/plots/components/comparisonTable/styles.module.scss
julieg18: > This looks really weird. Definitely need to fix.

Understood! I didn't think it looked that bad, but if you think it looks really weird, I'm happy to work on fixing it before merging.

extension/src/plots/paths/collect.test.ts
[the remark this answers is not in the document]
julieg18: Good idea! I'll move the multi plots images into the base, probably starting with stories in #4372.

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
```diff
@@ -0,0 +1,91 @@
...
+  setMultiPlotValue
+} from '../comparisonTableSlice'
+import { PlotsState } from '../../../store'
+import { zoomPlot } from '../../../util/messages'
+
+const getCurrentStep = (stateStep: number | undefined, imgsLength: number) => {
+  if (!stateStep) {
+    return 0
+  }
+  return stateStep > imgsLength - 1 ? imgsLength - 1 : stateStep
```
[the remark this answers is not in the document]
julieg18: >  This behaviour isn't very intuitive. I had no idea what was happening when running a new experiment. I would expect the image to appear as missing if the image for that step does not exist yet. This will probably simplify the code as well

I chose the new behavior since I was avoiding our sliders having mutatable max length. Example:

https://github.com/iterative/vscode-dvc/assets/43496356/32255c68-74dd-4297-b643-d98cb4b59a40

But if you think the behavior above is preferred though, I'm happy to change it. We also could look into finding the max length across all the plots in the backend so the max length doesn't get altered as much (probably in a followup).

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  const addDisabled = useCallback(() => {
+    dispatch(changeDisabledDragIds([path]))
+  }, [dispatch, path])
+
+  const removeDisabled = useCallback(() => {
+    dispatch(changeDisabledDragIds([]))
+  }, [dispatch])
+
+  let imageContent = (
```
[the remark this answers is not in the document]
julieg18: Realized I could actually reuse `ComparisonTableCell` inside of `ComparisonTableMultiCell`.

extension/src/plots/model/collect.ts
julieg18: Good idea! Updated to use an index collected from the filename!

webview/src/plots/components/comparisonTable/styles.module.scss
mattseddon: Can still be done in follow-up. Just make sure we don't move on before it is fixed 🙏🏻.

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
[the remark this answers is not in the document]
mattseddon: I think the slider should have the max number of available steps for all experiments but also show missing if that step is not available for that experiment.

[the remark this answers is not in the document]
julieg18: > I think the slider should have the max number of available steps for all experiments but also show missing if that step is not available for that experiment.

Ok, so just making sure I'm understanding things correctly. Let's say we have two experiments, one with 12 images, and one with 5 images. We want the plots to render two plots, each with a slider with a range of 0-11?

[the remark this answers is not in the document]
daavoo: > Maybe we are overcomplicating and shouldn't be syncing these sliders across experiments. It sounded nice, but it might occasionally be frustrating if the user actually wants to compare different steps. We could always further optimize later. Maybe @daavoo or others have ideas here.

I am ok with not syncing sliders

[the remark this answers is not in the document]
julieg18: Makes sense! I can remove the syncing logic for now and we can try out different ideas in a followup.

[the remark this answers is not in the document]
dberenbaum: Sorry for the flip flopping here @julieg18!

extension/src/cli/dvc/constants.ts
```diff
@@ -18,6 +18,8 @@ export const EXP_RWLOCK_FILE = join(TEMP_EXP_DIR, 'rwlock.lock')
 export const DEFAULT_NUM_OF_COMMITS_TO_SHOW = 3
 export const NUM_OF_COMMITS_TO_INCREASE = 2
 
+export const MULTI_IMAGE_PATH_REG = /\w+[/\\|]\d+\.[a-z]+$/i
```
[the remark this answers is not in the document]
julieg18: Added some tests and adjusted the regex!

julieg18:
Merging and taking care of the following comments in a followup:

* https://github.com/iterative/vscode-dvc/pull/4372/files#r1279987427
* https://github.com/iterative/vscode-dvc/pull/4372/files#r1279668019

PR #4374, opened 2023-07-28
Apply experiment filters to workspace

mattseddon:
This PR adds a new "all filtered" empty screen to the experiments webview and apply filters to the workspace.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/ccf84e61-c669-44d7-9d85-2974ac92293e

Note: As it is not possible to star the workspace I have excluded it from the `starred` filter (fixed after demo).

### DIFF
--- a/webview/src/experiments/components/emptyState/GetStarted.tsx
+++ b/webview/src/experiments/components/emptyState/GetStarted.tsx
@@ -1,8 +1,34 @@
 import React from 'react'
 import { Welcome } from './Welcome'
 import { AddColumns } from './AddColumns'
+import { RemoveFilters } from './RemoveFilters'
 import { EmptyState } from '../../../shared/components/emptyState/EmptyState'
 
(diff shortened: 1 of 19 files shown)


webview/src/experiments/components/emptyState/GetStarted.tsx
```diff
@@ -1,8 +1,34 @@
...
+    return (
+      <EmptyState>
+        <RemoveFilters />
+      </EmptyState>
+    )
+  }
+
+  return (
+    <EmptyState>
+      <Welcome />
```
mattseddon, line 31: [F] I don't think we should ever get here anymore (mainly because of the setup webview) but will leave "just in case"

shcheklein:
Tbh, I think it's better to keep the workspace for now. It's single entry and it's special in many ways. E.g it's not static (users can change value there, etc).

PR #4375, opened 2023-07-28
Use instance spies to test experiments webview messages

mattseddon:
This PR introduces a `buildExperimentsWebview` test util which can be used to get the appropriate classes and stubs to test message passing from the experiments webview to the extension.

### DIFF
--- a/extension/src/test/suite/experiments/util.ts
+++ b/extension/src/test/suite/experiments/util.ts
@@ -1,4 +1,4 @@
-import { stub } from 'sinon'
+import { spy, stub } from 'sinon'
 import { WorkspaceExperiments } from '../../../experiments/workspace'
 import { Experiments } from '../../../experiments'
 import { Disposer } from '../../../extension'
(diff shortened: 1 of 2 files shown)


extension/src/test/suite/experiments/util.ts
```diff
@@ -129,6 +130,35 @@ export const buildExperiments = ({
...
+  stageList?: string | null
+}) => {
+  const all = buildExperiments(inputs)
+  const { experiments, messageSpy } = all
+  await experiments.isReady()
+  const webview = await experiments.showWebview()
+  messageSpy.restore()
+  const instanceMessageSpy: typeof messageSpy = spy(webview, 'show')
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ;(experiments as any).webviewMessages.sendWebviewMessage()
```
mattseddon, line 150: [F] Ideally we will move towards this instance spy pattern and be able to remove the original spy altogether. This will however take a bit of work so we'll need to move there incrementally.

mattseddon: spying on the instance is more reliable than spying on `BaseWebview.prototype`

julieg18 (approved):

PR #4385, opened 2023-07-31
Watch a small directory for testing purposes

mattseddon:


extension/src/test/suite/setup/util.ts
```diff
@@ -162,6 +162,7 @@ export const buildSetupWithWatchers = async (disposer: Disposer) => {
   ensureDirSync(TEMP_DIR)
 
   stub(WorkspaceFolders, 'getFirstWorkspaceFolder').returns(TEMP_DIR)
+  stub(WorkspaceFolders, 'getWorkspaceFolders').returns([TEMP_DIR])
```
mattseddon, line 165: [F] Stubbing this means that we should be watching far less files and a timeout should be much less likely

julieg18 (approved):

PR #4396, opened 2023-08-01
Ensure experiment summary info (columns) is always available in the experiment table data

mattseddon:
When looking at tooltips for https://github.com/iterative/vscode-dvc/issues/4229 I found multiple ways to completely break the tooltips by having the columnOrder out of sync with the data in the table. Previously the only way to fix this was by moving a column. This PR should close the loop.

### DIFF
--- a/extension/src/experiments/columns/model.ts
+++ b/extension/src/experiments/columns/model.ts
@@ -5,13 +5,13 @@ import {
   collectRelativeMetricsFiles,
   collectParamsFiles
 } from './collect'
-import { EXPERIMENT_COLUMN_ID, timestampColumn } from './constants'
 import {
   MAX_SUMMARY_ORDER_LENGTH,
   SummaryAcc,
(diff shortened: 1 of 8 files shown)


extension/src/experiments/columns/model.ts
```diff
@@ -197,12 +172,18 @@ export class ColumnsModel extends PathSelectionModel<Column> {
 
     this.data = columns
 
-    if (this.columnOrderState.length === 0) {
```
mattseddon, line 200: [B] If `this.columnOrderState.length > 0` but `columnOrderState` does not contain the current params and metrics then `getSummaryColumnOrder` gives unexpected results.

mattseddon: This is actually still the case because there can be stale data in the status object. I will make an update.

julieg18 (approved):
Great work!

PR #4398, opened 2023-08-01
Add action menu button to zoomable plots

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/3ca49372-b3df-45a3-91f6-6576bdee7857

Part of #4246


webview/src/plots/components/styles.module.scss
```diff
@@ -58,6 +58,33 @@ $gap: 20px;
   }
 }
 
+.plotActions {
+  z-index: 500;
+  width: 25px;
```
julieg18, line 63: I aimed for a size that looked "ok" across the different plot sizes since it's easier than adjust the sizing based off the one of the multiple plots size variations.

webview/src/plots/components/App.test.tsx
```diff
@@ -1416,7 +1470,9 @@ describe('App', () => {
       template: complexTemplatePlotsFixture
     })
 
-    const plot = within(screen.getAllByTestId(/^plot_/)[0]).getByRole('button')
+    const plot = within(screen.getAllByTestId(/^plot_/)[0]).getByTestId(
+      'zoomable-plot'
```
julieg18, line 1474: There are now two possible buttons (plot itself and our actions button). Added an id to differentiate.

mattseddon: [C] could have added an `aria-label` and used that instead.

webview/src/plots/components/App.test.tsx
```diff
@@ -1387,13 +1391,63 @@ describe('App', () => {
...
+  it('should open a modal with the plot zoomed in and actions menu open when key pressing on a plot actions button', async () => {
+    renderAppWithOptionalData({
+      custom: customPlotsFixture
+    })
+
+    expect(screen.queryByTestId('modal')).not.toBeInTheDocument()
+
+    const plotActionsButton = within(
+      screen.getAllByTestId(/^plot-/)[0]
+    ).getByTestId('zoomable-plot-actions')
```
mattseddon, line 1436: [C] Would have preferred `getByLabelText('See Plot Export Options')` here.

mattseddon (approved):

webview/src/plots/components/App.test.tsx
mattseddon: https://derekndavis.com/posts/getbytestid-overused-react-testing-library

mattseddon:
Storybook snapshots are unhappy: 

image

webview/src/plots/components/App.test.tsx
julieg18: Great points made in the article! I'll adjust the code to avoid using test ids.

julieg18:
> Storybook snapshots are unhappy:
> 
> image

@mattseddon, are you referring to the fact that there are changes that must be accepted in almost every plot story? Or is there something broken in the stories that I'm missing?

mattseddon:
> > Storybook snapshots are unhappy:
> > image
> 
> @mattseddon, are you referring to the fact that there are changes that must be accepted in almost every plot story? Or is there something broken in the stories that I'm missing?

image

Does that overlap reflect how the button looks in the app?

julieg18:
> Does that overlap reflect how the button looks in the app?

Yes, it happens occasionally depending on the plot's contents or size: 

image

The button has partial opacity so it's see-through. I didn't see a problem with it, but if you think it looks odd, we could possibly increase the padding of the plot container 🤔

PR #4400, opened 2023-08-01
Update config tests

mattseddon:
This PR improves our `Config` integration tests.


extension/src/test/suite/config.test.ts
```diff
@@ -52,113 +79,79 @@ suite('Config Test Suite', () => {
       expect(config.getPythonBinPath()).to.equal(pythonBinPath)
     })
 
-    it('should fire an event if dvc.focusedProjects is changed', async () => {
```
mattseddon, line 55: [F] This test became flaky in the last couple days.

mattseddon: Makes sense (to me) to break it up.

PR #4402, opened 2023-08-02
Add toggle to show only changed columns in experiments table

### LINKED ISSUE #1994: Only show relevant columns in experiments table
Similar to `dvc exp show --only-changed`, the experiments table should be able to show (either by default or through some option) only the columns where there are differences between experiments.

mattseddon:
Revisited whilst looking at #4229

Closes #1994

This PR gives users the ability to switch the experiments table into a show-only changed columns mode. The behaviour closely follows the `--only-changed` flag for `exp show` (documented [here](https://dvc.org/doc/command-reference/exp/show#--only-changed)):

image

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/e2424061-84f8-44e0-8112-7b9e568e1b53


extension/src/experiments/columns/collect/index.ts
```diff
@@ -136,3 +139,63 @@ export const collectRelativeMetricsFiles = (
...
+  for (const { pathArray, path, hasChildren } of columns) {
+    if (!pathArray || hasChildren) {
+      continue
+    }
+    checkColumn(path, pathArray, changedPaths, records)
+  }
+  return changedPaths
+}
+
+export const collectColumnsWithChangedValues = (
```
mattseddon, line 181: [I] Needs a significant test effort

extension/src/experiments/webview/messages.ts
```diff
@@ -258,14 +262,34 @@ export class WebviewMessages {
...
+    this.sendWebviewMessage()
+    sendTelemetryEvent(
+      EventName.VIEWS_EXPERIMENTS_TABLE_REFRESH,
+      undefined,
+      undefined
+    )
+  }
+
   private getWebviewData(): TableData {
+    const filters = this.experiments.getFilters()
```
mattseddon, line 276: [I] Follow up by making all of this async

webview/src/experiments/components/emptyState/AddColumns.tsx
```diff
@@ -1,10 +1,28 @@
...
-)
+export const AddColumns: React.FC = () => {
+  const showOnlyChanged = useSelector(
+    (state: ExperimentsState) => state.tableData.showOnlyChanged
+  )
+  return (
+    <div>
+      <p>No Columns Selected.</p>
+      <StartButton onClick={selectColumns} text="Add Columns" />
+      {showOnlyChanged && (
```
mattseddon, line 17: [F] Good thing I remembered this because without it users would have been stuck.

webview/src/experiments/components/table/Indicators.tsx
```diff
@@ -98,8 +101,27 @@ export const Indicators = () => {
...
+    (state: ExperimentsState) => state.tableData.showOnlyChanged
+  )
+
   return (
     <div className={styles.tableIndicators}>
+      <CellHintTooltip
+        tooltipContent="Toggle Show Only Changed Columns"
+        delay={[1000, 0]}
+      >
+        <button
```
julieg18, line 114: Is there a reason why we're basically adding a new feature instead of building off of selected columns? Aka, "Show Only Changed" selects only changed columns and deselects none-changed columns?

julieg18: There's also a possibility of users being confused on why "selected" columns just aren't showing up in the table and assume something is broken 🤔

webview/src/experiments/components/emptyState/AddColumns.tsx
```diff
@@ -1,10 +1,28 @@
...
+        <IconButton
+          icon={Close}
+          appearance="secondary"
+          isNested={true}
+          onClick={toggleShowOnlyChanged}
+          text="Show Only Changed"
```
julieg18, line 23: Minor! The "X" does make it more clear but when I saw this I assumed that only changed was off since the text is "Show Only Changed". Maybe "Disable Only Changed", Turn Off Only Changed" or "Toggle Only Changed"?

webview/src/experiments/components/table/Indicators.tsx
mattseddon: > Is there a reason why we're basically adding a new feature instead of building off of selected columns? Aka, "Show Only Changed" selects only changed columns and deselects none-changed columns?

Here is a list of things to consider with the approach of building on top of selecting columns

1. User takes a long time setting up mono-repo only to display the columns related to the project they are working in. Toggles show only changed - all selection is lost.
2. What happens when show only changed is toggled off? Would we hold the previous state and return to that? Does the selection stay the same when show only changed is turned off? What happens when selection has changed when only changed is turned on?
3. Filtering can update the columns selected for show only changed. Would this also change the selection? 
4. After applying only changed I deselect a number of columns. What happens when the data updates? Do these columns reappear? Can other new changed columns appear/get selected?

> There's also a possibility of users being confused on why "selected" columns just aren't showing up in the table and assume something is broken 🤔

This is why I tried to make the indicator as obvious as possible.

webview/src/experiments/components/emptyState/AddColumns.tsx
mattseddon: Updated to "Disable".

webview/src/experiments/components/table/Indicators.tsx
mattseddon: so tl;dr - it gets complicated fast. We can discuss tomorrow when we catch up.

webview/src/experiments/components/table/styles.module.scss
```diff
@@ -210,6 +210,14 @@ $badge-size: 0.85rem;
   }
 }
 
+.onlyChanged {
+  background-color: $indicator-badge-background;
```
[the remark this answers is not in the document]
mattseddon: PTAL at the update. We can followup if needed.

BradyJ27:
Thanks for this! @mattseddon One small thing I noticed was when using the "show only changed" setting, the "columns" tab on the left no longer works to filter out columns. I wasn't sure if this was a limitation or a bug.

mattseddon:
@BradyJ27 show only changed is an overlay on top of the selected columns. It automatically hides unchanged columns that are selected but does not change the selected state of the columns. De-selecting columns that are still visible is possible:


https://github.com/iterative/vscode-dvc/assets/37993418/477915ba-5324-49f4-a83e-84a84c29889a


See https://github.com/iterative/vscode-dvc/pull/4402#discussion_r1282005925 above for more details on why this implementation was chosen.

BradyJ27:
> @BradyJ27 show only changed is an overlay on top of the selected columns. It automatically hides unchanged columns that are selected but does not change the selected state of the columns. De-selecting columns that are still visible is possible:
> 
>  Screen.Recording.2023-08-08.at.7.55.46.am.mov 
> See [#4402 (comment)](https://github.com/iterative/vscode-dvc/pull/4402#discussion_r1282005925) above for more details on why this implementation was chosen.

Sure, this makes sense. The columns tab will still show all columns, but selecting/de-selecting hidden columns does nothing when the "toggle only changed" is on. Thanks for clarifying!

PR #4405, opened 2023-08-03
Collect webview messages async

mattseddon:
Follow up to https://github.com/iterative/vscode-dvc/pull/4402#discussion_r1281384894

This PR makes the collection of experiments and plots webview messages async.


extension/src/test/suite/experiments/index.test.ts
```diff
@@ -1527,85 +1522,49 @@ suite('Experiments Test Suite', () => {
...
-      const messageSpy = spy(BaseWebview.prototype, 'show')
-
-      const resourceLocator = disposable.track(
-        new ResourceLocator(extensionUri)
-      )
-      stub(dvcReader, 'stageList').resolves('train')
-      stub(dvcReader, 'dag').resolves('')
-
-      const pipeline = buildExperimentsPipeline({
+      const { experiments, messageSpy } = await buildExperimentsWebview({
```
mattseddon, line 1525: [F] After reading this test I had no idea why it didn't use the builder so moved it across.

extension/src/test/suite/experiments/index.test.ts
```diff
@@ -997,21 +989,22 @@ suite('Experiments Test Suite', () => {
...
-        })
+        }
+      )
 
       messageSpy.resetHistory()
 
       const movedGroup = 'params:params.yaml'
 
-      const tableChangePromise = experimentsUpdatedEvent(experiments)
+      const messageSent = waitForSpyCall(messageSpy, messageSpy.callCount)
```
mattseddon, line 1002: [F] I had to add this new util as the data for the message is now gathered async.

mattseddon: Where possible we still rely on the old util (experimentsUpdatedEvent(experiments)`)

extension/src/plots/webview/messages.ts
```diff
@@ -61,18 +61,12 @@ export class WebviewMessages {
...
-      cliError: this.errors.getCliError()?.error || null,
-      comparison: this.getComparisonPlots(),
-      custom: this.getCustomPlots(),
-      hasPlots: !!this.paths.hasPaths(),
-      hasUnselectedPlots: this.paths.getHasUnselectedPlots(),
-      sectionCollapsed: this.plots.getSectionCollapsed(),
-      selectedRevisions,
-      template: this.getTemplatePlots(selectedRevisions)
-    })
+    const webview = this.getWebview()
```
mattseddon, line 64: [F] although this function is duplicated it is not worth the hassle that a base class would bring

julieg18 (approved):
Great work!

PR #4412, opened 2023-08-03
Fix plots not showing on first experiment run

julieg18:
## main
I shorted the demos since the experiment run takes over a minute to run on my computer at the moment 😅 

https://github.com/iterative/vscode-dvc/assets/43496356/fff27f4b-81ff-4440-8e98-f9088ed753a4

## PR

https://github.com/iterative/vscode-dvc/assets/43496356/d6b4e52d-a163-4066-bcf9-05f7476c01ca

Fixes bug mentioned in [slack thread](https://app.slack.com/client/T6Z5CT96G/C01APS0FHDM/thread/C01APS0FHDM-1691060619.944549)


extension/src/plots/paths/collect.ts
```diff
@@ -118,6 +109,46 @@ const collectPathRevisions = (data: PlotsData, path: string): Set<string> => {
...
+  acc.map(existing => {
+    const plotPath = { ...existing }
+
+    if (existing.path !== path) {
+      return plotPath
+    }
+
+    plotPath.revisions = new Set([...existing.revisions, ...revisions])
+
+    if (!plotPath.type) {
```
julieg18, line 145: Not sure how common this bug is but when adding the plots to dvc.yaml in example-get-started and running an experiment, `plot diff` will have plots with `file not found` errors. This causes us to build our plot paths without types that never get added since the code only updates revisions if the path doesn't exist. 

I updated it to check the existing path's type existence but there might be a better way to handle this.

mattseddon (approved):
Needs a regression test

mattseddon:
I'll review today.

extension/src/plots/paths/collect.ts
```diff
@@ -126,31 +157,20 @@ const collectOrderedPath = (
...
   if (acc.some(({ path: existingPath }) => existingPath === path)) {
-    return acc.map(existing =>
-      existing.path === path
-        ? {
-            ...existing,
-            revisions: new Set([...existing.revisions, ...revisions])
-          }
-        : existing
-    )
+    return updateExistingPlotPath(acc, data, hasChildren, revisions, path)
```
mattseddon, line 163: [I] In `updateExistingPlotPath` you are looping over the entire array again. This could be cut down to a single loop

extension/src/plots/paths/collect.ts
mattseddon: I think we can remove the `!plotPath.type` check.

extension/src/plots/paths/collect.ts
mattseddon: *I realise that this change is just expanding on what the old code was doing but we can still improve/update.

PR #4415, opened 2023-08-04
Add filter options to experiment table header context menu

mattseddon:
This PR adds "Filter By" and "Remove Filter(s)" options to the experiments table header context menu.

The options are only available for metric/param and Created leaf headers. Remove filters is only enabled when a filter is applied.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/104a09f6-457d-423b-925e-596bdac81858


julieg18 (approved):
Great work!

PR #4425, opened 2023-08-04
Followup on fixing broken plots on first exp run

julieg18:
* reduce array looping
* add regression test

Followup to #4412


extension/src/plots/paths/collect.ts
```diff
@@ -167,15 +170,17 @@ const collectOrderedPath = (
   const isPathLeaf = idx === pathArray.length
   const isMultiImgPlot = isMultiImgDir && isPathLeaf
 
-  if (acc.some(({ path: existingPath }) => existingPath === path)) {
-    return updateExistingPlotPath(
+  const existingPathInd = acc.findIndex(existing => existing.path === path)
+  if (existingPathInd >= 0) {
```
mattseddon, line 174: [Q] would `!== -1` be easier to grok at a glance (I had to look up what `findIndex` returned for a miss).

extension/src/plots/paths/collect.test.ts
```diff
@@ -134,6 +139,46 @@ describe('collectPaths', () => {
     }
   })
 
+  it('should update the existing path types when the paths are recollected', () => {
```
mattseddon, line 142: [N] could be more specific here as it is a regression test. E.g `should collect path types after an error is returned for a new path`.

mattseddon (approved):
👍🏻

PR #4426, opened 2023-08-04
Fix broken table on detached HEAD

julieg18:
## main 

image

## PR

image

Fixes bug mentioned in [slack thread](https://iterativeai.slack.com/archives/C01APS0FHDM/p1691060619944549)

### DIFF
--- a/extension/src/cli/git/reader.ts
+++ b/extension/src/cli/git/reader.ts
@@ -1,6 +1,6 @@
 import { resolve } from 'path'
 import { GitCli } from '.'
-import { Command, DEFAULT_REMOTE, Flag } from './constants'
+import { Command, Commit, DEFAULT_REMOTE, Flag } from './constants'
 import { getOptions } from './options'
 import { typeCheckCommands } from '..'
 import { trimAndSplit } from '../../util/stdout'
@@ -118,8 +118,20 @@ export class GitReader extends GitCli {
       env: { LANG: 'en_US.UTF-8' }
     })
     try {
-      const branches = await this.executeProcess(options)
-      return trimAndSplit(branches)
+      const branchesOutput = await this.executeProcess(options)
+      const branches = trimAndSplit(branchesOutput)
+      const parsedBranches = []
+      const noBranchText = '(no branch)'
+
+      for (const branch of branches) {
+        if (branch.includes(noBranchText)) {
+          const sha = await this.revParseHead(cwd)
+          parsedBranches.push(branch.replace(noBranchText, sha))
+          continue
+        }
+        parsedBranches.push(branch)
+      }
+      return parsedBranches
     } catch {
       return []
     }
@@ -162,5 +174,18 @@ export class GitReader extends GitCli {
 
   private getUris(repositoryRoot: string, relativePaths: string[]) {
     return relativePaths.map(path => resolve(repositoryRoot, path))
+  }
+
+  private async revParseHead(cwd: string) {
+    const options = getOptions({
+      args: [Command.REV_PARSE, Commit.HEAD],
+      cwd
+    })
+    try {
+      const sha = await this.executeProcess(options)
+      return sha.slice(0, 7)
+    } catch {
+      return ''
+    }
   }
 }
--- a/extension/src/cli/git/reader.ts
+++ b/extension/src/cli/git/reader.ts
@@ -181,11 +181,7 @@ export class GitReader extends GitCli {
       args: [Command.REV_PARSE, Commit.HEAD],
       cwd
     })
-    try {
-      const sha = await this.executeProcess(options)
-      return sha.slice(0, 7)
-    } catch {
-      return ''
-    }
+    const sha = await this.executeProcess(options)
+    return sha.slice(0, 7)
   }
 }
--- a/extension/src/cli/git/reader.test.ts
+++ b/extension/src/cli/git/reader.test.ts
@@ -68,6 +68,31 @@ describe('GitReader', () => {
       const cliOutput = await gitReader.getBranches(cwd)
       expect(cliOutput).toStrictEqual([])
     })
+
+    it('should replace a "detached HEAD" branch with the latest commit', async () => {
+      const cwd = __dirname
+      const branches = ['* (no branch)']
+      mockedCreateProcess.mockReturnValueOnce(
+        getMockedProcess(branches.join('\n'))
+      )
+      mockedCreateProcess.mockReturnValueOnce(
+        getMockedProcess('lengthy0sha0tag0name')
+      )
+
+      const cliOutput = await gitReader.getBranches(cwd)
+      expect(cliOutput).toStrictEqual(['* lengthy'])
+      expect(mockedCreateProcess).toHaveBeenNthCalledWith(1, {
+        args: ['branch'],
+        cwd,
+        env: { LANG: 'en_US.UTF-8' },
+        executable: 'git'
+      })
+      expect(mockedCreateProcess).toHaveBeenNthCalledWith(2, {
+        args: ['rev-parse', 'HEAD'],
+        cwd,
+        executable: 'git'
+      })
+    })
   })
 
   describe('version', () => {


extension/src/cli/git/reader.ts
```diff
@@ -163,4 +175,13 @@ export class GitReader extends GitCli {
...
     return relativePaths.map(path => resolve(repositoryRoot, path))
   }
+
+  private async revParseHead(cwd: string) {
+    const options = getOptions({
+      args: [Command.REV_PARSE, Commit.HEAD],
+      cwd
+    })
+    const sha = await this.executeProcess(options)
+    return sha.slice(0, 7)
```
mattseddon, line 185: [N] Could use `shortenForLabel` here.

extension/src/cli/git/reader.ts
```diff
@@ -118,8 +118,20 @@ export class GitReader extends GitCli {
...
-      return trimAndSplit(branches)
+      const branchesOutput = await this.executeProcess(options)
+      const branches = trimAndSplit(branchesOutput)
+      const parsedBranches = []
+      const noBranchText = '(no branch)'
+
+      for (const branch of branches) {
+        if (branch.includes(noBranchText)) {
+          const sha = await this.revParseHead(cwd)
+          parsedBranches.push(branch.replace(noBranchText, sha))
```
mattseddon, line 129: [C] We now have two sets of logic. One here and one at: https://github.com/iterative/vscode-dvc/blob/a2920aa84f4f25a6b91bb16b6d111ac8f4dbb5ba/extension/src/experiments/model/collect.ts#L497-L501

Maybe that is ok but it would be good to pull out constants like `HEAD` & `(no branch)` as they may serve as breadcrumbs later. We could also move `isCurrentBranch` into the `git` folder.

A good practice would be to (generally) encapsulate all of the logic in the same place. I've tried to leave the reader as dumb as possible previously and not put any transformation logic in here at all (an exception being trim and split). The reason for encapsulation is that in future when things inevitably break we won't be searching through the code to work out all of the possible places something has been changed.

With all of that said maybe the correct place for this logic is in:

https://github.com/iterative/vscode-dvc/blob/a2920aa84f4f25a6b91bb16b6d111ac8f4dbb5ba/extension/src/experiments/data/index.ts#L125-L132

after calling `setBranches` check the current branch and overwrite with `revParseHead` if it matches `(no branch)` this is a special case so we should still avoid calling `revParseHead` if necessary. 

Does all of that make sense?

mattseddon (approved):

extension/src/cli/git/reader.ts
mattseddon: Another (more involved) way of fixing the issue is assuming that these commands:

https://github.com/iterative/vscode-dvc/blob/a2920aa84f4f25a6b91bb16b6d111ac8f4dbb5ba/extension/src/experiments/data/index.ts#L85-L97

will work if we don't pass a branch to them (needs tested). We could then leave the detached HEAD/no branch strings as is (I think) and not pass a branch for the current branch. If this all worked as expected it should lead to a slightly better UX of showing the original branch name in the experiments table (e.g `(no branch)` so that they user knows that they currently don't have a branch to work with.

julieg18: > Does all of that make sense?

Yes, makes sense! I'll move this branch parsing logic to `updateBranches`.

> will work if we don't pass a branch to them (needs tested). We could then leave the detached HEAD/no branch strings as is (I think) and not pass a branch for the current branch. If

Did a quick test and `GIT_GET_NUM_COMMITS` needs a branch to work with.

julieg18 (commented):
Resolved the review comments and going to rerequest reviews since I've changed the PR quite a bit

extension/src/cli/git/reader.ts
mattseddon: `GIT_GET_NUM_COMMITS` <= could this use `HEAD`?

extension/src/experiments/model/index.ts
```diff
@@ -521,6 +518,12 @@ export class ExperimentsModel extends ModelWithPersistence {
     return this.availableBranchesToShow
   }
 
+  public getAvailableBranchesToSelect() {
+    return this.availableBranchesToShow.filter(
+      branch => branch !== this.currentBranch
```
mattseddon, line 523: [Q] Should this be done in `setBranches` or further up the change in `collectBranches`?

extension/src/cli/git/reader.ts
mattseddon: Making these updates seems to work (rough): 

```
  private async updateExpShow() {
    await this.updateBranches()
    const [currentBranch, ...branches] = this.experiments.getBranchesToShow()
    const availableNbCommits: { [branch: string]: number } = {}

    const promises = []

    promises.push(
      this.collectGitLogByBranch(currentBranch, availableNbCommits, true)
    )

    for (const branch of branches) {
      promises.push(this.collectGitLogByBranch(branch, availableNbCommits))
    }

    const branchLogs = await Promise.all(promises)
    const { args, gitLog, rowOrder } = this.collectGitLogAndOrder(branchLogs)

    const expShow = await this.internalCommands.executeCommand<ExpShowOutput>(
      AvailableCommands.EXP_SHOW,
      this.dvcRoot,
      ...args
    )

    this.collectFiles({ expShow })
    return { availableNbCommits, expShow, gitLog, rowOrder }
  }

  private async collectGitLogByBranch(
    branch: string,
    availableNbCommits: { [branch: string]: number },
    isCurrent?: boolean
  ) {
    const nbOfCommitsToShow = this.experiments.getNbOfCommitsToShow(branch)

    const [branchLog, totalCommits] = await Promise.all([
      this.internalCommands.executeCommand(
        AvailableCommands.GIT_GET_COMMIT_MESSAGES,
        this.dvcRoot,
        isCurrent ? 'HEAD' : branch,
        String(nbOfCommitsToShow)
      ),
      this.internalCommands.executeCommand<number>(
        AvailableCommands.GIT_GET_NUM_COMMITS,
        this.dvcRoot,
        isCurrent ? 'HEAD' : branch
      )
    ])

    availableNbCommits[branch] = totalCommits

    return {
      branch,
      branchLog: isCurrent ? branchLog.replace('HEAD', branch) : branchLog
    }
  }

...

  private async updateBranches() {
    const allBranches = await this.internalCommands.executeCommand<string[]>(
      AvailableCommands.GIT_GET_BRANCHES,
      this.dvcRoot
    )

    const { currentBranch, branches } = collectBranches(allBranches)

    this.experiments.setBranches(branches, currentBranch)
  }


```

image


Happy to pick up this second change if I won't be stepping on your toes.

mattseddon: note: also need to update `collectBranches`:

```
    const cleanBranch = branch.replace('* ', '')
    // .replace(new RegExp(`\\(${gitPath.DOT_GIT_HEAD}\\s\\w+\\s\\w+\\s`), '')
    // .replace(')', '')

```

julieg18: Ohhhhhh, ok, I understand your solution a lot better, thanks! I'll apply your changes + add tests.

extension/src/experiments/data/index.ts
```diff
@@ -78,21 +83,23 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
 
   private async collectGitLogByBranch(
     branch: string,
-    availableNbCommits: { [branch: string]: number }
+    availableNbCommits: { [branch: string]: number },
+    isCurrent?: boolean
   ) {
     const nbOfCommitsToShow = this.experiments.getNbOfCommitsToShow(branch)
+    const branchName = isCurrent ? gitPath.DOT_GIT_HEAD : branch
```
mattseddon, line 90: [F] An explanatory variable would have been good here. The reason being when we come back to this we would have some idea as to why we made this decision. Example would be

- `detachedSafeBranchName` 
- `safeBranchName`

The information won't be lost as we can always come back to this PR (I find the PR via Gitlens file/line history) but you are saving your future self some work by fighting to make the code as expressive as possible 👍🏻.

PR #4427, opened 2023-08-05
Reduce wait time in test utils

mattseddon:
This should (slightly) cut down the amount of time taken to run the integration tests.


extension/src/test/suite/util.ts
```diff
@@ -330,7 +330,7 @@ export const waitForSpyCall = async (
   messageSpy: SinonSpy,
   originalCallCount: number
 ): Promise<unknown> => {
-  await Time.delay(500)
+  await Time.delay(100)
```
mattseddon, line 333: [F] Could probably go as far as using process.nextTick here but this is good enough for now.

julieg18 (approved):

PR #4432, opened 2023-08-07
Fix sorts and filters for dep columns

mattseddon:
After considering excluding deps from sorts and filters I decided that fixing the functionality was a better option.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/2dc27aa6-e49c-4acb-ba4f-c4aec6a7a643

Note: I am pretty sure this has never worked as expected as it was missed when highlighting dep changes was first implemented.


julieg18 (approved):
Great work!

Issue #4422, opened 2023-08-04
Improve "Image By Step" plots


julieg18:
- [x] Remove fixed height on plots (https://github.com/iterative/vscode-dvc/pull/4372/files#r1279668019)
- [x] Save slider values across sessions
- [x] Review need for img check in `ComparisonTableRows` (https://github.com/iterative/vscode-dvc/pull/4372/files#r1279987427)

PR #4458, opened 2023-08-09
Fix Comparison Plot "Image By Step" height

julieg18:
## PR

https://github.com/iterative/vscode-dvc/assets/43496356/1dcebe74-77d6-49fa-b532-ddd097e37af0

## Main

https://github.com/iterative/vscode-dvc/assets/43496356/9fa00ab8-e12b-45de-a9a2-d71747310ac0

Part of #4422


webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -39,6 +44,31 @@ export const ComparisonTableRow: React.FC<ComparisonTableRowProps> = ({
     setIsShown(!isShown)
   }
 
+  const updateMultiImgHeight = (img: HTMLImageElement) => {
+    const aspectRatio = img.naturalWidth / img.naturalHeight
+    const width = img.clientWidth
+    const calculatedHeight = Number.parseFloat((width / aspectRatio).toFixed(2))
+    setMultiImgHeight(calculatedHeight)
```
julieg18, line 51: Anyone have a better idea on how to accomplish keeping the images the same height?

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -39,6 +44,31 @@ export const ComparisonTableRow: React.FC<ComparisonTableRowProps> = ({
...
+  }
+
+  useEffect(() => {
```
julieg18, line 54: Tried adding a test but it would fail:

```
  it('should update the multi image height when changing the width of plots', async () => {
    const store = renderAppWithOptionalData({
      comparison: comparisonTableFixture
    })
    setWrapperSize(store)

    const multiImage = screen.getAllByTestId('multi-image-cell')[0]

    const originalHeight =
      getComputedStyle(multiImage).getPropertyValue('--img-height')

    await waitFor(
      () => {
        const imgEl: HTMLImageElement = within(multiImage).getByRole('img')
        expect(Boolean(imgEl.complete)).toStrictEqual(true) // fails
      },
      { timeout: 1000 }
    )

    sendSetDataMessage({
      comparison: {
        ...comparisonTableFixture,
        width: 3
      }
    })

    await waitFor(
      () => {
        expect(
          getComputedStyle(multiImage).getPropertyValue('--img-height')
        ).not.toBe(originalHeight)
      },
      { timeout: 1000 }
    )
  })
```

I think it's due to how the virtual dom deals with images 🤔. Any idea on how to test for this?

mattseddon: Have to take a snapshot

mattseddon (approved):
This is fine, but let's double-check that we actually need it instead of simply removing the fixed height.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
julieg18: > Are you able to reproduce the broken behaviour from https://github.com/iterative/vscode-dvc/pull/4372 on main with height: 380px; removed from .multiImageWrapper?

Yes, I'm still able to reproduce. Here's a demo:

https://github.com/iterative/vscode-dvc/assets/43496356/3c7ad2d5-bd55-4aea-91f6-c2875421ecb5

Images need to be same height to slider doesn't "jump". 

>  I wanted to see if simply setting img.styles.height here and not using useState worked

Oh, that should work! I'll try it out.

julieg18: Tested setting img.styles.height on the images directly instead of using `useState`! Realized that we'd have to set img state directly on to all images in the row and we'd have to add an detection to see when a new image is added so we could update the height for that image which makes things more complicated.

But we could apply `--img-height` directly on to the multi image row wrapper instead of each individual cell. That way we can avoid using `useState` and not need to watch for image addition. I'll update the code :)

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
julieg18: Added an extra snapshot!

PR #4473, opened 2023-08-10
Improve efficiency of column transformation in experiments table

mattseddon:
When running the extension with lots of columns it becomes apparent that the experiments webview is completely tanking compared to everything else. I think we need to

1. cut down the amount of data being passed to the bare minimum
2. restructure some of the data to reduce the amount of calculations it has to make

### DIFF
--- a/webview/src/experiments/components/Experiments.tsx
+++ b/webview/src/experiments/components/Experiments.tsx
@@ -26,6 +26,8 @@ import styles from './table/styles.module.scss'
 import { ErrorState } from './emptyState/ErrorState'
 import { GetStarted } from './emptyState/GetStarted'
 import { RowSelectionProvider } from './table/RowSelectionContext'
+import { DateCellContents } from './table/content/DateCellContent'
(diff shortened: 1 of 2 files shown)


webview/src/experiments/components/Experiments.tsx
```diff
@@ -67,34 +69,55 @@ const getDefaultColumn = () =>
     size: 240
   })
 
+const collectColumnObj = (
+  columns: Column[]
+  // eslint-disable-next-line sonarjs/cognitive-complexity
+): { [parentPath: string]: Column[] } => {
+  const columnObj: { [parentPath: string]: Column[] } = {}
```
mattseddon, line 76: [I] Move this back into the extension. Only send minimal columns data to the webview.

[C] Consider limiting the amount of data we send for each experiment to the required columns.

mattseddon: Also need to work out how to virtualize columns. Rows seems easy enough

PR #4476, opened 2023-08-10
Save comparison multi plot image values across sessions

julieg18:
## To Do

- [ ] Lower amount of calls to backend being made
- [ ] Review solution
- [ ] Tests

Part of #4422

### DIFF
--- a/extension/src/plots/model/index.ts
+++ b/extension/src/plots/model/index.ts
@@ -33,7 +33,8 @@ import {
   DEFAULT_NB_ITEMS_PER_ROW,
   PlotHeight,
   SmoothPlotValues,
-  ImagePlot
+  ImagePlot,
+  ComparisonMultiPlotValues
 } from '../webview/contract'
 import {
   EXPERIMENT_WORKSPACE_ID,
@@ -80,6 +81,7 @@ export class PlotsModel extends ModelWithPersistence {
 
   private comparisonData: ComparisonData = {}
   private comparisonOrder: string[]
+  private comparisonMultiPlotValues: ComparisonMultiPlotValues = {}
   private smoothPlotValues: SmoothPlotValues = {}
 
   private revisionData: RevisionData = {}
@@ -111,6 +113,10 @@ export class PlotsModel extends ModelWithPersistence {
     this.customPlotsOrder = this.revive(PersistenceKey.PLOTS_CUSTOM_ORDER, [])
     this.smoothPlotValues = this.revive(
       PersistenceKey.PLOTS_SMOOTH_PLOT_VALUES,
+      {}
+    )
+    this.comparisonMultiPlotValues = this.revive(
+      PersistenceKey.PLOTS_COMPARISON_MULTI_PLOT_VALUES,
       {}
     )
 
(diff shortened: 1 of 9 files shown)


webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
```diff
@@ -1,16 +1,23 @@
...
+import { PlotsState } from '../../../store'
 
 export const ComparisonTableMultiCell: React.FC<{
   path: string
   plot: ComparisonPlot
 }> = ({ path, plot }) => {
-  const [currentStep, setCurrentStep] = useState<number>(0)
+  const values = useSelector(
+    (state: PlotsState) => state.comparison.multiPlotValues
+  )
```
julieg18, line 16: Another slightly more responsive but slightly more complex solution is keeping the slider/image step state being handled by React `useState` and only using `multiPlotValues` on mount. Example of how that looks:

```
const values = useSelector(
    (state: PlotsState) => state.comparison.multiPlotValues
  )
const [step, setCurrentStep] = useState(values?.[path]?.[plot.id] || 0)

  useEffect(() => {
    window.clearTimeout(changeDebounceTimer.current)
    changeDebounceTimer.current = window.setTimeout(() => {
      setComparisonMultiPlotValue(path, plot.id, currentStep)
    }, 500)
  }, [path, plot.id, currentStep])

return (
   <div data-testid="multi-image-cell" className={styles.multiImageWrapper}>
      <ComparisonTableCell
        path={path}
        plot={{ id: plot.id, imgs: [plot.imgs[currentStep]] }}
        imgAlt={`${currentStep} of ${path} (${plot.id})`}
      />
      <div
        className={styles.multiImageSlider}
        onMouseEnter={addDisabled}
        onMouseLeave={removeDisabled}
      >
        <label htmlFor={`${plot.id}-step`}>Step</label>
        <input
          name={`${plot.id}-step`}
          min="0"
          max={plot.imgs.length - 1}
          value={currentStep}
          type="range"
          onChange={event => {
            setCurrentStep(Number(event.target.value))
          }}
        />
        <p>{currentStep}</p>
      </div>
    </div>
)
```

julieg18: On further testing, we're going to need to apply this solution to ensure everything stays stable during experiment runs.

extension/src/plots/model/index.ts
```diff
@@ -307,6 +313,26 @@ export class PlotsModel extends ModelWithPersistence {
     this.persist(PersistenceKey.PLOT_COMPARISON_ORDER, this.comparisonOrder)
   }
 
+  public setComparisonMultiPlotValue(
+    revision: string,
+    path: string,
+    value: number
+  ) {
+    if (!this.comparisonMultiPlotValues[revision]) {
+      this.comparisonMultiPlotValues[revision] = {}
```
julieg18, line 322: Should we look into removing old revisions from the `comparisonMultiPlotValues` state? Updating the state every time a revision gets removed, for example?

webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
```diff
@@ -20,6 +27,13 @@ export const ComparisonTableMultiCell: React.FC<{
     dispatch(changeDisabledDragIds([]))
   }, [dispatch])
 
+  useEffect(() => {
+    window.clearTimeout(changeDebounceTimer.current)
+    changeDebounceTimer.current = window.setTimeout(() => {
+      setComparisonMultiPlotValue(path, plot.id, currentStep)
```
julieg18, line 33: Originally, I wanted to rely solely on `values` and remove `useState` but that
* was less performant (took a sec for slider change to update the image and only when the user let go of their mouse)
* made it more difficult to update the slider since the browser was handling the state instead of react

extension/src/plots/model/index.ts
mattseddon: If we were going to do that I would check to see if the revision is in the full list of experiments/commits before removing

mattseddon (approved):

PR #4478, opened 2023-08-10
Fix image by step plots breaking on nonexistent steps

julieg18:
Spotted this possible bug when working on #4476. See inline comment for details. 

Demos were made in an environment where the experiment run had less epochs than the workspace initially. I trimmed the videos towards the end.  

## Main

https://github.com/iterative/vscode-dvc/assets/43496356/99d53da0-1264-461a-b8f4-7037c0510e0e

## PR

https://github.com/iterative/vscode-dvc/assets/43496356/cd03930f-22ec-4a4e-982e-a317f0282b87


webview/src/plots/components/comparisonTable/cell/ComparisonTableMultiCell.tsx
```diff
@@ -24,7 +24,16 @@ export const ComparisonTableMultiCell: React.FC<{
     <div data-testid="multi-image-cell" className={styles.multiImageWrapper}>
       <ComparisonTableCell
         path={path}
-        plot={{ id: plot.id, imgs: [plot.imgs[currentStep]] }}
+        plot={{
+          id: plot.id,
+          imgs: [
+            plot.imgs[currentStep] || {
```
julieg18, line 30: Running experiments start off with the workspace's images than go on to hold their own created images (Is that normal or is there a bug with `plot diff`?). This means that an revision could start off with 20 images in the beginning of an experiment run if the `workspace` originally had 20 than less than 20 by the end. 

A user could set a multi image slider to a step that doesn't actually exist within the experiment during the beginning of the experiment run. This would cause the plots to crash once the images get updates to a smaller length. This change handles this scenario and stops plots from crashing.

mattseddon (approved):

PR #4481, opened 2023-08-11
Fix comparison table test

julieg18:
Fixes test that was sending mock data that was missing plot images

Part of https://github.com/iterative/vscode-dvc/issues/4422

### DIFF
--- a/webview/src/plots/components/comparisonTable/ComparisonTableRows.tsx
+++ b/webview/src/plots/components/comparisonTable/ComparisonTableRows.tsx
@@ -54,9 +54,7 @@ export const ComparisionTableRows: React.FC<ComparisonTableRowsProps> = ({
           path={path}
           plots={columns.map(column => ({
             id: column.id,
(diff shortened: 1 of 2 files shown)


webview/src/plots/components/comparisonTable/ComparisonTableRows.tsx
```diff
@@ -54,9 +54,7 @@ export const ComparisionTableRows: React.FC<ComparisonTableRowsProps> = ({
           path={path}
           plots={columns.map(column => ({
             id: column.id,
-            imgs: revs[column.id]?.imgs || [
-              { errors: undefined, loading: false, url: undefined }
-            ]
+            imgs: revs[column.id]?.imgs
```
julieg18, line 57: Reviewed the way we created images on the backend and did some testing on the frontend. I don't think it's possible for `imgs` to be missing on the backend.

mattseddon (approved):

Issue #4323, opened 2023-07-21
If dvc fails to initialize (CLI fails) show experiments doesn't get to setup


shcheklein:
(Sorry, typing this fast. Can give more details. This is first to not forget to discuss and fix this)

It open an empty page with a broken state. Which is also fine, after refresh button click it works (since DVC is fixed now). But it's now on the empty screen with some link to the docs.

We also need a refresh button in the side panel, data tree. (I had to open experiments to get to the refresh in the first place).

julieg18:
Not able to reproduce so far 🤔 

* What OS are you on?
* Any errors in the console or web developer tools?

shcheklein:
OS - I think it was Mac OS.

Errors - don't know in this case, sorry. 

Did you manage to reproduce a state where it's an error in the side panel? You would need to make DVC to throw some unexpected error may be.

shcheklein:
> But you're saying that after you clicked refresh, you saw an empty screen with doc links despite having experiment data? Did the webview stay on this doc link page or did it resolve itself?

I think I didn't have experiments in that repo. What I was expecting though is it to take me to the Setup page, the experiments section. As it usually happens when you don't have any experiments and trying to open the panel.

Error page is totally fine an expected (since if it's broken we don't even know if we have experiments or not).

PR #4483, opened 2023-08-11
Have experiment webview redirect to setup when there is no data

julieg18:
## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/ae03f8be-3e8b-445a-a559-37e8047a8b03~

https://github.com/iterative/vscode-dvc/assets/43496356/5b59993c-fa2c-4684-bfca-800de5fbf096

Part of #4323


extension/src/experiments/webview/messages.ts
```diff
@@ -265,6 +267,19 @@ export class WebviewMessages {
...
+    sendTelemetryEvent(
+      EventName.VIEWS_EXPERIMENTS_TABLE_REDIRECT_TO_SETUP,
+      undefined,
+      undefined
+    )
+
+    await commands.executeCommand(RegisteredCommands.SETUP_SHOW_EXPERIMENTS)
+
+    const webview = this.getWebview()
+    webview?.dispose()
```
julieg18, line 280: Is there a better way to handle a "redirect"?

webview/src/experiments/components/emptyState/GetStarted.tsx
```diff
@@ -1,8 +1,24 @@
...
+  useEffect(() => {
+    if (noExpsToDisplay) {
+      redirectToSetup()
+    }
+  }, [noExpsToDisplay])
+
+  return (
+    <EmptyState>
+      {noExpsToDisplay ? (
+        <span>No experiments to display. Redirecting to setup...</span>
```
julieg18, line 18: Could also leave the text empty since the redirect should happen quickly and users likely won't have time to see the message.

extension/src/experiments/webview/messages.ts
mattseddon: Why send the message to the webview to send a message back? could add code in `sendWebviewMessage` to close the webview/open setup if no data is detected. This would mean the logic in the webview could also be simplified as hasData would never be false (could remove it)

mattseddon: less code to maintain, less to test, etc

julieg18: Good idea!

julieg18: Done!

webview/src/stories/Table.stories.tsx
```diff
@@ -39,7 +39,6 @@ const tableData: TableDataState = {
   filters: ['params:params.yaml:lr'],
   hasBranchesToSelect: true,
   hasCheckpoints: true,
-  hasColumns: true,
```
mattseddon, line 42: [Q] Is there a story that needs to be removed?

mattseddon (approved):

PR #4486, opened 2023-08-12
Replace experiments table react context with redux state

mattseddon:


webview/src/experiments/components/table/body/Row.tsx
```diff
@@ -39,11 +41,16 @@ export const RowContent: React.FC<
         if (args?.mouse?.shiftKey) {
           batchRowSelection({ row })
```
mattseddon, line 42: [I] still need to try and get rid of this

julieg18 (approved):

PR #4500, opened 2023-08-14
Add refresh button to experiments tree cli errors

### LINKED ISSUE #4323: If dvc fails to initialize (CLI fails) show experiments doesn't get to setup

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/bafe09ff-4cfe-4780-b184-a185d2694466

Fixes #4323


extension/src/experiments/index.ts
```diff
@@ -585,6 +585,10 @@ export class Experiments extends BaseRepository<TableData> {
     return this.experiments.getAvailableBranchesToSelect()
   }
 
+  public refresh() {
+    return this.data.update()
```
mattseddon, line 589: [F] There are more instances of `this.data.update` in this file.

mattseddon (approved):

PR #4503, opened 2023-08-15
Avoid more unnecessary experiments table re-renders

mattseddon:
This PR changes the level that some props are at, adds `memo` to two components and updates drilling to avoid some unnecessary re-renders.


julieg18 (approved):
Looks good!

PR #4508, opened 2023-08-15
Fix tests

mattseddon:
See https://iterativeai.slack.com/archives/C01RB7BTG4C/p1692106121927349


extension/src/test/suite/experiments/model/filterBy/tree.test.ts
```diff
@@ -54,6 +55,7 @@ suite('Experiments Filter By Tree Test Suite', () => {
 
   afterEach(() => {
     disposable.dispose()
+    return closeAllEditors()
```
mattseddon, line 58: [F] This set of tests opens webviews and should close them when it is done with them.

mattseddon: same goes for the other suites where this is added

shcheklein (approved):

PR #4510, opened 2023-08-16
Add Scikit-learn and YOLO to other frameworks supported by DVCLive

mattseddon:
### Storybook 

image


daavoo (commented):
Up to you but, for the record, in Studio I went with dropping the detailed list in favor of a generic link https://github.com/iterative/studio/pull/7111
Not sure how frequent we will add frameworks but the list is likely to grow

julieg18 (approved):

PR #4511, opened 2023-08-16
Make redux utils generic

mattseddon:
Noticed that these comparison utils always had to cast values after working with them. Making them generic reduces the need to cast.


webview/src/experiments/state/rowSelectionSlice.ts
```diff
@@ -80,10 +80,7 @@ export const rowSelectionSlice = createSlice({
     },
     updateRowOrder(state, action: PayloadAction<SelectedRow[]>) {
       const newOrder = action.payload
-      state.rowOrder = keepEqualOldReferencesInArray(
-        state.rowOrder,
-        newOrder
-      ) as SelectedRow[]
+      state.rowOrder = keepReferenceIfEqual(state.rowOrder, newOrder)
```
mattseddon, line 83: [F] This was a bug (used the wrong util)

julieg18 (approved):

PR #4514, opened 2023-08-16
Update demo project and latest tested CLI version (3.15.3)

julieg18:


mattseddon (approved):
Can you please update the demo repo again before merging so that we get the DVCLive update without having to raise another PR in this repo 🙏🏻.

PR #4522, opened 2023-08-16
Fix possibly cuttoff cli errors in tree views

### LINKED ISSUE #4519: Experiments UI shows cutoff error message
I called `dvc destroy -f` and found that the `Experiments` menu shows only the last part of the path.
On mouse-over the full message is shown. Maybe This could be displayed differently?

image

julieg18:
* fix errors in tree views possibly getting cut off due to containing a forward slash

## Main

Screenshot 2023-08-16 at 5 47 22 PM

## PR

Screenshot 2023-08-16 at 5 47 57 PM

Fixes #4519


extension/src/tree/index.ts
```diff
@@ -61,6 +61,9 @@ export const getCliErrorTreeItem = (
     command: RegisteredCommands.EXTENSION_SHOW_OUTPUT,
     title: 'Show DVC Output'
   }
+
+  treeItem.label = label
```
julieg18: Any way to test for this issue? I couldn't come up with anything

extension/src/tree/index.ts
```diff
@@ -47,7 +47,8 @@ export const getCliErrorLabel = (msg: string): string =>
 export const getCliErrorTreeItem = (
   label: string,
   msg: string,
-  decoratableTreeItemScheme: DecoratableTreeItemScheme
+  decoratableTreeItemScheme: DecoratableTreeItemScheme,
+  isLabelPath = true
```
mattseddon, line 51: `isLabelPath` is hard to track. 

IIUC -
You are setting it to true for both `repository` and `experiments` but the default will also set it to true for `plots`.

then `!isLabelPath` will then never get hit.

mattseddon: can you add some tests around this please

extension/src/tree/index.ts
```diff
@@ -61,6 +62,11 @@ export const getCliErrorTreeItem = (
     command: RegisteredCommands.EXTENSION_SHOW_OUTPUT,
     title: 'Show DVC Output'
   }
+
+  if (!isLabelPath) {
+    treeItem.label = label
```
mattseddon, line 67: [C] If this is the only difference for the boolean flag why not create a wrapper function that calls `getCliErrorTreeItem` and updates the label of the resulting `treeItem`? The wrapper can get a descriptive name.

extension/src/tree/index.ts
julieg18: > You are setting it to true for both repository and experiments but the default will also set it to true for plots.

Whoops, forgot to push the latest commit that fixed that.

> isLabelPath is hard to track.

Makes sense, I'll rethink the solution. 

> can you add some tests around this please

I wasn't sure how to test for this, but after reviewing our tree tests again, maybe I can test with `getTreeItem`. I'll look into that!

extension/src/tree/index.ts
julieg18: Oh, looks like my original review never got sent in 🤣. As mentioned in https://github.com/iterative/vscode-dvc/pull/4522#discussion_r1296543485 going to look into testing this with `getTreeItem`.

extension/src/repository/model/tree.test.ts
```diff
@@ -406,12 +407,16 @@ describe('RepositoriesTree', () => {
 
       expect(mockedTreeItem).toHaveBeenCalledTimes(1)
       expect(treeItem).toStrictEqual({
-        ...mockedItem,
```
julieg18, line 409: I added tests for each tree that uses the `cliError` tree item and adjusted the check to not use `mockedItem` so we could confirm that `label` is or is not there.

julieg18 (commented):
Simplified the solution and added tests!

mattseddon (approved):

PR #4527, opened 2023-08-18
Restructure plots path tree for nested dvc.yaml paths

### LINKED ISSUE #4489: Wrong plots group in the side panel
On Codespaces. `dvc.yaml::dvclive` doesn't exist as a directory. Repo is this one (can be changing as do other things): https://github.com/shcheklein/ultralytics (branch `dvc` + run an experiment with a regular launch action, it doesn't have dvc.yaml in the root). 

Screenshot 2023-08-12 at 11 38 17 AM

but that's what I'm getting in the plots section:

Screenshot 2023-08-12 at 11 36 41 AM

## Bonus

A general UX issue. In the same section (and other section) - things should be copy-pastable. Either via context menu / or via some select + hot key or with some other mechanism.

## Environment

Extension: v1.0.38
DVC: 

```
DVC version: 3.15.1 (pip)
-------------------------
Platform: Python 3.10.12 on Linux-5.15.0-1042-azure-x86_64-with-glibc2.36
Subprojects:
        dvc_data = 2.13.1
        dvc_objects = 0.25.0
        dvc_render = 0.5.3
        dvc_task = 0.3.0
        scmrepo = 1.2.1
Supports:
        http (aiohttp = 3.8.5, aiohttp-retry = 2.8.3),
        https (aiohttp = 3.8.5, aiohttp-retry = 2.8.3)
Config:
        Global: /home/vscode/.config/dvc
        System: /etc/xdg/dvc
Cache types: hardlink, symlink
Cache directory: ext4 on /dev/loop3
Caches: local
Remotes: None
Workspace directory: ext4 on /dev/loop3
Repo: dvc, git
Repo.site_cache_dir: /var/tmp/dvc/repo/ae31ab50ca6008737617db73d448c512
```

mattseddon:
Closes #4489 

This PR updates the plot "paths" shown in the plots tree. The "paths" are now structured to have the full path to the relevant `dvc.yaml` as the first parent. All children underneath the `dvc.yaml` path are then put into the normal tree structure. 

The crux of the implementation is that the path of the tree nodes remains the same but the path to the `dvc.yaml` is squashed into a single tree item. A label has been added to each tree item to ensure that the plot IDs show up as expected. 

### Screenshot

Screenshot 2023-08-22 at 10 56 54 am

Note: If a plot's id is not prefixed by a `dvc.yaml` path then there is no change to the previous behaviour.


extension/src/plots/paths/collect.ts
```diff
@@ -165,7 +166,7 @@ const collectOrderedPath = (
   idx: number,
   isMultiImgDir: boolean
 ): PlotPath[] => {
-  const path = getPath(pathArray, idx)
+  const path = getPath(pathArray, idx).replace(`dvc.yaml${sep}`, 'dvc.yaml::')
```
mattseddon, line 169: [F] Need to separate out the path from the id (probably add a field)

extension/src/plots/paths/collect.ts
```diff
@@ -196,6 +197,15 @@ const collectOrderedPath = (
   return acc
 }
 
+const fun = (path: string): string[] => {
+  if (!path.includes('::')) {
+    return getPathArray(path)
+  }
+
+  const [dvcYamlPath, plotPath] = path.split('::')
+  return [dvcYamlPath, ...getPathArray(plotPath)]
```
mattseddon, line 206: The first entry will always be the path to the dvc.yaml probably needs another field as we use the information in the tree ([here](https://github.com/iterative/vscode-dvc/pull/4527/files#diff-69f2a655cd9dd9767fb87d0d4f5cfcaabd011d1a8212d0bb0b45691d9c4423edR64))

extension/src/plots/paths/collect.test.ts
```diff
@@ -230,56 +242,133 @@ describe('collectPaths', () => {
...
       },
       {
         hasChildren: false,
+        label: 'predictions.json',
         parentPath: undefined,
         path: 'predictions.json',
         revisions: new Set(revisions),
         type: new Set(['template-multi'])
+      },
+      {
```
mattseddon, line 324: [F] These next objects are examples of how we now handle `dvc.yaml` paths

extension/src/plots/paths/collect.ts
```diff
@@ -157,6 +159,46 @@ const updateExistingPlotPath = ({
   return acc
 }
 
+const separateDvcYamlPath = (path: string): string[] => {
```
mattseddon, line 162: [F] I've kept these three functions together to try and encapsulate the logic in a single place.

PR #4544, opened 2023-08-21
Add copy sha and name to experiment table row context menu

### LINKED ISSUE #4533: Context menu entry to copy experiment name / base commit SHA
I can highlight the short base commit SHA in the table and copy it in order to paste it into the Terminal to use with `dvc exp show --rev` for example, but it would be nice to have context menu entries (right click) to copy SHA and/or experiment name for a given row in the Experiments table.

For `params` and `metrics` columns, there is a "Copy cell contents" icon when hovering over it. Maybe this could be added also for the baseline commit SHA and experiment name / ID.

mattseddon:
Closes #4533 

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/6256ab15-fece-4d3f-88a1-c332895311be


extension/src/experiments/webview/messages.ts
```diff
@@ -571,4 +575,13 @@ export class WebviewMessages {
   private showPlots() {
     return commands.executeCommand(RegisteredCommands.PLOTS_SHOW, this.dvcRoot)
   }
+
+  private async copyToClipboard(text: string) {
+    await writeToClipboard(text)
```
julieg18, line 580: Should we be copying the full sha instead of the short version? When I see these context options, my first assumption is I'm copying the full sha, but on the other hand, the shortened version is probably all that's needed for most use cases.

julieg18 (approved):
Great work!

extension/src/experiments/webview/messages.ts
mattseddon: The short sha should work for now.

PR #4556, opened 2023-08-22
Attempt to fix setup-node cache issues

mattseddon:


mattseddon:
looks like manually deleting all of the current cache entries helped

Issue #4528, opened 2023-08-18
Experiments table not updating while experiments are running


aschuh-hf:
The metrics in the Experiments table are not following the results of multiple parallel experiment runs even though `dvc exp show` does show the different updated metrics. The Experiments table mainly contains the original base commit metrics.

This issue may be because it seems that the extension does not call `dvc exp show`. Instead, I see lots of `dvc stage list` commands in the "DVC" output window.

```
[version: 1.0.43, 2023-08-18T00:34:42.565Z, pid: 36047] > dvc stage list - INITIALIZED
[version: 1.0.43, 2023-08-18T00:34:43.435Z, pid: 36047] > dvc stage list - COMPLETED (912ms)
[version: 1.0.43, 2023-08-18T00:34:43.473Z, pid: 36050] > dvc stage list - INITIALIZED
[version: 1.0.43, 2023-08-18T00:34:43.487Z, pid: 35221] > dvc data status --granular --unchanged --json - COMPLETED (141986ms)
[version: 1.0.43, 2023-08-18T00:34:47.993Z, pid: 36050] > dvc stage list - COMPLETED (4556ms)
[version: 1.0.43, 2023-08-18T00:34:48.026Z, pid: 36054] > dvc stage list - INITIALIZED
[version: 1.0.43, 2023-08-18T00:34:49.929Z, pid: 36054] > dvc stage list - COMPLETED (1935ms)
[version: 1.0.43, 2023-08-18T00:34:49.981Z, pid: 36073] > dvc stage list - INITIALIZED
[version: 1.0.43, 2023-08-18T00:34:50.667Z, pid: 36073] > dvc stage list - COMPLETED (737ms)
[version: 1.0.43, 2023-08-18T00:34:50.715Z, pid: 36076] > dvc stage list - INITIALIZED
[version: 1.0.43, 2023-08-18T00:34:51.405Z, pid: 36076] > dvc stage list - COMPLETED (736ms)
[version: 1.0.43, 2023-08-18T00:34:51.452Z, pid: 36079] > dvc stage list - INITIALIZED
[version: 1.0.43, 2023-08-18T00:34:52.142Z, pid: 36079] > dvc stage list - COMPLETED (735ms)
[version: 1.0.43, 2023-08-18T00:34:52.196Z, pid: 36082] > dvc stage list - INITIALIZED
[version: 1.0.43, 2023-08-18T00:34:52.857Z, pid: 36082] > dvc stage list - COMPLETED (713ms)
[...]

PR #4564, opened 2023-08-24
Update collection of files which trigger exp show updates

mattseddon:
Related to #4528

We could be missing files we should watch for `exp show` updates because we previously only checked the workspace record for the params and metrics keys. I've updated the collection code to account for all of the records in the `exp show` output.

### DIFF
--- a/extension/src/experiments/data/collect.test.ts
+++ b/extension/src/experiments/data/collect.test.ts
@@ -6,9 +6,13 @@ import { generateTestExpShowOutput } from '../../test/util/experiments'
 
 describe('collectFiles', () => {
   it('should collect all of the available files from the test fixture', () => {
-    expect(collectFiles(expShowFixture, [])).toStrictEqual([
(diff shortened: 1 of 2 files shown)


extension/src/experiments/data/collect.test.ts
```diff
@@ -6,9 +6,13 @@ import { generateTestExpShowOutput } from '../../test/util/experiments'
 
 describe('collectFiles', () => {
   it('should collect all of the available files from the test fixture', () => {
-    expect(collectFiles(expShowFixture, [])).toStrictEqual([
-      'params.yaml',
+    const files = collectFiles(expShowFixture, [])
+    files.sort()
+
+    expect(files).toStrictEqual([
+      'metrics.json',
```
mattseddon, line 13: [F] Wasn't previously captured

julieg18 (approved):

PR #4580, opened 2023-08-28
Call Studio API will all refs available on the Git remote

mattseddon:
This PR is based on https://github.com/iterative/studio/pull/7196#issuecomment-1691023697

### DIFF
--- a/extension/src/experiments/studio.ts
+++ b/extension/src/experiments/studio.ts
@@ -19,7 +19,10 @@ export class Studio extends Disposable {
     return !!this.baseUrl
   }
 
-  public async setBaseUrl(studioToken: string | undefined) {
+  public async setBaseUrl(
+    studioToken: string | undefined,
+    remoteExpRefs: string[]
+  ) {
     if (!studioToken) {
       this.baseUrl = undefined
       return
@@ -30,7 +33,7 @@ export class Studio extends Disposable {
       this.dvcRoot
     )
 
-    this.baseUrl = await this.getBaseUrl(studioToken, gitRemote)
+    this.baseUrl = await this.getBaseUrl(studioToken, gitRemote, remoteExpRefs)
   }
 
   public getLink(sha: string) {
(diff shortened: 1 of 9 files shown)


extension/src/experiments/studio.ts
```diff
@@ -30,7 +33,11 @@ export class Studio extends Disposable {
       this.dvcRoot
     )
 
-    this.baseUrl = await this.getBaseUrl(studioToken, gitRemote)
+    this.baseUrl = await this.getBaseUrl(
+      studioToken,
+      gitRemote,
+      remoteExpRefs.length > 0 ? remoteExpRefs : [null]
```
mattseddon, line 39: [F] Once there is a `GET` endpoint we won't call the `POST` endpoint if there are no new refs.

extension/src/experiments/studio.ts
```diff
@@ -40,12 +47,16 @@ export class Studio extends Disposable {
...
+    studioToken: string,
+    gitRemoteUrl: string,
+    remoteExpRefs: string[] | [null]
+  ) {
     try {
       const response = await fetch(`${STUDIO_URL}/webhook/dvc`, {
         body: JSON.stringify({
           client: 'vscode',
-          refs: [null],
+          refs: remoteExpRefs,
```
skshetry, line 59: @mattseddon, dvc uses `refs` as an object that supports `pushed` and `removed` keys, which are `string[]`.

See https://github.com/iterative/studio/issues/5095#issuecomment-1489777935.

mattseddon: are `pushed`/`removed` optional?

skshetry: yes, both are optional, but any one of it is required to pass the `not refs` checks.

So, you could just pass `pushed: []` instead of `[null]` that would work today until you have a new API.

extension/src/experiments/studio.ts
```diff
@@ -40,12 +43,16 @@ export class Studio extends Disposable {
...
   }
 
-  private async getBaseUrl(studioToken: string, gitRemoteUrl: string) {
+  private async getBaseUrl(
+    studioToken: string,
+    gitRemoteUrl: string,
+    remoteExpRefs: string[]
+  ) {
     try {
       const response = await fetch(`${STUDIO_URL}/webhook/dvc`, {
```
skshetry, line 52: Do you have plans to support `DVC_STUDIO_URL` envvar or `studio.url` config?

julieg18 (approved):

extension/src/experiments/studio.ts
mattseddon: Only when a user specifically asks for that functionality.

skshetry: It’s useful for testing with local Studio servers.

PR #4586, opened 2023-08-29
Add top-level plots wizard

### LINKED ISSUE #3539: Extend/Reuse `Custom Plots` for building `Top-Level` plots
I really like the existing UI for `Custom Plots` and it feels that it could be extended/reused to allow users to build custom [Top-Level](https://dvc.org/doc/command-reference/plots/show#multiple-data-series-plot) plots. 

The current UX for `Top-Level` plots requires manually editing the `dvc.yaml`, which requires understanding the syntax, and sometimes you would want to build the top-level plot on demand for exploration from a UI without having to serialize it in the dvc.yaml (i.e. this is how the MLFlow UI for compare works).

For example, the plots used in the `example-get-started`:

https://github.com/iterative/example-get-started/blob/471811e866f414ffa08c5acd9eed380ec5d165b1/dvc.yaml#L48-L66

Could be built from a context menu by selecting a `name`, `template`, `x`, and `y` sources.

For the `x` and `y` sources, VSCode should display all the existing columns in all existing plot files.

---

It would be nice to have feature to be able to serialize the definition of the generated top-level plot to the `dvc.yaml` file.

julieg18:
# 1/3 `main` <= this <= #4627 <= #4628 

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/e1252676-556c-44e0-bc50-bb263c432fb2

Fixes #3539


extension/src/fileSystem/index.ts
```diff
@@ -199,16 +200,30 @@ stages:
...
 
+export const addPlotToDvcYamlFile = (cwd: string, plotObj: PlotConfigData) => {
+  const dvcYamlFile = `${cwd}/dvc.yaml`
+  const { dataFile, ...plot } = plotObj
+  const plotName = relative(cwd, dataFile)
+  const dvcYamlLines = readFileSync(dvcYamlFile, 'utf8').split('\n')
+  const plotYaml = yaml.stringify({ plots: [{ [plotName]: plot }] }).split('\n')
+
+  // TBD this only works correctly for yaml with 2 space indent and no plots
+  // will adjust for other possibilities in another pr
```
julieg18, line 211: Taken care of in #4628 (still has a couple bugs I need to fix).

extension/src/pipeline/util.ts
```diff
@@ -0,0 +1,103 @@
...
+  dataFile: string
+  template: string
+  x: string
+  y: string
+}
+
+export const pickPlotConfiguration = async (): Promise<
+  PlotConfigData | undefined
+> => {
+  // TBD data file validation will be in next pr
```
julieg18, line 91: Taken care of in #4627.

extension/package.json
```diff
@@ -423,6 +423,12 @@
         "category": "DVC",
         "icon": "$(symbol-class)"
       },
+      {
+        "title": "Add Plot",
```
julieg18, line 427: Maybe "Add Top-Level Plot" to better differentiate between "Add Custom Plot"?

extension/package.json
```diff
@@ -903,6 +909,10 @@
           "command": "dvc.showPipelineDAG",
           "when": "dvc.commands.available && dvc.project.available"
         },
+        {
+          "command": "dvc.addTopLevelPlot",
+          "when": "dvc.commands.available && dvc.project.available"
```
julieg18, line 914: Is currently available in the command pallette.

extension/src/fileSystem/index.ts
mattseddon: [C] Let's not leave comments TODO in the code/merge before fixing obvious/known issue. It would have been enough to add a `do not merge` label to this PR and leave a comment linking to #4628 instead.

extension/src/fileSystem/index.ts
```diff
@@ -199,16 +200,30 @@ stages:
...
+
 export const getFileExtension = (filePath: string) => parse(filePath).ext
 
 export const relativeWithUri = (dvcRoot: string, uri: Uri) =>
   relative(dvcRoot, uri.fsPath)
 
 export const removeDir = (path: string): void => removeSync(path)
 
-export const loadYaml = <T>(path: string): T | undefined => {
+export const loadYamlAsJs = <T>(path: string): T | undefined => {
```
mattseddon, line 224: [Q] Don't all of these functions `loadYamlAsJs`, `loadCsv`, `loadTsv` take a file and convert it into a Javascript object? Do we need the `AsJs` suffix?

extension/src/pipeline/index.ts
```diff
@@ -107,6 +112,22 @@ export class Pipeline extends DeferredDisposable {
     return this.addPipeline()
   }
 
+  public async addTopLevelPlot() {
+    const cwd = await this.getCwd()
```
mattseddon, line 116: [Q] What does the workflow look like if there is no `dvc.yaml` available in the project? I.e. I am in "dvclive-only" mode (e.g. [dvclive-exp-tracking](https://github.com/iterative/dvclive-exp-tracking)) and I try to add a plot... what happens? This will add a root level `dvc.yaml` but will that work unless the train script is added as well? Not a blocker but need to check/update if it is broken.

extension/src/pipeline/util.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    expect(mockedPickFile).toHaveBeenCalledWith(Title.SELECT_PLOT_DATA, {
+      filters: {
+        'Data Formats': ['json', 'csv', 'tsv', 'yaml']
+      },
+      openLabel: 'Select'
+    })
+  })
+
+  it('should parse chosen data file', async () => {
```
mattseddon, line 68: [N] `the chosen`

extension/src/pipeline/util.test.ts
```diff
@@ -0,0 +1,237 @@
...
+    await pickPlotConfiguration()
+
+    expect(mockedLoadTsv).toHaveBeenCalledWith('file.tsv')
+
+    await pickPlotConfiguration()
+
+    expect(mockedLoadYaml).toHaveBeenCalledWith('file.yaml')
+  })
+
+  it('should let user pick a template, x field, and y field', async () => {
```
mattseddon, line 97: [N] `the user` => check all the `it` statements

extension/src/pipeline/util.test.ts
```diff
@@ -0,0 +1,237 @@
...
+        'bar_horizontal'
+      ],
+      'Pick a Plot Template'
+    )
+    expect(mockedQuickPickOne).toHaveBeenNthCalledWith(
+      2,
+      ['actual', 'prob'],
+      'Pick a Metric for X'
+    )
+    expect(mockedQuickPickOne).toHaveBeenNthCalledWith(
```
mattseddon, line 230: [Q] Do we need these `toHaveBeenNthCalledWith` assertions in the early return tests? Could we just have a `toHaveBeenCalledTimes` and avoid the duplication?

extension/src/pipeline/util.ts
```diff
@@ -0,0 +1,103 @@
...
+  loadYamlAsJs
+} from '../fileSystem'
+import { quickPickOne } from '../vscode/quickPick'
+import { pickFile } from '../vscode/resourcePicker'
+import { Title } from '../vscode/title'
+
+const parseDataFile = (file: string) => {
+  const ext = getFileExtension(file)
+
+  if (ext === '.json') {
```
mattseddon, line 15: [Q] Would a switch statement be appropriate here?

extension/src/pipeline/util.ts
```diff
@@ -0,0 +1,103 @@
...
+    [
+      'simple',
+      'linear',
+      'confusion',
+      'confusion_normalized',
+      'scatter',
+      'scatter_jitter',
+      'smooth',
+      'bar_horizontal_sorted',
+      'bar_horizontal'
```
mattseddon, line 54: Let's move this array into `extension/src/cli/dvc/contract.ts` as a constant. Gives us a better chance of catching the need to make updates if/when the list changes in `dvc-render`

extension/src/pipeline/workspace.ts
```diff
@@ -63,6 +63,16 @@ export class WorkspacePipeline extends BaseWorkspace<Pipeline> {
     )
   }
 
+  public async addTopLevelPlot() {
+    const cwd = this.focusedProject || (await this.getOnlyOrPickProject())
```
mattseddon, line 67: [N] Could be a private method now it is used in `showDag` and here

extension/src/pipeline/util.ts
```diff
@@ -0,0 +1,103 @@
+import {
```
mattseddon, line 1: [F] For other parts of the codebase I've grouped `quickPick` based functions into their own file (avoiding generic `util` file where possible). E.g. 

- `extension/src/experiments/model/quickPick.ts`
- `extension/src/path/selection/quickPick.ts`
- `extension/src/setup/quickPick.ts`

IMO `parseDataFile` is more of a generic function that could be used elsewhere. I'd probably move that to `extension/src/fileSystem/index.ts` so that other people can find it.

extension/src/pipeline/util.ts
```diff
@@ -0,0 +1,103 @@
...
+    return loadYamlAsJs<Record<string, unknown>>(file)
+  }
+}
+
+const pickDataFile = () => {
+  return pickFile(Title.SELECT_PLOT_DATA, {
+    filters: {
+      'Data Formats': ['json', 'csv', 'tsv', 'yaml']
+    },
+    openLabel: 'Select'
```
mattseddon, line 37: [Q] Should this `openLabel` option be set to "Select" whenever we use `pickFile`?

mattseddon (approved):
👍🏻

extension/package.json
mattseddon: This is ok for now.

extension/src/pipeline/util.ts
mattseddon: I'd always recommend moving functions into a standardized file structure across the project because then we should have a common understanding of where to find things. Not only when we are debugging how the code works but also when we are looking to reuse functions/old pieces of functionality.

extension/src/pipeline/index.ts
julieg18: It automatically runs the "create a stage" workflow instead. While technically works, we might want to add a toast with an action to run the "create a stage" workflow instead 🤔.

extension/src/pipeline/util.test.ts
julieg18: Noticing a test description habit I need to break 🤣

extension/src/fileSystem/index.test.ts
```diff
@@ -436,6 +518,38 @@ describe('findOrCreateDvcYamlFile', () => {
...
+  it('should add plots item with created plot if dvc.yaml file has no plots', () => {
+    const mockDvcYamlContent = [
+      'stages:',
+      '  train:',
+      '    cmd: python train.py'
+    ].join('\n')
+    const mockPlotYamlContent = [
+      '',
+      'plots:',
+      `  - ${relative('/', 'data.json')}:`,
```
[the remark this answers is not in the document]
julieg18: Removed!

extension/src/fileSystem/index.ts
julieg18: Makes sense! Removed the unneeded comments.

extension/src/pipeline/util.ts
julieg18: Makes sense, thanks! Created a `quickPick.ts` instead of `util.ts` and moved the `parseDataFile` to `fileSystem`!

PR #4607, opened 2023-08-31
Fix plots paths breaking on windows

julieg18:
* standardizes paths in plots output before plot models process it

## Bug

https://github.com/iterative/vscode-dvc/assets/43496356/adef843f-49e5-44ee-8508-6f45f29cb542

Fixes issue mentioned in [slack thread](https://iterativeai.slack.com/archives/C01RB7BTG4C/p1693507752341379)


extension/src/fileSystem/util.ts
```diff
@@ -1,6 +1,7 @@
 import { sep, parse } from 'path'
 
-export const getPathArray = (path: string): string[] => path.split(sep)
+export const getPathArray = (path: string): string[] =>
+  path.replace(/\\/g, '/').split('/')
```
julieg18, line 4: On further inspection, while this fixes image by step plots, it breaks regular image plots on windows due to the manually joined paths (using `sep`) are different from the paths inside the plots data (using `/` or `\`). I think adjusting how our `getType` works should fix the issue. 

Will get back to this tomorrow.

julieg18: Ended up adjusting the output data as a whole to make sure that plots data and path data matched up correctly.

mattseddon: I think it would be safer to leave this function (revert the behaviour) and more `path.replace(/\\/g, '/').split('/')` into `standardisePath` only.

The function also needs a different name to help differentiate it from `standardizePath` here: https://github.com/iterative/vscode-dvc/blob/4cdac53936bf4555996d901f5fc56cf132e4afd9/extension/src/fileSystem/path.ts#L3

mattseddon: [Q] How about `ensureOsFileSep` or something similar?

extension/src/plots/util.test.ts
```diff
@@ -0,0 +1,58 @@
+import { sep } from 'path'
+import { standardisePlotsDataPaths } from './util'
+import { PlotsType } from './webview/contract'
+import { FIELD_SEPARATOR } from '../cli/dvc/constants'
+import { PlotsOutput } from '../cli/dvc/contract'
+
+const join = (pathArr: string[], slash = sep) => pathArr.join(slash)
```
mattseddon, line 7: [N] Naming this `join` threw me off, I thought `join(['plots', 'heatmap.png'], slash)` was using some magical undocumented path for `path.join`

mattseddon (approved):

PR #4619, opened 2023-09-04
Patch Studio API update timinig issue

mattseddon:
Related to https://github.com/iterative/vscode-dvc/issues/3574

After pushing an experiment there is a delay before it becomes available via the new Studio GET endpoint. 

There were two possible solutions that came to mind.

1. Continually ping the API until the experiment becomes available.
2. Assume that if the experiment is on the Git remote it will be available in Studio by the time the link icon is shown.

I went with option 2 as option 1 seemed to be more brittle.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/366f4807-af28-4132-baa4-e9f6ad49d54c

Note: We can update this later if we run into issues. The code is clearly signposted.

### DIFF
--- a/extension/src/experiments/index.ts
+++ b/extension/src/experiments/index.ts
@@ -10,6 +10,7 @@ import omit from 'lodash.omit'
 import { ColumnLike, addStarredToColumns } from './columns/like'
 import { setContextForEditorTitleIcons } from './context'
 import { ExperimentsModel } from './model'
+import { collectRemoteExpShas } from './model/collect'
 import {
   pickExperiment,
   pickExperiments,
@@ -229,9 +230,10 @@ export class Experiments extends BaseRepository<TableData> {
   }
 
   public async unsetPushing(ids: string[]) {
-    await this.update()
+    await Promise.all([this.update(), this.patchStudioApiTimingIssue(ids)])
+
     this.experiments.unsetPushing(ids)
-    return this.notifyChanged()
+    return this.webviewMessages.sendWebviewMessage()
   }
 
   public hasCheckpoints() {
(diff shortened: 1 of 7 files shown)


extension/src/experiments/index.ts
```diff
@@ -759,4 +760,36 @@ export class Experiments extends BaseRepository<TableData> {
     const columnLikes = addStarredToColumns(columns)
     return pickColumnToFilter(columnLikes)
   }
+
+  private async patchStudioApiTimingIssue(ids: string[]) {
+    if (!this.studio.isConnected()) {
+      return
+    }
+
+    await this.waitForStudioUpdate()
```
mattseddon, line 769: [F] If we don't wait for this the `assumePushed` ids are overwritten.

extension/src/experiments/index.ts
```diff
@@ -759,4 +761,56 @@ export class Experiments extends BaseRepository<TableData> {
...
+    for (const { id, sha } of this.experiments.getExperiments()) {
+      if (ids.includes(id) && sha && remoteExpShas.has(sha)) {
+        shas.push(sha)
+      }
+    }
+
+    this.experiments.assumePushed(shas)
+  }
+
+  private waitForStudioUpdate() {
```
mattseddon, line 787: [F] I tried to combine the `waitForStudioUpdate` & `waitForRemoteUpdate` methods but the code was a lot harder to read.

extension/src/experiments/index.ts
```diff
@@ -759,4 +761,56 @@ export class Experiments extends BaseRepository<TableData> {
...
     return pickColumnToFilter(columnLikes)
   }
+
+  private async patchStudioApiTimingIssue(ids: string[]) {
+    if (!this.studio.isConnected()) {
+      return
+    }
+
+    const [, { lsRemoteOutput }] = await Promise.all([
+      this.waitForStudioUpdate(),
```
mattseddon, line 771: [F] If we don't wait for this the `assumePushed` shas can be overwritten.

julieg18 (approved):
Looks good!

PR #4621, opened 2023-09-05
fix: show current max depth value instead of 0

yathomasi:
<table>
<tr>
 <th> Current
 <th> PR
<tr>
 <td> 

https://github.com/iterative/vscode-dvc/assets/20840228/d19f93c9-118a-4e15-943f-c0dc2ecbfc3e

 <td>

https://github.com/iterative/vscode-dvc/assets/20840228/7acfa69f-a691-4f43-b5ba-9a921b25c633

</table>

### DIFF
--- a/extension/src/experiments/webview/messages.ts
+++ b/extension/src/experiments/webview/messages.ts
@@ -19,7 +19,7 @@ import { ExperimentsModel } from '../model'
 import { SortDefinition } from '../model/sortBy'
 import { getPositiveIntegerInput } from '../../vscode/inputBox'
 import { Title } from '../../vscode/title'
-import { ConfigKey, setConfigValue } from '../../vscode/config'
+import { ConfigKey, getConfigValue, setConfigValue } from '../../vscode/config'
 import { NUM_OF_COMMITS_TO_INCREASE } from '../../cli/dvc/constants'
 import { Pipeline } from '../../pipeline'
 import { collectColumnsWithChangedValues } from '../columns/collect'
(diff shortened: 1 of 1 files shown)


extension/src/experiments/webview/messages.ts
```diff
@@ -376,9 +376,10 @@ export class WebviewMessages {
   }
 
   private async setMaxTableHeadDepth() {
+    const currentValue = getConfigValue(ConfigKey.EXP_TABLE_HEAD_MAX_HEIGHT)
     const newValue = await getPositiveIntegerInput(
       Title.SET_EXPERIMENTS_HEADER_HEIGHT,
-      { prompt: 'Use 0 for infinite height.', value: '0' },
+      { prompt: 'Use 0 for infinite height.', value: currentValue },
```
mattseddon, line 382: At the start of the "PR demo" there is no config option set but "5" is shown for the value. How did that happen? Should this be `currentValue ?? '0'` or similar?

mattseddon: Oh, the default is set in `package.json`

```
        "dvc.experimentsTableHeadMaxHeight": {
          "description": "Maximum height of Experiment table head rows. Use 0 for infinite height.",
          "type": "number",
          "default": 5
        }
```

yathomasi: Actually, I kind of assumed `5` must have been set as the default value somewhere as it's the same in Studio. But, I didn't know where, so thanks for sharing 😄.

PR #4626, opened 2023-09-06
Fix custom plots breaking with non-numeric values

### LINKED ISSUE #4625: `Custom plots`: Fail to render when selected parameter is not numeric
I created a custom plot by selecting a parameter containing different string values. The plot is empty:

Screenshot 2023-09-06 130254

However, I can export with `Save as JSON` and see the expected values:

```json
[
    {
        "id": "main",
        "metric": 0.9340795388735517,
        "param": "shufflenet_v2_x2_0"
    },
    {
        "id": "309c133",
        "metric": 0.8923946870030989,
        "param": "shufflenet_v2_x2_0"
    },
    {
        "id": "rimed-acre",
        "metric": 0.9288800744178878,
        "param": "squeezenet1_1"
    },
    {
        "id": "cadgy-loup",
        "metric": 0.9355338898522014,
        "param": "resnet34"
    },
    {
        "id": "tawie-moxa",
        "metric": 0.9305068735503278,
        "param": "alexnet"
    },
    {
        "id": "parky-many",
        "metric": 0.9340795388735517,
        "param": "shufflenet_v2_x2_0"
    }
]
```

Maybe we could create an index of unique values and use it as the horizontal axis while displaying the associated string as label?

julieg18:
## `main`

image

## PR

image

~Looks buggy due to the length of the param string, but that would not be a quick fix since I'd need to do some extra research on what we could do.~

Fixes #4625

### DIFF
--- a/extension/src/plots/model/custom.ts
+++ b/extension/src/plots/model/custom.ts
@@ -69,7 +69,16 @@ export const checkForCustomPlotOptions = (
   return false
 }
 
-export const createSpec = (title: string, metric: string, param: string) =>
+const getSpecDataType = (type: string) =>
+  type === 'number' ? 'quantitative' : 'nominal'
+
+export const createSpec = (
+  title: string,
+  metric: string,
+  param: string,
+  metricType: string,
+  paramType: string
+) =>
   ({
     $schema: 'https://vega.github.io/schema/vega-lite/v5.json',
     data: { name: 'values' },
@@ -80,7 +89,7 @@ export const createSpec = (title: string, metric: string, param: string) =>
           zero: false
         },
         title: param,
-        type: 'quantitative'
+        type: getSpecDataType(paramType)
       },
       y: {
         field: 'metric',
@@ -88,7 +97,7 @@ export const createSpec = (title: string, metric: string, param: string) =>
           zero: false
         },
         title,
-        type: 'quantitative'
+        type: getSpecDataType(metricType)
       }
     },
     height: 'container',
--- a/extension/src/plots/model/custom.ts
+++ b/extension/src/plots/model/custom.ts
@@ -84,6 +84,10 @@ export const createSpec = (
     data: { name: 'values' },
     encoding: {
       x: {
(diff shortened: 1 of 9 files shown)


extension/src/plots/model/custom.ts
```diff
@@ -69,7 +69,16 @@ export const checkForCustomPlotOptions = (
   return false
 }
 
-export const createSpec = (title: string, metric: string, param: string) =>
+const getSpecDataType = (type: string) =>
+  type === 'number' ? 'quantitative' : 'nominal'
```
julieg18, line 73: When custom plots were first created, I had them set to `nominal` by default due to the fact that none-number values break on `quantitive`. Somewhere down the line, it must have changed back to `quantitive` 🤔.  I think it was because only `quantitive` orders the number values correctly.

extension/src/test/fixtures/expShow/base/customPlots.ts
```diff
@@ -9,7 +9,7 @@ import { Experiment } from '../../../../experiments/webview/contract'
 export const customPlotsOrderFixture: CustomPlotsOrderValue[] = [
   {
     metric: 'summary.json:loss',
-    param: 'params.yaml:dropout'
+    param: 'params.yaml:log_file'
```
julieg18, line 12: I updated the fixture to include a string param for testing but this did lead me to touching practically every custom plots test, despite this being a rather small spec change 😅

mattseddon (approved):
Is it possible to change the angle of this axis label or truncate it?:

image

mattseddon:
Take a look here: https://vega.github.io/vega-lite/docs/axis.html#labels

julieg18:
> Is it possible to change the angle of this axis label or truncate it?:
> Take a look here: https://vega.github.io/vega-lite/docs/axis.html#labels

Thanks, I now see I can adjust label and title padding, fixing the overlap issue! I'll go ahead and add it here instead of following up like originally planned.

PR #4627, opened 2023-09-06
Add data file validation to plots wizard

julieg18:
# 2/3 `main` <= #4586 <= this <= #4628

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/48244e71-0b2d-4fb3-b457-59c424fc6714~

https://github.com/iterative/vscode-dvc/assets/43496356/f5f1c032-87e1-4322-85ec-4be01e289beb

Part of https://github.com/iterative/vscode-dvc/issues/3539


extension/src/pipeline/util.test.ts
```diff
@@ -94,6 +98,79 @@ describe('pickPlotConfiguration', () => {
...
+
+    const result = await pickPlotConfiguration()
+
+    expect(result).toStrictEqual(undefined)
+  })
+
+  const failedToParseMessage =
+    'Failed to find field options for plot data. Is your file following DVC plot guidelines for [JSON/YAML](https://dvc.org/doc/command-reference/plots/show#example-hierarchical-data) or [CSV/TSV](https://dvc.org/doc/command-reference/plots/show#example-tabular-data) files?'
+
+  it('should show a toast message if file fails to parse', async () => {
```
mattseddon, line 112: [N] `the file` or `a file` => check these `it` statements 😬

extension/src/pipeline/util.test.ts
```diff
@@ -94,6 +98,79 @@ describe('pickPlotConfiguration', () => {
...
+      ],
+      { field1: [{ field1: 1, field2: 2 }] }
+    ]
+
+    for (const [ind, val] of invalidValues.entries()) {
+      mockedLoadYaml.mockReturnValueOnce(val)
+
+      await pickPlotConfiguration()
+
+      expect(mockedShowError).not.toHaveBeenCalledTimes(1)
```
mattseddon, line 169: [N] `expect(mockedShowError).not.toHaveBeenCalled()` (if it was called twice this would pass)

mattseddon: [I] ... wait, I am confused. `invalidValues` is a very misleading const name. Should be `validValues`. Right?

mattseddon: [Q] Will the file be parsed if the records (lines) have different keys? Can we add a test for that?

extension/src/pipeline/util.ts
```diff
@@ -85,13 +87,45 @@ export type PlotConfigData = {
...
-  const keys = Object.keys(data[0])
+  const file = await pickDataFile()
+
+  if (!file) {
+    return
+  }
+
+  const data = await parseDataFile(file)
+  const failedToParseMessage =
+    'Failed to find field options for plot data. Is your file following DVC plot guidelines for [JSON/YAML](https://dvc.org/doc/command-reference/plots/show#example-hierarchical-data) or [CSV/TSV](https://dvc.org/doc/command-reference/plots/show#example-tabular-data) files?'
```
mattseddon, line 118: [I] I would split this and show a different message when parsing throws an error/returns nothing vs the data has < 2 keys. On first read `field options for plot data` is ambiguous/needs more context. 

As an end goal we would want to tell the user:

1. If the file did not parse successfully
2. If the file was empty
3. There are not enough keys to generate a plot

As splitting 1/2 out would require a large refactor I'd suggest breaking out the message to 1+2 or 3.

E.g. 

`'Failed to parse the requested file. Does the file contain data and follow the DVC plot guidelines for [JSON/YAML](https://dvc.org/doc/command-reference/plots/show#example-hierarchical-data) or [CSV/TSV](https://dvc.org/doc/command-reference/plots/show#example-tabular-data) files?'`
and 
`'The request file does not contain enough keys (columns) to generate a plot. Does the file follow the DVC plot guidelines for [JSON/YAML](https://dvc.org/doc/command-reference/plots/show#example-hierarchical-data) or [CSV/TSV](https://dvc.org/doc/command-reference/plots/show#example-tabular-data) files?'`

extension/src/util/object.ts
```diff
@@ -11,3 +11,6 @@ export const createTypedAccumulator = <T extends string>(
   }
   return acc
 }
+
+export const isObject = (value: unknown): value is Record<string, unknown> =>
+  !!(value && !Array.isArray(value) && typeof value === 'object')
```
mattseddon, line 16: [F] This looks like a duplicate of `isValueTree`. 

[Q] Not massively important but can we reuse this there?

i.e. 
```
export const isValueTree = (
  value: Value | ValueTree
): value is NonNullable<ValueTree> =>
  isObject(value)
```

mattseddon (approved):

extension/src/pipeline/util.test.ts
julieg18: > [I] ... wait, I am confused. invalidValues is a very misleading const name. Should be validValues. Right?

Yes, invalidValues is a typo, whoops!

> [Q] Will the file be parsed if the records (lines) have different keys? Can we add a test for that?

File will pass anyway since I was trying to keep file validation on the simpler side. We can go ahead and check for values throughout the array though!

extension/src/pipeline/util.ts
julieg18: Good idea! Update the messaging!

extension/src/util/object.ts
julieg18: Reused `isValueTree`

PR #4628, opened 2023-09-06
Correctly parse dvc.yaml on plot creation

julieg18:
# 3/3 `main` <= #4586 <= #4627 <= this

https://github.com/iterative/vscode-dvc/assets/43496356/24ae3040-465c-4522-b78b-90d549ac3bff

Part of #3539


extension/src/fileSystem/index.ts
```diff
@@ -200,14 +200,60 @@ stages:
...
+  } catch {
+    Logger.error(`failed to load yaml ${path}`)
+  }
+}
+
+const getYamlFileIndent = (lines: string[]) => {
+  const firstLineWithInd = lines.find(line => line.startsWith(' ')) || ''
+  const spacesMatches = firstLineWithInd.match(/^( +)[^ ]/)
+  const spaces = spacesMatches?.[1]
+
```
julieg18, line 221: This appeared to be the simplest way to get the indent length.

extension/src/fileSystem/index.ts
```diff
@@ -200,14 +200,60 @@ stages:
...
+  }
+
+  const { doc, lineCounter } = dvcYamlDoc
   const { dataFile, ...plot } = plotObj
   const plotName = relative(cwd, dataFile)
   const dvcYamlLines = readFileSync(dvcYamlFile, 'utf8').split('\n')
-  const plotYaml = yaml.stringify({ plots: [{ [plotName]: plot }] }).split('\n')
+  const indent = getYamlFileIndent(dvcYamlLines)
+
+  const plots = doc.get('plots', true) as yaml.YAMLSeq | undefined
```
julieg18, line 239: We use the `yaml` package's [Document API](https://eemeli.org/yaml/#documents) to check if the plots exist and get the last position of the plots item for inserting lines into the yaml file.

mattseddon: As a follow up can you PTAL at `vscode`'s `WorkspaceEdit` API (should be able to use `insert`). We don't have to use it but it could simplify things.

extension/src/fileSystem/index.ts
mattseddon: Potentially different sections could have different indentations. Can we match the plots section as a priority (if it exists)

mattseddon (approved):

extension/src/fileSystem/index.ts
julieg18: Updated the code to search the plots section for indent spaces if plots exists!

PR #4629, opened 2023-09-07
Add ability to zoom and pan plots when zoomed

### LINKED ISSUE #4530: Plots: Ability to zoom and maybe pan
Is it possible to zoom the metrics plots within a selected local rectangular window as in TensorBoard? I couldn't figure out how if. Selecting a rectangular zoom window and zooming out again fully via a simple interaction (e.g., double click as in TensorBoard) would be quite useful. Also being able to then pan that window would be nice (SHIFT + left click drag in TensorBoard).

mattseddon:
Closes #4530
Related to https://github.com/iterative/dvc-render/issues/7

This PR adds the ability to zoom (using the mouse wheel) and pan plots when they are focused. I have limited the change to only focused because it mitigates the unexpected behaviour of the plot being zoomed when the user is trying to scroll between plots.

Types covered by this change?:

1. Scatter plots
2. Smooth plots
3. Simple plots
4. Horizontal bar plots
5. Custom plots

What's not covered:

- Confusion matrices (and probably other multiview plots)

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/63dc7c1f-4585-4fa7-bda0-f99a58105200

Note: has the potential to break custom templates.

### DIFF
--- a/webview/src/plots/components/ZoomedInPlot.tsx
+++ b/webview/src/plots/components/ZoomedInPlot.tsx
@@ -3,7 +3,10 @@ import VegaLite, { VegaLiteProps } from 'react-vega/lib/VegaLite'
 import { Config } from 'vega-lite'
 import merge from 'lodash.merge'
 import cloneDeep from 'lodash.clonedeep'
-import { reverseOfLegendSuppressionUpdate } from 'dvc/src/plots/vega/util'
+import {
+  makePlotZoomOnWheel,
+  reverseOfLegendSuppressionUpdate
+} from 'dvc/src/plots/vega/util'
 import { TemplateVegaLite } from './templatePlots/TemplateVegaLite'
 import styles from './styles.module.scss'
 import { getThemeValue, ThemeProperty } from '../../util/styles'
@@ -40,6 +43,12 @@ export const ZoomedInPlot: React.FC<ZoomedInPlotProps> = ({
   isTemplatePlot,
   openActionsMenu
 }: ZoomedInPlotProps) => {
+  const isCustomPlot = !isTemplatePlot
+  const hasSmoothing =
+    isTemplatePlot &&
+    (props.spec as { params?: { name: string }[] }).params?.[0]?.name ===
+      'smooth'
+
   const zoomedInPlotRef = useRef<HTMLDivElement>(null)
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


webview/src/plots/components/ZoomedInPlot.tsx
```diff
@@ -40,6 +43,12 @@ export const ZoomedInPlot: React.FC<ZoomedInPlotProps> = ({
   isTemplatePlot,
   openActionsMenu
 }: ZoomedInPlotProps) => {
+  const isCustomPlot = !isTemplatePlot
+  const hasSmoothing =
+    isTemplatePlot &&
+    (props.spec as { params?: { name: string }[] }).params?.[0]?.name ===
+      'smooth'
```
mattseddon, line 50: [C] This is pretty brittle but will do for now. Will discuss at the next cross-team whether or not we want to push this back in to `dvc-render`.

mattseddon: worst case scenario - if the spec changes then `makePlotZoomOnWheel` will fall through to the default. The zoom/pan behaviour won't work but the template will still render.

julieg18 (approved):
Great work!

shcheklein:
Looks great! Let go for it (and can we add it to Studio?)

QQ - are there other ways to control it? A square box selection? (can be then easier to enable on the main dashboard?)

mattseddon:
> Looks great! Let go for it (and can we add it to Studio?)

Do you want me to add this to Studio?

> QQ - are there other ways to control it? A square box selection? (can be then easier to enable on the main dashboard?)

I have been through the docs and this seems like the only option without trying to directly access d3 (like tensorboard does) so there is not a simple way to "do the same thing as tensorboard". Brush zoom looks like [this](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoAWOgCtY3AHaQANOCgATdHkw5gkVgCcx2LTvR9CAVgAMtuvFgVIAX3URIFeMvg6zANoaEKAQYVAA7oyaLGbSABy2HuFQGEYA1maQ6EbImMnhkMhKPpqMStRmoSmeAB5VwTVQAGaMyGKaWdp4yGqNNZB4AJ6EvTiDaITcRuimBU2QCLNjYAaa3BjlVYvtyPB4PCrj7EassIhqUMWl5ZXj9a7u-SnZtYyw2wd4YiuQbs8QJ5NKBDBrAzytdqdcaEIyMeC9eYDYajLIAR1Y6CUXx0jCofXBUCWP22602R1WOx++0OWROZwuqiuJW4ZQqWVBjwBYFcz154SBYWqhSiMQuOASSWekBQjGo+DMADYpQM0plxjk8gSXrBdjTlNt6edPiNfuUekYKLNLkUWWzqB8cAFIPUmZBQQBdR5I5k3dn6bku7aQjpdHSIqAo349VBTGamQUDUEBwmQEPQqCw+GI7meKPozHY5i4-E+wroN6OykHeDpADC3FY2LMAGY3dQ4RnmrNdfymn2BY0+x6QK4gA) or [this](https://vega.github.io/vega-lite-v4/docs/zoom.html) (note v4) as usual there is an open issue https://github.com/vega/vega-lite/issues/4742.

Issue #4620, opened 2023-09-05
Apply sorting to the entire table


Danila89:
According to this [message](https://discord.com/channels/485586884165107732/1110155971482886195/1111364895116632195) sorting and filtering in vscode extension is only applied to the experiments under commit.
I'm using the extension as a leaderboard and it would be pretty convenient to be able to sort/filter the entire table.
<img width="750" alt="image" src="https://github.com/iterative/vscode-dvc/assets/18012372/54bad9e3-2946-4bc4-b7b5-4459ef5471f9">

julieg18:
Filters were applied to commits in https://github.com/iterative/vscode-dvc/pull/4363, though sorting is currently only applied to experiments under commits.

julieg18:
> I'm using the extension as a leaderboard and it would be pretty convenient to be able to sort/filter the entire table.

@Danila89, do custom plots work for your use case instead of table sorting? You can use them to compare a metric/param across all commits and experiments:

https://github.com/iterative/vscode-dvc/assets/43496356/ee65e4ac-292f-48f3-9328-87bca4679180

Danila89:
These plots are really helpful, but it is essential for me to use some filtering of the experiments before plotting them. I didn't find the way to filter experiments before plotting them in the vscode extension.
Plots are nice in addition to sorting, not instead of it)

julieg18:
> These plots are really helpful, but it is essential for me to use some filtering of the experiments before plotting them. I didn't find the way to filter experiments before plotting them in the vscode extension.

Filtering experiments in custom plots would be simple enough to implement! Would that enhancement work for your use case or would table sorting still be needed?

Danila89:
It is OK like a workaround (and a very useful feature by the way), but table sorting is also nice to have

julieg18:
Did some research for table sorting and tried a couple ideas.

### 1. Flattening then sorting the table

By flattening the table and turning off nested rows entirely, we could sort all rows from biggest to smallest and vice versa. While I don't think this would be too hard to implement on the backend, we'd need to rethink how our table would look while flattened (how do we tell what row belongs to what commit/branch). This would require some design work and adjusting our table components to handle the possibility. 

### 2. Sorting commits along with experiments

[...]

PR #4647, opened 2023-09-08
Apply selected experiment colors to custom plots

### LINKED ISSUE #4624: `Custom Plots`: honor selected revisions and colors.
When adding a `Custom Plot`, _all_ the experiments in the table are included in the plot, but there is no signal of which revision is which (except for the tooltip).

Screenshot 2023-09-06 130013

I was expecting the `Custom Plot` to be linked to the selected revisions and to include color and legend, matching the behavior of `Data Series` plots:

Screenshot 2023-09-06 130003

mattseddon:
Closes #4624 

This PR applies the colors of selected revision to custom plots.  The legend is suppressed unless the plots is zoomed (as per the normal template plot behaviour).

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/e4978a7b-b706-4bfc-bc3a-4bb6c57c3f64

### DIFF
--- a/extension/src/plots/model/collect.ts
+++ b/extension/src/plots/model/collect.ts
@@ -42,6 +42,7 @@ import {
   getParent,
   getPathArray
 } from '../../fileSystem/util'
+import { Color } from '../../experiments/model/status/colors'
 
 export const getCustomPlotId = (metric: string, param: string) =>
   `custom-${metric}-${param}`
@@ -52,7 +53,8 @@ const getValueFromColumn = (path: string, experiment: Experiment) =>
 const getValues = (
   experiments: Experiment[],
   metricPath: string,
-  paramPath: string
+  paramPath: string,
+  renderLastIds: Set<string> = new Set<string>()
 ): CustomPlotValues => {
   const values: CustomPlotValues = []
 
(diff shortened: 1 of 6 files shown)


extension/src/plots/model/collect.ts
```diff
@@ -104,21 +115,54 @@ const getCustomPlotData = (
...
 }: {
+  colorScale: ColorScale | undefined
   plotsOrderValues: CustomPlotsOrderValue[]
   experiments: Experiment[]
   height: number
   nbItemsPerRow: number
 }): CustomPlotData[] => {
   const plots = []
 
+  const completeColorScale = fillColorScale(colorScale, experiments)
```
mattseddon, line 153: [F] If we want to respect (only show) selected revision then we just need to drop this function.

extension/src/plots/model/collect.ts
```diff
@@ -52,21 +53,28 @@ const getValueFromColumn = (path: string, experiment: Experiment) =>
...
+      continue
     }
+
+    const value = {
+      id: experiment.id,
+      metric: metricValue,
+      param: paramValue
+    }
+
+    renderLastIds.has(experiment.id)
```
mattseddon, line 75: [F] We want the selected revisions to be rendered last so they show up.

extension/src/plots/model/collect.ts
daavoo: No strong opinion but could we default to show only selected and have some sort of toggle to "show all"?

julieg18 (approved):
Great work!

extension/src/plots/model/collect.ts
mattseddon: That's a bigger change. Will come back to it after we make some changes for #4620.

PR #4657, opened 2023-09-11
Apply table filters to custom plot data

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/bb76fb01-414c-40bc-89a0-cbadc39f14b5

Related to #4620


extension/src/test/suite/plots/index.test.ts
```diff
@@ -172,6 +179,42 @@ suite('Plots Test Suite', () => {
...
+        .onFirstCall()
+        .returns([
+          {
+            operator: Operator.EQUAL,
+            path: 'params:params.yaml:epochs',
+            value: 2
+          }
+        ])
+
+      plotsModel.getCustomPlots()
```
julieg18, line 202: Is there a better way to test for this?

extension/src/experiments/model/index.ts
```diff
@@ -365,6 +365,14 @@ export class ExperimentsModel extends ModelWithPersistence {
     return [...this.getWorkspaceAndCommits(), ...this.getExperiments()]
   }
 
+  public getUnfilteredCommitsAndExperiments() {
```
julieg18, line 368: I didn't see a function that already collected all unfiltered experiments/commits but I might have missed something.

extension/src/plots/model/index.ts
```diff
@@ -154,9 +154,7 @@ export class PlotsModel extends ModelWithPersistence {
   }
 
   public getCustomPlots(): CustomPlotsData | undefined {
-    const experiments = this.experiments
-      .getWorkspaceCommitsAndExperiments()
-      .filter(({ id }) => id !== EXPERIMENT_WORKSPACE_ID)
+    const experiments = this.experiments.getUnfilteredCommitsAndExperiments()
```
mattseddon, line 157: [I] Need to handle the case where there are no unfiltered experiments left. IIRC we will currently show an empty plot.

extension/src/experiments/model/index.ts
mattseddon: This will give slightly different results to what is shown in the table as we have to show commits that have > 0 unfiltered experiments

extension/src/test/suite/plots/index.test.ts
mattseddon: this is fine

mattseddon (approved):
I think we need to add/update the empty screen for custom plots before we merge this.

extension/src/experiments/model/index.ts
julieg18: Yes, I figured we didn't need to show those commits since the plots data isn't hierarchical like the table. But if we want it to reflect the table exactly, I can update the function?

extension/src/plots/model/index.ts
julieg18: If I understand your statement correctly, when I tested for this possibility I got the "No Plots To Display" message which I thought worked fine. 

image

Though we could extend the description for more clarity 🤔. For example, "No Plots to Display. Are any of your experiments unfiltered?"

julieg18: Updated the description:

image

extension/src/experiments/model/index.ts
mattseddon: The current implementation is fine.

extension/src/plots/model/index.ts
mattseddon: There should be a different state depending on whether or not there are custom plots created (there will always be data unless all of the experiments are filtered).

mattseddon: I.e no plots created is different from no unfiltered data

mattseddon:
merging `main` to test code climate

extension/src/plots/model/index.ts
julieg18: Right now we have two states, "no data" (No Plots to Display. Are any...) and "no plots created"(No Plots Added). But you're saying we should introduce a third separate state for when there are no unfiltered experiments? Would be easy enough :)

mattseddon: What I am trying to avoid is this being brittle/getting forgotten about when we make changes because we have had issues with stale empty states previously.

```
  if (!hasData) {
    return <EmptyState isFullScreen={false}>No Plots to Display</EmptyState>
  }

  if (order.length === 0) {
    return <EmptyState isFullScreen={false}>No Plots Added</EmptyState>
  }
```

Previously `!hasData` was a placeholder because we could never hit that state. This was because there always had to be at least one commit in the table which would produce a single datapoint for the custom plots. This is almost an error state. 

IMO we should be explicit to the user what is going on. I.e. if we know there are plots but no data to put in them we should state something like `There is no data to Plot`. 

You cannot separate out the messages in the current state, i.e. `!hasData` will happen if all of the experiments are filtered whether plots have been added or not. We need to perform the `order.length` check before sending the webview message and send that information instead of `undefined`.

You could replace the `No Plots to Display.` text with `There is No Data to Plot` but you would need to add tests on both sides of the message to document that behaviour and make sure that in future when we change the code we know what the expected flow is. Relying on the assumption of "if there is no data then all the experiments must be filtered" seems optimistic at best.

julieg18: OHHHH, ok I see what you're saying now 🤦‍♀️ . Thanks for the further explanation and apologies, I don't know why my brain was struggling so much here 😅. 

I'll update the backend to send a different kind of data when there is no unfiltered values so that the webview can detect the difference and updates tests accordingly.

extension/src/plots/model/index.ts
```diff
@@ -154,44 +153,40 @@ export class PlotsModel extends ModelWithPersistence {
...
-    const plotsOrderValues = this.getCustomPlotsOrder()
-    const plots: CustomPlotData[] = collectCustomPlots({
-      colorScale,
-      experiments,
-      height,
-      nbItemsPerRow,
-      plotsOrderValues
-    })
-
-    if (plots.length === 0 && plotsOrderValues.length > 0) {
```
julieg18, line 183: It's actually not possible for this if statement to be true. `collectCustomPlots` creates an array that matches the order of `plotsOrderValues`. Aka they will always be the same length. I went ahead and removed it.

extension/src/plots/webview/contract.ts
```diff
@@ -102,6 +102,8 @@ export type CustomPlotsData = {
   nbItemsPerRow: number
   enablePlotCreation: boolean
   height: PlotHeight
+  hasUnfilteredExperiments: boolean
+  hasAddedPlots: boolean
```
julieg18, line 106: We need to add `hasAddedPlots` because `plots` will be an empty array if there is no table data OR no added plots. 

Another option is to have `plots` only be empty when there are no added plots and fill `plots` with technically broken plots when there is no table data. These broken plots wouldn't be rendered of course since we don't render any plots when `hasUnfilteredExperiments` is false. I chose `hasAddedPlots` since I felt like that solution has more clarity.

julieg18 (commented):
Updated code to resolve https://github.com/iterative/vscode-dvc/pull/4657/files#r1322159906. Since it did add an extra 100 lines of code, going to rerequest reviews! 

cc @mattseddon

mattseddon (approved):

Issue #4654, opened 2023-09-11
Improve Plots Wizard


julieg18:
- [x] Make the reasoning behind the first file selector more obvious. Consolidate our "Create Custom Plot" and "Create Plot" commands, having a single "Create Plot" command with an extra quick pick. (https://github.com/iterative/vscode-dvc/pull/4664#pullrequestreview-1625680505)
- [x] Add an action into the plots webview (~add a "placeholder" plot that is button for adding plots~ (breaks drag and drop) add a plus button to the Data Series inputs)
- [x] Look into using `WorkspaceEdit` API (https://github.com/iterative/vscode-dvc/pull/4628#discussion_r1320892829). Could be used but wouldn't simplify anything. 
- [x] Look into allowing multiple files being selected for a plot (https://dvc.org/doc/command-reference/plots/show#sourcing-data-from-different-files)
- [x] Add option for choosing the plot title
- [x] Allow multiple metrics to be selected for y 
- [x] Allow multiple metrics to be selected for x (https://github.com/iterative/vscode-dvc/pull/4787#discussion_r1350813007)
- [x] Group tests in `/pipeline/quickPick.test.ts`.

Followup to #4628

julieg18:
> @julieg18 Was there at some point an extra plots pane in the circled area with a + to add a new plot? I'm wondering if we can make the plots wizard more obvious.

Yes, we wanted to do this originally, but adding one breaks our plot drag-and-drop code. It is most likely doable but would complicate our drag-and-drop code (which is already pretty complex as is) so we decided to not do this for now.

Issue #4661, opened 2023-09-13
Stale secret deletion


0x2b3bfa0:
It looks like the following GitHub Actions secrets aren't being used in any of your workflows, at least on the default branch:
* CC_TEST_REPORTER_ID

Please ping @iterative/security before 2023-09-22 if you'd like to preserve them.

mattseddon:
It is used here: https://github.com/iterative/vscode-dvc/blob/main/.github/workflows/continuous-integration.yml#L47

mattseddon:
@iterative/security do not delete that secret.

0x2b3bfa0:
Sorry, @mattseddon! I saw the issue “closed as completed” and thought I closed it myself. 🙈

0x2b3bfa0:
Can you please set that secret again?

mattseddon:
Yep done.

PR #4664, opened 2023-09-13
Add "Add Top-level Plot" command to plots webview

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/9ccd1a8d-5eeb-4e11-89a4-35c9fcead8f8

image

Part of #4654

### DIFF
--- a/extension/src/test/suite/plots/index.test.ts
+++ b/extension/src/test/suite/plots/index.test.ts
@@ -1124,6 +1124,23 @@ suite('Plots Test Suite', () => {
       )
     })
 
+    it('should handle a add template plot mmessage from the webview', async () => {
+      const { mockMessageReceived } = await buildPlotsWebview({
+        disposer: disposable,
+        plotsDiff: plotsDiffFixture
+      })
+
+      const executeCommandSpy = spy(commands, 'executeCommand')
+
+      mockMessageReceived.fire({
+        type: MessageFromWebviewType.ADD_PIPELINE_PLOT
+      })
+
+      expect(executeCommandSpy).to.be.calledWithExactly(
+        RegisteredCommands.PIPELINE_ADD_PLOT
+      )
+    })
+
     it('should handle the CLI throwing an error', async () => {
       const { data, errorsModel, mockPlotsDiff, plots, plotsModel } =
         await buildPlots({ disposer: disposable, plotsDiff: plotsDiffFixture })
--- a/extension/src/test/suite/plots/index.test.ts
+++ b/extension/src/test/suite/plots/index.test.ts
@@ -1137,7 +1137,8 @@ suite('Plots Test Suite', () => {
       })
 
       expect(executeCommandSpy).to.be.calledWithExactly(
-        RegisteredCommands.PIPELINE_ADD_PLOT
+        RegisteredCommands.PIPELINE_ADD_PLOT,
+        dvcDemoPath
       )
     })
 
--- a/extension/src/test/suite/plots/index.test.ts
+++ b/extension/src/test/suite/plots/index.test.ts
@@ -1124,7 +1124,7 @@ suite('Plots Test Suite', () => {
       )
     })
 
-    it('should handle a add template plot mmessage from the webview', async () => {
+    it('should handle an add pipeline plot message from the webview', async () => {
       const { mockMessageReceived } = await buildPlotsWebview({
         disposer: disposable,
         plotsDiff: plotsDiffFixture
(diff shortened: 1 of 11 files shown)


webview/src/plots/components/templatePlots/TemplatePlotsWrapper.tsx
```diff
@@ -20,6 +21,7 @@ export const TemplatePlotsWrapper: React.FC = () => {
       changeSize={changeSize}
       hasItems={hasItems}
       height={height}
+      addPlotsButton={{ onClick: addPipelinePlot }}
```
julieg18, line 24: Tried to add a plot button to the grid, but couldn't get past the drag and drop bugs (I needed the button to stay at the end of the plots but trying to drag and drop plots around the button would break things). There is most likely a way to implement it, but I was definitely having a hard time understanding the plots drag and drop code enough to figure out how. 

I'll give it another try as a followup later.

julieg18:
Code climate is failing with 

```
Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag
```

Does it have to do with the issue we had with [code climate keys](https://iterativeai.slack.com/archives/C01R00PPQ1L/p1694453513665289)?

mattseddon:
> Code climate is failing with
> 
> ```
> Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag
> ```
> 
> Does it have to do with the issue we had with [code climate keys](https://iterativeai.slack.com/archives/C01R00PPQ1L/p1694453513665289)?

https://github.com/iterative/vscode-dvc/issues/4661 - disappointing

mattseddon:
@0x2b3bfa0

0x2b3bfa0:
* https://github.com/iterative/vscode-dvc/issues/4661#issuecomment-1718507766 😳 

Can you plese restore that secret? 🙈

webview/src/plots/components/templatePlots/TemplatePlotsWrapper.tsx
mattseddon: Quick feedback on the UX (which I assume I missed in a previous PR).

Clicking on this button presents the user with this screen:

image

It is not obvious from this screen (unless you have context like you and I do) that the user is looking to add a file that contains data that can be used to generate a plot.

mattseddon: Also in the demo the tooltip says "Add Plot**s**" but should be "Add Plot":

image

extension/src/test/suite/plots/index.test.ts
```diff
@@ -1124,6 +1124,23 @@ suite('Plots Test Suite', () => {
...
+      })
+
+      const executeCommandSpy = spy(commands, 'executeCommand')
+
+      mockMessageReceived.fire({
+        type: MessageFromWebviewType.ADD_PIPELINE_PLOT
+      })
+
+      expect(executeCommandSpy).to.be.calledWithExactly(
+        RegisteredCommands.PIPELINE_ADD_PLOT
```
mattseddon, line 1140: [B] We know which `dvcRoot` we are calling `RegisteredCommands.PIPELINE_ADD_PLOT` from so we should not use `getCwd` again. Otherwise when a user is using a mono-repo they will click on "Add Plot" from a specific repos plots' webview and then be asked to choose which repository they want to add the plot to. There are plenty of examples of how we normally do this in `extension/src/experiments/webview/messages.ts` (e.g. `RegisteredCliCommands.EXPERIMENT_VIEW_PUSH`)

mattseddon (approved):
Good iteration. One bug to fix and we need to follow up on making the expected behaviour in the wizard's steps more obvious.

Consider combining/moving the add plot(s) buttons into the ribbon to make them more obvious/visible. I.e somewhere around here:

Screenshot 2023-09-14 at 11 27 36 am

webview/src/plots/components/templatePlots/TemplatePlotsWrapper.tsx
julieg18: > It is not obvious from this screen (unless you have context like you and I do) that the user is looking to add a file that contains data that can be used to generate a plot.

Oh, good point! First idea that comes to mind is opening a quick pick that lets user type in a relative path or select a file (we do something similar when creating a train stage). The quick pick can add a sentence that would clarify things further. I'll update #4654.

extension/src/test/suite/plots/index.test.ts
julieg18: Good catch! Updating the command to check for a `dvcRoot` being passed.

PR #4680, opened 2023-09-18
Merge create plot commands

julieg18:
* combine "Add Custom Plot" and "Add Plot" command into one. This adds an explanation for the "file selector" that appears when you add a top-level plot and simplifies some logic. 

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/62dad7cc-df0d-4051-a79e-981fe02f68fe~

~https://github.com/iterative/vscode-dvc/assets/43496356/5e50305c-c30a-4406-a27f-5c5af7e6f922~

https://github.com/iterative/vscode-dvc/assets/43496356/086406e2-9c29-4517-a0a6-3d3667fd42e4

image

image

image

## To Do

- [x] Rename "top-level" to "Data Series" 

Part of #4654


extension/src/commands/util.ts
julieg18, line 1: Is there a better place for these two commands that are registered in `extension.ts`? I couldn't think of a place that didn't feel confusing 🤔

webview/src/plots/components/ribbon/Ribbon.tsx
```diff
@@ -75,6 +76,14 @@ export const Ribbon: React.FC = () => {
           text={`${revisions.length} of ${MAX_NB_EXP}`}
         />
       </li>
+      <li className={styles.buttonWrapper}>
```
julieg18, line 79: Example with primary colors:

Screenshot 2023-09-18 at 9 44 19 AM

webview/src/plots/util/messages.ts
```diff
@@ -3,16 +3,15 @@ import { PlotsSection } from 'dvc/src/plots/webview/contract'
...
 
-export const addCustomPlot = () =>
+export const addPlot = () =>
   sendMessage({
-    type: MessageFromWebviewType.ADD_CUSTOM_PLOT
+    type: MessageFromWebviewType.ADD_PLOT
   })
 
-export const addPipelinePlot = () => {
+export const addCustomPlot = () =>
```
julieg18, line 11: I chose to keep the add plot button in the custom section since we still have the trash icon there, so I thought it made sense to have the add button next to it. 

image

But we can remove it if preferred.

webview/src/plots/components/ribbon/Ribbon.tsx
mattseddon: As per the [guidelines](https://github.com/microsoft/vscode-webview-ui-toolkit/blob/main/src/button/README.md#best-practices) there should only be one primary button. I think Having `Add Plot` as the primary button now makes more sense than the revision selector but up to you.

webview/src/plots/util/messages.ts
mattseddon: If you think the UX is better then it is ok to leave and we can even keep the add plot button above the `Data Series` section but do keep in mind that we will to maintain this code. Any code that runs can eventually cause bugs.

extension/src/commands/util.ts
```diff
@@ -18,3 +20,40 @@ export const showSetupOrExecuteCommand =
...
+  const result: PLOT_TYPE | undefined = await quickPickValue(
+    [
+      {
+        description: 'Create a dvc.yaml plot based off a chosen data file',
+        label: 'Top-Level',
+        value: PLOT_TYPE.TOP_LEVEL
+      },
+      {
+        description:
+          'Create an extension-only plot based off a chosen metric and param',
```
mattseddon, line 39: [C] `Create an extension-only plot by selecting one metric and one param from the experiments table`

extension/src/commands/util.ts
```diff
@@ -18,3 +20,40 @@ export const showSetupOrExecuteCommand =
...
+const enum PLOT_TYPE {
+  CUSTOM = 'custom',
+  TOP_LEVEL = 'top-level'
+}
+
+export const addPlotCommand = async (context: Context) => {
+  const result: PLOT_TYPE | undefined = await quickPickValue(
+    [
+      {
+        description: 'Create a dvc.yaml plot based off a chosen data file',
```
mattseddon, line 33: [C] `Create a top-level plot definition by selecting data from a file`

extension/src/commands/util.ts
mattseddon: Both of these are suggestions and can be dismissed/improved. What I am trying to do is standardise on simple terminology that helps users to connect the dots between concepts.

mattseddon: [C] I think we need to revisit the `Custom` name now (not in this PR but we should discuss/consider updating).

extension/src/commands/util.ts
```diff
@@ -18,3 +20,40 @@ export const showSetupOrExecuteCommand =
...
     return callback(context)
   }
+
+const enum PLOT_TYPE {
+  CUSTOM = 'custom',
+  TOP_LEVEL = 'top-level'
+}
+
+export const addPlotCommand = async (context: Context) => {
+  const result: PLOT_TYPE | undefined = await quickPickValue(
```
mattseddon, line 30: [N] usually named `response`

extension/src/commands/util.ts
```diff
@@ -18,3 +20,40 @@ export const showSetupOrExecuteCommand =
 
     return callback(context)
   }
+
+const enum PLOT_TYPE {
+  CUSTOM = 'custom',
+  TOP_LEVEL = 'top-level'
+}
+
+export const addPlotCommand = async (context: Context) => {
```
mattseddon, line 29: [I] These should go in a new `plots/quickPick` file. Can you also please check to see if we will get two analytics events each time this is called (I don't think we want that).

extension/src/extension.ts
```diff
@@ -193,6 +193,10 @@ class Extension extends Disposable {
     )
     registerPipelineCommands(this.pipelines, this.internalCommands)
     registerPlotsCommands(this.plots, this.internalCommands, this.setup)
+    this.internalCommands.registerExternalCommand(
```
mattseddon, line 196: [I] Plots commands should be registered in `extension/src/plots/commands/register.ts`.

mattseddon (approved):

extension/src/commands/util.ts
julieg18: > Can you also please check to see if we will get two analytics events each time this is called (I don't think we want that).

Oh, I think we would get two since `addPlot` fires one and then "add custom"/"add pipeline" would fire another. We could remove the "add custom" and "add pipeline" commands and add a `type` argument to the "add plot" command 🤔

julieg18:
Resolved comments and re-requesting reviews since quite a bit of the solution has been updated! cc @mattseddon

webview/src/plots/util/messages.ts
mattseddon: Are we still keeping this?

mattseddon (approved):

webview/src/plots/util/messages.ts
julieg18: I thought we agreed to keep it for now, but it wouldn't be the first time I misunderstood what was said in a meeting 😊. I'll go ahead and take off the `+` button. The `addCustomPlot` util would stay since we use it in the error screen.

julieg18: Actually, after testing it, we should keep it for now until we add a button to our "No Plots Added" screen.

PR #4685, opened 2023-09-18
Flatten table on sort

### LINKED ISSUE #4620: Apply sorting to the entire table

julieg18:
# 1/2 `main` <= this <= #4691 

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/1c3d0a62-6ece-47a6-a6c4-b72bbfc47405

## To Do

- [x] Remove possible duplicate rows (https://github.com/iterative/vscode-dvc/pull/4685#discussion_r1338040254) (done in #4735)

Fixes #4620


extension/src/experiments/model/index.ts
```diff
@@ -437,28 +438,21 @@ export class ExperimentsModel extends ModelWithPersistence {
...
+    const workspaceBranch = {
+      branch: WORKSPACE_BRANCH,
+      ...this.addDetails(this.workspace)
     }
+    const rowsAcc: Commit[] = []
+    const sorts = this.getSorts()
+    const flattenRows = sorts.length > 0
 
-    return rows
+    for (const { branch, sha } of this.rowOrder) {
```
mattseddon, line 452: [I] If we have sorts. Let's not collect the data with `subrows` let's have a separate branch of logic which collects the data and returns it. If we don't have sorts leave the logic as it was previously.

(As discussed on a call) This gives us:

1. Breadcrumbs for the table being flat when sorts are applied (think method names)
2. Opportunity to filter correctly when the table is flat (simplify logic for filtering as don't have to retain "filtered parents" if children as unfiltered.
3. No need to collect subrows then mutate the data further (simpler logic overall).

extension/src/test/fixtures/expShow/sorted/rows.ts
```diff
@@ -0,0 +1,348 @@
+import { join } from '../../../util/path'
```
julieg18, line 1: image

The large numbers are mainly due to:
* adding a new code fixture for Storybook (`fixtures/expShow/sorted`)
* Moving the `getRowModel` mock function value outside of a describe block inside of `TableContent.tsx`

extension/src/experiments/model/index.ts
julieg18: Moved the logic to `getFlattenedRowData` that gets called inside of `getRowData` if there are sorts.

extension/src/test/fixtures/expShow/sorted/tableData.ts
```diff
@@ -0,0 +1,19 @@
+import columns from './columns'
+import rows from './rows'
+import defaultData from '../base/tableData'
+import { TableData } from '../../../../experiments/webview/contract'
+
+const data: TableData = {
+  ...defaultData,
```
julieg18, line 7: `sorted/tableData` is a slightly simplified (no deps and no selected exps) version of `base/tableData`. We could simplify further though if preferred.

webview/src/stories/Table.stories.tsx
```diff
@@ -61,10 +62,7 @@ const tableData = getTableState({
   selectedBranches: [],
   selectedForPlotsCount: 2,
   showOnlyChanged: false,
-  sorts: [
```
julieg18, line 64: Sorts used to be a default addition to the storybooks but since `tableData` rows are different when they are sorted, I removed the default sorts array (sorts are still shown in our new sorted fixture and in the `deeplyNested` fixture).

extension/src/test/fixtures/expShow/sorted/rows.ts
julieg18: Probably should have moved the storybook changes in a separate chain pr 😅

extension/src/experiments/model/index.ts
```diff
@@ -829,4 +837,42 @@ export class ExperimentsModel extends ModelWithPersistence {
...
+
+    return commitsBySha
+  }
+
+  private getFlattenedRowData(workspaceRow: Commit): Commit[] {
+    const commitsBySha: { [sha: string]: Commit[] } =
+      this.applyFiltersToFlattenedCommits()
+    const rows = []
+
+    for (const { branch, sha } of this.rowOrder) {
```
mattseddon, line 865: [C] Once the table is flattened we could have duplicate entries right next to each other. We will need a way to collapse these rows:

image

mattseddon: The only thing that will differ between records will be the branch/tags. I think we should consider concatenating the information on multiple lines into a single cell.

webview/src/experiments/components/table/body/TableContent.test.tsx
```diff
@@ -19,210 +22,64 @@ jest.mock('../../../../shared/api')
...
-                  }
-                },
-                selected: false,
-                sha: 'ae4100a4b4a972c3ceffa3062739845d944b3ddf',
-                starred: false
-              }
-            ]
+const mockedGetIsExpanded = jest.fn()
+const mockedGetAllCells = jest.fn().mockReturnValue([1, 2, 3, 4, 5]) // Only needed as length
+const mockRowModel = {
```
mattseddon, line 27: [Q] Can we move this out of this file so that any further changes to this fixture aren't mixed in with test changes?

mattseddon (approved):
Good work.

One item for follow up.

webview/src/experiments/components/table/body/TableContent.test.tsx
julieg18: Moved into a separate file!

extension/src/experiments/model/index.ts
mattseddon: This needs to be fixed before moving on.

julieg18: > This needs to be fixed before moving on.

Oh, ok! Apologies, I misunderstood [your comment](https://github.com/iterative/vscode-dvc/pull/4685#pullrequestreview-1645603063).

tapadipti:
> @tapadipti, @dberenbaum we might want to consider flattening the table for sorting in Studio:

In the flattened table, some of the existing functionality may become a bit confusing. Eg, when a parent commit is hidden, all its exps get thidden. This will not be easy to visualize in a flat table. 

We recently included experiments in sorting, and atm it applies it hierarchically. That is, exps within a commit are sorted (similar to how it is in vs code before this pr?). We have a [follow-up issue](https://github.com/iterative/studio/issues/7639) to improve this, and we could include flattening the table as one of the options. But I think we should limit some functions in this mode.

@julieg18 Will vs code have different ways to sort or will this be the default and only way to sort the table?

julieg18:
> @julieg18 Will vs code have different ways to sort or will this be the default and only way to sort the table?

Currently flattening the rows is the only way to sort the table, but that could change depending on user feedback.

tapadipti:
> Currently flattening the rows is the only way to sort the table, but that could change depending on user feedback.

Thanks @julieg18. I've update the issue description for https://github.com/iterative/studio/issues/7639 and included this option (to flatten and sort the entire table).

mattseddon:
> In the flattened table, some of the existing functionality may become a bit confusing. Eg, when a parent commit is hidden, all its exps get thidden. This will not be easy to visualize in a flat table.

The extension uses different filtering logic in that it will keep a parent commit in the table if any of its children are unfiltered. We do however change the filtering logic when the table is flat. Each record is filtered individually.

> We recently included experiments in sorting, and atm it applies it hierarchically. That is, exps within a commit are sorted (similar to how it is in vs code before this pr?). We have a https://github.com/iterative/studio/issues/7639 to improve this, and we could include flattening the table as one of the options. But I think we should limit some functions in this mode.

That is also how the extension sorts experiments. All the user feedback that we have gotten is that this does not really help.

extension/src/experiments/model/index.ts
julieg18: Fixed in #4735

julieg18 (commented):
Alright, going to merge this! Last things left to do are:

*  Remove possible duplicate rows (https://github.com/iterative/vscode-dvc/pull/4685#discussion_r1338040254) (done in https://github.com/iterative/vscode-dvc/pull/4735)
* Make "Parent" and "Branch/Tags" columns sticky (https://github.com/iterative/vscode-dvc/pull/4691#discussion_r1338145437) 

I'll open a followup issue and start looking into the second task this sprint :)

PR #4687, opened 2023-09-19
Simplify data structure used to collect multisource encoding update

mattseddon:
Noticed that we were collecting data in a more complicated format than is needed/expected whilst working through the logic needed for https://github.com/iterative/dvc-render/pull/142

### DIFF
--- a/extension/src/plots/multiSource/collect.ts
+++ b/extension/src/plots/multiSource/collect.ts
@@ -119,31 +119,15 @@ const collectMultiSourceValues = (variations: Variations): Values => {
   return values
 }
 
-const sortDifferentVariations = (
-  differentVariations: { field: string; variations: number }[],
-  expectedOrder: string[]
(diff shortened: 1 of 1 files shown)


extension/src/plots/multiSource/collect.ts
```diff
@@ -265,16 +248,14 @@ const collectUnmergedShapeEncoding = (
...
-  lessValuesThanVariations: { field: string; variations: number }[]
+  lessValuesThanVariations: string[]
 ): void => {
   acc[path] = {
     ...acc[path],
     shape: collectEncodingFromValues(Shape, values, lessValuesThanVariations)
   }
 }
 
-// this is how we can collect the encoding for multi-source plots
```
mattseddon, line 276: [F] This was not meant to be committed.

julieg18 (approved):

PR #4691, opened 2023-09-19
Add "Parents" Columns

julieg18:
# `main` <= #4685 <= this

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/b1b8a945-b4e3-47de-bf58-a4331efe0af0

Part of #4620


extension/src/experiments/columns/constants.ts
```diff
@@ -11,3 +11,12 @@ export const timestampColumn: Column = {
 }
 
 export const EXPERIMENT_COLUMN_ID = 'id'
+
+export const COMMIT_COLUMN_ID = 'commit'
+export const BRANCH_COLUMN_ID = 'branch'
+
+export const DEFAULT_COLUMN_IDS = [
```
julieg18, line 18: Added two more default columns (ids are commt and branch). These always stay in the order on the backend and appear on the frontend if the table is sorted. Similar to the `id` column, they are not draggable and can't be unselected.

webview/src/experiments/components/Experiments.tsx
```diff
@@ -108,7 +112,9 @@ export const ExperimentsTable: React.FC = () => {
     toggleAllRowsExpanded()
   }, [toggleAllRowsExpanded])
 
-  const hasOnlyDefaultColumns = columns.length <= 1
+  const hasOnlyDefaultColumns = columns.every(
+    ({ id }) => id && isDefaultColumn(id)
```
julieg18, line 116: Updated this to check for the column ids since there are now 1 or 3 default columns depending on if the table is sorted.

webview/src/experiments/components/App.test.tsx
```diff
@@ -25,7 +26,7 @@ import { vsCodeApi } from '../../shared/api'
 import {
   commonColumnFields,
   expectHeaders,
-  tableData as sortingTableDataFixture
+  tableData as simplifiedSortedTableDataFixture
```
julieg18, line 29: Renamed this to differentiate from the `sortedTableDataFixture` which has data based on the `expShow` data. I did consider replacing `simplifiedSortedTableData` with the backend one, but that would complicate multiple tests so I decided to keep it.

webview/src/experiments/components/Experiments.tsx
mattseddon: [Q] Does `every` return `true` if the array is empty?

mattseddon (approved):

webview/src/experiments/components/table/content/CommitHeader.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import styles from '../styles.module.scss'
+import { Icon } from '../../../../shared/components/Icon'
+import { Info } from '../../../../shared/components/icons'
+import Tooltip from '../../../../shared/components/tooltip/Tooltip'
+
+export const CommitHeader: React.FC = () => (
```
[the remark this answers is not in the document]
julieg18: Yes, we could use the same component!

webview/src/experiments/util/columns.ts
```diff
@@ -80,3 +84,8 @@ export const leafColumnIds = (
 
 export const isExperimentColumn = (id: string): boolean =>
   id === EXPERIMENT_COLUMN_ID
+
+export const isDefaultColumn = (id: string) =>
```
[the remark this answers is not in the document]
julieg18: > [Q] Couldn't we just use DEFAULT_COLUMN_IDS instead of importing all of the individual IDs?

Yes, thanks for catching that! 

> [Q] Won't the addition of these two columns break <WithExpColumnNeedsShadowUpdates>?

Forgot to mention this in a review comment, For now, the columns aren't sticky:

https://github.com/iterative/vscode-dvc/assets/43496356/10adf0f9-5f48-411f-b1b5-1d00449eb98c

I wanted to save it for a follow-up since I was having some difficulty making three columns sticky and have the experiment column conditionally have a shadow.

PR #4694, opened 2023-09-20
Add action buttons to empty plot sections

julieg18:
### Custom

~https://github.com/iterative/vscode-dvc/assets/43496356/23f4a1db-a22d-4adf-bf1e-90aa0725ab7a~

https://github.com/iterative/vscode-dvc/assets/43496356/3a6f37c5-4740-4e5d-8cc7-4a5f634dc17e

### Template

https://github.com/iterative/vscode-dvc/assets/43496356/8d069507-fb3a-4809-8a55-35305687f4a8

Followup to #4680


webview/src/plots/components/templatePlots/NoPlotsToDisplay.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <EmptyState isFullScreen={false}>
+      <p>No Plots to Display</p>
+      {hasUnselectedPlots ? (
+        <div>
+          <StartButton
+            isNested={true}
+            onClick={selectPlots}
+            text="Select Plots"
+          />
+          <StartButton
```
julieg18, line 22: The custom section "Add Plot" button is a bit different than the template one since it can take you to right to custom plot creation, unlike the data series which needs that extra description for more context.

We could keep as is, but if we wanted them to function the same we could:

1. Have the custom plot section open the "Select Plot Type" quick pick as well. Would mean an extra click for the user but would reduce some code. 
2. Update our "Add Data Series Plot" action to open a quick pick first instead of the file selector, allowing the user to get context and type in a relative file if preferred. Would get rid of the extra click a user needs to make, but would make data series plot creation a bit more complex.

shcheklein:
Demo look perfect to me. Thanks!

extension/src/plots/model/custom.ts
```diff
@@ -52,24 +52,6 @@ export const getCustomPlotPathsFromColumns = (
   return { metrics, params }
 }
 
-export const checkForCustomPlotOptions = (
```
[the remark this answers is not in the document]
mattseddon: If the option fails under these conditions with a clear message then that is ok.

I think the chances of real users hitting this is low.

webview/src/plots/components/templatePlots/NoPlotsToDisplay.tsx
mattseddon: I vote for option 1.

mattseddon (approved):

extension/src/plots/model/custom.ts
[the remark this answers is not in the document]
julieg18: > If the option fails under these conditions with a clear message then that is ok.

Yes, option fails with a error toast message.

webview/src/plots/components/templatePlots/NoPlotsToDisplay.tsx
julieg18: Yes, "Add Data Series" opens with the dvc.yaml being opened. Makes sense! I'll update the code to use the add plot command instead.

Issue #4725, opened 2023-09-27
custom plots issues


dberenbaum:
I ran into a couple unexpected issues with custom plots. Here's what I saw using https://github.com/iterative/example-get-started-experiments/:

<img width="994" alt="Screenshot 2023-09-27 at 8 22 07 AM" src="https://github.com/iterative/vscode-dvc/assets/2308172/6f9b39b8-e045-4563-8ef8-cd120a9bb9c0">

There are two unexpected behaviors:

1. It shows unselected revisions (all of the revs with the same color were not selected for plots).
2. The x-axis only shows one value for this string/categorical parameter. Here are the actual values for that parameter:

<img width="638" alt="Screenshot 2023-09-27 at 8 26 47 AM" src="https://github.com/iterative/vscode-dvc/assets/2308172/a03b8271-6716-42b0-9b05-3d4b05a11ea6">

julieg18:
> It shows unselected revisions (all of the revs with the same color were not selected for plots).

This is expected. Custom plots are supposed to show all experiments in the table. 

<img width="482" alt="image" src="https://github.com/iterative/vscode-dvc/assets/43496356/91715d5a-fc4e-4b69-bc78-bf6d12cf4ff0">

> The x-axis only shows one value for this string/categorical parameter. Here are the actual values for that parameter:

Looks to be a bug, let me see if I can reproduce 👀

julieg18:
> Looks to be a bug, let me see if I can reproduce 👀

Hmmmm, x-axis is showing correctly on my end:

<img width="1718" alt="image" src="https://github.com/iterative/vscode-dvc/assets/43496356/6ec8729b-52b5-41b9-8a02-85064a269317">

@dberenbaum, can you:
* check what the `dice_multi` metric value is for your `podgy-sima` experiment? 
* export the plot as json? I'd like to see what experiments/values are inside the `json` file.

dberenbaum:
My mistake @julieg18! I think I must have chosen the wrong `dice_multi` column that only appeared in one experiment. Now it looks like expected:

<img width="990" alt="Screenshot 2023-09-27 at 9 25 44 AM" src="https://github.com/iterative/vscode-dvc/assets/2308172/ad63ad23-cc07-4615-976d-26640eec0edb">

PR #4729, opened 2023-09-27
Ensure custom plots legend matches plot values

julieg18:
* our custom plot legend will always show all experiments/commits even if they are not in the plot values. This pr ensures that the legend only shows experiments/commits that are in the plot

## main

image

## PR

image

Related to #4725


extension/src/plots/model/collect.ts
```diff
@@ -93,6 +114,7 @@ const getCustomPlotData = (
   const paramPath = getFullValuePath(ColumnType.PARAMS, param)
 
   const values = getValues(experiments, metricPath, paramPath, renderLastIds)
+  const filteredColorScale = filterColorScale(completeColorScale, values)
```
mattseddon, line 117: [Q] Is there a reason to do this after collecting `completeColorScale`? Would be good to have all of the transformation/collection in the same place.

julieg18: Originally, I was trying to avoid looping over all the experiments for each plot, but I now see we're still looping over `completeColorScale` anyway 😄. I'll update the code to create a single color scale per plot.

extension/src/plots/model/collect.ts
```diff
@@ -80,19 +80,65 @@ const getValues = (
...
+
+  return filteredColorScale
+}
+
+const fillColorScale = (
+  experiments: Experiment[],
+  colorScale: ColorScale | undefined,
+  valueIds: string[]
+) => {
+  const completeColorScale = filterColorScale(colorScale, valueIds)
```
julieg18, line 112: We need to filter through the original color scale values first so they retain their order in the legend. Maybe there's a simpler way of doing this though 🤔

mattseddon: something like 
```
const removeSelectedExperiment = (
  orderedColorScale: ColorScale,
  hasValues: boolean,
  idx: number
) => {
  const isSelectedExperiment = idx !== -1
  if (!isSelectedExperiment || hasValues) {
    return
  }

  orderedColorScale.domain.splice(idx, 1)
  orderedColorScale.range.splice(idx, 1)
}

const fillColorScale = (
  experiments: Experiment[],
  colorScale: ColorScale | undefined,
  valueIds: string[]
) => {
  const orderedColorScale = {
    domain: [...(colorScale?.domain || [])],
    range: [...(colorScale?.range || [])]
  }

  for (const experiment of experiments) {
    const { id } = experiment
    const idx = orderedColorScale.domain.indexOf(id)
    const isSelectedExperiment = idx !== -1
    const hasValues = valueIds.includes(id)

    if (!hasValues || isSelectedExperiment) {
      removeSelectedExperiment(orderedColorScale, hasValues, idx)
      continue
    }

    orderedColorScale.domain.push(id)
    orderedColorScale.range.push('#4c78a8' as Color)
  }

  return orderedColorScale
}
```

mattseddon (approved):

extension/src/plots/model/collect.ts
```diff
@@ -80,19 +80,65 @@ const getValues = (
...
+    domain: [],
+    range: []
+  }
+
+  if (!colorScale) {
+    return filteredColorScale
+  }
+
+  for (const [ind, id] of colorScale.domain.entries()) {
+    if (!valueIds.includes(id)) {
```
mattseddon, line 97: [N] this would probably be better as a `Set`

extension/src/plots/model/collect.ts
julieg18: Works great, thanks!

PR #4748, opened 2023-10-02
Allow Multi File Select on Plot Wizard

julieg18:
## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/6a98c334-b10e-4c48-8ad2-40178a9d62e9~

https://github.com/iterative/vscode-dvc/assets/43496356/720cfb94-c65f-4e22-ac64-ed2824c58b28

Part of #4654


dberenbaum:
Looks great @julieg18!

Some thoughts on what else we could do:
- Can we add `title` (you could skip inferring a name for the plot if we do this), `x_label`, and `y_label`?
- Does it allow for a single `x` and a single `y` right now? Most of the value is likely when we allow multiple entries for each.

julieg18:
> Can we add title (you could skip inferring a name for the plot if we do this), x_label, and y_label (see https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#available-configuration-fields)?

We could, though we might want to keep in mind that each of these options would involve adding an extra quick pick. If we have too many quick picks, it could lead to plot creation being tedious. 

> Does it allow for a single x and a single y right now? Most of the value is likely when we allow multiple entries for each.

Yes, a single x and single y field. I wasn't 100% sure if every plot type would work with multiple entries so I decided to keep it as single for now. We could allow multiple entries in a followup :)

extension/src/vscode/resourcePicker.ts
```diff
@@ -34,3 +28,26 @@ export const pickResources = (title: Title): Thenable<Uri[] | undefined> => {
     title
   })
 }
+
+export const pickFiles = async (
```
julieg18, line 32: Had to create a new picker since `pickResources` allows folder selection.

extension/src/pipeline/quickPick.ts
```diff
@@ -79,36 +92,79 @@ const getFieldOptions = (data: UnknownValue): string[] => {
...
 
-  const data = await loadDataFile(file)
+  const fileExts = new Set(files.map(file => getFileExtension(file)))
 
-  if (!data) {
-    return Toast.showError(
-      'Failed to parse the requested file. Does the file contain data and follow the DVC plot guidelines for [JSON/YAML](https://dvc.org/doc/command-reference/plots/show#example-hierarchical-data) or [CSV/TSV](https://dvc.org/doc/command-reference/plots/show#example-tabular-data) files?'
-    )
+  if (fileExts.size > 1) {
+    return Toast.showError('Files must of the same type.')
```
julieg18, line 144: Our error handling has gotten more complex with the addition of multiple files.

**Original:**
1. Check if file can be parsed.
2. Check if file holds at least two field options
3. Go on to plot template picker

**Multiple Files:**
1. Check if all files have the same extension
2. Check if all files can be parsed
3. Check if all files contain at least one field
4. Check if we have at least two fields total among all files
5. Go on to plot template picker

We could also filter out invalid files instead of failing entirely but I chose to just fail since I thought it would be less confusing. Any ideas for simplifying the logic are welcome!

julieg18: Also, all the files need to have an array of the same length but I chose not to check for that. DVC fails with this error:

image

We could add a check for it though if we think it would be needed.

extension/src/pipeline/quickPick.ts
```diff
@@ -1,40 +1,54 @@
...
   }
 
-  const x = await quickPickOne(fields, 'Pick a Metric for X')
+  const items = []
+
+  for (const [file, keys] of Object.entries(fields)) {
+    items.push(
+      {
+        kind: QuickPickItemKind.Separator,
+        label: getFileName(file),
```
julieg18, line 36: Now looking at this, the filenames could be the same (`train/values.json` and `test/values.json` for example). I'll update it to show the relative path.

extension/src/vscode/resourcePicker.ts
sroy3: Maybe `pickResources` could take two more parameters `canSelectFolders` defaulting to `true` and `canSelectMany` defaulting to `true` also. That way you could remove the duplication between `pickFile`, `pickResources` and `pickFiles`.

extension/src/pipeline/quickPick.ts
```diff
@@ -79,36 +95,79 @@ const getFieldOptions = (data: UnknownValue): string[] => {
...
   }
 
-  const keys = getFieldOptions(data as UnknownValue)
+  const filesData = await loadDataFiles(files)
 
-  if (keys.length < 2) {
+  if (!filesData) {
     return Toast.showError(
-      'The requested file does not contain enough keys (columns) to generate a plot. Does the file follow the DVC plot guidelines for [JSON/YAML](https://dvc.org/doc/command-reference/plots/show#example-hierarchical-data) or [CSV/TSV](https://dvc.org/doc/command-reference/plots/show#example-tabular-data) files?'
+      'Failed to parse the requested file(s). Does the file or files contain data and follow the DVC plot guidelines for [JSON/YAML](https://dvc.org/doc/command-reference/plots/show#example-hierarchical-data) or [CSV/TSV](https://dvc.org/doc/command-reference/plots/show#example-tabular-data) files?'
```
sroy3, line 154: It feels a little weird to have both `file(s)` and `file or files` so close to one another. I'd stick with only one formulation.

extension/src/pipeline/quickPick.ts
```diff
@@ -110,5 +151,5 @@ export const pickPlotConfiguration = async (): Promise<
     return
   }
 
-  return { ...templateAndFields, dataFile: file }
+  return { ...templateAndFields }
```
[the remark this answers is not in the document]
sroy3: This function looks like it's doing many things too. It might be good to break it down into subfunctions like `validateExtensions`, `validateFilesData`... It would also clear this error.

sroy3 (approved):

mattseddon:
image

It is not obvious from the workflow that multiple files can be selected for the y-axis. I think we should update this text. If the multi-select is shift+click then that needs to be stated.

extension/src/fileSystem/index.ts
```diff
@@ -214,21 +214,38 @@ const loadYamlAsDoc = (
   }
 }
 
+const getPlotYamlObj = (cwd: string, plot: PlotConfigData) => {
+  const { x, y, template } = plot
+  const usesSingleFile = x.file === y.file
```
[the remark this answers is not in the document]
mattseddon: [Q] Do we need this split? Can we make the creation of `plots` entries a bit more opinionated from the wizard and reduce complexity? Otherwise, when we come to add in custom titles we will have to complicate the differentiation further.

From our demo `dvc.yaml`:

```
  - Loss:
      x: step
      y:
      y_label: loss
  - Confusion matrix:
      template: confusion
      x: actual
      y:
        training/plots/sklearn/confusion_matrix.json: predicted
  - hist.csv:
      x: preds
      y: digit
      template: bar_horizontal
      title: Histogram of Predictions
```

Maybe we want to stick to the first two entry types.

extension/src/pipeline/quickPick.test.ts
```diff
@@ -47,39 +49,51 @@ const mockValidData = [
...
+    const result = await pickPlotConfiguration('/')
 
     expect(result).toStrictEqual(undefined)
   })
 
-  it('should show a toast message if the file fails to parse', async () => {
-    mockedPickFile.mockResolvedValueOnce('file.csv')
-    mockedLoadDataFile.mockReturnValueOnce(undefined)
+  it('should show a toast message if the files are not the same data type', async () => {
+    mockedPickFiles.mockResolvedValueOnce(['file.json', 'file.csv'])
```
mattseddon, line 73: [Q] Is this a dvc constraint?

extension/src/pipeline/quickPick.ts
```diff
@@ -79,36 +95,79 @@ const getFieldOptions = (data: UnknownValue): string[] => {
...
+) => {
+  const keys: {
+    [file: string]: string[]
+  } = {}
+  let keysAmount = 0
+
+  for (const { file, data } of dataArr) {
+    const fields = getFieldsFromValue(data)
+
+    if (fields.length === 0) {
```
mattseddon, line 114: [Q] Should we do this check after we've collected all of the keys? If not I think we need to specify which of the files was the one that caused the process to reject.

extension/src/pipeline/quickPick.ts
mattseddon: IMO we should be checking for things that make DVC fail and providing as much information as possible up front. Remember the main reason we're doing this is for onboarding to plots/simplification of the process.

And (as previously stated) if we fail the process because of a single file then we need to let the user know which file made the process fail and why.

extension/src/pipeline/quickPick.ts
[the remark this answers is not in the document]
mattseddon: Yep, we could also branch the logic depending on the number of files selected.

mattseddon (approved):

extension/src/fileSystem/index.ts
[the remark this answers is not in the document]
mattseddon: See also https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#available-configuration-fields:

image

extension/src/pipeline/quickPick.test.ts
julieg18: Yes, DVC needs the files to be the same type.

extension/src/pipeline/quickPick.ts
julieg18: > [Q] Should we do this check after we've collected all of the keys? If not I think we need to specify which of the files was the one that caused the process to reject.

Makes sense! I can adjust the toast message to mention which file is failing.

extension/src/vscode/resourcePicker.ts
julieg18: Done. Good idea :)

extension/src/fileSystem/index.ts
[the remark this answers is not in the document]
julieg18: Updated the two plot types to be more similar:

```yaml
plots:
  # two files
  - scatter_plot:
      template: scatter
      x:
        props.json: acc
      y:
        values.json: prob
  # single files
  - simple_plot:
      template: simple
      x: actual
      y:
        probs.json: prob
```

extension/src/pipeline/quickPick.ts
julieg18: Used `file or files` since the grammar looked off with `file(s)` :)

julieg18 (commented):
Alright, resolved some of the comments. What's left:

- [ ] Improve `pickPlotConfiguration` by breaking into sub functions and branching logic (https://github.com/iterative/vscode-dvc/pull/4748#discussion_r1342928821)
- [ ] Add check for file array lengths (https://github.com/iterative/vscode-dvc/pull/4748#discussion_r1344877457)
- [ ] State what files are causing failures in Toast messages (https://github.com/iterative/vscode-dvc/pull/4748#discussion_r1344876077)

Taking care of these in #4770, but I think this pr is good enough to merge as a first iteration :)

PR #4616, opened 2023-09-03
exp rename: Add exp rename to experiments table

BradyJ27:
cc @mattseddon 

Add rename capability to the experiments table view. 

dvc-rename-vscode

Fixes: #4330 
Upstream Dependencies: https://github.com/iterative/dvc/pull/9812 docs - https://github.com/iterative/dvc.org/pull/4805

_Note_: To test you'll need to pull the dvc changes from 9812 above.

PR #2644, opened 2022-10-20
Add authorize step to both PR workflows

mattseddon:
More information: https://github.com/iterative/cml/pull/574#issuecomment-1188650440

image

### Internal

image

### External

image

PR #4762, opened 2023-10-04
Handle pull requests from forks

mattseddon:
Needed for #4616

The continuous integration workflow will currently always fail for forked pull requests. This is due to secrets not being passed to forks.

Comments inline.

If we start getting a lot of outside contributions then we can revisit.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -42,16 +42,25 @@ jobs:
 
       - run: yarn run lint
 
-      - uses: paambaati/codeclimate-action@v5.0.0
+      - name: Run Coverage Report
+        if: '!github.event.pull_request.head.repo.fork'
+        uses: paambaati/codeclimate-action@v5.0.0
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -42,16 +42,25 @@ jobs:
 
       - run: yarn run lint
 
-      - uses: paambaati/codeclimate-action@v5.0.0
+      - name: Run Coverage Report
```
mattseddon, line 45: This is the easiest way to bypass uploading a coverage report/storybook for pull requests from a fork. There would be value in getting both reports but the effort required to set that up along with the potential security risk (if we added extra secrets/forget) makes this the logical choice for now IMO.

mattseddon: [source](https://github.com/orgs/community/discussions/25217#discussioncomment-3246904)

mattseddon:
The other route is along the lines of https://github.com/iterative/vscode-dvc/pull/2644

julieg18 (approved):
Looks good!

sroy3 (approved):

PR #4768, opened 2023-10-04
Add remove as inline SCM action for added resources

mattseddon:
# 2/2 `main` <- #4769 <- this

Based on feedback provided in a [customer call](https://iterativeai.slack.com/archives/C01SR9Q12LB/p1696347929524469).

We did already have the move command wired up in the `DVC Tracked` tree but it not only removed the file/folder but deleted it as well. 

I have removed the part of the action that deleted the file after removing it from DVC and added the action inline to the SCM tree for "added" resources.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/1dc61fd2-c99e-46fb-9698-69a661fcbe09

https://github.com/iterative/vscode-dvc/assets/37993418/b108cba2-7e19-4658-956f-22a42854de3d

cc @dberenbaum

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -345,7 +345,7 @@
         "title": "Remove",
         "command": "dvc.removeTarget",
         "category": "DVC",
-        "icon": "$(trash)"
+        "icon": "$(discard)"
       },
       {
         "title": "Rename",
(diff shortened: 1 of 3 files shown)


julieg18 (approved):
Looks good!

sroy3 (approved):

PR #4770, opened 2023-10-04
Improve Plot Wizard Error Handling

julieg18:
* breaks up functions further
* adds filenames to most toast error messages
* adds a check for the same array length in multi file selection

## Demo

WIP

Followup to #4748

### DIFF
--- a/extension/src/fileSystem/index.ts
+++ b/extension/src/fileSystem/index.ts
@@ -342,17 +342,14 @@ const loadDataFile = (file: string): unknown => {
 
 export const loadDataFiles = async (
   files: string[]
-): Promise<{ file: string; data: unknown }[] | undefined> => {
+): Promise<{ file: string; data: unknown }[]> => {
   const filesData: { file: string; data: unknown }[] = []
+
   for (const file of files) {
     const data = await loadDataFile(file)
-
-    if (!data) {
-      return undefined
-    }
-
     filesData.push({ data, file })
   }
+
   return filesData
 }
 
--- a/extension/src/pipeline/quickPick.ts
+++ b/extension/src/pipeline/quickPick.ts
@@ -13,6 +13,16 @@ import { pickFiles } from '../vscode/resourcePicker'
 import { Title } from '../vscode/title'
 import { Toast } from '../vscode/toast'
 
+export type PlotConfigData = {
+  x: { file: string; key: string }
+  template: string
+  y: { file: string; key: string }
+}
+
+type UnknownValue = Value | ValueTree
+
+type fileFields = { [file: string]: string[] }
+
 const pickDataFiles = (): Thenable<string[] | undefined> =>
   pickFiles(Title.SELECT_PLOT_DATA, {
     'Data Formats': ['json', 'csv', 'tsv', 'yaml']
@@ -20,9 +30,7 @@ const pickDataFiles = (): Thenable<string[] | undefined> =>
 
 const pickTemplateAndFields = async (
   cwd: string,
-  fields: {
-    [file: string]: string[]
-  }
+  fields: fileFields
 ): Promise<PlotConfigData | undefined> => {
   const template = await quickPickOne(PLOT_TEMPLATES, 'Pick a Plot Template')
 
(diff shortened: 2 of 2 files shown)


extension/src/fileSystem/index.ts
```diff
@@ -214,31 +214,27 @@ const loadYamlAsDoc = (
...
 
-const getPlotYamlObj = (cwd: string, plot: PlotConfigData) => {
+const getPlotYamlObj = (plot: PlotConfigData) => {
   const { x, y, template } = plot
   const plotName = `${template}_plot`
   return {
     [plotName]: {
       template,
-      x: x.file === y.file ? x.key : { [relative(cwd, x.file)]: x.key },
-      y: { [relative(cwd, y.file)]: y.key }
```
julieg18, line 224: Updated `file` to be the relative path instead of the absolute path since we need the relative path for error messages and for quick pick labels in `pickPlotConfiguration`.

extension/src/pipeline/quickPick.ts
```diff
@@ -61,109 +68,160 @@ const pickTemplateAndFields = async (
...
   )
-  return objsHaveSameKeys ? fieldObjKeys : []
+  if (!objsHaveSameKeys) {
+    return
+  }
+  return { arrLength: dataArr.length, fields: fieldObjKeys }
 }
 
-const getFieldsFromValue = (data: UnknownValue): string[] => {
+const getMetricInfoFromValue = (
```
julieg18, line 113: Instead of just fields, we need to get the length of the array found inside of the file data as well. I renamed the function from `getFieldsFromValue` to `getMetricInfoFromValue` since `getFieldsAndArrLengthFromValue` looked pretty lengthy :)

extension/src/pipeline/quickPick.ts
```diff
@@ -61,109 +68,160 @@ const pickTemplateAndFields = async (
...
+
+    relativeFilesData.push({ data, file: relativeFile })
+  }
+
+  return relativeFilesData.length === 1
+    ? validateSingleFileData(relativeFilesData[0])
+    : validateMultiFilesData(relativeFilesData)
+}
+
 export const pickPlotConfiguration = async (
```
julieg18, line 208: Summary of error handling:

* Check if file(s) have the same extension
*  Check if file(s) have data
*  If Single File
*  * Check if file has at least two fields to pick from
* If Multiple Files
* * Check if each file has at least one field to pick from (toast is thrown at first invalid file found)
* * Check if all files have the same array length

extension/src/pipeline/quickPick.test.ts
```diff
@@ -70,30 +70,50 @@ describe('pickPlotConfiguration', () => {
...
+    expect(mockedShowError).toHaveBeenNthCalledWith(
+      2,
+      'Found files with .json, .csv, and .tsv extensions. Files must be of the same type.'
+    )
   })
 
-  it('should show a toast message if the file or files fail to parse', async () => {
-    mockedPickFiles.mockResolvedValueOnce(['data.csv'])
-    mockedLoadDataFiles.mockResolvedValueOnce(undefined)
+  it('should show a toast message if a file to parse', async () => {
```
mattseddon, line 99: [C] parsing a file failed

extension/src/pipeline/quickPick.ts
```diff
@@ -61,109 +68,160 @@ const pickTemplateAndFields = async (
...
+  filesData: { data: UnknownValue; file: string }[]
+) => {
+  const filesArrLength: Set<number> = new Set()
+  const keys: FileFields = []
+
+  for (const { file, data } of filesData) {
+    const metricInfo = getMetricInfoFromValue(data)
+    if (!metricInfo) {
+      void Toast.showError(
+        `${file} does not contain any keys (columns) to generate a plot. ${dvcPlotGuidelinesQuestion}`
```
mattseddon, line 163: [C] I don't think you need to split out no keys from < 2 keys. `${file} does not contain enough keys (columns) to generate a plot. ${dvcPlotGuidelinesQuestion}` is enough for both situations.

mattseddon (approved):

extension/src/pipeline/quickPick.ts
julieg18: Done :)

PR #4786, opened 2023-10-09
Add title option to plot wizard

julieg18:


julieg18:
> The test coverage on the diff in this pull request is 86.3% (85% is the threshold).

Not sure why the test coverage is so low, everything seems to be tested for unless I'm missing something 🤔

extension/src/pipeline/quickPick.ts
```diff
@@ -28,15 +30,27 @@ const pickDataFiles = (): Promise<string[] | undefined> =>
...
+const pickTemplateAndTitle = async (): Promise<
+  Omit<PlotConfigData, 'x' | 'y'> | undefined
+> => {
   const template = await quickPickOne(PLOT_TEMPLATES, 'Pick a Plot Template')
 
   if (!template) {
     return
   }
 
+  const title = await getInput(Title.ENTER_PLOT_TITLE, `${template}_plot`)
```
mattseddon, line 42: [C] If you move this to after the selection of files/variables then we could provide a more meaningful default.

mattseddon: i.e `${x} vs ${y}` or `${filename}`.

julieg18: > [C] If you move this to after the selection of files/variables then we could provide a more meaningful default.
> i.e ${x} vs ${y} or ${filename}.

We could do that, but wouldn't we have to add extra checks if the user has selected multiple files or variables? Seems simpler to just keep a basic default, but if you think it would be valuable to have the name be more meaningful, I can update the code to use an updated title if the user has only chosen two metrics 🤔

mattseddon: IMO `${x[0]} vs ${y[0]}` would still be a better default and not too hard to do. `${template}_plot` really isn't meaningful.

mattseddon: An alternative would be to not provide a default and let the user skip the title step by entering empty string. Would need to make this clear in the UI.

mattseddon (approved):

sroy3 (approved):

PR #4795, opened 2023-10-09
Remove banner from README

mattseddon:
This is what users currently see when they visit the marketplace listing/GH:

Screenshot 2023-10-10 at 9 19 07 am

image

Removing the banner will bring the GIF into view. LMK what you think.


julieg18 (approved):
Makes sense to me! Most marketplace extensions I see don't have a banner at the top either.

demo
sroy3, line 1: I don't think this change is meant to be here.

sroy3 (approved):

PR #4797, opened 2023-10-10
Add Multiple X Field Selection to Plot Wizard

julieg18:
# 2/3 `main` <- #4787 <- this <- #4798

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/52b9a5ac-5c1e-4732-9a83-78bd01f9a4f7

Part of #4654


extension/src/vscode/quickPick.ts
```diff
@@ -219,6 +220,55 @@ export const quickPickLimitedValues = <T>(
...
+      orderedSelection.push(item.value)
+    }
+    itemValues.push(item.value)
+  }
+  return orderedSelection.filter(item =>
+    itemValues.some(val => isEqual(item, val))
+  )
+}
+
+export const quickPickUserOrderedValues = <T>(
```
julieg18, line 239: How could I test this function? Looked into creating a vscode test but can't figure out a way to trigger `onDidChangeSelection` in the test 🤔

julieg18: Figured it out! Took way to long to find the `workbench.action.quickPickManyToggle` command :woman_facepalming:

extension/src/pipeline/quickPick.ts
```diff
@@ -79,16 +75,16 @@ const pickFields = async (
...
 
-  if (!xValue) {
+  if (!xValues || xValues.length === 0) {
     return
   }
 
-  const yValues = (await quickPickManyValues(
-    items.filter(item => !isEqual(item.value, xValue)),
+  const yValues = (await quickPickUserOrderedValues(
+    items.filter(item => xValues.every(val => !isEqual(val, item.value))),
```
julieg18, line 87: Is there a simpler way to filter out items that are in `xValues`? Right now we've basically got a nested loop 🤔

extension/src/pipeline/quickPick.test.ts
```diff
@@ -482,40 +560,44 @@ describe('pickPlotConfiguration', () => {
...
     mockedLoadDataFiles.mockResolvedValue([
       { data: mockValidData, file: 'file.json' }
     ])
     mockedQuickPickOne.mockResolvedValue('simple')
     mockedGetInput.mockResolvedValue('simple_plot')
-    mockedQuickPickValue.mockResolvedValue('actual')
-    mockedQuickPickValues
+    mockedQuickPickUserOrderedValues
       .mockResolvedValueOnce(undefined)
       .mockResolvedValueOnce([])
```
julieg18, line 592: This test is actually incorrectly written, fixed in #4798

extension/src/pipeline/quickPick.ts
```diff
@@ -8,18 +8,14 @@ import {
...
-} from '../vscode/quickPick'
+import { quickPickOne, quickPickUserOrderedValues } from '../vscode/quickPick'
 import { pickFiles } from '../vscode/resourcePicker'
 import { Title } from '../vscode/title'
 import { Toast } from '../vscode/toast'
 import { getInput } from '../vscode/inputBox'
 
 export type PlotConfigData = {
-  x: { [file: string]: string }
+  x: { [file: string]: string[] | string }
```
mattseddon, line 18: [Q] Why do we have a mixed data type here? Why not always have a string array?

extension/src/vscode/quickPick.ts
```diff
@@ -219,6 +220,55 @@ export const quickPickLimitedValues = <T>(
     quickPick.show()
   })
 
+const getUserOrderedSelection = <T>(
+  orderedSelection: T[],
+  newSelection: readonly QuickPickItemWithValue<T>[]
+) => {
+  const itemValues: T[] = []
+  for (const item of newSelection) {
```
mattseddon, line 228: [C] This function is really dense. Can we break into `addItems`/`removeItems`?

I.e something like
```
itemValues = addNewItems(...)
return removeMissingItems(...)
```

mattseddon: (definitely try to improve on the code provided at speed as a comment in a pull request)

mattseddon (approved):

extension/src/pipeline/quickPick.ts
julieg18: I chose to have a mixed data type so we could need to loop over `x`/`y` fields twice, one over here and again in the yaml plot obj. But I see that it would simplify things if we just looped over it twice. I'll update the code to just use a string array.

sroy3 (approved):

PR #4798, opened 2023-10-10
Add error handling for Plot Wizard plots with x fields

julieg18:
# 3/3 `main`<- #4787 <- #4797 <- this

* If the user selects more than one x field, we check if there amount of x fields and y fields chosen are the same amount and we check that the user hasn't selected more than one field in a file

### Check for only metric per file

https://github.com/iterative/vscode-dvc/assets/43496356/cc342b69-9647-44b1-a647-1cae75059063

### Check for x and y having the same amount of metrics

https://github.com/iterative/vscode-dvc/assets/43496356/21beb866-db78-4345-8fe3-7eac69ce6cb6

Part of #4654


extension/src/pipeline/quickPick.test.ts
```diff
@@ -588,17 +617,109 @@ describe('pickPlotConfiguration', () => {
     mockedQuickPickOne.mockResolvedValue('simple')
     mockedGetInput.mockResolvedValue('simple_plot')
     mockedQuickPickUserOrderedValues
+      .mockResolvedValueOnce([
```
julieg18, line 620: This test was actually incorrectly written. See https://github.com/iterative/vscode-dvc/pull/4797/files#r1355869332

extension/src/pipeline/quickPick.test.ts
```diff
@@ -580,6 +580,35 @@ describe('pickPlotConfiguration', () => {
     expect(mockedQuickPickUserOrderedValues).toHaveBeenCalledTimes(2)
     expect(noFieldsResult).toStrictEqual(undefined)
   })
+
+  it('should show a toast if user selects more than one key in a file for an x field', async () => {
```
julieg18, line 584: At this point, we've got around 20 different tests in this `describe` block. I'd like to break group these tests to make things more readable but I'll do it in a separate pr since that would involve touching the whole file. Updated #4654.

mattseddon: [Q] What is the problem with having 20 tests in a block?

extension/src/pipeline/quickPick.ts
```diff
@@ -52,6 +52,69 @@ const formatFieldQuickPickValues = (
   return formattedFields
 }
 
+const verifyFilesHavingSingleField = (files: {
```
mattseddon, line 55: [N] `verifyFilesHaveSingleField`

extension/src/pipeline/quickPick.ts
```diff
@@ -79,26 +142,31 @@ const pickFields = async (
...
   return {
-    fields: {
-      x: formatFieldQuickPickValues(xValues),
-      y: formatFieldQuickPickValues(yValues)
-    },
-    firstXKey: xValues[0].key,
-    firstYKey: yValues[0].key
+    fields: { x, y },
+    firstXKey,
+    firstYKey
```
mattseddon, line 169: [Q] Do we really need to return `firstXKey` and `firstYKey` from this function? Could we simplify the type and extract the first keys in the caller?

extension/src/pipeline/quickPick.ts
```diff
@@ -52,6 +52,69 @@ const formatFieldQuickPickValues = (
   return formattedFields
 }
 
+const verifyFilesHavingSingleField = (files: {
+  [file: string]: string[] | string
+}) => {
+  for (const [file, fields] of Object.entries(files)) {
+    if (Array.isArray(fields)) {
+      void Toast.showError(
+        `${file} cannot have more than one metric selected. See [an example](https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#available-configuration-fields) of a plot with multiple x metrics.`
```
mattseddon, line 61: [C] `See [an example](https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#available-configuration-fields) of a plot with multiple x metrics` is repeated.

extension/src/pipeline/quickPick.ts
```diff
@@ -52,6 +52,69 @@ const formatFieldQuickPickValues = (
   return formattedFields
 }
 
+const verifyFilesHavingSingleField = (files: {
+  [file: string]: string[] | string
+}) => {
+  for (const [file, fields] of Object.entries(files)) {
+    if (Array.isArray(fields)) {
```
mattseddon, line 59: [C] If the data type was consistent (always an array) then we could check the length here.

extension/src/pipeline/quickPick.ts
```diff
@@ -52,6 +52,69 @@ const formatFieldQuickPickValues = (
...
+
+  const y = formatFieldQuickPickValues(yValues)
+
+  if (xFieldsLength === 1) {
+    return { firstYKey: yValues[0].key, y }
+  }
+
+  if (yValues.length !== xFieldsLength) {
+    void Toast.showError(
+      'The amount of y metrics needs to match the amount of x metrics. See [an example](https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#available-configuration-fields) of a plot with multiple x metrics.'
```
mattseddon, line 104: [C] When there are multiple x metrics selected there must be an equal number of y metrics.

[Q] Also, did you consider using a version of `quickPickLimitedValues` when there are multiple xs? That would avoid the need for this validation. Thinking out loud => maybe we could have a quick pick that validates the number of items selected in the onDidAccept callback?

mattseddon (approved):

extension/src/pipeline/quickPick.ts
```diff
@@ -52,6 +52,69 @@ const formatFieldQuickPickValues = (
...
+    )
+    return
+  }
+
+  const doFilesHaveOneField = verifyFilesHavingSingleField(y)
+
+  if (!doFilesHaveOneField) {
```
mattseddon, line 111: [B] Whilst I believe this is correct for x values this is not correct for ys. This is a valid plot:

```
  - Loss:
      x: step
      y:
        training/plots/metrics/test/acc.tsv: [acc, step]
      y_label: loss
```

extension/src/pipeline/quickPick.ts
mattseddon: [F] The CLI error is `WARNING: In 'workspace', cannot have different number of x and y data sources. found 2 x and 4 y data sources.`

extension/src/pipeline/quickPick.ts
mattseddon: so is this 
```
  - Loss:
      x: step
      y:
        training/plots/metrics/train/acc.tsv: [acc, step]
```

extension/src/pipeline/quickPick.test.ts
julieg18: Not necessarily a problem but I find it easier to navigate/update a test file when tests are grouped. Looking at this file, we could group file picking and field picking into its own blocks. Though, of course, no strong preference, I'm happy to leave it as is if preferred.

extension/src/pipeline/quickPick.ts
julieg18: > [B] Whilst I believe this is correct for x values this is not correct for ys. 

We only check for this when there are multiple `x` values. When we have an x dict, the `y` and `x` need to match format-wise.

extension/src/pipeline/quickPick.ts
```diff
@@ -52,6 +52,69 @@ const formatFieldQuickPickValues = (
...
+
+const verifyYFields = (
+  yValues: { file: string; key: string }[] | undefined,
+  xFieldsLength: number
+) => {
+  if (!yValues || yValues.length === 0) {
+    return
+  }
+
+  const y = formatFieldQuickPickValues(yValues)
```
sroy3, line 96: There is a lot of repetition between `verifyXFields`and `verifyYFields` that could be extracted into a helper function.

extension/src/pipeline/quickPick.ts
```diff
@@ -79,26 +142,31 @@ const pickFields = async (
...
+  const { x, firstXKey } = xFieldInfo
+
   const yValues = (await quickPickUserOrderedValues(
-    items.filter(item => xValues.every(val => !isEqual(val, item.value))),
+    items.filter(item => xValues?.every(val => !isEqual(val, item.value))),
     { title: Title.SELECT_PLOT_Y_METRIC }
   )) as { file: string; key: string }[] | undefined
 
-  if (!yValues || yValues.length === 0) {
+  const yFieldInfo = verifyYFields(yValues, xValues?.length || 0)
```
sroy3, line 158: `verifyYFields` should have this signature instead: `(yValues: { file: string; key: string }[] | undefined, xFieldsLength = 0)`

sroy3 (approved):

extension/src/pipeline/quickPick.ts
julieg18: > When we have an x dict, the y and x need to match format-wise. I'll adjust this code to make this more clear :)

Actually, after testing `x dict`s further, this doesn't appear to be true 🤦‍♀️... I'm getting mixed results when I try to test out `x dict` plots and I'm no longer sure what we actually need to check for 😅. I'll just leave in a check for the same amount of fields and followup once I understand plots with multiple x fields better.

julieg18: Discussed with Matt synchronously, we're going to keep the extra check for y files having a single metric when there are multiple x metrics.

julieg18:
Going to merge and improve validation for the y metrics quick pick (https://github.com/iterative/vscode-dvc/pull/4798#discussion_r1355908616) in a followup pr!

PR #4808, opened 2023-10-11
Fix error when user chooses no Y fields in the Plot Wizard

julieg18:
Realized that a user could possibly select 0 fields when it comes to choosing y fields. I added a extra check so that the extension doesn't throw an error:

Screenshot 2023-10-11 at 11 07 12 AM

### DIFF
--- a/extension/src/pipeline/quickPick.ts
+++ b/extension/src/pipeline/quickPick.ts
@@ -92,7 +92,7 @@ const pickFields = async (
     { title: Title.SELECT_PLOT_Y_METRIC }
   )) as { file: string; key: string }[] | undefined
 
-  if (!yValues) {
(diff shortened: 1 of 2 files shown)


extension/src/pipeline/quickPick.ts
```diff
@@ -92,7 +92,7 @@ const pickFields = async (
     { title: Title.SELECT_PLOT_Y_METRIC }
   )) as { file: string; key: string }[] | undefined
 
-  if (!yValues) {
+  if (!yValues || yValues.length === 0) {
```
julieg18, line 95: We could also add a toast message if the user selects 0 y metrics if preferred but I thought it would be obvious enough :)

sroy3 (approved):

PR #4810, opened 2023-10-11
Deeply nested repo

### LINKED ISSUE #4775: Allow for deeply nested DVC repositories inside of monorepos
Some monorepos have DVC repositories more deeply nested in the structure than the few levels where we currently are looking for them. 

Currently, the extension will only check for sub-projects that are direct children of the workspace root. This is because visiting all subfolders would be too slow, potentially making the extension feel buggy.

Some possible solutions would include using the `findFiles` function, which should relies on VS Code's `ripgrep` under the hood, leveraging the git index to find the directories, or possibly having a setting to indicate deeply nested DVC repositories.

sroy3:
https://github.com/iterative/vscode-dvc/assets/3683420/1f9c848e-2f20-45e8-9410-f2c40a8d38f5

Closes #4775


extension/src/fileSystem/index.test.ts
```diff
@@ -302,15 +305,44 @@ describe('findDvcRootPaths', () => {
...
+    }
+    const mockSecondDvcRoot = {
+      fsPath: join(mockSecondDvcRootPath, DOT_DVC, 'config')
+    }
+
+    mockedFindFiles.mockResolvedValue([mockFirstDvcRoot, mockSecondDvcRoot])
+
+    const dvcRoots = await findDvcRootPaths(parentDir)
+
+    expect([...dvcRoots]).toStrictEqual([
```
mattseddon, line 340: [Q] What's the reason for this spread?

extension/src/fileSystem/index.ts
```diff
@@ -76,6 +77,18 @@ export const findDvcRootPaths = async (cwd: string): Promise<string[]> => {
...
     dvcRoots.push(...subRoots)
+  } else {
+    const nested = await findFiles(FULLY_NESTED_DVC)
+    if (definedAndNonEmpty(nested)) {
+      dvcRoots.push(
+        ...nested.map(nestedRoot =>
+          parse(nestedRoot)
+            .dir.split(sep)
+            .filter(part => part !== DOT_DVC)
+            .join(sep)
```
mattseddon, line 88: [Q] Could you use `dirname(dirname(nestedRoot))` here instead or `resolve(nestedRoot,'..', '..')?

The code would look something like this:

```
const nestedConfigs = await findFiles(FULLY_NESTED_DVC)
if (definedAndNonEmpty(nested)) {
  dvcRoots.push(
    ...nestedConfigs.map(config => resolve(config,'..','..')
  )
}
```
I think we also want to call `standardizePath` on the result because of 🪟s.

mattseddon: [Q] Do we just want to call this `else` instead of the rest of the function? Will it capture everything? What stops us from ending up with duplicates at the moment?

mattseddon (approved):

julieg18 (approved):

extension/src/fileSystem/index.ts
sroy3: About duplication, `findDvcRootPaths` is only used by  `Setup`'s `findWorkspaceDvcRoots` method and duplication is handled there:

```
const workspaceFolderRoots = await findDvcRootPaths(workspaceFolder)
if (definedAndNonEmpty(workspaceFolderRoots)) {
  dvcRoots = new Set([...dvcRoots, ...workspaceFolderRoots])
  continue
}
```
It is handled there in a loop over workspace folders.

sroy3: I'll rework the whole thing because I think I can remove some more code by going that route.

extension/src/fileSystem/index.test.ts
sroy3: Must have been because it was a `Set` before.

PR #4819, opened 2023-10-12
Ignore every WebdriverIO dependency in Renovate

sroy3:
Since #4806 did not work, I added every dependency instead.


mattseddon (approved):

mattseddon:
Does not look like this worked either 😕

PR #4827, opened 2023-10-13
Improve validation of Plot Wizard Y metrics quick pick

julieg18:
* If there are multiple x metrics, the quick pick has a different title and has a max number of items

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/570efac8-f635-4706-9f72-c247b0339823

Followup to [#4789 comment](https://github.com/iterative/vscode-dvc/pull/4798#discussion_r1355908616)

### DIFF
--- a/extension/src/pipeline/quickPick.ts
+++ b/extension/src/pipeline/quickPick.ts
@@ -8,7 +8,11 @@ import {
   isValueTree
 } from '../cli/dvc/contract'
 import { getFileExtension, loadDataFiles } from '../fileSystem'
-import { quickPickOne, quickPickUserOrderedValues } from '../vscode/quickPick'
+import {
+  QuickPickItemWithValue,
+  quickPickOne,
+  quickPickUserOrderedValues
+} from '../vscode/quickPick'
 import { pickFiles } from '../vscode/resourcePicker'
 import { Title } from '../vscode/title'
 import { Toast } from '../vscode/toast'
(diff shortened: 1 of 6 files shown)


extension/src/pipeline/quickPick.ts
```diff
@@ -86,14 +90,27 @@ const verifyXFields = (xValues: QuickPickFieldValues | undefined) => {
...
+  yItems: QuickPickItemWithValue<{ file: string; field: string } | undefined>[],
   xFieldsLength: number
 ) => {
-  if (yValues.length !== xFieldsLength) {
+  const yValues = (await quickPickUserOrderedValues(
+    yItems,
+    {
+      title: `Select ${xFieldsLength} Metrics for Y` as Title
+    },
+    xFieldsLength
```
julieg18, line 102: One issue with the quick pick is that we lose separators (aka file names) when the items are collapsed:

image

extension/src/pipeline/quickPick.ts
```diff
@@ -160,28 +172,29 @@ const pickFields = async (
...
+  const xValuesLength = xValues?.length || 0
 
-  const y = verifyYFields(yValues, xValues?.length)
+  const yItems = items.filter(
+    item => item.value === undefined || !xValues?.includes(item.value)
+  )
+  const y =
+    xValuesLength > 1
+      ? await pickYFieldsWithMultiXFields(yItems, xValuesLength)
+      : await pickYFields(yItems)
```
julieg18, line 183: With this change, y fields are picked and validated differently depending on if there are multiple x values.

sroy3 (approved):

PR #4840, opened 2023-10-16
Add a tooltip to plots with long titles that are cut by Vega

sroy3:
### Demo

https://github.com/iterative/vscode-dvc/assets/3683420/63cd1c63-08ec-47a6-9a14-9f7a0abb3f3f

### DIFF
--- a/webview/src/plots/components/ZoomablePlotWrapper.tsx
+++ b/webview/src/plots/components/ZoomablePlotWrapper.tsx
@@ -1,0 +1,21 @@
+import React, { ReactElement, PropsWithChildren } from 'react'
+import Tooltip from '../../shared/components/tooltip/Tooltip'
+
+interface ZoomablePlotWrapperProps {
+  title?: string
+  id: string
+}
+
+export const ZoomablePlotWrapper: React.FC<
+  PropsWithChildren<ZoomablePlotWrapperProps>
+> = ({ title, id, children }) => {
+  const isTitleCut = title?.indexOf('…') === 0
+
+  return isTitleCut ? (
+    <Tooltip content={id} placement="top" interactive>
+      {children as ReactElement}
+    </Tooltip>
+  ) : (
+    children
+  )
+}
--- a/webview/src/plots/components/App.test.tsx
+++ b/webview/src/plots/components/App.test.tsx
@@ -2637,6 +2637,78 @@ describe('App', () => {
       expect(clickEvent.stopPropagation).toHaveBeenCalledTimes(1)
     })
 
+    it('should have a tooltip on the plot if the title is cut', () => {
+      const title = 'Plot with a long title'
+      renderAppWithOptionalData({
+        template: {
+          ...templatePlotsFixture,
+          plots: [
+            {
+              entries: [
+                {
+                  ...templatePlotsFixture.plots[0].entries[0],
+                  content: {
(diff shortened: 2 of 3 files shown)


webview/src/plots/components/ZoomablePlotWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface ZoomablePlotWrapperProps {
+  title?: string
+  id: string
+}
+
+export const ZoomablePlotWrapper: React.FC<
+  PropsWithChildren<ZoomablePlotWrapperProps>
+> = ({ title, id, children }) => {
+  const isTitleCut = title?.indexOf('…') === 0
```
mattseddon, line 12: TIL we aren't doing this in code.

mattseddon: no wait.... we are doing this in code but you can bypass because the id is the title, got it

mattseddon (approved):
We should be able to extend this for any truncation when the big template update lands.

I.e tooltip could contain

Title: ${title}
y-axis: ${yLabel}
...
if any are truncated.

webview/src/plots/components/ZoomablePlotWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  return isTitleCut ? (
+    <Tooltip content={id} placement="top" interactive>
```
mattseddon, line 15: Please confirm that having a tooltip in place does not break the ability to zoom the plot 🙏🏻

webview/src/plots/components/ZoomablePlotWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+      {children as ReactElement}
+    </Tooltip>
```
julieg18, line 17: The cursor changes to a grab icon when you hover over the tooltip: 

https://github.com/iterative/vscode-dvc/assets/43496356/d778d6a0-d8c7-46d0-b8da-f31fac51d704

julieg18 (approved):
Great work!

tapadipti:
Thanks @sroy3 for addressing this so promptly.
Quick comment: I feel like it would be helpful if all the plot titles (including short ones) have tooltips. Just so that the user realizes that the titles have tooltips. wdyt?

webview/src/plots/components/ZoomablePlotWrapper.tsx
sroy3: Fixed! Good catch!

sroy3:
> We should be able to extend this for any truncation when the big template update lands.
> 
> I.e tooltip could contain
> 
> ```
> Title: ${title}
> y-axis: ${yLabel}
> ...
> ```
> 
> if any are truncated.

I can do a follow up for this.

sroy3:
> Thanks @sroy3 for addressing this so promptly. Quick comment: I feel like it would be helpful if all the plot titles (including short ones) have tooltips. Just so that the user realizes that the titles have tooltips. wdyt?

I feel like this would pollute the UI without adding any real value. Especially if all your plots have short titles.

The tooltip appears instantly when hovering a plot with a title that's been cut off. It is a common UX pattern to hover to view missing information, I don't think people will feel lost about this.

Issue #4836, opened 2023-10-16
DVCLive API code snippet in notebooks


tibor-mach:
In the extension -> setup -> experiments we have the following example under Python API:

```
import random
import sys
from dvclive import Live

with Live(save_dvc_exp=True) as live:
    epochs = int(sys.argv[1])
    live.log_param("epochs", epochs)
    for epoch in range(epochs):
        live.log_metric("train/accuracy", epoch + random.random())
        live.log_metric("train/loss", epochs - epoch - random.random())
        live.log_metric("val/accuracy",epoch + random.random() )
        live.log_metric("val/loss", epochs - epoch - random.random())
        live.next_step()
```

This works nicely if you copy it to a script and run it from the command line. But if people expect to use this in a jupyter notebook (which they might with DVCLive) and then copy it to a cell and run it, this will fail since no arguments are received.

Maybe it would make sense to create an example which can be run in notebooks also? Just replacing `sys.argv[1]` with a positive integer constant should do it.

mattseddon:
See also https://dvc.org/doc/start/experiments/experiment-tracking?tab=General-Python-API#running

mattseddon:
The same snippet is also in Studio: https://github.com/iterative/studio/blob/492ba930d81cf8460c368286b26df4cb67957c07/frontend/src/composite/ProjectExperimentingGuide/GetStartedTabs.tsx#L30

mattseddon:
@tibor-mach there is more context/discussion in https://github.com/iterative/studio/issues/6422 / https://github.com/iterative/dvclive/pull/635 PTAL. If you have strong opinions contributions are welcome. Code for the extension is here: https://github.com/iterative/vscode-dvc/blob/c0cc75c9487913136355f9dd015965065cf27869/webview/src/setup/snippets/pythonApi.py#L1

and `dvc.org` is here:

https://github.com/iterative/dvc.org/blob/main/content/docs/start/experiments/experiment-tracking.md#L92

Studio link is shown above.

tibor-mach:
Thanks @mattseddon !

I agree with @omesser ([this](https://github.com/iterative/studio/issues/6422#issuecomment-1627822395) comment) especially on the sys.argv.

The first thing I tried without thinking about the snipped (or even really reading it, frankly) was to copy-paste it into a notebook and run it.

Obviously, it fails and with an error which is confusing at first (only at this point did I start actually reading the code of the snippet).

[...]

PR #4853, opened 2023-10-17
Fix DVCLive snippet in walkthrough

mattseddon:
Stumbled across this because of #4836 / #4836


extension/resources/walkthrough/live-plots.md
```diff
@@ -27,7 +27,7 @@ for epoch in range(NUM_EPOCHS):
     for metric_name, value in metrics.items():
         live.log(metric_name, value)
 
- live.next_step()
+    live.next_step()
```
mattseddon, line 30: [F] If this is not set in the loop metrics will be overwritten

sroy3 (approved):

PR #4857, opened 2023-10-18
Add created branch (from `dvc exp branch`) to the experiments table

julieg18:
* If a user creates a new branch using the extension, we add the newly created branch (if it exists) to the table

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/d9b2d9af-a707-4b5f-9fdc-259d904a5475

Fixes https://iterativeai.slack.com/archives/C01APS0FHDM/p1697466359466489?thread_ts=1697456789.149749&cid=C01APS0FHDM


extension/src/experiments/webview/messages.ts
```diff
@@ -248,6 +245,29 @@ export class WebviewMessages {
...
+    const branchName = await commands.executeCommand(
+      RegisteredCliCommands.EXPERIMENT_VIEW_BRANCH,
+      { dvcRoot: this.dvcRoot, id: expId }
+    )
+
+    if (!branchName) {
+      return
+    }
+
+    await this.update()
```
julieg18, line 258: We have to run `update` twice (once to make sure the branch is in `availableBranchesToSelect` and another time to update the table). 

We could run only branch updates on the first run (aka run only `data.updateBranches()` which is currently `private`) though this would mean more code to maintain.

extension/src/experiments/workspace.ts
```diff
@@ -250,7 +250,8 @@ export class WorkspaceExperiments extends BaseWorkspaceWebviews<
     if (!input) {
       return
     }
-    return runCommand(...args, input)
+    await runCommand(...args, input)
+    return input
```
julieg18, line 254: Updated `getInputAndRun` to return the input instead of the command result. I checked commands that use this util and they aren't using the result of the command anywhere.

extension/src/experiments/webview/messages.ts
```diff
@@ -248,6 +245,29 @@ export class WebviewMessages {
...
+
+    const selectedBranches = this.experiments.getSelectedBranches()
+    const branchesWithNewBranch = this.experiments
+      .getAvailableBranchesToSelect()
+      .filter(
+        branch => selectedBranches.includes(branch) || branch === branchName
+      )
+
+    this.experiments.setSelectedBranches(branchesWithNewBranch)
```
julieg18, line 267: We only update the table if the user runs the context menu action. If the user runs the command in the command palette or the tree, we don't update the table. Figured it only made sense to update the table if the user could see the table, but we could also always update if preferred.

extension/src/experiments/webview/messages.ts
mattseddon: I don't think you need this. You can get the current list of selected branches with `this.experiments.getSelectedBranches()`
and push the new branch in and use `setSelectedBranches` before calling `update`. If it does not exist then it will be filtered out, if it is there it should be fine.

extension/src/experiments/webview/messages.ts
mattseddon: Consistent behaviour is better, what if the tree is open, etc.

extension/src/experiments/workspace.ts
mattseddon: [Q] Can you return both? In this instance, we know that `runCommand` returns a string. I'd want to check if the response from `dvc` was non-empty before calling for updates.

extension/src/experiments/webview/messages.ts
mattseddon:
```
  public setSelectedBranches(branches: string[]) {
    this.selectedBranches = branches.filter(
      branch => branch !== this.currentBranch
    )
    this.persistBranchesToShow()
  }
```
☝🏻 there is nothing to filter the new branch here

```
  public setBranches(
    branches: string[],
    branchesToSelect: string[],
    currentBranch: string
  ) {
    this.availableBranchesToShow = branches
    this.availableBranchesToSelect = branchesToSelect
    this.currentBranch = currentBranch

    this.selectedBranches = this.selectedBranches.filter(
      branch => branches.includes(branch) && branch !== this.currentBranch
    )
    this.persistBranchesToShow()
  }
```
☝🏻 but there is here

```
  private async updateBranches() {
    const allBranches = await this.internalCommands.executeCommand<string[]>(
      AvailableCommands.GIT_GET_BRANCHES,
      this.dvcRoot
    )

    const { currentBranch, branches, branchesToSelect } =
      collectBranches(allBranches)

    this.experiments.setBranches(branches, branchesToSelect, currentBranch)
  }
```
so inserting the branch before calling update (which calls the above) should be fine.

.... please nail that down with a test

mattseddon (approved):
I'm not sold on the utility of this. A few months ago we started moving away from auto-magical behaviour as it unnecessarily complicates the code and leads to unexpected behaviour for users.

sroy3 (approved):
Good job! 

I'm not expecting what I will say next to be added to this PR. It could be a follow-up for later on if we agree on it.
I feel like there is a good amount of time passing between when we create the branch, when we show the toast and when we add the branch. 
It is a lot more advanced, but having some sort of loading at the bottom of the table from the time the user creates the branch to when we add the branch would feel less confusing for the user. Some message that says "`branch-name` is being added to the table."

extension/src/experiments/webview/messages.ts
julieg18: > You can get the current list of selected branches with this.experiments.getSelectedBranches()
and push the new branch in and use setSelectedBranches before calling update. If it does not exist then it will be filtered out, if it is there it should be fine.

We could do that! However we will need to `.sort()` the branch list to ensure that the branches are in the right order.

extension/src/experiments/webview/messages.ts
julieg18: Updated the behavior to always update the table!

extension/src/experiments/workspace.ts
julieg18: `getInputAndRun` no longer needs to be updated since we no longer need the `WebviewMessages` to have the input.

extension/src/experiments/webview/messages.ts
julieg18: Adjusted the code to just add the new branch with `setBranches` and `getSelectedBranches`!

julieg18:
> I feel like there is a good amount of time passing between when we create the branch, when we show the toast and when we add the branch.

@sroy3 , with the latest change (only needing to run an update once), the table runs a lot faster. Plus, my computer tends to run on the slower side. If you have a chance, could you test this feature out with the newest changes? If it feels slow, I can work on adding a Toast message that says something like `Adding branch to table...` at the very least.

sroy3:
> > I feel like there is a good amount of time passing between when we create the branch, when we show the toast and when we add the branch.
> 
> @sroy3 , with the latest change (only needing to run an update once), the table runs a lot faster. Plus, my computer tends to run on the slower side. If you have a chance, could you test this feature out with the newest changes? If it feels slow, I can work on adding a Toast message that says something like `Adding branch to table...` at the very least.

It does work almost instantly. Great work!

PR #4734, opened 2023-09-28
Bump min DVC version to 3.33.0 (standardise plots across products)

mattseddon:
Part of https://github.com/iterative/dvc/issues/9940

This PR makes the changes required to accept the new templates provided by `dvc-render` in https://github.com/iterative/dvc-render/pull/142. Made available through `dvc` in https://github.com/iterative/dvc/pull/9931.

The main difference in the approach is that extra anchors have been added to the default templates. These anchors will all be filled in the webview so that the values can be manipulated depending on whether or not the plot is being shown in the zoom modal. We will no longer be overwriting information in templates that do not have the appropriate anchors.

I have consolidated our plot's test fixtures as part of the change. I have been able to drop a few files as the bulk of the complexity has been moved into `dvc-render`.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/47d2f39a-7719-4f82-a313-65c3902a6f96

PR #4858, opened 2023-10-18
Plot tooltip for truncated titles

sroy3:
### Demo

https://github.com/iterative/vscode-dvc/assets/3683420/af210480-5c55-4423-ba3d-2333b0fc92d4

I'm unsure whether the tooltip should contain the whole thing every time and be there on every plot or not. That would certainly make the code simpler if we removed the part of the code that checks whether or not it is truncated.


extension/src/plots/vega/util.ts
```diff
@@ -262,11 +262,23 @@ export const truncateVerticalTitle = (
 const isEndValue = (valueType: string) =>
   ['string', 'number', 'boolean'].includes(valueType)
 
+type SpecTitle = {
```
sroy3, line 265: This could be removed if we always showed everything in the tooltip

mattseddon (approved):
Nice.

If you look at #4734 most of this logic gets deleted. I'll have to move the tooltip creation to the webview when I get back to it 🙏🏻.

PR #4869, opened 2023-10-20
Warn user before trying to add a stage to their pipeline

mattseddon:
Address [this feedback](https://iterativeai.slack.com/archives/C01APS0FHDM/p1697789220288419?thread_ts=1697667391.944309&cid=C01APS0FHDM):

"
When you click Add Plot -> Data Series  it opens the Finder where you can select files. In my case doesn’t work  :disappointed:  I see the next prompt like below. Do you know why is it difference?
"

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/9c06c820-2af9-4fbd-a1c2-000d9f18dc3d

Note: As demonstrated above this modal will be shown any time that a user tries to perform an action without a stage except adding the stage from the experiments webview.

### DIFF
--- a/extension/src/pipeline/index.ts
+++ b/extension/src/pipeline/index.ts
@@ -17,6 +17,8 @@ import { getInput, getValidInput } from '../vscode/inputBox'
 import { Title } from '../vscode/title'
 import { quickPickOne, quickPickOneOrInput } from '../vscode/quickPick'
 import { pickFile } from '../vscode/resourcePicker'
+import { Response } from '../vscode/response'
+import { Modal } from '../vscode/modal'
 
 export enum ScriptCommand {
   JUPYTER = 'jupyter nbconvert --to notebook --inplace --execute',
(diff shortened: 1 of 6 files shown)


extension/src/pipeline/index.ts
```diff
@@ -150,6 +152,16 @@ export class Pipeline extends DeferredDisposable {
   }
 
   private async addPipeline() {
+    const response = await Modal.showInformation(
```
mattseddon, line 155: [I] Actually need to move things around or bypass this for plots

mattseddon: I think the best thing to do would be break up `getCwd` (which checks for the stage) and move the check upfront for adding plots, along with a modal that tells the user what is going on ... come back to it on Monday

sroy3 (approved):

julieg18 (approved):
Looks good!

PR #4879, opened 2023-10-24
Fix `quickPickUserOrderedValues` windows test timing out

julieg18:
One of tests for `quickPickUserOrderedValues` is failing frequently with a timeout error in windows:

```
         quickPickUserOrderedValues
           should limit the number of values that can be selected to the max selected items:
       Error: Timeout of 6000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (d:\a\vscode-dvc\vscode-dvc\extension\dist\test\suite\vscode\quickPick.test.js)
```


extension/src/test/suite/vscode/quickPick.test.ts
```diff
@@ -200,11 +200,7 @@ suite('Quick Pick Test Suite', () => {
         { label: 'B', value: 2 },
         { label: 'C', value: 3 },
         { label: 'D', value: 4 },
-        { label: 'E', value: 5 },
```
julieg18, line 203: Looking at the test, I don't think there is a bug inside the test but there are a lot of commands being run in `selectMultipleQuickPickItems` due to needing to repeatedly loop over the items. Possibly why the test keeps hitting a timeout 🤔 

I've decreased the number of quick pick items and added `2000`ms to the timeout.

Issue #4867, opened 2023-10-20
Improve plots performance when there are many plots


sroy3:
With 200+ plots, updating in real-time takes a long time and even feels very buggy. Any other action that applies to all plots (resizing, changing the order...) takes more time than it should and makes things confusing to the user.

In the case of resizing, the problem seems to affect the virtualized grid (lots of re-calculating and re-rendering), while updating blocks even before rendering.

- [X] Remove drag and drop when not dragging (#4934)
- [x] Set the default number of plots selected to 20 (#5008 )
- [x] Remove drag and drop container and replace it with a hook (no element creation on the fly) (#5062)
- [x] Remove drag and drop when not dragging for the comparison table(#5017 )
- [ ] Remove drag and drop when not dragging for the comparison table rows
- [x] Do not render a section if it is closed (#5025)

sroy3:
The main bottleneck comes from `DragDropContainer`. We create a lot of functions in there and re-create all plots to add these functions. This causes a lot of re-rendering whenever we try to do something (update the plots, rearrange them, resize them...).

There is a good cleanup to be done in there. That and all the usual tricks to limit re-rendering (`useCallback`, `useMemo`, stop passing down complex props, stop re-creating functions...) should help a lot.

Cutting the performance lag completely while still using drag and drop will be challenging, though. One solution is to have a drag-and-drop mode. In this mode, the details of the plots will be hidden, showing only their title. It would be possible to move the plots freely in this mode. In normal mode, it would be impossible to change the order of the plots. It would be possible to toggle modes from the section header. This solution should increase the speed of actions on plots dramatically.

sroy3:
As @mattseddon mentioned in our last planning meeting, enabling the drag-and-drop when gripping the plot and disabling it on drop should work as a drag-and-drop toggle and enable us to render plots without drag and drops components most of the time for added performance.

sroy3:
[...]

PR #4880, opened 2023-10-24
Split drag and drop container into different files

sroy3:
This is to make #4867 easier

Demo to show that every thing is the same:

https://github.com/iterative/vscode-dvc/assets/3683420/29907c53-63ca-4e86-8c09-403eacc66873


mattseddon (approved):

julieg18 (approved):
Looks good!

Issue #4878, opened 2023-10-24
Feature request: Filter the experiment table by tag


francesco086:
## Short description

In the context of an model/artifact registry set with gto, it would be very handful to be able to filter the experiment table by tag.

## Example

For example, imagine that I have a model registry with the following artifacts

- base-churn-time-series
- base-churn-hierarchical-model
- acquisition-time-series

and therefore I have git tags created by gto, like

- base-churn-time-series@v1.0.0
- base-churn-time-series@v1.0.1
- base-churn-time-series#prod#1
- base-churn-hierarchical-model@v1.0.0
- base-churn-hierarchical-model@v2.0.0
- acquisition-time-series@v1.0.0
- acquisition-time-series#prod#1

and when I look at the experiment table, I would like to focus on type of artifact at a time. It would be nice to be able to filter by a regex or substring, e.g. "base-churn", or "acquisition", or "base-churn-time-series".

## Workarounds

Of course I could get around this issue by adding this information in the params.yaml, but it is somewhat redundant, I think.

Perhaps there are other ways I could not think of...

mattseddon:
Hi @francesco086,

Thanks for raising. I've opened #4882 to address this.

Here is a demo of the feature:

https://github.com/iterative/vscode-dvc/assets/37993418/1ab9a3e7-dd26-40f3-b8d9-7f170cf58b61

Please let me know if you have any feedback or would like any changes.

Thanks

francesco086:
Wow! That was fast!
And very good too! 

I have only one small suggestion for something to check, perhaps it's already implemented but I don't know the base code enough to tell. If you open the "Command Palette", and select "DVC: Add filter to experiments table", will you be prompted to filter by tag too?

mattseddon:
> If you open the "Command Palette", and select "DVC: Add filter to experiments table", will you be prompted to filter by tag too?

Yep, that is covered.

francesco086:
Then that's it, I'd say! 

Thank you so much, really impressive to see how fast this was addressed.

mattseddon:
The feature went out with `1.1.11`

PR #4882, opened 2023-10-24
Add tags filters to experiments table

mattseddon:
Closes https://github.com/iterative/vscode-dvc/issues/4878

This PR adds equals and contains filters to the experiments table for tags.

When the equals filter is applied each of the commits tags will be checked for equality when the contains filter is applied each of the commits tags will be checked for a substring.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/1ab9a3e7-dd26-40f3-b8d9-7f170cf58b61

### DIFF
--- a/extension/src/experiments/columns/like.ts
+++ b/extension/src/experiments/columns/like.ts
@@ -2,25 +2,35 @@ import { Column } from '../webview/contract'
 
 export type ColumnLike = {
   firstValueType?: string
+  description?: string
   label: string
   path: string
 }
 
-const starredColumnLike: ColumnLike = {
+export const starredColumnLike: ColumnLike = {
+  description: '$(star-full)',
   firstValueType: 'boolean',
-  label: '$(star-full)',
+  label: 'Starred',
   path: 'starred'
 }
 
-export const addStarredToColumns = (
-  columns: Column[] | undefined
+export const tagsColumnLike: ColumnLike = {
+  description: '$(git-commit)',
+  firstValueType: 'tags',
+  label: 'Git Tag',
+  path: 'Git Tag'
+}
+
+export const addToColumns = (
+  columns: Column[] | undefined,
+  ...columnLikes: ColumnLike[]
 ): ColumnLike[] | undefined => {
   if (!columns?.length) {
     return
   }
 
   return [
-    starredColumnLike,
+    ...columnLikes,
     ...columns.map(({ label, path, firstValueType }) => ({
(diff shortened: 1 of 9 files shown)


extension/src/experiments/index.ts
```diff
@@ -291,7 +296,7 @@ export class Experiments extends BaseRepository<TableData> {
 
   public async addSort() {
     const columns = this.columns.getTerminalNodes()
-    const columnLikes = addStarredToColumns(columns)
+    const columnLikes = addToColumns(columns, starredColumnLike)
```
mattseddon, line 299: [F] Don't want to allow sorting by a tags array so had to change the function

extension/src/experiments/columns/like.ts
```diff
@@ -2,25 +2,35 @@ import { Column } from '../webview/contract'
...
+  label: 'Starred',
   path: 'starred'
 }
 
-export const addStarredToColumns = (
-  columns: Column[] | undefined
+export const tagsColumnLike: ColumnLike = {
+  description: '$(git-commit)',
+  firstValueType: 'tags',
+  label: 'Tag',
```
julieg18, line 20: Should we say `Git Tag` or would it be obvious that the tags are related to git?

extension/src/experiments/model/filterBy/index.ts
```diff
@@ -105,12 +107,31 @@ const evaluate = <T extends string | number | boolean>(
...
+      evaluate<typeof value>(value, filter.operator, filter.value as string)
+    )
+  }
+}
+
 const buildFilter =
   (
     filterDefinitions: FilterDefinition[]
   ): ((experiment: Experiment) => boolean) =>
   experiment => {
```
julieg18, line 129: I noticed that experiments will always be hidden since they can't have tags. Is the behavior expected or should we be showing the experiments if their parent commit has a right tag?

julieg18 (approved):
Great work!

sroy3 (approved):
Good job! 
Maybe later we can add that clicking a tag filters to only that tag also.

extension/src/experiments/model/filterBy/index.ts
mattseddon: The user seems to be happy with what has been implemented: https://github.com/iterative/vscode-dvc/issues/4878#issuecomment-1778510087. Let's release this and see what feedback we get.

PR #4900, opened 2023-10-25
Fix unstable `quickPickUserOrderedValues` test in windows

julieg18:
Despite changes in #4879, test is still timing out:

```
    2) Quick Pick Test Suite
         quickPickUserOrderedValues
           should limit the number of values that can be selected to the max selected items:
       Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (d:\a\vscode-dvc\vscode-dvc\extension\dist\test\suite\vscode\quickPick.test.js)
    	at listOnTimeout (node:internal/timers:569:17)
    	at process.processTimers (node:internal/timers:512:7)
```

Saw this error message once as well:
```
    2) Quick Pick Test Suite
         quickPickUserOrderedValues
           should limit the number of values that can be selected to the max selected items:
  
        the max number of items are selected
        + expected - actual
  
        -2
        +3
        
    	at d:\a\vscode-dvc\vscode-dvc\extension\dist\test\suite\vscode\quickPick.test.js:147:105
    	at Generator.next (<anonymous>)
    	at fulfilled (d:\a\vscode-dvc\vscode-dvc\extension\dist\test\suite\vscode\quickPick.test.js:5:58)
```


extension/src/test/suite/vscode/quickPick.test.ts
```diff
@@ -161,7 +161,7 @@ suite('Quick Pick Test Suite', () => {
       const result = await resultPromise
 
       expect(result).to.deep.equal([5, 2, 1])
-    }).timeout(8000)
+    }).timeout(10000)
```
julieg18, line 164: Increased the timeout again, but I wonder if there is an issue in `selectMultipleQuickPickItems` that I could be missing: 

```javascript
const toggleQuickPickItem = async (number: number, itemsLength: number) => {
  for (let itemInd = 1; itemInd <= itemsLength; itemInd++) {
    await commands.executeCommand('workbench.action.quickOpenSelectNext')

    if (itemInd === number) {
      await commands.executeCommand('workbench.action.quickPickManyToggle')
    }
  }
}

export const selectMultipleQuickPickItems = async (
  numbers: number[],
  itemsLength: number,
  acceptItems = true
) => {
  for (const number of numbers) {
    await toggleQuickPickItem(number, itemsLength)
  }
  if (acceptItems) {
    return commands.executeCommand(
      'workbench.action.acceptSelectedQuickOpenItem'
    )
  }
}
```

extension/src/test/suite/vscode/quickPick.test.ts
```diff
@@ -208,8 +208,16 @@ suite('Quick Pick Test Suite', () => {
         { title: 'select up to 3 values' as Title },
         maxSelectedItems
       )
+      const selectionEvent = new Promise(resolve =>
+        quickPick.onDidChangeSelection(items => {
+          if (items.length === 3) {
```
julieg18, line 213: Going by this error:

```
    2) Quick Pick Test Suite
         quickPickUserOrderedValues
           should limit the number of values that can be selected to the max selected items:
  
        the max number of items are selected
        + expected - actual
  
        -2
        +3
        
    	at d:\a\vscode-dvc\vscode-dvc\extension\dist\test\suite\vscode\quickPick.test.js:147:105
    	at Generator.next (<anonymous>)
    	at fulfilled (d:\a\vscode-dvc\vscode-dvc\extension\dist\test\suite\vscode\quickPick.test.js:5:58)
```

`await executeCommand` could possibly not be resolving when the command is 100% finished. So I've tried to add an extra await event that checks for when there are three selected items.

julieg18: Or maybe the quick pick select commands aren't being run or being run correctly for some reason 🤔

julieg18 (commented):
Any ideas on why this test keeps failing are welcome!

extension/src/test/suite/vscode/quickPick.test.ts
mattseddon: I would wait for an update event each time you run `await commands.executeCommand('workbench.action.quickPickManyToggle')` otherwise I think the results will always be unstable

julieg18: > I would wait for an update event each time you run await commands.executeCommand('workbench.action.quickPickManyToggle') otherwise I think the results will always be unstable

Good point, updated the code to wait for an update event when `commands.executeCommand('workbench.action.quickPickManyToggle')` is called!

mattseddon (approved):

PR #4904, opened 2023-10-25
Remove check for stages in plot wizard

julieg18:
* updates the plot wizard to just create a missing dvc.yaml instead of having the user setup a stage first

## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/00287d37-808a-47bc-bab3-eb31713b8b7d


extension/src/fileSystem/index.ts
```diff
@@ -276,8 +277,9 @@ export const addPlotToDvcYamlFile = (cwd: string, plotObj: PlotConfigData) => {
   if (!plots?.range) {
     const plotYaml = getPlotsYaml(plotObj, dvcYamlLines)
     dvcYamlLines.push(...plotYaml)
-    writeFileSync(dvcYamlFile, dvcYamlLines.join('\n'))
-    return
```
julieg18, line 280: This "no plots found" route is actually missing the `openFileInEditor` line. I've added it and updated the tests to check for this function being called.

extension/src/fileSystem/index.ts
```diff
@@ -276,8 +277,9 @@ export const addPlotToDvcYamlFile = (cwd: string, plotObj: PlotConfigData) => {
...
+
+    void openFileInEditor(dvcYamlFile)
+    return writeFileSync(dvcYamlFile, dvcYamlLines.join('\n'))
```
julieg18, line 282: There is an empty line on top of the dvc.yaml in the case of there being an empty file:

```

plots:
  - step vs acc:
      template: confusion
      x: step
      y:
        training/plots/metrics/train/acc.json: acc

```

If we wanted to get rid of the top line, we could need to add an extra `if` check. I decided to just keep it to avoid extra complexity.

extension/src/pipeline/index.ts
```diff
@@ -158,11 +138,7 @@ export class Pipeline extends DeferredDisposable {
   }
 
   public async addDataSeriesPlot() {
-    const cwd = await this.getCwd()
-
-    if (!cwd) {
-      return
-    }
+    const cwd = (await this.findPipelineCwd()) || this.dvcRoot
```
julieg18, line 141: We now default to `this.dvcRoot` if a `cwd` isn't found.

extension/src/pipeline/index.ts
```diff
@@ -177,6 +153,29 @@ export class Pipeline extends DeferredDisposable {
...
   }
 
+  private findPipelineCwd() {
+    const focusedPipeline = this.getFocusedPipeline()
+    if (focusedPipeline) {
+      return focusedPipeline
+    }
+
+    const pipelines = this.model.getPipelines()
+    if (!pipelines?.size) {
```
sroy3, line 163: Out of scope, but since we have a lint rule to test zero length explicitly (`something.length === 0` instead of `!something.length`). Should we have a similar rule for `size` as well?

sroy3 (approved):

extension/src/pipeline/index.ts
julieg18: Makes sense to me. I can look into a eslint rule for it!

extension/src/pipeline/index.ts
```diff
@@ -83,28 +83,8 @@ export class Pipeline extends DeferredDisposable {
...
-    }
-    if (pipelines.size === 1) {
-      return [...pipelines][0]
-    }
-
-    return quickPickOne(
-      [...pipelines],
-      'Select a Pipeline to Run Command Against'
-    )
+    return this.findPipelineCwd()
```
julieg18, line 87: Going to mention it here since it was discussed asynchronously with Matt, but it would simplify things to just have the Plot Wizard automatically create a dvc.yaml in the root vs looking for one elsewhere. 

While this means that it wouldn't work for people with workspaces or nested dvc.yamls, users who do those are most likely advanced DVC users who don't have a need for the plot wizard. 

I chose to keep it like it is for now but I'm happy to update if preferred.

extension/src/pipeline/index.ts
julieg18: Looks like the lint rule we have (`explicit-link-check`) works for both arrays and size. This is allowed though since we have an undefined check there as well.

PR #4930, opened 2023-10-30
Remove unused dependencies

sroy3:
depcheks will still return:
```
Unused devDependencies
* @vscode/codicons
* eslint-plugin-n
* eslint-plugin-promise
* lint-staged
* prettier
```
Those are all used dependencies


package.json
```diff
@@ -61,7 +60,6 @@
     "lint-staged": "14.0.1",
     "npm-run-all": "4.1.5",
     "nyc": "15.1.0",
-    "postcss": "8.4.31",
```
mattseddon, line 64: [Q] Does stylelint definitely not require this? @julieg18 was the one who added it, maybe she remembers

mattseddon (approved):

package.json
sroy3: I'm not getting any errors or warnings when running `yarn lint:css`. I'll wait on @julieg18 to merge.

julieg18: I don't think stylelint requires it. It should work correctly without it. My guess is I accidentally added it when I was running get started scripts or I added it when I was troubleshooting.

julieg18 (approved):
Great work!

PR #4934, opened 2023-10-31
Add drag and drop mode to plots

sroy3:
# 1/3 main <- this <- #4942 <- #4960

Part of #4867 

Demo 

https://github.com/iterative/vscode-dvc/assets/3683420/3f51aa6f-7b91-467b-8468-930b24bf9448

I'll iterate on this PR before merging it. This is big enough as is without adding more functionality. To be added in future PRs merging back into this one:

- [ ] Style for the plot while dragging (I think the titles for custom plots aren't the right ones either, to verify)
- [x] Toggle modes while dragging (seamless with  what it used to be) (#4942 )

PR #4942, opened 2023-11-02
Seamless drag and drop mode

sroy3:
# 2/3 main <- #4934 <- this <- #4960 

Part of #4867 

Demo:

https://github.com/iterative/vscode-dvc/assets/3683420/72ae9e13-93ba-41c0-906e-465613d0160f


mattseddon (approved):

webview/src/plots/components/ZoomablePlot.tsx
```diff
@@ -73,8 +75,11 @@ export const ZoomablePlot: React.FC<ZoomablePlotProps> = ({
       <button
         className={styles.zoomablePlot}
         onClick={() => handleOnClick()}
+        onMouseDown={() => changeDragAndDropMode(section, dispatch, false)}
```
julieg18, line 78: Probably already a known issue that will get worked on later but I currently cannot interact with the plot without activating drag and drop mode:

https://github.com/iterative/vscode-dvc/assets/43496356/d87374ef-2f22-4fcf-ad07-987792da42ae

julieg18 (approved):

webview/src/plots/components/ZoomablePlot.tsx
sroy3: I totally missed that. Thanks for noticing, that was a big fail on my part. It is fixed now.


https://github.com/iterative/vscode-dvc/assets/3683420/a45f4af6-2441-4fd1-b0f7-058f09cff903

sroy3:
There is one last thing that need to be fixed in this PR: the drag end event does not work for some reason.

PR #4943, opened 2023-11-02
Add Setup Studio module

julieg18:
* Moved studio-related setup code into its own module in preparation for #4931

### DIFF
--- a/extension/src/setup/index.ts
+++ b/extension/src/setup/index.ts
@@ -16,7 +16,7 @@ import { WebviewMessages } from './webview/messages'
 import { validateTokenInput } from './inputBox'
 import { findPythonBinForInstall } from './autoInstall'
 import { run, runWithRecheck, runWorkspace } from './runner'
-import { isStudioAccessToken } from './token'
+import { Studio } from './studio'
 import {
   PYTHON_EXTENSION_ACTION,
   pickFocusedProjects,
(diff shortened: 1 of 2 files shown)


extension/src/setup/studio.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Args, ConfigKey, Flag } from '../cli/dvc/constants'
+import { ContextKey, setContextValue } from '../vscode/context'
+
+export class Studio {
+  protected studioConnectionChanged: EventEmitter<void>
+  private studioAccessToken: string | undefined = undefined
+  private studioIsConnected = false
+  private shareLiveToStudio: boolean | undefined = undefined
+  private readonly getCwd: () => string | undefined
+  private readonly internalCommands: InternalCommands
```
julieg18, line 14: This class would be handling token requests and responses when we implement Studio token authentication.

extension/src/setup/studio.ts
```diff
@@ -0,0 +1,129 @@
+import { EventEmitter } from 'vscode'
+import { isStudioAccessToken } from './token'
```
julieg18, line 2: I wanted to move `isStudioAccessToken` into `Studio` but it's being used in `setup/InputBox`.

extension/src/setup/index.ts
```diff
@@ -160,6 +151,12 @@ export class Setup
     this.setCommandsAvailability(false)
     this.setProjectAvailability()
 
+    this.studio = new Studio(
+      internalCommands,
+      this.studioConnectionChanged,
```
mattseddon, line 156: [F] Unless you are planning to use `this.studioConnectionChanged` to fire events outside of the Studio class I would create the event emitter insider of the class and wire up the listener with 
```
this.onDidChangeStudioConnection = this.studio.onDidChangeStudioConnection
```

extension/src/setup/studio.ts
mattseddon: it could still be a function in this file

mattseddon (approved):

sroy3 (approved):

extension/src/setup/studio.ts
julieg18: Oh, right, I've seen that pattern in other modules as well!

extension/src/setup/index.ts
julieg18: Good idea!

PR #4955, opened 2023-11-06
Fix e2e tests (SCM entry removed by insiders)

mattseddon:


extension/src/test/e2e/extension.test.ts
```diff
@@ -192,7 +192,6 @@ describe('Source Control View', function () {
   // eslint-disable-next-line sonarjs/cognitive-complexity
   it('should show the expected changes after running an experiment', async function () {
     const expectedScmItemLabels = [
-      'demo DVC',
```
mattseddon, line 195: This was removed in the lastest VS Code insiders release.

mattseddon: We also shouldn't need this to be present for the test to pass. My bad.

julieg18 (approved):

sroy3 (approved):

PR #4960, opened 2023-11-08
Add style to plots in drag and drop mode

sroy3:
# 3/3 main <- #4934  <- #4942 <- this

Demo

https://github.com/iterative/vscode-dvc/assets/3683420/e1a2d775-ebc9-4d19-8f98-309cbedeb0a0

Part of #4867

### DIFF
--- a/extension/src/experiments/columns/paths.ts
+++ b/extension/src/experiments/columns/paths.ts
@@ -1,16 +1,14 @@
 import { join } from 'path'
+import {
+  DECODE_METRIC_PARAM_REGEX,
+  ENCODED_METRIC_PARAM_SEPARATOR,
+  ENCODE_METRIC_PARAM_REGEX,
+  FILE_SEPARATOR,
+  FILE_SPLIT_REGEX,
+  METRIC_PARAM_SEPARATOR
+} from './constants'
 import { definedAndNonEmpty } from '../../util/array'
 import { ColumnType } from '../webview/contract'
-
-export const METRIC_PARAM_SEPARATOR = '.'
-const ENCODED_METRIC_PARAM_SEPARATOR = '%2E'
-const ENCODE_METRIC_PARAM_REGEX = /\./g
-const DECODE_METRIC_PARAM_REGEX = /%2E/g
-
-export const FILE_SEPARATOR = ':'
-const FILE_SPLIT_REGEX = new RegExp(
-  `([^${FILE_SEPARATOR}]*)(?:${FILE_SEPARATOR}([^${FILE_SEPARATOR}]*))?(?:${FILE_SEPARATOR}(.*))?`
-)
 
 const encodeColumn = (segment: string) =>
   segment.replace(ENCODE_METRIC_PARAM_REGEX, ENCODED_METRIC_PARAM_SEPARATOR)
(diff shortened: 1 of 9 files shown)


extension/src/experiments/columns/paths.ts
```diff
@@ -1,17 +1,15 @@
 import { join } from 'path'
```
sroy3, line 1: This import is what why I need to move out all constants to use them in the webview.

webview/src/plots/components/util.ts
```diff
@@ -43,3 +48,22 @@ export const changeDragAndDropMode = (
       : toggleCustomDragAndDrop
   return dispatch(toggleMode(!isDragAndDropMode))
 }
+
+const cleanTitlePart = (title: string) => {
```
sroy3, line 52: There is a bit of duplication between this and `splitColumnPath`, but it's not possible to import this function inside the webview.

mattseddon:
Found a bug with what seems like the drag end event. Clicking without moving leaves the plots in d'n'd mode:

https://github.com/iterative/vscode-dvc/assets/37993418/9814f31d-58e0-496a-ba4b-94e929a8b715

mattseddon:
WDYT about adding a border to the plot whilst in d'n'd mode? Something like:

image

Although, I think having a border in-place of where the canvas was would look better.

Just a suggestion.

sroy3:
> Found a bug with what seems like the drag end event. Clicking without moving leaves the plots in d'n'd mode:
> 
>  Screen.Recording.2023-11-10.at.8.52.39.am.mov

I stumbled upon it too and forgot to fix it, thanks.

sroy3:
> WDYT about adding a border to the plot whilst in d'n'd mode? Something like:
> 
> image
> Although, I think having a border in-place of where the canvas was would look better.
> 
> Just a suggestion.
> 
> 

I'll play with it a little more.

For the border around the whole plot was making the whole thing way too crowded. It might be subtle with your theme, but they're not all like that. It also makes it harder to spot the drop target as it's not popping as much as before.

I'm also not sure if it's a good idea to give more importance to decorative elements rather than the identification of the plot itself.

webview/src/plots/components/DragAndDropPlot.tsx
```diff
@@ -0,0 +1,52 @@
...
+  plot,
+  sectionKey
+}) => {
+  const { spec, isTemplatePlot } = useGetPlot(sectionKey, plot)
+
+  let title = spec?.titles.main.normal as unknown as string
+  let subtitle = ''
+
+  if (!isTemplatePlot) {
+    const yTitle = spec?.titles.y.normal as unknown as string
```
mattseddon, line 26: [F] We'll be able to get these for all plots with the change in #4734 (more conflicts 😭)

mattseddon (approved):

webview/src/plots/components/styles.module.scss
```diff
@@ -178,6 +197,7 @@ $gap: 20px;
 
 .smallDropIcon {
```
julieg18, line 198: Probably just me staring at things too long, but the icon seemed... out of place 🤔. One idea I played with was having the drag and drop icon be centered like targets, but with less opacity, keeping the title the focus: 

https://github.com/iterative/vscode-dvc/assets/43496356/fe2d6908-3cd3-493d-af3a-abce05bf5d27

What do you think?

julieg18 (approved):
Great work!

sroy3:
New demo


https://github.com/iterative/vscode-dvc/assets/3683420/7383f349-0b27-4a9e-a4c8-17ba6d6907ee

PR #4989, opened 2023-11-14
Add constant to verify if there is a custom selection of plots

stephr0y:
Part of #4867 

This does nothing on its own, but it'll be used to display only 20 plots by default prior to modifying the plot selection.


mattseddon (approved):

extension/src/plots/paths/model.test.ts
```diff
@@ -502,4 +504,75 @@ describe('PathsModel', () => {
 
     expect(model.getTerminalNodes()).toStrictEqual([])
   })
+
+  it('should not change hasCustomSelection when checking for it if it is already defined', () => {
```
julieg18, line 508:
```suggestion
  it('should not change hasCustomSelection when checking for if it is already defined', () => {
```

extension/src/plots/paths/model.ts
```diff
@@ -144,6 +151,30 @@ export class PathsModel extends PathSelectionModel<PlotPath> {
...
+
+  public checkIfHasPreviousCustomSelection() {
+    if (this.hasCustomSelection === undefined) {
+      const statuses = this.getTerminalNodeStatuses()
+      const plotsLength = statuses.length
+
+      const hasCustomSelection =
+        plotsLength > 20
+          ? plotsLength -
+              statuses.filter(nodeStatus => nodeStatus !== Status.SELECTED)
```
julieg18, line 170: Minor! Should we move all or part of `plotsLength -
              statuses.filter(nodeStatus => nodeStatus !== Status.SELECTED).length !==
            20` into a variable? Personally, I had to reread the statement multiple times to figure how what it was doing :)

julieg18 (approved):
Great work!

PR #5008, opened 2023-11-17
Add a message when we are showing only 20 plots

sroy3:


webview/src/plots/components/TooManyPlots.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const TooManyPlots: React.FC = () => {
+  const shouldShow = useSelector(
+    (state: PlotsState) => state.webview.shouldShowTooManyPlotsMessage
+  )
+
+  return shouldShow ? (
+    <MessageBand icon={Info} id="too-many-plots-message">
+      We are only showing 20 plots by default. To view other plots, you can
+      toggle their visibility in the sidebar. You can also group your plots by
+      adding a path prefix to their IDs (
```
mattseddon, line 21: [C] In the `dvc.yaml` the ID for the plot can be a file or a string. The following are all valid (from demo project):

```
  - Loss:
      x: step
      y:
      y_label: loss
      title: "Loss"
  - Confusion matrix:
      template: confusion
      x: actual
      y:
        training/plots/sklearn/confusion_matrix.json: predicted
  - hist.csv:
      x: preds
      y: digit
      template: bar_horizontal
      title: Histogram of Predictions
  - probs.json:
      x: actual
      y: prob
      template: scatter
      title: Predicted Probabilities
  - training/plots/images
```

This means the instructions to group need to be updated slightly. In the case of the key being a file the user would need to group the underlying data. If it is just a string then it can be done in the suggested way.

webview/src/plots/components/Plots.tsx
```diff
@@ -88,7 +90,8 @@ const PlotsContent = () => {
   }
 
   return (
-    <div ref={wrapperRef}>
+    <div ref={wrapperRef} className={styles.plotsContent}>
+      <TooManyPlots />
```
mattseddon, line 94: [Q] Is it a normal pattern to have the band at the bottom of the screen instead of the top?

mattseddon (approved):

webview/src/plots/components/Plots.tsx
sroy3: It does not really matter. At first, I was building it to be at the top and decided to move it down later. I'll move it in the tree, which will help with reading the code.

webview/src/plots/components/TooManyPlots.tsx
sroy3: I changed the message. Does that look like complete instructions to you now @mattseddon ?

sroy3:
Because this has already been reviewed and there were a lot of changes requested in the planning yesterday, I will leave this up and merge the changes back into this PR through another PR.

Changes requested:

- 20 plots per section
- Static message per section

extension/src/plots/paths/model.test.ts
```diff
@@ -575,4 +578,144 @@ describe('PathsModel', () => {
...
+      buildMockErrorsModel(),
+      buildMockMemento()
+    )
+
+    model.setHasCustomSelection(false)
+
+    const nodes = [{}]
+    for (let i = 0; i < 19; i++) {
+      nodes.push({})
+    }
```
julieg18, line 606: Minor! I think we can create `nodes` for this test and the next one in a single line: 

```js
const nodes = Array.from({ length: 20 }).fill({})
```

julieg18 (approved):

mattseddon (approved):

PR #5020, opened 2023-11-23
Select 20 plots by default per section

sroy3:
# 2/2 main <- #5008 <- this

Part of #4867 

Demo

https://github.com/iterative/vscode-dvc/assets/3683420/11beeb76-3c42-4b08-881e-ca5a9a6eece9


webview/src/plots/components/TooManyPlots.tsx
```diff
@@ -25,5 +19,5 @@ export const TooManyPlots: React.FC = () => {
       shared folder (and updating the path in your{' '}
       <PathHighlight>dvc.yaml</PathHighlight> file).
     </MessageBand>
```
julieg18, line 21: Why not create too different messages so we can simplify the message for each section? 
Example Template: `Only 20 plots are shown by default. To view other plots, you can toggle their visibility in the sidebar. To toggle multiple plots simultaneously, you can group your plots by adding a path prefix to their IDs (group1/plot1, group1/plot2, group2/plot1).`

Example Comparison: `Only 20 plots are shown by default. To view other plots, you can toggle their visibility in the sidebar. To toggle multiple plots simultaneously, you can group your plots by placing your data inside a shared folder and updating the path in your dvc.yaml file.`

julieg18: Another thing we could do to simplify the message is to direct them to the `Select Plots to Display` command. Technically, that allows you to toggle multiple plots by name without updating dvc.yaml.

webview/src/plots/components/TooManyPlots.tsx
```diff
@@ -10,11 +8,7 @@ const PathHighlight: React.FC<PropsWithChildren> = ({ children }) => (
 )
 
 export const TooManyPlots: React.FC = () => {
-  const shouldShow = useSelector(
-    (state: PlotsState) => state.webview.shouldShowTooManyPlotsMessage
-  )
-
-  return shouldShow ? (
+  return (
     <MessageBand icon={Info} id="too-many-plots-message">
```
julieg18, line 12: Minor! Have we considered adding a border with extra spacing to the message? 

image

Personally, I think it looks better but up to you!

webview/src/plots/components/TooManyPlots.tsx
sroy3: > Another thing we could do to simplify the message is to direct them to the `Select Plots to Display` command. Technically, that allows you to toggle multiple plots by name without updating dvc.yaml.

We discussed this in the planning, we decided to go as is for now and and the action later if needed.

extension/src/plots/index.ts
```diff
@@ -172,8 +178,32 @@ export class Plots extends BaseRepository<TPlotsData> {
...
 
-  private setHasCustomSelection() {
-    this.paths.setHasCustomSelection(this.paths.getTerminalNodes().length > 20)
+  private setHasCustomSelection(types: Set<PathType> | undefined) {
+    if (!types || types.has(PathType.COMPARISON)) {
+      const comparisonNodes = this.paths.getTerminalNodesByType(
+        PathType.COMPARISON
+      )
+      this.paths.setHasCustomSelection(
+        comparisonNodes.length > 20,
```
mattseddon, line 187: [Q] Could we have this as a boolean value that is only set to true once? As a path is toggled (on or off) we could set the boolean to false. Would that work/help to simplify the code? Something like `hasUserSelectedPlots`?

mattseddon: It might actually make everything easier to follow/simplify the logic if we invert the flag. 
Change the flag to `showingDefaultNumberPlots` and only set it to `true` when previously there were < 20 plots available in a section and now there are > 20 plots. Any subsequent change in selection could then set the flag to `false`.

☝🏻 might be easier to chat about this on a call.

sroy3: This definitely needs to be simplified because I try to explain my reasoning and I keep backtracking.

PR #4673, opened 2023-09-16
Fix Vega plots disappearing

mattseddon:
Closes #4671

I chased this one around for a few hours. Initially, I thought the problem was with smooth plots but I could recreate with a confusion matrix as well. In the end, the problem seemed to be setting the `<PlotsContainer>`'s `children` to `null` when the section was collapsed.

### Demo

https://github.com/iterative/vscode-dvc/assets/37993418/6d85fc25-1e27-453b-9135-7d8d190861a2

PR #5025, opened 2023-11-27
Do not render section if it is not open

sroy3:
Part of #4867 

https://github.com/iterative/vscode-dvc/assets/3683420/d9ee9c1b-b615-4caf-81ca-f8b5643224d3


webview/src/plots/components/App.test.tsx
```diff
@@ -700,42 +700,6 @@ describe('App', () => {
     expect(screen.queryByText('Images')).not.toBeInTheDocument()
   })
 
-  it('should hide plots when their section is collapsed (setting to null can break some Vega plots)', async () => {
```
sroy3, line 703: I think this might have been there because we used to animate the height of the section. With the changes it re-renders everything when we open the section. If anything, it should solve any previous rendering problems.

julieg18 (approved):

sroy3:
> From memory: we had to "always render" the template sections because the plots can end up empty under some weird circumstances. I will do my best to find the PR(s) shortly.
> 
> Edit: #4673

This is now fixed by clearing the state when closing the section.
Demo:

https://github.com/iterative/vscode-dvc/assets/3683420/52291d1c-a281-4982-aa0f-fb73d4a49875

PR #5045, opened 2023-11-29
Handle project having experiments with no data

mattseddon:
Fixes: 

```
When I'm trying to show experiments/plots, that "You project contains no data".
But I have some... experiments and those plots
Any ideas why is VSCode behave like this?  Also, no clear what should I do here
```

From https://iterativeai.slack.com/archives/CB41NAL8H/p1701257451702629

### DIFF
--- a/extension/src/experiments/webview/messages.ts
+++ b/extension/src/experiments/webview/messages.ts
@@ -84,10 +84,11 @@ export class WebviewMessages {
     }
     const data = await this.getWebviewData()
 
-    const hasNoRows = data.rows.length === 0
-    const hasNoData = !this.columns.hasNonDefaultColumns() || hasNoRows
+    const hasNoData = !this.experiments.hasData(
+      this.columns.hasNonDefaultColumns()
+    )
 
-    if (hasNoData && !data.cliError) {
+    if (hasNoData) {
       await commands.executeCommand(RegisteredCommands.SETUP_SHOW_EXPERIMENTS)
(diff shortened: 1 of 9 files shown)


extension/src/experiments/webview/messages.ts
```diff
@@ -85,7 +85,10 @@ export class WebviewMessages {
     const data = await this.getWebviewData()
 
     const hasNoRows = data.rows.length === 0
-    const hasNoData = !this.columns.hasNonDefaultColumns() || hasNoRows
+    const hasNoExperiments =
+      this.experiments.getExperimentsAndQueued().length === 0
+    const hasNoData =
+      (!this.columns.hasNonDefaultColumns() && hasNoExperiments) || hasNoRows
```
mattseddon, line 91: [I] Need to handle the tree as well, make an `ExperimentsModel` helper.

julieg18 (commented):
This updates the experiments tree but the webview is still showing the "has no data" message:

Screenshot 2023-11-30 at 8 19 43 AM

I updated `Experiments.getHasData()` to see if the webviews could work without none default columns but no luck: 

Screenshot 2023-11-30 at 8 18 24 AM

mattseddon:
> This updates the experiments tree but the webview is still showing the "has no data" message:

@julieg18 what data/setup are you using?

julieg18 (approved):
Single plots project is working now, great work!

PR #5049, opened 2023-11-30
Improve Studio Token Authentication URI Handler

julieg18:
Details are inline. 

Related to #3864

### DIFF
--- a/extension/src/setup/studio.ts
+++ b/extension/src/setup/studio.ts
@@ -109,7 +109,7 @@ export class Studio extends Disposable {
   }
 
   public async requestStudioTokenAuthentication() {
-    this.studioAccessTokenUriHandler?.dispose()
+    this.resetAccessTokenUriHandler()
 
     const response = await this.fetchFromStudio(
       `${STUDIO_URL}/api/device-login`,
@@ -145,6 +145,11 @@ export class Studio extends Disposable {
     )
   }
 
+  private resetAccessTokenUriHandler() {
+    this.studioAccessTokenUriHandler?.dispose()
+    this.studioAccessTokenUriHandler = undefined
+  }
+
   private fetchFromStudio(reqUri: string, body: Record<string, unknown>) {
     return fetch(reqUri, {
       body: JSON.stringify(body),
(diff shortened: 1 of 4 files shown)


extension/src/vscode/external.ts
```diff
@@ -10,12 +10,11 @@ export const getCallBackUrl = async (path: string) => {
   return uri.toString()
 }
 
-export const waitForUriResponse = (path: string, onResponse: () => unknown) => {
+export const waitForUriResponse = (path: string, onResponse: () => unknown) =>
   window.registerUriHandler({
     handleUri(uri: Uri): ProviderResult<void> {
-      if (uri.path.startsWith(path)) {
+      if (uri.path === path) {
```
julieg18, line 16: Studio is now decoding the uri so we can use the equals check.

extension/src/setup/studio.ts
```diff
@@ -134,9 +137,12 @@ export class Studio extends Disposable {
     verificationUrlWithParams.searchParams.append('code', userCode)
 
     await openUrl(verificationUrlWithParams.toString())
-    void waitForUriResponse('/studio-complete-auth', () => {
-      void this.requestStudioToken(deviceCode, tokenUri)
-    })
+    this.studioAccessTokenUriHandler = waitForUriResponse(
```
julieg18, line 140: Realized we aren't disposing the uri handler after use. I updated the code to dispose of the listener if:

* we get a response from Studio
* the user wants to try authentication from the start again

extension/src/setup/studio.ts
```diff
@@ -108,6 +109,8 @@ export class Studio extends Disposable {
   }
 
   public async requestStudioTokenAuthentication() {
+    this.studioAccessTokenUriHandler?.dispose()
```
mattseddon, line 112: [C] personal preference but I always set class members to `undefined` after disposing.

mattseddon: could create a `resetAccessTokenUriHandler` method

sroy3 (approved):

extension/src/setup/studio.ts
julieg18: Done!

Issue #5035, opened 2023-11-28
Use correct URL for Studio Prem Environments


julieg18:
Currently, we use `https://studio.iterative.ai` for making calls to Studio but [self-hosted instances](https://dvc.org/doc/studio/self-hosting/installation) would use a different URL. 

We can use the DVC config's [`studio.url`](https://dvc.org/doc/user-guide/project-structure/configuration#studio) to get the URL to call to and default to `https://studio.iterative.ai` if it doesn't exist. 

- [x] Use `studio.url` for calls to Studio, defaulting to `https://studio.iterative.ai`
- [x] Add commands for updating the URL in the command pallette
- [ ] Look into adding reference to studio prem URL in setup section

PR #5061, opened 2023-12-01
Get studio url from setup

julieg18:
Has fetch requests get the studio url from setup instead of a constant in preparation for #5035

### DIFF
--- a/extension/src/experiments/commands/index.ts
+++ b/extension/src/experiments/commands/index.ts
@@ -9,7 +9,7 @@ import {
   getConfigValue,
   setUserConfigValue
 } from '../../vscode/config'
-import { STUDIO_URL } from '../../setup/webview/contract'
+import { DEFAULT_STUDIO_URL } from '../../setup/webview/contract'
 import { RegisteredCommands } from '../../commands/external'
 
 export const getBranchExperimentCommand =
@@ -48,7 +48,7 @@ export const getRenameExperimentCommand =
 
 const promptToAddStudioToken = async () => {
   const response = await Toast.askShowOrCloseOrNever(
-    `Experiments can be automatically shared to [Studio](${STUDIO_URL}) by setting the studio.token in your config.`
+    `Experiments can be automatically shared to [Studio](${DEFAULT_STUDIO_URL}) by setting the studio.token in your config.`
   )
 
   if (!response || response === Response.CLOSE) {
--- a/extension/src/setup/studio.ts
+++ b/extension/src/setup/studio.ts
@@ -1,6 +1,6 @@
 import { Event, EventEmitter, Disposable as VSCodeDisposable } from 'vscode'
 import fetch from 'node-fetch'
-import { STUDIO_URL } from './webview/contract'
+import { DEFAULT_STUDIO_URL } from './webview/contract'
 import { AvailableCommands, InternalCommands } from '../commands/internal'
 import { getFirstWorkspaceFolder } from '../vscode/workspaceFolders'
 import { Args, ConfigKey, Flag } from '../cli/dvc/constants'
@@ -28,6 +28,7 @@ export class Studio extends Disposable {
   private studioIsConnected = false
   private shareLiveToStudio: boolean | undefined = undefined
   private studioAccessTokenUriHandler: VSCodeDisposable | undefined = undefined
+  private studioUrl: string = DEFAULT_STUDIO_URL
 
   constructor(
     internalCommands: InternalCommands,
@@ -50,6 +51,10 @@ export class Studio extends Disposable {
 
   public getShareLiveToStudio() {
     return this.shareLiveToStudio
+  }
+
+  public getStudioUrl() {
+    return this.studioUrl
   }
 
   public async removeStudioAccessToken(dvcRoots: string[]) {
@@ -112,7 +117,7 @@ export class Studio extends Disposable {
     this.resetAccessTokenUriHandler()
 
     const response = await this.fetchFromStudio(
-      `${STUDIO_URL}/api/device-login`,
+      `${this.getStudioUrl()}/api/device-login`,
       {
         client_name: 'VS Code'
       }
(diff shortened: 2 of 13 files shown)


extension/src/experiments/commands/index.ts
```diff
@@ -48,7 +48,7 @@ export const getRenameExperimentCommand =
 
 const promptToAddStudioToken = async () => {
   const response = await Toast.askShowOrCloseOrNever(
-    `Experiments can be automatically shared to [Studio](${STUDIO_URL}) by setting the studio.token in your config.`
+    `Experiments can be automatically shared to [Studio](${DEFAULT_STUDIO_URL}) by setting the studio.token in your config.`
```
julieg18, line 51: We only use the url found in the config for fetch requests to studio (aka requesting a token or studio experiment data). 

For links in the webview or Toasts, we just use the default one.

extension/src/setup/studio.ts
```diff
@@ -205,30 +210,47 @@ export class Studio extends Disposable {
...
 
-    const [studioAccessToken, shareLiveToStudio] = await Promise.all([
-      this.accessConfig(cwd, ConfigKey.STUDIO_TOKEN),
-      (await this.accessConfig(cwd, ConfigKey.STUDIO_OFFLINE)) !== 'true'
-    ])
+    const [studioAccessToken, shareLiveToStudio, studioUrl] = await Promise.all(
+      [
+        this.accessConfig(cwd, ConfigKey.STUDIO_TOKEN),
+        (await this.accessConfig(cwd, ConfigKey.STUDIO_OFFLINE)) !== 'true',
+        (await this.accessConfig(cwd, ConfigKey.STUDIO_URL)) ||
```
julieg18, line 241: We default to `https://studio.iterative.ai` if the url is not set found in the config.

webview/src/setup/components/studio/Connect.tsx
```diff
@@ -30,7 +30,8 @@ export const Connect: React.FC = () => {
           />
         </div>
         <p>
-          Don&apos;t Have an account? <a href={STUDIO_URL}>Get Started</a>
+          Don&apos;t Have an account?{' '}
+          <a href={DEFAULT_STUDIO_URL}>Get Started</a>
```
julieg18, line 34: Currently, next steps were adding a Command Pallete command for updating the URL and maybe adding text about it in the studio section somewhere. Not sure where would be good spot yet 🤔

extension/src/experiments/commands/index.ts
mattseddon: why?

mattseddon: Do you mean for "generic" links when the user doesn't have an account or when they don't have a token setup? I thought that "links in the webview" meant "shareable experiment links".

extension/src/experiments/studio.ts
```diff
@@ -1,12 +1,14 @@
...
 import { DeferredDisposable } from '../class/deferred'
+import { DEFAULT_STUDIO_URL } from '../setup/webview/contract'
 
 export class Studio extends DeferredDisposable {
   private readonly dvcRoot: string
   private readonly internalCommands: InternalCommands
 
-  private baseUrl: string | undefined = undefined
+  private url: string = DEFAULT_STUDIO_URL
+  private baseViewUrl: string | undefined = undefined
```
mattseddon, line 11: [Q] Why change the name from `baseUrl`?

mattseddon: I think `viewUrl` would be better. `url` should probably be `instanceUrl` or something like that

extension/src/setup/studio.ts
mattseddon: [Q] Does `dvc config` return an empty string when there is no value present?

extension/src/test/suite/setup/index.test.ts
```diff
@@ -985,6 +988,58 @@ suite('Setup Test Suite', () => {
       expect(delayProgressClosingSpy).to.be.calledTwice
     }).timeout(WEBVIEW_TEST_TIMEOUT)
 
+    it('should handle a message from the webview to request a token from a self hosted studio', async () => {
```
mattseddon, line 991: [C] Studio instance

extension/src/setup/studio.ts
```diff
@@ -205,30 +210,47 @@ export class Studio extends Disposable {
...
 
+  private resetStudioValues(
+    previousStudioUrl: string,
+    previousStudioAccessToken: string | undefined
+  ) {
+    this.studioAccessToken = undefined
+    this.shareLiveToStudio = undefined
+    this.studioUrl = DEFAULT_STUDIO_URL
+
+    if (previousStudioAccessToken || previousStudioUrl !== DEFAULT_STUDIO_URL) {
```
mattseddon, line 221: [Q] what about `shareLiveToStudio`?

mattseddon (approved):

extension/src/experiments/commands/index.ts
julieg18: > Do you mean for "generic" links when the user doesn't have an account or when they don't have a token setup

Yes, apologies, I didn't explain that very well 🤦‍♀️

extension/src/setup/studio.ts
julieg18: Yes, it returns an empty string.

extension/src/experiments/studio.ts
julieg18: I changed `baseUrl` to `baseViewUrl` to better differentiate between the two URLs, but I like your idea better :)

julieg18: Updated!

extension/src/setup/studio.ts
julieg18: Updated to check for `shareLiveToStudio`!

PR #5062, opened 2023-12-01
Use hook for drag and drop

sroy3:
Part of #4867 

Demo

https://github.com/iterative/vscode-dvc/assets/3683420/384835b6-c124-4331-93b9-2015cd7da4aa

Things should be as before, but faster


webview/src/plots/components/App.test.tsx
```diff
@@ -1733,8 +1733,7 @@ describe('App', () => {
     it('should show a drop target at the end of the custom plots when moving a plot inside the section but not over any other plot', () => {
       renderAppWithOptionalDataInDragAndDropMode(
         {
-          custom: customPlotsFixture,
-          template: templatePlotsFixture
```
mattseddon, line 1737: [Q] From the test title it sounds like these plots are needed. Do the statements need to be updated?

webview/src/plots/components/comparisonTable/ComparisonTable.test.tsx
```diff
@@ -535,14 +535,9 @@ describe('ComparisonTable', () => {
...
-      // eslint-disable-next-line testing-library/no-node-access
-      expect(headerWrapper.childElementCount).toBe(2)
       expect(
-        // eslint-disable-next-line testing-library/no-node-access
-        Object.values(headerWrapper.children)
-          .map(child => child.id)
-          .includes(endingNode.id)
-      ).toBe(true)
+        within(headerWrapper).getByTestId('comparison-drop-target')
+      ).toBeInTheDocument()
```
mattseddon, line 540: [Q] How do these assertions differ/relate to one another?

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -13,19 +20,29 @@ import { isSelecting } from '../../../util/strings'
...
 
 export interface ComparisonTableRowProps {
   path: string
   plots: ComparisonPlot[]
   nbColumns: number
   pinnedColumn: string
+  onLayoutChange: () => void
+  setOrder: (order: string[]) => void
+  order: string[]
+  bodyRef?: React.RefObject<HTMLTableSectionElement>
```
mattseddon, line 34: [Q] Should this be `RefObject`? (I am asking because that is what I've been using)

webview/src/test/dragDrop.ts
```diff
@@ -50,7 +50,7 @@ export const dragEnter = (
   })
 
   act(() => {
-    jest.advanceTimersByTime(1)
+    jest.advanceTimersByTime(501)
```
mattseddon, line 53: [Q] Why is this needed?

webview/src/util/objects.ts
```diff
@@ -9,7 +9,7 @@ export type BaseType =
   | undefined
   | null
 
-export type Any = BaseType | BaseType[]
+type Any = BaseType | BaseType[]
```
mattseddon, line 12: 🎉

mattseddon (approved):

.eslintrc.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
             cases: {
               pascalCase: true
             },
-            ignore: [/.*\.stories.tsx$/, /.*\.test\.tsx$/]
+            ignore: [/.*\.stories.tsx$/, /.*\.test\.tsx$/, /.*use.*\.tsx$/]
```
julieg18, line 64: What's the reasoning for adding this? I'm not seeing any new files with this syntax 🤔

julieg18 (approved):
Great work!

.eslintrc.js
sroy3: `useDragAndDrop.tsx` there is some `jsx in that file

webview/src/test/dragDrop.ts
sroy3: It isn't. I added `dragLeave` in there to test something and the `500` was needed then. Turns out that wasn't how to solve the test and I forgot to revert this.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
sroy3: Copied and pasted without verifying. I'll fix it.

webview/src/plots/components/comparisonTable/ComparisonTable.test.tsx
sroy3: It's not wrapped the exact same way, but the test checks the same thing, it was just not written in a clear way before.

webview/src/plots/components/App.test.tsx
sroy3: It was never testing the other part. I guess we need to update the statement.

PR #5066, opened 2023-12-04
Do not mutate plot props when zooming

sroy3:


webview/src/plots/components/templatePlots/TemplateVegaLite.tsx
sroy3, line 1: Using the full Vega API for signals and views (https://vega.github.io/vega/docs/api/view/#signals) to make things simpler

mattseddon (approved):

PR #5071, opened 2023-12-05
Cancel Studio Token Request after five minutes

julieg18:
Disposes of the uri listener if the user does not authorize with Studio and get redirected after ~fifteen~ five minutes.

Part of #3864


extension/src/setup/studio.ts
```diff
@@ -142,17 +143,31 @@ export class Studio extends Disposable {
...
+    this.accessTokenUriHandlerTimeout = undefined
+
+    this.accessTokenUriHandler?.dispose()
+    this.accessTokenUriHandler = undefined
+  }
+
+  private cancelUriHandlerAfterTimeout() {
+    const waitTime = 15 * 60000
+    this.accessTokenUriHandlerTimeout = setTimeout(
+      () => this.resetAccessTokenUriHandler(),
```
julieg18, line 168: I chose to fail silently, but we could also put up some kind of Toast message.

mattseddon: Feels like 15 minutes is too long. I think 5 minutes would be a very reasonable amount of time to wait.

mattseddon: Undecided on the toast message. I would lean towards showing it but I am conflicted.

mattseddon (approved):

sroy3 (approved):

extension/src/setup/studio.ts
julieg18: I'll set the time for five minutes. As for the toast message, I don't think it's needed for now. Most if not all users should just be authenticating immediately.

mattseddon:
getting failures for the test on windows: https://github.com/iterative/vscode-dvc/actions/runs/7120127578/job/19389801179?pr=5074

PR #5072, opened 2023-12-05
Add commands for updating `studio.url` in DVC configuration

julieg18:
## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/3327633a-e747-42c9-84d2-2909175afafe~

https://github.com/iterative/vscode-dvc/assets/43496356/c2886dea-8f51-4931-b273-ed7455c5f68d

Part of #5035


extension/package.json
```diff
@@ -396,6 +401,11 @@
         "command": "dvc.selectFocusedProjects",
         "category": "DVC"
       },
+      {
+        "title": "Set Self-Hosted Studio URL",
```
julieg18, line 405: Decided on just using one command instead of `Update` and `Add` commands.

extension/src/setup/index.ts
```diff
@@ -325,6 +325,30 @@ export class Setup
...
+  public removeStudioUrl() {
+    if (!this.getCliCompatible()) {
+      return
+    }
+
+    return this.studio.removeStudioUrl(this.dvcRoots)
+  }
+
+  public async saveStudioUrl() {
+    const cwd = this.getCwd()
```
julieg18, line 337: Did look into using `getCwd()` for `saveStudioAccessToken` and it does look like we can use it without issues. Will update (plus maybe remove the dvc compatible check as well) in a followup!

extension/package.json
julieg18: Ohhhh, ok. Well, if we want all commands to be consistent, `Add` makes more sense to me in other commands contexts. I'll go ahead and just use `Update` and `Add` commands 😄 for this too.

extension/src/test/suite/setup/index.test.ts
```diff
@@ -1213,6 +1211,121 @@ suite('Setup Test Suite', () => {
       )
     }).timeout(WEBVIEW_TEST_TIMEOUT)
 
+    it("should be able to save the user's self-hosted url", async () => {
```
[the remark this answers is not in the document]
mattseddon: [C] When you are making multiple chained assertions it is good to add a description to the `expect` statements. E.g.

```
      expect(
        executeCommandSpy,
        'should call executeCommand with the args required to select a file to compare against'
      ).to.be.calledWith('selectForCompare', baseline)
```

`expect(saveStudioUrlSpy).not.to.be.called` would be a good place to add one as this has nothing to do with the `it` statement 

and/or split the test

mattseddon (approved):

extension/src/test/suite/setup/index.test.ts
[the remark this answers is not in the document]
julieg18: > [C] When you are making multiple chained assertions it is good to add a description to the expect statements.

Good idea! I forget that we can add descriptions to expect statements, thanks!

extension/src/setup/index.ts
julieg18: Discussed synchronously, but we'll keep the dvc compatible check for failsafe reasons.

PR #5085, opened 2023-12-08
Use resize observer to simplify use get plot

mattseddon:
Matches part of the change that I made in https://github.com/iterative/studio/pull/8490.

@sroy3 when you're back I'd like to get your opinion on performance implications of the change 🙏🏻. 

The thread at https://github.com/iterative/studio/pull/8490#discussion_r1419809476 is relevant.

### DIFF
--- a/webview/src/plots/hooks/useGetPlot.ts
+++ b/webview/src/plots/hooks/useGetPlot.ts
@@ -1,5 +1,5 @@
 import { PlotsSection } from 'dvc/src/plots/webview/contract'
-import { RefObject, useCallback, useEffect, useState } from 'react'
+import { RefObject, useEffect, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { VisualizationSpec } from 'react-vega'
 import { plotDataStore } from '../components/plotDataStore'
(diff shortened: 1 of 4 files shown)


julieg18 (approved):
Looks good to me!

webview/src/plots/hooks/useGetPlot.ts
```diff
@@ -9,49 +9,49 @@ import { fillTemplate } from '../components/vegaLite/util'
 export const useGetPlot = (
   section: PlotsSection,
   id: string,
-  plotRef: RefObject<HTMLButtonElement | HTMLDivElement>
+  plotRef: RefObject<HTMLButtonElement | HTMLDivElement>,
+  plotFocused: boolean
```
sroy3, line 13: Was this needed in the end?

sroy3:
> @sroy3 when you're back I'd like to get your opinion on performance implications of the change 🙏🏻.

Since this applies to only one element and resizing is quite rare or happens at one moment only, the performance optimization here is really, really small, if there is one at all. But we should favour `ResizeObserver` and make it best practice as it is generally more performant and better suited for most uses.

webview/src/plots/hooks/useGetPlot.ts
mattseddon: Yes, for suppressing the legend

PR #5089, opened 2023-12-08
Add self-hosted url info in Studio setup section

### LINKED ISSUE #5035: Use correct URL for Studio Prem Environments

julieg18:
## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/0c9cfbb5-c820-4760-acf2-fdb3f80541c5~

https://github.com/iterative/vscode-dvc/assets/43496356/9f9f53c4-69bd-46ee-be88-2eec3a96b5cf

Fixes #5035


webview/src/setup/components/studio/Studio.tsx
```diff
@@ -22,9 +27,28 @@ export const Studio: React.FC<{
     )
   }
 
+  const children = selfHostedStudioUrl && (
```
mattseddon, line 30: [I] We should always show this section of the table so that users know the option is available.

When there is no URL set we can present "None" (or match what we show elsewhere on the setup page) in place of the URL.

julieg18: Good point! I'll update the section.

sroy3 (approved):

extension/src/test/suite/setup/index.test.ts
```diff
@@ -1332,6 +1336,68 @@ suite('Setup Test Suite', () => {
       )
     })
 
+    it("should handle a message from the webview to update the user's self hosted url", async () => {
+      const { setup, mockExecuteCommand } = buildSetup({
+        disposer: disposable
+      })
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      stub(Setup.prototype as any, 'getCliCompatible').returns(true)
```
mattseddon, line 1345: [C] Maybe this should be moved into `buildSetup`

mattseddon: and should be stubbed on the instance (if possible)

webview/src/setup/components/studio/Studio.tsx
```diff
@@ -22,9 +27,32 @@ export const Studio: React.FC<{
...
+                {
+                  onClick: saveStudioUrl,
+                  text: 'Update'
+                },
+                {
+                  onClick: removeStudioUrl,
+                  text: 'Remove'
+                }
+              ]
+            : [{ onClick: saveStudioUrl, text: 'Add Url' }]
```
mattseddon, line 47: [Q] Should this be "Add"? Does the text appear too short?

mattseddon (approved):

webview/src/setup/components/studio/Studio.tsx
julieg18: > [Q] Should this be "Add"? Does the text appear too short?

Yes, the text appeared too short :)

extension/src/test/suite/setup/index.test.ts
julieg18: I was thinking about that as well since we are running this line in multiple tests. I'll look into moving it in a followup.

PR #5097, opened 2023-12-11
Reduce number of plots resize observers

mattseddon:
Follow up from #5085

Reduces the number of `ResizeObserver`s that we need to 2 (1 for each plots grid).

### DIFF
--- a/webview/src/plots/components/customPlots/CustomPlots.tsx
+++ b/webview/src/plots/components/customPlots/CustomPlots.tsx
@@ -1,11 +1,10 @@
-import React, { DragEvent, useEffect, useState } from 'react'
-import { useSelector } from 'react-redux'
-import { PlotsSection } from 'dvc/src/plots/webview/contract'
+import React, { DragEvent, useEffect, useRef, useState } from 'react'
 import cx from 'classnames'
+import { useSelector } from 'react-redux'
 import { NoPlotsAdded } from './NoPlotsAdded'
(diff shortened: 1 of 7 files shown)


julieg18 (approved):

webview/src/plots/components/customPlots/CustomPlots.tsx
```diff
@@ -82,15 +83,15 @@ export const CustomPlots: React.FC<CustomPlotsProps> = ({ plotsIds }) => {
       onDragLeave={() => setOnSection(false)}
       onDragOver={handleDragOver}
       onDrop={handleDropAtTheEnd}
+      ref={gridRef}
     >
-      <Grid
-        setOrder={setPlotsIdsOrder}
+      <CustomPlotsGrid
+        gridRef={gridRef}
```
sroy3, line 89: Why does this need to be passed down? Can't the observer be initialized in here?

sroy3 (approved):

webview/src/plots/components/customPlots/CustomPlots.tsx
mattseddon: discussed IRL.

tl;dr - ref must be passed down or the grid will appear to have 0 dimensions. The alternative is to use `useLayoutEffect`.

sroy3: I just tried it, and it works correctly without changing anything else. https://github.com/iterative/vscode-dvc/pull/5111

PR #5105, opened 2023-12-12
Fix Toast messages showing a successful message after failed remote command

### LINKED ISSUE #5095: Error is not handled in remote add operation in non DVC repo
When I'm trying to add a remote in a repo that is not DVC in the setup DVC workflow
I'm getting Operation Successful, while actually DVC is complaining in the DVC Output
We should show a proper error message there

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/9750e863-4752-4b08-9dc5-ceeb70bf5025

Fixes #5095


extension/src/cli/dvc/config.ts
```diff
@@ -21,6 +24,28 @@ export class DvcConfig extends DvcCli {
     return this.executeSafeProcess(cwd, Command.REMOTE, ...args)
```
julieg18, line 24: `executeSafeProcess` catches the error, stopping the external command code from [showing the error toast](https://github.com/iterative/vscode-dvc/blob/add-error-toasts-to-remote-commands/extension/src/commands/internal.ts#L83) and allowing the [Toast.showOutput](https://github.com/iterative/vscode-dvc/pull/5105/files#diff-d85871702cda3fd295552b55cb8c3131e5abbbe3c96f7b5beec7491905dd82daR156) used in `setup/index.ts` to get `undefined` (aka `Operation successful`). 

I've fixed that by adding functions for adding and modifying the remote that don't catch the error thrown.

extension/src/setup/commands/index.ts
```diff
@@ -275,25 +272,21 @@ const confirmAndRemove = async (
...
 
-  try {
-    await internalCommands.executeCommand(
-      AvailableCommands.REMOTE,
-      dvcRoot,
-      SubCommand.REMOVE,
-      Flag.PROJECT,
-      remote
-    )
-  } catch {}
```
julieg18, line 286: I don't think we don't need these try/catch blocks since `executeSafeProcess` already catches errors.

sroy3 (approved):

extension/src/cli/dvc/config.test.ts
```diff
@@ -105,4 +105,78 @@ describe('DvcConfig', () => {
...
+        args: ['remote', 'rename', '--local', oldName, newName],
+        cwd,
+        env: mockedEnv,
+        executable: 'dvc'
+      })
+    })
+  })
+
+  describe('remoteModifyUrl', () => {
+    it('should call createProcess with the correct parameters to modify a remote in the config', async () => {
```
mattseddon, line 157: [C] might want to state "url" in this description.

extension/src/setup/commands/index.ts
mattseddon: is there a test for this?

extension/src/cli/dvc/config.ts
mattseddon: any other commands we need to move off safe process?

mattseddon (approved):

extension/src/setup/commands/index.ts
julieg18: I can add a test to ensure `remote` catches errors.

extension/src/cli/dvc/config.ts
julieg18: I don't think so. Remote list commands are run periodically and remote remove commands are guaranteed to almost always fail since we run them twice in different scopes.

PR #5111, opened 2023-12-13
Do not pass down the grid ref

sroy3:
While testing https://github.com/iterative/vscode-dvc/pull/5097/files#r1424088937

For performance, the least amount of props we pass down (especially complex objects), the better it is.

Demo

https://github.com/iterative/vscode-dvc/assets/3683420/d4c9c36d-41e3-4902-9a38-b0d8e220714a


webview/src/plots/components/templatePlots/TemplatePlotGroups.tsx
sroy3, line 1: I'm adding `react/no-multi-comp` as a lint rule next (more changes needed)

webview/src/plots/components/customPlots/CustomPlotsGrid.tsx
```diff
@@ -1,10 +1,8 @@
-import React, { RefObject } from 'react'
```
julieg18, line 1: When I tested the demo for the first time, the custom plots didn't work:

https://github.com/iterative/vscode-dvc/assets/43496356/4fce24fd-97f1-47d0-accf-f565e4542c9c

But when I switched to `main` then back to this pr, it was working correctly and I can no longer recreate the plots breaking. 

Not sure if the plots breaking was due to this pr or just something weird happening on my end 🤔

sroy3: I added `useLayoutEffect` after all. I reset my project a couple of times to make sure it does not happen again.

mattseddon (approved):
Can you update the description, please (useLayoutEffect is used)

PR #5115, opened 2023-12-14
Upgrade Storybook to v7.6

mattseddon:
Comments inline


webview/.storybook/preview.ts
```diff
@@ -1,6 +1,5 @@
 import { Preview } from '@storybook/react'
 import { InternalVsCodeApi } from '../src/shared/api'
-import { action } from '@storybook/addon-actions'
```
mattseddon, line 3: [F] This is the source of the bug. Having `action` wrapped around `postMessage`/`setState` in the VS Code API meant that we could record the payload sent to the extension. I have no used this in a long time and I think we should be testing that functionality differently (i.e. automated tests). I have removed it so that we can move forwards but we can park this for a while if you feel differently.

webview/.storybook/main.ts
```diff
@@ -10,8 +10,6 @@ export const config: StorybookConfig = {
   staticDirs: ['../../extension/src/test/fixtures/plotsDiff/staticImages'],
   addons: [
     'storybook-addon-themes',
-    '@storybook/addon-essentials',
-    '@storybook/addon-interactions',
```
mattseddon, line 14: [F] Unless chromatic uses this I don't think we actually use it at all.

webview/.storybook/preview.ts
mattseddon: (`@storybook/addon-actions` tries to import `@storybook/globals` which is no longer available)

sroy3 (approved):

PR #5125, opened 2023-12-15
Remove Studio data request failed Toast

### LINKED ISSUE #5120: Studio: if project doesn't exist, don't keep showing error
If I am connected to Studio but the project doesn't exist on Studio, I get an error that I can't get rid of like this:

https://github.com/iterative/vscode-dvc/assets/2308172/a24363a6-b50f-474b-bfa7-5d74dcbbd516

Can we stop repeatedly showing the error?

julieg18:
Removes the "Unable to fetch data from Studio" Toast since we try to fetch data periodically (meaning user could get this multiple times in a session)

Fixes #5120

### DIFF
--- a/extension/src/experiments/data/index.ts
+++ b/extension/src/experiments/data/index.ts
@@ -26,7 +26,6 @@ import { COMMITS_SEPARATOR, gitPath } from '../../cli/git/constants'
 import { getGitPath } from '../../fileSystem'
 import { ExperimentsModel } from '../model'
 import { Studio } from '../studio'
-import { Toast } from '../../vscode/toast'
 
 export class ExperimentsData extends BaseData<ExperimentsOutput> {
   private readonly experiments: ExperimentsModel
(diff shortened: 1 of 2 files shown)


extension/src/experiments/data/index.ts
```diff
@@ -157,9 +156,6 @@ export class ExperimentsData extends BaseData<ExperimentsOutput> {
       })
     } catch {
       this.notifyChanged(defaultData)
-      void Toast.showError(
```
julieg18, line 160: We could add some logic to check if the user has already received the Toast and has a self-hosted URL, but that would add some extra complexity that I don't think is needed since we probably have few users using self-hosted URLs.

sroy3 (approved):

extension/src/experiments/data/index.ts
shcheklein: We can have a health status in the Setup screen instead I think. If connection is failing we can a message, or even change it to "yellow" in the Setup screen. WDYT?

PR #5126, opened 2023-12-15
Fix "Add Plot" breaking dvc.yaml when there is no list indentation

### LINKED ISSUE #5122: plots wizard: indentation issue
When I used the `Add a plot` functionality, I ended up with an invalid dvc.yaml due to wrong indentation:

https://github.com/iterative/vscode-dvc/assets/2308172/c3298657-1031-4de0-9f9f-fb96a09dc1bd

julieg18:
## `main`

https://github.com/iterative/vscode-dvc/assets/43496356/9f789725-66ee-4f5f-8dc5-4c53766ecc3a

## PR

https://github.com/iterative/vscode-dvc/assets/43496356/9e501fb9-c96f-4600-92c6-564a4d9f87d3

Fixes #5122


extension/src/fileSystem/index.ts
```diff
@@ -250,14 +250,20 @@ const getPlotsYaml = (plotObj: PlotConfigData, indentSearchLines: string[]) => {
...
-    )
+  const newPlotLines = yaml
+    .stringify({ plots: [getPlotYamlObj(plotObj)] }, { indent: spaces })
     .split('\n')
+
+  const doesYamlListItemHaveNoIndent = indentSearchLines.find(line =>
+    line.startsWith('-')
+  )
+  return doesYamlListItemHaveNoIndent
+    ? newPlotLines.map(line =>
```
julieg18, line 263: We check for if the lines have a list item with no indentation and trim our new plot lines to ensure that the new plot also has 0 indentation for its list item.

julieg18 (commented):
This solution feels hacky but I'm not seeing a better way 🤔. At least I don't think this route will get hit very often since list items with no indentation are not the norm in my experience.

extension/src/fileSystem/index.test.ts
```diff
@@ -828,6 +828,42 @@ describe('addPlotToDvcYamlFile', () => {
...
+  it('should add a new plot when plot list items have no indent', () => {
+    const mockDvcYamlContent = [
+      'plots:',
+      '- eval/importance.png',
+      '- Precision-Recall:',
+      '    x: recall',
+      '    y:',
+      '        eval/prc/train.json: precision',
+      '        eval/prc/test.json: precision'
+    ].join('\n')
```
sroy3, line 841: You can use string literals to test for format (`) instead of creating lines and joining them.

sroy3 (approved):

extension/src/fileSystem/index.test.ts
julieg18: Yes, I prefer creating lines since I find it easier to keep track of the exact spacing on each line.

PR #5131, opened 2023-12-17
Update "Studio" to "DVC Studio" in titles

julieg18:


webview/src/setup/components/studio/Settings.tsx
```diff
@@ -19,7 +19,7 @@ export const Settings: React.FC<
   return (
     <EmptyState isFullScreen={false}>
       <div>
-        <h1>Studio Settings</h1>
+        <h1>DVC Studio Settings</h1>
```
julieg18, line 22: I updated title-like text but I'm not sure if we should be referring to DVC Studio everywhere. Other places:

* in Command Pallete commands
* link description in experiments table
* toasts
* description tooltip in setup

shcheklein (approved):

webview/src/setup/components/studio/Settings.tsx
shcheklein: can you please share screenshots to review those places?

julieg18: Sure! Here are other places where we mention "Studio":  

image

image

image

image

image

image

sroy3 (approved):
Out of scope, but this could be a variable (`const STUDIO_PRODUCT_NAME = 'DVC Studio'`) that way we only have to change it once.

julieg18:
> Out of scope, but this could be a variable (const STUDIO_PRODUCT_NAME = 'DVC Studio') that way we only have to change it once.

Not sure about needing a variable. Product names don't tend to change often 🤔

webview/src/setup/components/studio/Settings.tsx
shcheklein: I don't have a strong opinion on this tbh. I think it's clear from the context already. @jendefig wdyt? :)

Issue #5087, opened 2023-12-08
Improve UX for plots debugging 


mnrozhkov:
## Why it's important?
I spent some time helping a few teams with plot debugging.  I helped multiple teams debug their plots. When users manually configure plots in `dvc.yaml`, two types of errors are likely to appear:
1. Provided data and configuration are not compatible with the plot template (e.g pass `str` to `numerical` data type) 
2. Typos in file names 

In every case, both DVC/DVCLive are "silent". We discussed here https://github.com/iterative/dvclive/issues/750 and plan to improve.

Eventually, users open VSCode extension and don't see plots at all!   This issue aims to find a better way to notify users about problems with plots and metrics visualization in VSCode. 

## AS IS
(I don't consider VSCode extension logs, because they are not appropriate for end users) 

**1. Error messages on hovering top-level tiles** 
At this moment we provide little means for understanding the error and debugging. As an example, out [internal discussion in Slack](https://iterativeai.slack.com/archives/C01APS0FHDM/p1702045563682519?thread_ts=1702043693.214339&cid=C01APS0FHDM)  

It seems that the best means we have now is error messages in tiles with branch/exp name: 
<img width="696" alt="image" src="https://github.com/iterative/vscode-dvc/assets/11489539/d2b83df5-315b-42d6-9176-19a0e6712bef">

However 
- It's difficult to find (found this feature accidentally when trying to find smth) 
- it was not expected that tiles would show additional information for `!` sign
- for a customer who used VSCode via Dev Container to run/debug code in the remote machine, there was no message on hover `!` - but only `copy` icon. Is it expected?

**2. Error messages on hovering side menu** 
- I was not aware of them
- Takes too long to show the message on hover 
<img width="343" alt="image" src="https://github.com/iterative/vscode-dvc/assets/11489539/a92a1473-c2ac-4caf-a6c1-3f94784c8e8a">

## DISCUSSION
1. Can we show an "empty" plot for every buggy plot specified in `dvc.yaml` and show an error message on it? 
<img width="935" alt="image" src="https://github.com/iterative/vscode-dvc/assets/11489539/7bd9e561-a455-4cd4-a045-e1045e1eac78">

2. Is there another way to help with plot debugging? 
3. Can we improve documentation to mention existing features for error messages? 
4. When people open Show Plots for the first time, there is nothing inside because no experiments are selected 
   - can we select "workspace" by default? I guess it's needed most of the time.

julieg18:
[...]

PR #5145, opened 2023-12-22
Make ribbon block error indicators more obvious

julieg18:
## Demo

Screenshot 2023-12-21 at 6 25 37 PM

Part of #5087

### DIFF
--- a/webview/src/plots/components/App.test.tsx
+++ b/webview/src/plots/components/App.test.tsx
@@ -49,6 +49,7 @@ import {
   toggleDragAndDropMode as toggleCustomPlotsDragAndDropMode
 } from './customPlots/customPlotsSlice'
 import { comparisonTableInitialState } from './comparisonTable/comparisonTableSlice'
+import plotRibbonStyles from './ribbon/styles.module.scss'
 import { plotsReducers, plotsStore } from '../store'
 import { vsCodeApi } from '../../shared/api'
 import {
@@ -2763,11 +2764,11 @@ describe('App', () => {
       })
     })
 
-    it('should show an error indicator for each revision with an error', () => {
+    it('should show an error icon and highlight the title for each revision with an error', () => {
       renderAppWithOptionalData({
         comparison: comparisonTableFixture,
         selectedRevisions: plotsRevisionsFixture.map(rev => {
-          if (rev.label === 'main') {
+          if (rev.label === 'main' || rev.description === '[exp-e7a67]') {
             return {
               ...rev,
               errors: ['error']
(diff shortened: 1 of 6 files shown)


webview/src/plots/components/ribbon/RibbonBlock.tsx
```diff
@@ -48,14 +53,14 @@ export const RibbonBlock: React.FC<RibbonBlockProps> = ({
             </div>
           </>
         ) : (
-          <div className={styles.title}>
+          <div className={cx(styles.title, hasError && styles.errorIndicator)}>
```
julieg18, line 56: If we think making the titles red is too noisy, we could just update the icons:

image

webview/src/plots/components/ribbon/RibbonBlockIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const RibbonBlockIcon: React.FC<{
+  hasError: boolean
+}> = ({ hasError }) =>
+  hasError ? (
+    <Error
+      width={14}
+      height={14}
+      className={cx(styles.blockIcon, styles.error)}
+      aria-label="Error Icon"
```
julieg18, line 14: Technically, you can label svgs but we don't really do in the extension. I added labels for testing but there might be a better way of handling this.

webview/src/plots/components/App.test.tsx
```diff
@@ -49,6 +49,7 @@ import {
   toggleDragAndDropMode as toggleCustomPlotsDragAndDropMode
```
julieg18, line 49: Next steps after this would be updating section title tooltips to include an error. Full design ideas list is in https://github.com/iterative/vscode-dvc/issues/5087#issuecomment-1859295790

shcheklein:
On this demo, what does red exactly mean? I feel ribbon alone is not enough, and if we do something else (see below, and you also suggested it in the ticket) we won't need the ribbon changes.

Can we take the idea from Studio (@sroy3 is working on errors / warnings now) - show errors in the header of the section, show a sign (yellow, green, red) nearby the section title (similar to setup screen). If we do this I think we don't need any changes in the ribbon, wdyt?

julieg18:
> On this demo, what does red exactly mean? I feel ribbon alone is not enough, and if we do something else (see below, and you also suggested it in the ticket) we won't need the ribbon changes.
> 
> Can we take the idea from Studio (@sroy3 is working on errors / warnings now) - show errors in the header of the section, show a sign (yellow, green, red) nearby the section title (similar to setup screen). If we do this I think we don't need any changes in the ribbon, wdyt?

Makes sense! I'll close this pr and work on showing errors in the header of the section.

PR #5152, opened 2023-12-23
Add error message to "Data Series" section 

julieg18:
## Demo

https://github.com/iterative/vscode-dvc/assets/43496356/fc25e368-50d7-4028-9388-cc7f6dbd4c79

## To Do

- [ ] Fix bug where errors don't show unless dvc.yaml changes in session
- [ ] Tests

Part of #5087

### DIFF
--- a/extension/src/plots/webview/messages.ts
+++ b/extension/src/plots/webview/messages.ts
@@ -380,6 +380,7 @@ export class WebviewMessages {
     }
 
     return {
+      errors: {},
       height: this.plots.getHeight(PlotsSection.TEMPLATE_PLOTS),
       nbItemsPerRow: this.plots.getNbItemsPerRowOrWidth(
         PlotsSection.TEMPLATE_PLOTS
--- a/extension/src/plots/webview/messages.ts
+++ b/extension/src/plots/webview/messages.ts
@@ -380,7 +380,10 @@ export class WebviewMessages {
     }
 
     return {
-      errors: {},
+      errors: this.errors.getTemplateErrors(
+        paths.flatMap(({ paths }) => paths),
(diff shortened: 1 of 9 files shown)


extension/src/plots/webview/messages.ts
```diff
@@ -380,6 +380,7 @@ export class WebviewMessages {
     }
 
     return {
+      errors: {},
```
julieg18, line 383: Currently, we don't send errors with paths + revs to the front end. We will need to update our plots error model to collect error information in the form of something like `{ [path: string]: { [rev: string]: string[] } }` for us to show in our error messages. 

Unfortunately, I'm off next week so I won't be able to get back to this till January 2nd.

julieg18:
> - [ ] Fix bug where errors don't show unless dvc.yaml changes in current session

Figured out the issue. We actually don't know what type a plot is unless it's built at least once in the current session. If the user starts a new session or reloads the window with a broken plot, the extension will not know the broken plot type and we can't show the error in the `Data Series` section: 

https://github.com/iterative/vscode-dvc/assets/43496356/2fcd5c67-41b5-495c-9909-bdeb8fb380c7

One option to fix this is to show the error messaging above the sections below the ribbon: 

image

We also could just leave it as is but it seems buggy 🤔 Any other ideas? cc @iterative/vs-code, @shcheklein

julieg18:
> One option to fix this is to show the error messaging above the sections below the ribbon:
> We also could just leave it as is but it seems buggy 🤔 Any other ideas? cc @iterative/vs-code, @shcheklein

Going to open a new pr that adds the error message above the sections and close this one for now. We could at least have a generic error message above all sections as a first iteration and come back to this later if we find a solution for the bug.

PR #5165, opened 2024-01-03
Add plots error message below ribbon

julieg18:


webview/src/plots/components/ZoomedInPlot.tsx
```diff
@@ -43,15 +44,7 @@ export const ZoomedInPlot: React.FC<ZoomedInPlotProps> = ({
   openActionsMenu
 }: ZoomedInPlotProps) => {
   const zoomedInPlotRef = useRef<HTMLDivElement>(null)
-
-  useEffect(() => {
```
julieg18, line 47: Moved this `useEffect` into a hook so we could use it in both `ZoomedInPlot` and `ErrorsModal`

webview/src/plots/components/ribbon/Errors.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <li className={styles.errors}>
+      <p className={styles.errorsTitle}>
+        <Error className={styles.errorsIcon} width="16" height="16" />
+        Errors Found
+      </p>
+      <button
+        className={styles.errorsButton}
+        onClick={() => dispatch(setShowErrorsModal(true))}
+      >
+        {errorsTotal === 1 ? 'Show error' : `Show ${errorsTotal} errors`}
```
julieg18, line 35: Is there a text we could use for both single and plural errors? Maybe `Show` or `Show More` would be better?

webview/src/plots/components/ribbon/Ribbon.tsx
```diff
@@ -95,6 +96,7 @@ export const Ribbon: React.FC = () => {
           onClear={() => removeRevision(revision.id)}
         />
       ))}
+      <Errors />
```
julieg18, line 99: As always, any comments/ideas for improving the design are welcome!

webview/src/stories/Plots.stories.tsx
```diff
@@ -83,6 +86,29 @@ const Template: StoryFn<{
 export const WithData = Template.bind({})
 WithData.parameters = CHROMATIC_VIEWPORTS_WITH_DELAY
 
+export const WithPlotsErrors = Template.bind({})
```
julieg18, line 89: Added a storybook for both revision and plot errors.

shcheklein:
@julieg18 what happens when you scroll it? can you show on a demo please? also it seems it jumps a bit when you show the popup / close it.

julieg18:
> @julieg18 what happens when you scroll it? can you show on a demo please? 

It's currently fixed to the ribbon:

https://github.com/iterative/vscode-dvc/assets/43496356/783c490b-0199-4aee-aeee-6275dd2206a8

Though looking at it when we're scrolling, I think we can decrease its height a bit. It's currently feels taller than the ribbon 🤔 

> also it seems it jumps a bit when you show the popup / close it.

Yes, I believe it's due to the fact that we are removing the scroll bar when a popup opens. Our plot modals work the same:

https://github.com/iterative/vscode-dvc/assets/43496356/1234c5cf-6522-458f-a68c-1d7be75b13e8

shcheklein:
Thanks @julieg18 , I think it looks good. One suggestion - get rid of the Errors Found and just move "Show 2 errors" there instead. I don't think we need both in this case.

webview/src/plots/components/App.tsx
```diff
@@ -81,6 +82,9 @@ export const feedStore = (
         case PlotsDataKeys.HAS_UNSELECTED_PLOTS:
           dispatch(updateHasUnselectedPlots(!!data.data[key]))
           continue
+        case PlotsDataKeys.PLOT_ERRORS:
```
sroy3, line 85: This is a note to self, we should reduce the number of branches in this switch/case by having an enum instead:

```
const actionToDispatch = {
  ...
  [PlotsDataKeys.PLOT_ERRORS]: updatePlotErrors
  [PlotsDataKeys.SELECTED_REVISIONS]: updateSelectedRevisions
  ...
}
```

and a simple `dispatch(actionToDispatch[key](data.data.key))`

sroy3: I'll open an issue for that. And don't worry, it's totally out of scope and has nothing to do with your PR.

webview/src/plots/components/ErrorsModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <tbody>
+          {errors.map(({ path, revs }, index) => (
+            <React.Fragment key={index}>
+              <tr>
+                <th colSpan={2} className={styles.errorsModalPlot}>
+                  {path}
+                </th>
+              </tr>
+              {revs.map(({ rev, msg }, index) => (
+                <tr key={index}>
```
sroy3, line 28: In this context, `index` is not a good  idea for a `key`. 

What happens if you have four errors and the 2nd gets removed? The 3rd and 4th will re-render because they do not have the same key anymore (more than re-rendering is actually happening, but let's keep this simple).  It isn't a big deal because errors are just text, and the list is short, but the longer it gets and the more complex the node is, the bigger of a problem it will be.

Keys do not need to be clean as IDs. As long as they are text and can be compared, it's okay. You could potentially have a block of text, and it wouldn't be a problem (it would be a bad idea, though).

In this case, what makes the error unique is the combination of `rev` and `msg`. These should not change for a specific error.

You can simply do `<tr key={`${rev}-${msg}`}>`.

webview/src/plots/components/ribbon/Errors.tsx
```diff
@@ -0,0 +1,36 @@
...
+    }
+    setErrorsTotal(total)
+  }, [errors])
+
+  if (errors.length === 0) {
+    return
+  }
+
+  return (
+    <li className={styles.errors}>
```
sroy3, line 26: It does not make much sense for this to be an `<li/>` element. I know that it is inside of the ribbon, but it does not fit with the rest of the list. The ribbon should be a wrapper that wraps the list and the errors.

sroy3 (approved):

webview/src/plots/components/ribbon/Errors.tsx
julieg18: Makes sense! I'll take care of this in a followup since the CSS will need to be adjusted and I don't want to accidentally break the ribbon.

webview/src/plots/components/ErrorsModal.tsx
julieg18: Good catch! Updated the keys.

PR #5194, opened 2024-01-15
Replace switch statements that feed the webviews stores

### LINKED ISSUE #5182: Reduce the number of branches in the switch / case that feed the webviews stores
We should reduce the number of branches in the switch/case by having an enum instead:

```
const actionToDispatch = {
  ...
  [PlotsDataKeys.PLOT_ERRORS]: updatePlotErrors
  [PlotsDataKeys.SELECTED_REVISIONS]: updateSelectedRevisions
  ...
}
```

and a simple `dispatch(actionToDispatch[key](data.data.key))`

_Originally posted by @sroy3 in https://github.com/iterative/vscode-dvc/pull/5165#discussion_r1446163741_

mattseddon:
Closes #5182


webview/src/experiments/components/App.tsx
```diff
@@ -1,19 +1,19 @@
-import React, { useCallback } from 'react'
-import { useDispatch } from 'react-redux'
+import { TableData } from 'dvc/src/experiments/webview/contract'
```
mattseddon, line 1: During the creation of this PR I moved the reducers in and out of this file. Rather than struggle to revert the exact order I resorted to using VS Code's "Organize Imports":

image

Rather than fight the new order I think we should start using this (means I have to think less to make things consistent).

webview/src/shared/dispatchActions.ts
```diff
@@ -0,0 +1,29 @@
+import { UnknownAction } from '@reduxjs/toolkit'
+import { TableData } from 'dvc/src/experiments/webview/contract'
+import { PlotsData } from 'dvc/src/plots/webview/contract'
+import { SetupData } from 'dvc/src/setup/webview/contract'
+import type { ExperimentsActions } from '../experiments/components/App'
+import type { ExperimentsDispatch } from '../experiments/store'
+import type { PlotsActions } from '../plots/components/App'
+import type { PlotsDispatch } from '../plots/store'
+import type { SetupActions } from '../setup/components/App'
+import type { SetupDispatch } from '../setup/store'
```
mattseddon, line 10: [F] Using `type` here to try and avoid including unnecessary code in each webview's bundle.

webview/src/experiments/components/App.tsx
sroy3: I used to do this, but it would clash with some linting we have. There might be a config that was needed.

sroy3 (approved):

PR #5195, opened 2024-01-15
Update filesystem watcher integration tests

mattseddon:
Follow up to https://github.com/iterative/vscode-dvc/pull/5147 (saw the same test fail yesterday, edit: and today)


extension/src/test/suite/experiments/data/index.test.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from '../../util'
 import { dvcDemoPath, getTestWorkspaceFolder } from '../../../util'
 import { ExperimentsData } from '../../../../experiments/data'
-import * as Watcher from '../../../../fileSystem/watcher'
```
mattseddon, line 15: [F] I think at some point this was stubbed somewhere in this suite but it is not anymore.

julieg18 (approved):

Issue #4917, opened 2023-10-26
plots: Interactive plots with bounding boxes 


BradyJ27:
In computer vision, specifically object detection, it is common for a pipeline to output images with bounding boxes displaying the area of interest for specific objects. When the image(s) are relatively small or packed with multiple objects, it can be hard to view these images. 

It would be nice to have some sort of interactive plots where the user can toggle on/off different objects based on labels. 

This may require some dvc or dvc-render changes first, but just opening here because it would be beneficial to have this implemented within the VS-Code extension.

mattseddon:
To clarify:

Are there multiple images created by the pipeline or do you have an original image that you want to compare to the output?

BradyJ27:
Usually it is multiple images. For example, we would print out the labels for the detections that we have made on the validation set. 

The following example is from YoloV8s default output. The dvclive yolo demo notebook is a good place to reproduce this. 

![77bbb4d8-97c4-4166-b55b-64ecaad1a4a1.jpeg](https://github.com/iterative/vscode-dvc/assets/26517764/521fd75f-8ec8-40ac-9e3e-2680172a6afd)

mattseddon:
Let me clarify the question. In the example above are there multiple images available for `000000000042.jpg`? Do you have an image available with each of the combinations of labels available? I.e 1 for each of 

- baseline 
- dog 
- motorcycle 
-  dog + motorcycle

I am not an expert in image manipulation but AFAIK removing these labelling boxes from an image is not a trivial task. 

Have you seen this done elsewhere?

BradyJ27:
So the image is just a copy of the training or validation image with bounding boxes added using some library (usually matplotlib). The bounding boxes are stored in a formatted file (xml, csv, json, or some custom format) normally like "label,x1,y1,x2,y2". 

So the approach would not be to manipulate the image with the boxes already on it, but rather set the image as the original image from the validation set, then place the interactive bounding boxes over the original image.

dberenbaum:
See https://docs.wandb.ai/guides/track/log/media#image-overlays for ideas on how others do this

BradyJ27:
One option for implementation would be a custom plot template, right?

Or is this something that's a little more in depth and actually a bigger feature?

mattseddon:
> One option for implementation would be a custom plot template, right?

No, I do not believe that you could shoe-horn the required data/image into the current DVC plots engine.

> Or is this something that's a little more in depth and actually a bigger feature?

[...]

PR #5227, opened 2024-01-22
Add bounding boxes plot frontend components

julieg18:
# `main` <- this <- https://github.com/iterative/vscode-dvc/pull/5241 <- #5250 <- #5305 <- #5313

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/68bf681b-d6e3-4069-818f-d3d5d2052633~

https://github.com/iterative/vscode-dvc/assets/43496356/f3b9d0aa-8ccf-452f-b4ae-89fe25fb5606

## To Do
- [ ] Add score filtering 
- [ ] Handle bounding box zoomed in plots
- [ ] Hide other box text labels when user is hovering over a particular box (https://github.com/iterative/vscode-dvc/pull/5227#discussion_r1477823743)
- [ ] Add way to deselect (or deselect/select) all boxes at once

A start to #4917


extension/src/plots/webview/contract.ts
```diff
@@ -50,8 +50,13 @@ export type SectionCollapsed = typeof DEFAULT_SECTION_COLLAPSED
...
 export type ComparisonRevisionData = { [revision: string]: ComparisonPlot }
 
+export type ComparisonBoundingBoxLabels = {
+  [label: string]: { selected: boolean; color: string }
+}
+
 export type ComparisonPlots = {
   path: string
+  boundingBoxLabels: ComparisonBoundingBoxLabels
   revisions: ComparisonRevisionData
```
julieg18, line 60: I've added two values to the `PlotsComparisonData`. `ComparisonBoundingBoxLabels` inside of `ComparisonPlots` (holds the labels information that exist in that plot/image row) and `ComparisonPlotBoundingBox` (holds box labels/coordinates for each single image).

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -128,21 +134,47 @@ export const ComparisonTableRow: React.FC<ComparisonTableRowProps> = ({
...
+                        className={styles.boundingBoxLabelsButton}
+                        style={{ background: color }}
+                        htmlFor={color.slice(1)}
+                      >
+                        {label}
+                      </label>
+                    </React.Fragment>
+                  ))}
+                </div>
+              )}
```
julieg18, line 177: Originally I was aiming for the labels to be on the far right like this:

image

but I realized that the CSS to get that working is more complicated than I originally thought. As an alternative, I've moved them to the left and made both the path and classes sticky:

Screenshot 2024-01-23 at 3 17 05 PM

but there's probably a better way to make this design. Any other ideas?

webview/src/test/boundingBoxesFixture.ts
```diff
@@ -0,0 +1,102 @@
...
+      boxes: [
+        { h: 30, w: 30, x: 190, y: 310 },
+        { h: 50, w: 60, x: 300, y: 320 }
+      ],
+      label: 'car'
+    }
+  ]
+}
+
+export const addBoundingBoxes = (
```
julieg18, line 67: I've created a util in the frontend that adds the bounding boxes to a comparison fixture for testing and storybooks. Once we've got the backend part of this working (https://github.com/iterative/dvc/issues/10198), this will all be removed and bounding box data will be part of our original comparison fixture.

webview/src/test/boundingBoxesFixture.ts
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable sonarjs/no-duplicate-string */
+import {
+  ComparisonBoundingBoxLabels,
+  ComparisonPlotBoundingBoxes,
+  ComparisonRevisionData,
+  PlotsComparisonData
+} from 'dvc/src/plots/webview/contract'
+
+const boxColors = [
```
julieg18, line 9: Looking at other `yolo` and other bounding box examples, they seem to rely on a variation of a [Tableau color palette](https://matplotlib.org/stable/gallery/color/named_colors.html#tableau-palette). I've grabbed the set of colors that `yolo` uses for the boxes for now.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
dberenbaum: Only thing I would warn is that it's possible there could be many classes, so we might want to consider how to handle that.

julieg18: Thanks for the reminder! I was planning to update the labels to have some kind of  "Show More" button when there are too many. I'll add it to the todo list in this issue.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -128,21 +134,47 @@ export const ComparisonTableRow: React.FC<ComparisonTableRowProps> = ({
...
+                    delay={NORMAL_TOOLTIP_DELAY}
+                  >
+                    <span className={styles.pathText}>{path}</span>
+                  </Tooltip>
+                </button>
+                <CopyButton value={path} className={styles.copyButton} />
+              </div>
+              {boundingBoxLabelsArr.length > 0 && (
+                <div className={styles.boundingBoxLabels}>
+                  <p className={styles.boundingBoxLabelsTitle}>Labels</p>
```
julieg18, line 156: I just realized I've been using the term "labels" instead of classes which makes more sense 🤦‍♀️. I'll need to update the term in variables and types as well as the text.

dberenbaum: I think either is acceptable @julieg18. I also used "labels" in https://github.com/iterative/dvclive/issues/766 and have seen both used in other products.

extension/src/plots/webview/contract.ts
julieg18: Currently, I was thinking that our backend would manage the "selected" state of the classes, but we could also manage that in redux instead.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
julieg18: Oh, good to know! I've already renamed everything to classes so I suppose we'll just keep it as classes 😄.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
```diff
@@ -128,21 +134,47 @@ export const ComparisonTableRow: React.FC<ComparisonTableRowProps> = ({
...
+                        type="checkbox"
+                        id={color.slice(1)}
+                        name="labels"
+                        value={label}
+                        defaultChecked={selected}
+                        className={styles.hiddenInput}
+                      />
+                      <label
+                        className={styles.boundingBoxClassesButton}
+                        style={{ background: color }}
```
sroy3, line 169: Can we ever get a background color on which to foreground color makes the text not accessible? If so, we can maybe calculate whether it should use a dark or light foreground color.

webview/src/plots/components/comparisonTable/cell/ComparisonTableCell.tsx
```diff
@@ -29,12 +35,21 @@ export const ComparisonTableCell: React.FC<{
...
-      <img
-        className={styles.image}
-        draggable={false}
-        src={plotImg.url}
-        alt={imgAlt || `Plot of ${path} (${plot.id})`}
-      />
+      {plotImg.url && plotImg.boundingBoxes ? (
+        <ComparisonTableBoundingBoxImg
+          src={plotImg.url}
+          boxCoords={plotImg.boundingBoxes}
```
sroy3, line 41: Instead of passing down complex objects (here and further up), can we just pass down an id and go grab the coordinates where they are used instead?

webview/src/plots/components/comparisonTable/styles.module.scss
```diff
@@ -137,6 +122,89 @@ $gap: 4px;
...
+  &:hover .copyButton {
+    opacity: 1;
+  }
+}
+
+.rowToggler {
+  border: none;
+  background: none;
+  color: $fg-color;
+  text-align: left;
```
sroy3, line 151: Is this property needed?

webview/src/plots/components/comparisonTable/styles.module.scss
```diff
@@ -137,6 +122,89 @@ $gap: 4px;
...
+.hiddenInput {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  border: 0;
+}
```
sroy3, line 190: What is the reason for hiding the input this way? (instead of changing its visibility, opacity, or display...)

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
sroy3: This might be problematic when pinning columns 

Screenshot 2024-01-24 at 10 31 46 AM

julieg18: > This might be problematic when pinning columns

Good catch! I forgot to check on pinned columns. Hmmm, we need to either find a better place or just tweak the z-index 🤔 

Screenshot 2024-01-24 at 10 07 21 AM
Screenshot 2024-01-24 at 9 56 26 AM

sroy3: I think the path shouldn't be longer than the first column.

One option would be to wrap the labels below the path and make sure it fits in the first column. The drawback would be that it would take more space.

Another option would be to use a dropdown multi-select menu. The biggest problem here is that always visible labels also act as a legend.

Could they be below the image instead? Like as another row? They could toggle with the image when clicking on the path and they'd have all of the path width more before wrapping.

webview/src/plots/components/comparisonTable/styles.module.scss
sroy3: That's fine. I think just making it `position: absolute` and `opacity: 0` is equally accessible, but as long as it isn't shown at all, there's no need to change it.

julieg18: > I think just making it position: absolute and opacity: 0 is equally accessible

While unlikely to happen, it would still be possible to click on the inputs in that case:

https://github.com/iterative/vscode-dvc/assets/43496356/9a9de37f-132e-48d9-830d-f7055be3b3ce

I think I can just set it to `display: none` for simplicity's sake.

webview/src/plots/components/comparisonTable/styles.module.scss
julieg18: Removed!

webview/src/plots/components/comparisonTable/cell/ComparisonTableCell.tsx
julieg18: Good point on the fact that we're passing down complex objects! I'll look into moving the bounding box data into an object that `ComparisonTableBoundingBoxImg` could grab from.

extension/src/plots/webview/contract.ts
```diff
@@ -151,11 +156,22 @@ export interface TemplatePlotsData {
   smoothPlotValues: SmoothPlotValues
 }
 
+export type ComparisonPlotBoundingBox = {
```
julieg18, line 159: @mattseddon, I've updated this type to not be plural, but, apologies, but I cannot, for the life of me, recall the other suggestion you made about the types during our mob programming session 🤦‍♀️. I think it was about `ComparisonBoundingBoxClasses`? 

I'll be sure to remember to bring my notebook for the next meeting 😅

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
julieg18: Thanks for the ideas! 

> I think the path shouldn't be longer than the first column.

Out of curiosity, why shouldn't the path be more than the first column? 

> Another option would be to use a dropdown multi-select menu. The biggest problem here is that always visible labels also act as a legend.

Agreed, I like keeping the labels visible for legend purposes. I am planning to put in a dropdown but it will function as a "Show More" button that will hold classes when there are too many. 

> Could they be below the image instead? Like as another row? They could toggle with the image when clicking on the path and they'd have all of the path width more before wrapping.

That could work! Example:

Screenshot 2024-01-25 at 7 54 01 AM

It does look a little weird when it comes to image by step plots. But maybe the styles just need to tweaked a bit. 

> One option would be to wrap the labels below the path and make sure it fits in the first column. The drawback would be that it would take more space.

Here's an example of placing the classes row above the images but separate from the path:

Screenshot 2024-01-25 at 8 06 01 AM

@sroy3, what do you think?

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
julieg18: If I understood the libraries that I got the colors from, they always use white as the text color. While testing, while I think a couple colors could benefit from a darker color, white 
does seem to work well enough:

Screenshot 2024-01-25 at 8 29 08 AM

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
sroy3: > Out of curiosity, why shouldn't the path be more than the first column?

There is a line at the right of the pinned column and it would overflow otherwise

> Here's an example of placing the classes row above the images but separate from the path
I think that's what would work best. Do you think it would now be possible to align the classes right now? There'd be more room for them. If not, I think it's a great start until we figure out a better solution (if there is one).

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
sroy3: Yes, apart from the neon green they all look fine. And I don't think there is anything we can do about that neon green anyway.

webview/src/plots/components/comparisonTable/ComparisonTableRow.tsx
julieg18: > I think that's what would work best. Do you think it would now be possible to align the classes right now? There'd be more room for them. If not, I think it's a great start until we figure out a better solution (if there is one).

Great! Yes, it shouldn't be difficult. I'll open a second chain PR that moves the classes below the path for easier reviewing and viewability.

julieg18: Updated design and fixed overflow issue in #5240

webview/src/plots/components/comparisonTable/cell/ComparisonTableBoundingBoxImg.tsx
```diff
@@ -0,0 +1,69 @@
...
+          selected && (
+            <ComparisonTableBoundingBoxColorFilter key={label} color={color} />
+          )
+      )}
+      <image href={src} width={naturalWidth} height={naturalHeight} />
+      {boxCoords.map(({ label, boxes }) => {
+        const labelColor = classes[label].color
+        return boxes.map(({ h, w, x, y }) => (
+          <React.Fragment key={label + h + w + x + y}>
+            <text
```
mattseddon, line 46: With the current design, I don't think we need this text as we are repeating a lot of information:

image

I.e. there is a big box that shows traffic light objects are yellow. I would drop this text unless we include the threshold along with the text. 

Once there are a lot of boxes this will become even more crowded:

image

WDYT @dberenbaum?

mattseddon: Let's also consider how we can filter the boxes by a provided threshold (i.e. "Objects detected with prediction threshold >= 0.5")

mattseddon:
Something else that we will need to revisit/implement:

How are bounding boxes displayed on/with a zoomed image? In VS Code we currently open the image in an editor and that is it.

webview/src/plots/components/comparisonTable/cell/ComparisonTableBoundingBoxImg.tsx
julieg18: > I.e. there is a big box that shows traffic light objects are yellow. I would drop this text unless we include the threshold along with the text.
> Once there are a lot of boxes this will become even more crowded

Good point! I chose to add labels since all classes toggle buttons may not be showing (if there are too many, we're going to add a "Show More" button that shows the hidden classes).  

No strong preference though, I'm happy to remove them!

extension/src/plots/webview/contract.ts
julieg18: Great points! I've updated the types to be `ComparisonClassDetails` and `ComparisonPlotClass`.

webview/src/plots/components/comparisonTable/cell/ComparisonTableBoundingBoxImg.tsx
```diff
@@ -0,0 +1,69 @@
...
+    >
+      {Object.entries(classes).map(
+        ([label, { color, selected }]) =>
+          selected && (
+            <ComparisonTableBoundingBoxColorFilter key={label} color={color} />
+          )
+      )}
+      <image href={src} width={naturalWidth} height={naturalHeight} />
+      {boxCoords.map(({ label, boxes }) => {
+        const labelColor = classes[label].color
```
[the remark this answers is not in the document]
julieg18: Done!

sroy3 (approved):

webview/src/plots/components/comparisonTable/cell/ComparisonTableBoundingBoxImg.tsx
julieg18: @AlexandreKempf, what do you think about keeping text labels on each individual box or would it be better to remove them?

AlexandreKempf: Working on object detection, we get used to overlapping text ^^", so we can keep them it is not a problem.
In my opinion, the best behavior would be to have the image with all the labels when not hovered. When hovered, only the bounding box below the mouse gets its text displayed. It is pretty tricky to get which bounding boxes to select based on mouse location, especially with lots of overlapping cases. So generally, what we end up doing is filtering the bounding boxes based on label or score ;)

AlexandreKempf:
Images shown by Matt on your work looks super nice ! Congrats :tada:

webview/src/plots/components/comparisonTable/cell/ComparisonTableBoundingBoxImg.tsx
julieg18: > In my opinion, the best behavior would be to have the image with all the labels when not hovered. When hovered, only the bounding box below the mouse gets its text displayed. It is pretty tricky to get which bounding boxes to select based on mouse location, especially with lots of overlapping cases. 

Sounds good! I'll add this feature to our todo list.

extension/src/cli/dvc/contract.ts
```diff
@@ -233,10 +233,16 @@ export type TemplatePlotOutput = {
...
+export type BoundingBox = {
+  box: { left: number; right: number; top: number; bottom: number }
+  score: number
+}
+
 export type ImagePlotOutput = {
   revisions: string[]
   type: PlotsType
   url: string
+  annotations?: { [label: string]: BoundingBox[] }
```
julieg18, line 245: DVC spec has been changed from `boxes` to `annotations`. I've updated the code accordingly.

mattseddon:
Will get back to this later

PR #5241, opened 2024-01-26
Store bounding box plot coordinates in single obj

julieg18:


webview/src/plots/components/comparisonTable/cell/ComparisonTableCell.tsx
```diff
@@ -17,6 +22,17 @@ export const ComparisonTableCell: React.FC<{
 }> = ({ path, plot, imgAlt, boundingBoxClasses }) => {
   const plotImg = plot.imgs[0]
 
+  const boundingBoxCoords = useSelector(
+    (state: PlotsState) => state.comparison.boundingBoxPlotCoords
+  )
+  const getCellBoundingBoxCoords = createSelector(
```
julieg18, line 28: Discussed with @sroy3 in Slack, and we can use [`createSelector`](https://redux-toolkit.js.org/api/createSelector) to reduce rerendering when the coords change.

sroy3: This selector should be created outside of the component and used with `useSelector`.

`const cellBoundingBoxCoordsSelector = createSelector((state: PlotsState) => state.comparison.boundingBoxPlotCoords....)`

`const cellBoundingBoxCoords = useSelector((state) => cellBoundingBoxCoordsSelector(state, plot, path))`

julieg18: Done!

webview/src/plots/components/comparisonTable/cell/ComparisonTableCell.tsx
```diff
@@ -1,13 +1,24 @@
...
 } from 'dvc/src/plots/webview/contract'
 import { ComparisonTableLoadingCell } from './ComparisonTableLoadingCell'
 import { ComparisonTableMissingCell } from './ComparisonTableMissingCell'
 import { ComparisonTableBoundingBoxImg } from './ComparisonTableBoundingBoxImg'
 import styles from '../styles.module.scss'
 import { zoomPlot } from '../../../util/messages'
+import { PlotsState } from '../../../store'
+
+const selectPlotClasses = (state: PlotsState) => state.comparison.plotClasses
+const selectClasses = createSelector(
```
sroy3, line 17: It's good practice to add the suffix `selector` (`classesSelector`, `plotClassesSelector`)

webview/src/plots/components/comparisonTable/cell/ComparisonTableCell.tsx
```diff
@@ -35,10 +49,10 @@ export const ComparisonTableCell: React.FC<{
       onClick={() => zoomPlot(plotImg.url)}
       data-testid="image-plot-button"
     >
-      {plotImg.url && plotImg.classes ? (
+      {plotImg.url && classes ? (
         <ComparisonTableBoundingBoxImg
           src={plotImg.url}
-          classes={plotImg.classes}
+          classes={classes}
```
sroy3, line 55: Why not select the classes inside of `ComparisonTableBoundingBoxImg` to avoid passing this as a prop? It can always return `null` if there are no classes.

sroy3 (approved):

webview/src/plots/components/comparisonTable/cell/ComparisonTableCell.tsx
julieg18: Done!

webview/src/plots/components/comparisonTable/cell/ComparisonTableCell.tsx
julieg18: Good point! Updated

PR #5250, opened 2024-01-30
Add backend logic for bounding box plots

julieg18:
# `main` <- #5227 <- #5241 <- this <- #5305

This PR updates our comparison dvc output fixture with bounding box info and updates the backend accordingly. It also adds logic for a user toggling boxes on/off.


extension/src/plots/model/collect.ts
```diff
@@ -376,6 +477,80 @@ export const collectSelectedComparisonPlots = ({
...
+    if (selectedState === false) {
+      continue
+    }
+
+    imgClasses.push({
+      boxes: boxes.map(({ x_min, x_max, y_min, y_max }) => ({
+        h: y_max - y_min,
+        w: x_max - x_min,
+        x: x_min,
+        y: y_min
```
julieg18, line 507: We could also perform this calculation (getting `h`, `w`, `x`, `y`) when we're collecting `comparisonData` from the DVC output.

extension/src/cli/dvc/contract.ts
```diff
@@ -233,10 +233,18 @@ export type TemplatePlotOutput = {
   type: PlotsType
 }
 
+export type BoundingBox = {
+  x_min: number
+  x_max: number
+  y_min: number
+  y_max: number
+}
```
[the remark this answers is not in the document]
julieg18: Updated to `top`, `bottom`, `right`, `left` based off the top-left corner after a discussion with Alex ([see his comment](https://github.com/iterative/dvclive/issues/766#issuecomment-1920978138)):

image

extension/src/plots/model/collect.ts
```diff
@@ -205,10 +211,26 @@ const getMultiImageInd = (path: string) => {
...
 }
 
+const collectImageBoundingBoxes = (
+  boundingBoxes: { label: string; box: BoundingBox }[]
+): { [label: string]: BoundingBox[] } => {
+  const acc: { [label: string]: BoundingBox[] } = {}
+
+  for (const { label, box } of boundingBoxes) {
+    if (!acc[label]) {
+      acc[label] = []
```
sroy3, line 221: You don't need the box in the array yet?

extension/src/plots/model/collect.ts
```diff
@@ -312,60 +342,131 @@ export const collectData = (output: PlotsOutput): DataAccumulator => {
...
+  '#00c2ff',
+  '#344593',
+  '#6473ff',
+  '#0018ec',
+  '#8438ff',
+  '#520085',
+  '#cb38ff',
+  '#ff95c8',
+  '#ff37c7'
+]
```
sroy3, line 378: There is also the file `extension/src/model/status/color.ts` that has a lot of colors inside. We should probably try to merge these two list (trying to reuse the same colors as much as possible) or at least have them in the same file so that it is easier to find.

extension/src/plots/model/collect.ts
```diff
@@ -312,60 +342,131 @@ export const collectData = (output: PlotsOutput): DataAccumulator => {
...
+            imgs,
+            getComparisonPlotImg,
+            id,
+            path
+          )
         : [{ errors: undefined, loading: false, url: undefined }]
     }
   }
+
+  for (const [ind, label] of [...boundingBoxClassLabels].entries()) {
```
sroy3, line 440: `.entries()` should work directly on `Set`  without conversion.

extension/src/plots/model/collect.ts
```diff
@@ -376,6 +477,80 @@ export const collectSelectedComparisonPlots = ({
...
+  comparisonData: ComparisonData
+  paths: string[]
+  selectedRevisionIds: string[]
+}) => {
+  const acc: ComparisonPlotClasses = {}
+
+  for (const path of paths) {
+    for (const id of selectedRevisionIds) {
+      for (const img of comparisonData[id][path]) {
+        collectSelectedImgComparisonPlotClasses({
```
sroy3, line 540: Is there a way to refactor `collectSelectedImgComparisonPlotClasses` so that it does not have to be nested inside three loops?

extension/src/test/fixtures/plotsDiff/index.ts
```diff
@@ -936,10 +1004,60 @@ const getIndFromComparisonMultiImgPath = (path: string) => {
...
+}) => {
+  const classAcc: {
+    [label: string]: {
+      label: string
+      boxes: {
+        h: number
+        w: number
+        x: number
+        y: number
+      }[]
```
sroy3, line 1030: Why isn't this reusing [this](https://github.com/iterative/vscode-dvc/pull/5250/files#diff-97ae15b18dd87d9190444f391683d3bc0d68c9187d1af5af01870e7631ea323dR247) type or the `BoundingBox ` one?

extension/src/test/fixtures/plotsDiff/index.ts
```diff
@@ -936,10 +1004,60 @@ const getIndFromComparisonMultiImgPath = (path: string) => {
...
+
+    if (!classAcc[label]) {
+      classAcc[label] = { label, boxes: [] }
+    }
+
+    classAcc[label].boxes.push({
+      x: box.left,
+      y: box.top,
+      w: box.right - box.left,
+      h: box.bottom - box.top
```
sroy3, line 1045: Isn't this doing the same as [this](https://github.com/iterative/vscode-dvc/pull/5250/files#diff-b8591ca0f7ff9633208614960ef48ca0fd0bcc1fb055531bbf943a8250a14d86R503)

julieg18: Yes, it is doing the same thing. Is there another way to about this? I think we want to keep the webview message creation logic separate from our extension's collect logic?

extension/src/plots/model/collect.ts
julieg18: Apologies, I'm not sure what you mean 🤔 Are you saying we should just leave the `comparisonData` boxes as `{ label: string, box: BoundingBox }[]`, and only convert them to the new format (`{ label: BoundingBox[] }` when we are grabbing the selected plot data?

extension/src/test/fixtures/plotsDiff/index.ts
julieg18: Good catch!

extension/src/plots/model/collect.ts
sroy3: Below, you do `acc[label].push(box)`, but here you initialize an empty array. What is the reason for not doing `acc[label] = [box]`

extension/src/plots/model/collect.ts
```diff
@@ -312,60 +342,131 @@ export const collectData = (output: PlotsOutput): DataAccumulator => {
...
+  imgs: ImagePlot[],
+  getComparisonPlotImg: GetComparisonPlotImg,
+  id: string,
+  path: string
+) => {
+  const plotImgs = []
+  for (const img of imgs) {
+    plotImgs.push(getComparisonPlotImg(img, id, path))
+
+    if (img.boundingBoxes) {
```
sroy3, line 391: This is a side effect of a function that returns something. It feels unexpected. `boundingBoxClassLabels` should be created inside this function and returned as well and recollected on the other side.

extension/src/plots/model/collect.ts
julieg18: > Below, you do acc[label].push(box), but here you initialize an empty array. What is the reason for not doing acc[label] = [box]

Oh, ok, thanks for the clarification! Looking at the two methods for collection one or more boxes: 

```js
  for (const { label, box } of boundingBoxes) {
    if (!acc[label]) {
      acc[label] = [box]
      continue;
    }

    acc[label].push(box)
```

```js
  for (const { label, box } of boundingBoxes) {
    if (!acc[label]) {
      acc[label] = []
    }

    acc[label].push(box)
```

I didn't think there was much of a difference, but I'm happy to update if you prefer the first example!

extension/src/test/fixtures/plotsDiff/index.ts
sroy3: Can't you change `w`, `x`, `h` and `y` on the frontend as well so that we have a continuity throughout the extension?

julieg18: > Can't you change w, x, h and y on the frontend as well so that we have a continuity throughout the extension?

Good point, that would simplify things! I'll update the code so that instead of the backend performing the calculations, the calculations are done in `ComparisonTableBoundingBoxImg`.

extension/src/plots/model/collect.ts
julieg18: I vote we keep the colors separate since I don't want certain class colors to feel like they correlate with an particular experiments. 

Good idea on having both color sets in a file though! I'll save this for a followup since that would involve touching some experiment files and this PR is already pretty big.

extension/src/plots/model/collect.ts
julieg18: Good catch! I've adjusted the code so that the class collection is in a separate function outside of the img collection to reduce complexity.

extension/src/test/fixtures/plotsDiff/index.ts
julieg18: Done!

extension/src/test/fixtures/plotsDiff/index.ts
julieg18: Done!

extension/src/plots/model/collect.ts
julieg18: Updated the code to no longer have three nested loops!

extension/src/plots/model/collect.ts
julieg18: We now perform this calculation inside of just the react img file.

extension/src/test/fixtures/plotsDiff/index.ts
mattseddon: Can we leave left, right, top and bottom as variable names for as long as possible in the system? Also, how is x box.left?

julieg18: > Can we leave left, right, top and bottom as variable names for as long as possible in the system?

I believe we're [already doing that ](https://github.com/iterative/vscode-dvc/pull/5250/files#diff-9602466799fe1c9d6f9748ba405f6e319440fcaccc9473a1fcf274c133868adbL63). We currently use left, right, top and bottom terms throughout the frontend and the backend, getting x, y, width, height when we are creating the SVGs. Unless I'm misunderstanding what you mean?

extension/src/plots/model/collect.ts
```diff
@@ -376,6 +466,93 @@ export const collectSelectedComparisonPlots = ({
...
+}: {
+  comparisonClassesSelected: ComparisonClassesSelected
+  comparisonData: ComparisonData
+  paths: string[]
+  selectedRevisionIds: string[]
+}) => {
+  const acc: ComparisonPlotClasses = {}
+
+  for (const id of selectedRevisionIds) {
+    for (const path of paths) {
```
sroy3, line 542: What I meant by removing nested loops was not moving one inside of another function (I know realize we have 4 nested loops total), but can we change the code to drop some or at least make them more performant. 

I see that in the fourth nested loop, we check for selected state. Maybe for a start, we could loop only through `paths` that are also in `comparisonClassesSelected`, since we only care about  labels in `comparisonClassesSelected[path]`.

There might be other optimizations to be made as well.

extension/src/plots/model/index.ts
```diff
@@ -266,6 +274,20 @@ export class PlotsModel extends ModelWithPersistence {
...
   }
 
+  public getComparisonPlotClasses(paths: string[]) {
+    const selectedRevisionIds = this.getSelectedRevisionIds()
+    if (!definedAndNonEmpty(selectedRevisionIds)) {
+      return {}
+    }
+
+    return collectSelectedComparisonPlotClasses({
+      comparisonClassesSelected: this.getComparisonClassesSelected(),
```
sroy3, line 284: If this is empty, should this return `{}` as well? This would avoid having to loop through `selectedRevisionIds` for nothing.

extension/src/plots/webview/contract.ts
```diff
@@ -134,8 +134,11 @@ export type CustomPlotsData = {
 export const isVegaPlot = (plot: Plot): plot is TemplatePlot =>
   plot.type === PlotsType.VEGA
 
-export type ImagePlot = ImagePlotOutput & {
+export type ImagePlot = Omit<ImagePlotOutput, 'boundingBoxes'> & {
   ind?: number
+  boundingBoxes?: {
+    [label: string]: BoundingBox[]
+  }
```
mattseddon, line 141: [Q] Why not get data in this format from DVC? (We are being asked to define the shape of the API output)

julieg18: > [Q] Why not get data in this format from DVC? (We are being asked to define the shape of the API output)

Good point! That would reduce transforming on both Studio and VSCode's side. I'll mention that in my response tomorrow.

julieg18: Schema has been decided to be an object: 

```
{
  "boxes": {
    "cat": [
      { "left": 100, "right": 110, "top": 5, "bottom": 20 },
      { "left": 30, "right": 55, "top": 75, "bottom": 90 }
    ],
    "dog": [
      { "left": 80, "right": 100, "top": 25, "bottom": 50 }
    ]
  }
}
```

extension/src/plots/model/collect.ts
julieg18: > What I meant by removing nested loops was not moving one inside of another function (I know realize we have 4 nested loops total), but can we change the code to drop some or at least make them more performant.

Thanks for the clarification! 

> I see that in the fourth nested loop, we check for selected state. Maybe for a start, we could loop only through paths that are also in comparisonClassesSelected, since we only care about labels in comparisonClassesSelected[path].

`comparisonClassesSelected` only holds classes that the user has toggled directly, it doesn't hold every visible class. Unless I'm missing something, we can't rely on it to detect which paths need to be looped over to collect classes.

The four loops are:
1. looping over each revision (up to 7)
2. looping over each path (could be a lot depending on how many plots the user has)
3. looping over each path img (will usually be just one, unless it's an image step plot)
4. looping over each class, gathering only the selected class boxes

I don't see a way to get rid of the first three loops, since for us to gather a plot's classes, we need to check for classes in each single image and collect the data. Atleast the third loop will likely just be a single item array. 

We could get rid of the fourth loop by passing all image box data to the frontend and have `ComparisonTableBoundingBoxImg` filter out the boxes, but that means passing larger arrays that the component needs to loop over. Seems like we would want to limit the data passed to the frontend as much as possible but please correct me if I'm wrong.

Another alternative is having `collectSelectedComparisonPlots` gather class information along with the selected comparison plot data. I chose not to do this since that would make the function more complex, but if we think it's worth it for performance, I'm alright with doing that. 

### TL;DR

I'm not seeing a good way to get rid of any of our four loops. I'll review the loops and see what we can do to optimize them for now. 

@sroy3, what do you think?

extension/src/plots/model/index.ts
julieg18: On further inspection, `comparisonClassesSelected` is only going to hold classes that the user has deselected/selected, it doesn't hold every visible class so we are unable to use it as a check for if there are selected classes or not.

julieg18: This variable probably needs to avoid confusion. Maybe something like `comparisonClassesSelectedState` 🤔

extension/src/plots/model/collect.ts
julieg18: Great idea! Updated to use a variable for getting the index.

extension/src/plots/model/collect.ts
julieg18: Moved colors into `src/common/colors.ts` but maybe there is a better spot.

extension/src/plots/model/collect.ts
julieg18: Reviewed optimizing the loops! Instead of looping over every revision/path in `comparisonData`, we could start by looping over the already formed `plots` array. The `plots` already have a collected `classDetails` that tells you if a path has selected classes or not, lettings us skip early. Aka:

1. looping over each `plots` path (skip over if `classDetails` is empty)
2. looping over each revision
3. looping over each path img (will usually be just one, unless it's an image step plot)
4. looping over each class, gathering only the selected class boxes

Not seeing any other optimizations besides the two ideas I mentioned in my previous comment (have `collectSelectedComparisonPlots` collect box data, have `ComparisonTableBoundingBoxImg` do the box filtering).

sroy3: One thing I'm seeing is this:
```
if (selectedState === false) {
  continue
}
```
```
if (imgClasses.length === 0) {
  return
}
```
    
Why do we have to go through four loops to return nothing when we know it before hand? I would make this the first loop (the selected classes). 

If there are no other possible optimizations, then I guess it's fine to keep as is.

sroy3 (approved):

extension/src/common/colors.ts
```diff
@@ -0,0 +1,41 @@
...
+  '#4299e1',
+  '#ed8936',
+  '#f56565'
+] as const
+
+export type Color = (typeof colorsList)[number]
+
+export const copyOriginalColors = (): Color[] => [...colorsList]
+
+const boundingBoxColorsList = [
```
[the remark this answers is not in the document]
julieg18: Apologies, I thought we agreed that we don't want the experiment colors (`colorsList`) to be used to represent a bounding box class, because it might seem if that class is related to a particular selected experiment? 

Did I misunderstand your previous comment? If you would like us to use the experiment colors as possible bounding box class colors, I'm happy to update :) 

Or are you meaning something else entirely that I'm missing?

extension/src/plots/model/collect.ts
julieg18: That's strange, I could have sworn I updated this. Must have reverted it accidentally when I was adding a stash in one of the last commits I added. Apologies, I'll update this!

extension/src/plots/model/index.ts
julieg18: No, `collectSelectedComparisonPlotClasses` could still return selected class boxes. Currently, we select all classes in a plot by default and `comparisonClassesSelected` only gets updated with a class (setting it as true or false) when a user clicks directly on a class button. Hope that makes sense :)

extension/src/plots/model/collect.ts
julieg18: > if (selectedState === false) {
  continue
}
> Why do we have to go through four loops to return nothing when we know it before hand? I would make this the first loop (the selected classes).

Aren't we already getting the selected classes [in the first loop](https://github.com/iterative/vscode-dvc/pull/5250/files#diff-b8591ca0f7ff9633208614960ef48ca0fd0bcc1fb055531bbf943a8250a14d86R519)? In fact the line `if (selectedState === false) {...` no longer exists in the code. Did Github direct you the latest changes or is there something I'm missing here? 

> One thing I'm seeing is this:
> if (imgClasses.length === 0) {
  return
}

Good catch on `if (imgClasses.length === 0) {...`! I didn't spot the fact that we might be needlessy looping over selected classes if an img has no attached boxes to begin with. I'll update the code to check for `img.boxes` before running `getSelectedImgComparisonPlotClasses`.

extension/src/plots/model/collect.ts
julieg18: > That's strange, I could have sworn I updated this. Must have reverted it accidentally when I was adding a stash in one of the last commits I added. Apologies, I'll update this!

Updated!

extension/src/common/colors.ts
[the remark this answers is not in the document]
sroy3: You're right, that makes sense. Sorry, I forgot about that fact.

julieg18:
Going to go ahead and merge this and #5305 into our first PR since they've been sitting for a while and are approved.

PR #5251, opened 2024-01-31
Do not brick the extension when trying to get user_id for telemetry

mattseddon:
Should fix [this issue](https://discord.com/channels/485586884165107732/842220310585147452/1201884720225587270) from Discord

### DIFF
--- a/extension/src/telemetry/uuid.ts
+++ b/extension/src/telemetry/uuid.ts
@@ -40,30 +40,31 @@ const writeMissingConfigs = (
 }
 
 const readOrCreateConfig = (): string | undefined => {
-  const dvcAppDir = getDVCAppDir()
-  const iterativeAppDir = getIterativeAppDir()
+  try {
+    const dvcAppDir = getDVCAppDir()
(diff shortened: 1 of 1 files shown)


extension/src/telemetry/uuid.ts
```diff
@@ -40,30 +40,31 @@ const writeMissingConfigs = (
 }
 
 const readOrCreateConfig = (): string | undefined => {
-  const dvcAppDir = getDVCAppDir()
-  const iterativeAppDir = getIterativeAppDir()
+  try {
```
mattseddon, line 43: [F] This is called by `getUserId` which is in turn called here:

```
  reporter.sendTelemetryEvent(
    eventName as string,
    { ...sanitizedProperties, user_id: getUserId() },
    measurements
  )
```

☝🏻 this is essentially a `noop` if the user has disabled telemetry but the above code is being called regardless

julieg18 (approved):

sroy3 (approved):

PR #5305, opened 2024-02-16
Improve bounding box frontend components

julieg18:
# `main` <- #5227 <- https://github.com/iterative/vscode-dvc/pull/5241 <- https://github.com/iterative/vscode-dvc/pull/5250 <- this <- #5313

Updates our class logic to handle different use cases and adds other various improvements I spotted while working in Studio (details inline). 

## Demo

~https://github.com/iterative/vscode-dvc/assets/43496356/50a95839-ce6f-47f5-819a-d8a1ea953cb6~

https://github.com/iterative/vscode-dvc/assets/43496356/438f6766-774b-47ef-bd73-e9030ce90459

PR #5313, opened 2024-02-19
Fix image by step plots with bounding boxes

julieg18:
# `main` <- #5227 <- https://github.com/iterative/vscode-dvc/pull/5241 <- #5250 <- #5305 <- this

Fixes a bug with bounding boxes not being shown in every image in multi img plots.

### DIFF
--- a/extension/src/plots/webview/contract.ts
+++ b/extension/src/plots/webview/contract.ts
@@ -94,7 +94,9 @@ export type ComparisonPlotClass = {
 }
 
 export type ComparisonPlotClasses = {
-  [revision: string]: { [path: string]: ComparisonPlotClass[] }
+  [revision: string]: {
+    [path: string]: { [imgInd: number]: ComparisonPlotClass[] }
+  }
 }
 
 export interface PlotsComparisonData {
--- a/extension/src/test/suite/plots/index.test.ts
+++ b/extension/src/test/suite/plots/index.test.ts
@@ -259,6 +259,55 @@ suite('Plots Test Suite', () => {
       })
     })
 
+    it('should handle a plots diff output with a comparison multi plot containing bounding boxes', async () => {
+      const multiImgPath = join('plots', 'image', '5.jpg')
+      const mockBoxes = {
+        car: [{ box: { bottom: 0, left: 0, right: 0, top: 0 }, score: 0.5 }]
+      }
+      const imgDataWithBoundingBoxes = plotsDiffFixture.data[multiImgPath].map(
+        img => {
+          const [rev] = img.revisions
+          if (rev === 'main' || rev === 'exp-e7a67') {
+            return {
+              ...img,
+              boxes: mockBoxes
+            }
+          }
+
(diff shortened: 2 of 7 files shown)


extension/src/plots/webview/contract.ts
```diff
@@ -94,7 +94,9 @@ export type ComparisonPlotClass = {
 }
 
 export type ComparisonPlotClasses = {
-  [revision: string]: { [path: string]: ComparisonPlotClass[] }
+  [revision: string]: {
+    [path: string]: { [imgInd: number]: ComparisonPlotClass[] }
```
julieg18, line 98: In image by step plots case, path is a directory and could hold multiple images' class coordinates. I've amended this by adding a `ind` key. 

Another option that we could do is an nested array::

```
[revision: string]: { [path: string]: Array<ComparisonPlotClass[] | null>  }
```

But it felt like collection code would be a bit more complicated (would have to filter out created arrays that were filled with only `null` while keeping the ones that have at least one img with classes while ensuring we don't break accidentally break the order of the imgs' classes).

extension/src/test/suite/plots/index.test.ts
```diff
@@ -259,6 +260,158 @@ suite('Plots Test Suite', () => {
...
     })
 
+    describe('Bounding Box Plots', () => {
+      it('should handle image by step plots with bounding boxes', async () => {
+        const multiImgDirPath = join('plots', 'image')
+        const multiImgPath = join('plots', 'image', '5.jpg')
+        const mockBoxes = {
+          car: [{ box: { bottom: 0, left: 0, right: 0, top: 0 }, score: 0.5 }]
+        }
+        const imgDataWithBoundingBoxes = plotsDiffFixture.data[
```
julieg18, line 270: Another option is adding a new image by step plot with bounding boxes plot to our base comparison fixture.

extension/src/plots/webview/contract.ts
sroy3: You know the context better than I do, so don't change it if it doesn't make sense or if it's still more complicated, but would using an empty array instead of null make thing simpler? So:
```
[revision: string]: { [path: string]: ComparisonPlotClass[][] }
```

sroy3 (approved):

extension/src/plots/webview/contract.ts
julieg18: Using an empty array to represent an image that has no classes is doable and I'm happy to change if it benefits code simplicity or code optimization! But we would still face the same issue of needing to add an extra check for seeing if an array is just filled with empty arrays... 

Though now that I look over this code again, I think the check would be doable without adding an extra loop (what I originally thought and was trying to avoid). I'll give it a shot and see how it goes!

julieg18: Updated to use a nested array! @sroy3, can you take another look at this PR when you have some time since I've basically altered most of my original changes 😅.

PR #5340, opened 2024-02-26
Fix repeatedly failing windows test timeouts

julieg18:
~A couple tests are repeatedly failing in windows. Attempting to fix by adjusting the timeouts.~ Tests are now running correctly, but there are still some issues that can be corrected: 

* `timeout` being set on `describe` blocks
* `describe` blocks being nested inside of `describe` blocks 
* `should be able to create a background process that does not exit when the parent process exits` needing a larger timeout

### DIFF
--- a/extension/src/test/suite/experiments/workspace.test.ts
+++ b/extension/src/test/suite/experiments/workspace.test.ts
@@ -152,8 +152,8 @@ suite('Workspace Experiments Test Suite', () => {
 
       await testFile('params.yaml')
       await testFile('dvc.yaml')
-    })
-  }).timeout(WEBVIEW_TEST_TIMEOUT)
+    }).timeout(WEBVIEW_TEST_TIMEOUT)
+  })
 
   describe('dvc.modifyWorkspaceParamsAndQueue', () => {
     it('should be able to queue an experiment using an existing one as a base', async () => {
--- a/extension/src/test/suite/cli/index.test.ts
+++ b/extension/src/test/suite/cli/index.test.ts
@@ -69,7 +69,7 @@ suite('CLI Suite', () => {
         })
       )
       expect(killed).to.be.true
-    })
+    }).timeout(10000)
 
     it('should cleanup all non-background processes on dispose', async () => {
       const processStarted = disposable.track(new EventEmitter<CliEvent>())
--- a/extension/src/test/suite/experiments/index.test.ts
+++ b/extension/src/test/suite/experiments/index.test.ts
@@ -2036,8 +2036,8 @@ suite('Experiments Test Suite', () => {
         ['main', 'other-branch']
       ])
       expect(sorts).to.deep.equal([{ descending: true, path: paramPath }])
-    })
-  }).timeout(WEBVIEW_TEST_TIMEOUT)
+    }).timeout(WEBVIEW_TEST_TIMEOUT)
+  })
 
   describe('persisted state', () => {
     const firstSortDefinition = {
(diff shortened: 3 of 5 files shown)


extension/src/test/suite/experiments/workspace.test.ts
```diff
@@ -152,8 +152,8 @@ suite('Workspace Experiments Test Suite', () => {
 
       await testFile('params.yaml')
       await testFile('dvc.yaml')
-    })
-  }).timeout(WEBVIEW_TEST_TIMEOUT)
+    }).timeout(WEBVIEW_TEST_TIMEOUT)
```
julieg18, line 155: There seems to be an issue with tests inside of describe blocks not getting the timeout. This test was failing with a timeout of `6000` until I moved the `.timeout`. This PR is also failing with:

```
  1) Experiments Test Suite
       Sorting
         should not show duplicate rows when sorted:
     Error: Timeout of 6000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (d:\a\vscode-dvc\vscode-dvc\extension\dist\test\suite\experiments\index.test.js)
  	at listOnTimeout (node:internal/timers:569:17)
  	at process.processTimers (node:internal/timers:512:7)
```

This sorting test is also relying on its parent describe block for its set webview timeout.

julieg18: Reviewed `mocha` prs and issues, but not seeing anything. It's also strange that we're hitting all the timeouts today when we haven't merged anything recently related to `mocha` (as far as I can tell).

mattseddon: I would guess there shouldn't be multiple `describe` blocks nested (see below)

julieg18: Why shouldn't there be multiple `describe` blocks in a block? The mocha docs show examples of this. Or am I misunderstanding something?

mattseddon: > Why shouldn't there be multiple describe blocks in a block? The mocha docs show examples of this. Or am I misunderstanding something?

Have a look at [this](https://stackoverflow.com/questions/55415555/nested-describe-and-behaviour-of-dynamically-created-its) and [this](https://mochajs.org/#run-cycle-overview). For our purposes, we want to avoid extra complexity/parallelization. I have very little faith that `suite -> describe -> describe -> it` acts the same way as `suite -> describe -> it`

> This sorting test is also relying on its parent describe block for its set webview timeout.

`describe` blocks should not have `.timeout` added to them. If you see it that is an error.

julieg18: > Have a look at [this](https://stackoverflow.com/questions/55415555/nested-describe-and-behaviour-of-dynamically-created-its) and [this](https://mochajs.org/#run-cycle-overview). For our purposes, we want to avoid extra complexity/parallelization. I have very little faith that suite -> describe -> describe -> it acts the same way as suite -> describe -> it

Makes sense, thanks for the clarifications! I'm only seeing one example of `suite -> describe -> describe ->` which is our `Custom Plots Creation`. None of those tests have failed that I can see, but it would still be quick to remove the extra nesting there.

> describe blocks should not have .timeout added to them. If you see it that is an error.

Oh, ok that explains the issue. I actually couldn't find documentation on `.timeout` which explains how I missed this 🤦‍♀️. 

The tests appear to be running normally again but I'll go ahead and fix the issues mentioned above to ensure this doesn't happen again.

mattseddon (approved):

PR #5561, opened 2024-06-12
fix vscode integration tests - altered quickpick behaviour

mattseddon:


extension/src/test/suite/util.ts
```diff
@@ -108,9 +108,8 @@ const toggleQuickPickItem = async (
   itemsLength: number,
   quickPick: NumberQuickPick
 ) => {
-  for (let itemInd = 1; itemInd <= itemsLength; itemInd++) {
-    await commands.executeCommand('workbench.action.quickOpenSelectNext')
-
+  for (let itemInd = 1; itemInd <= itemsLength + 1; itemInd++) {
+    await commands.executeCommand('workbench.action.quickOpenNavigateNext')
```
mattseddon, line 112: [F] The quick pick no longer selects the first item by default so we have to navigate past the input bar on each loop

PR #5626, opened 2024-07-16
rename iterative.ai to dvc.ai in Studio domain

shcheklein:
Otherwise I was not able to connect the Studio:

Screenshot 2024-07-15 at 9 20 11 PM

probably this is the issue. But even if it's not it's better to update these domains.


shcheklein:
@mattseddon is this CI failure caused by the change (can you take a brief look please)?

mattseddon (approved):

Issue #5849, opened 2024-12-03
Unresponsive extension


bartvanerp:
Hi folks!

I was using the extension last week and it was a pleasure! Now this week I noticed that the extension did not load anymore. In VSCode it seems to never stop loading, both in the bottom bar there is a loading circle and in the DVC tab it also keeps on loading. I checked the Developer Tools and there I get the message `INFO UNRESPONSIVE extension host: starting to profile NOW`. It only happens with the DVC extension enabled, however, with DVC activated it also blocks the other extensions, such as the built-in git tree. 

I checked my local DVC installation and all seems fine. Updated everything to the latest versions and command line interface also seems to behave correctly. Any ideas what could cause this?

shcheklein:
As a first step to see what is happening here, could you please share this:

![Image](https://github.com/user-attachments/assets/6fe3d40a-88b1-4e5d-955a-0d304d2c81df)
![Image](https://github.com/user-attachments/assets/f02043d6-7ae7-4fd9-8a2e-8a8d53188cb4)

bartvanerp:
Hi Ivan, please find below the log:
```
[version: 1.2.23, 2024-12-04T09:33:00.830Z, pid: 15121] > dvc --version - INITIALIZED
[version: 1.2.23, 2024-12-04T09:33:00.970Z, pid: 15121] > dvc --version - COMPLETED (141ms)
[version: 1.2.23, 2024-12-04T09:33:00.971Z, pid: 15123] > dvc config studio.token - INITIALIZED
[version: 1.2.23, 2024-12-04T09:33:00.972Z, pid: 15125] > dvc config studio.offline - INITIALIZED
[version: 1.2.23, 2024-12-04T09:33:00.973Z, pid: 15127] > git rev-parse --show-toplevel - INITIALIZED
[version: 1.2.23, 2024-12-04T09:33:00.974Z, pid: 15129] > dvc data status --granular --unchanged --json - INITIALIZED
[version: 1.2.23, 2024-12-04T09:33:00.974Z, pid: 15131] > git ls-files --others --exclude-standard - INITIALIZED
[version: 1.2.23, 2024-12-04T09:33:00.975Z, pid: 15133] > git ls-files --others --exclude-standard --directory --no-empty-directory - INITIALIZED
[version: 1.2.23, 2024-12-04T09:33:00.976Z, pid: 15135] > git rev-parse --show-toplevel - INITIALIZED
[...]

PR #5853, opened 2024-12-07
Upgrade extension telemetry package

mattseddon:
The previous attempt to upgrade this package to `0.9.0` resulted in all events being dropped before they arrived in application insights but we didn't try to use the connection string last time (see https://github.com/microsoft/vscode-extension-telemetry/issues/192 for some more details).

https://github.com/iterative/vscode-dvc/issues/5849 may be caused by the fact we are running an old version of Telemetry that is no longer compatible with the current version of VS Code.

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -1609,7 +1609,7 @@
   },
   "dependencies": {
     "@hediet/std": "0.6.0",
-    "@vscode/extension-telemetry": "0.8.5",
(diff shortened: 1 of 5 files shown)


shcheklein (approved):

mattseddon:
I built and installed the extension locally and it seems to be sending events.

